PR #1, opened 2020-12-09
Use go modules.

hpidcock:
Use go modules for project to pin correct versions + allow out of GOPATH building.


benhoyt:
@niemeyer Could you please review/approve this? I have basically this same thing in my local version for testing.

Though I'd recommend Go 1.14 for now, to match what we're currently using for Juju.

PR #2, opened 2021-01-27
Makefile for building pebble

martinrusev:
This PR includes the following changes:

- A Makefile for building the project locally, based on https://github.com/juju/juju/blob/develop/Makefile
- go.mod with the project dependencies, set to 1.14 to match Juju

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+_build/
 cmd/pebble/pebble
 cmd/version_generated.go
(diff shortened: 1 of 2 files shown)


niemeyer:
Thanks for the PR, Martin, but on quick look this doesn't look like an improvement over just using the "go" tool itself directly. Is there something specific you're looking to achieve with it?

PR #4, opened 2021-01-28
Add boot id to system-info + public client + gofmt changes.

hpidcock:
### BootID
BootID is normally sourced from procfs, but inside _containerd_ this value is source from the host.
On LXD it appears LXD mounts a tmpfs over /proc/sys/kernel/random/boot_id to emulate the behavior of random id for boot.
If Pebble is PID 1, we consider it running as the entrypoint of a container, this logic should change at some point.

### Public Client Package
This PR makes the client package public for use outside of the pebble root package. This is inline with snapd and useful for juju.

### Gofmt + import changes
A few files needed to be run through gofmt + some import cleanup.


internal/daemon/api_test.go
```diff
@@ -80,13 +80,18 @@ func (s *apiSuite) TestSysInfo(c *check.C) {
 
 	d := s.daemon(c)
 	d.Version = "42b1"
+	state := d.overlord.State()
+	state.Lock()
+	state.VerifyReboot("ffffffff-ffff-ffff-ffff-ffffffffffff")
+	state.Unlock()
```
niemeyer, line 86: Just for the record, I've been in touch with the snapd team so that we hopefully change this at some point, as this sort of functionality doesn't make much sense on the state itself, but it should be easy to fix it when the time comes.

internal/overlord/overlord.go
```diff
@@ -140,13 +141,23 @@ func (o *Overlord) addManager(mgr StateManager) {
...
 	}
+	// If pebble is PID 1 we don't care about /proc/sys/kernel/random/boot_id
+	// as we are most likely running in a container. LXD mounts it's own boot_id
+	// to correctly emulate the boot_id behaviour of non-containerized systems.
+	// Within containerd/docker, boot_id is consistent with the host, which provides
+	// us no context of restarts.
+	if os.Getpid() == 1 {
+		curBootID, err = strutil.UUID()
+		if err != nil {
+			return nil, fmt.Errorf("fatal: failed to generate psuedo boot-id: %v", err)
```
niemeyer, line 158: As a design convention I've tried to keep, for consistency in wording all variants such as "failed to", "unable to", "could not", "can not", "error while", etc, are all always spelled consistently as _"cannot"_. But this is nitpick-level detail, and if this is the only thing in this PR I'll just fix it later.

internal/overlord/state/state.go
```diff
@@ -316,12 +315,18 @@ func (s *State) VerifyReboot(curBootID string) error {
 	return nil
 }
 
+// BootID returns the current boot id set by VerifyReboot.
+func (s *State) BootID() string {
```
niemeyer, line 319: We'll need to move this as well at some point, but doing the right thing should be obvious once the API is fixed as described above.

niemeyer (approved):
A few comments FYI, but all looks good, thanks!

internal/overlord/overlord.go
hpidcock: I like this, thanks for the suggestion.

hpidcock:
@niemeyer could you please merge this if you are happy with it.

PR #7, opened 2021-02-10
Ignore pid=0 in socket remote address.

hpidcock:
A UNIX socket connection between process namespaces _appears_ to always have a pid of 0.
I have tested with non-root users and they are passed correctly by the kernel, I only tested with containerd in microk8s, so this circumstance might be just related to that environment.

This change ignores `pid == ucrednetNoProcess` and solely relies on uid.
The remote address still has to be properly formed, this is enforced by the strict regex.

### DIFF
--- a/internal/daemon/ucrednet.go
+++ b/internal/daemon/ucrednet.go
@@ -27,7 +27,9 @@ import (
 var errNoID = errors.New("no pid/uid found")
 
 const (
-	ucrednetNoProcess = int32(0)
+	// ucrednetNoProcess is -1 to avoid the case when pid=0 is returned from the kernel
+	// because the connecting process is in an unrelated process namespace.
+	ucrednetNoProcess = int32(-1)
 	ucrednetNobody    = uint32((1 << 32) - 1)
 )
 
--- a/internal/daemon/ucrednet_test.go
+++ b/internal/daemon/ucrednet_test.go
@@ -194,3 +194,11 @@ func (s *ucrednetSuite) TestGetSneak(c *check.C) {
 	c.Check(uid, check.Equals, ucrednetNobody)
 	c.Check(socket, check.Equals, "")
 }
+
+func (s *ucrednetSuite) TestGetWithZeroPid(c *check.C) {
+	pid, uid, socket, err := ucrednetGet("pid=0;uid=42;socket=/run/snap.socket;")
+	c.Check(err, check.IsNil)
+	c.Check(pid, check.Equals, int32(0))
+	c.Check(uid, check.Equals, uint32(42))
+	c.Check(socket, check.Equals, "/run/snap.socket")
+}


internal/daemon/ucrednet.go
```diff
@@ -49,7 +49,9 @@ func ucrednetGet(remoteAddr string) (pid int32, uid uint32, socket string, err e
 		}
 		socket = subs[3]
 	}
-	if pid == ucrednetNoProcess || uid == ucrednetNobody {
+	// Socket connections between process namespaces, pid appears to always be 0,
+	// but uid is visible.
+	if uid == ucrednetNobody {
```
niemeyer, line 54: Do you have more details on this?  Is the pid coming from the other side as zero, and if so is that expected in the environment we're using it?  The comment feels a bit vague ("appears to") for something we want a strict understanding as it might bite us in the future. Also, a pid zero here could mean both an explicit pid=0, or a wrong match above, so we'll want to make a distinction here, and treat each case accordingly after we understand it properly.

niemeyer (changes requested):

internal/daemon/ucrednet.go
hpidcock: In kernel src:
net/core/socket.c cred_to_ucred call to
kernel/pid.c pid_vnr call to
kernel/pid.c pid_nr_ns which returns 0 if the process is not visible from the current process namespace.

Since the connecting process is in a non-related process namespace, the daemon process can't see the connecting process and therefore unable to resolve the pid.

hpidcock: I will fix this tomorrow for the invalid case vs explicit zero (the case we care about here)

niemeyer: Thanks, yeah that would be great. We just need a slightly more clear and strict comment explaining that then, and have a proper difference between bad/unset, and good that includes zero in such cases.

For bad data -1 would be okay, assuming the overflow is outside of the data range for the particular variable type (IOW, any valid pid falls under the max for the data type).

internal/daemon/ucrednet.go
```diff
@@ -27,7 +27,9 @@ import (
 var errNoID = errors.New("no pid/uid found")
 
 const (
-	ucrednetNoProcess = int32(0)
+	// ucrednetNoProcess is -1 to avoid the case when pid=0 is returned from the kernel
+	// because the connecting process is in an unrelated process namespace.
+	ucrednetNoProcess = int32(-1)
```
niemeyer, line 32: Is this safe to use or are PIDs in a positive 32bit namespace? /me checks
```
$ cat /proc/sys/kernel/pid_max
4194304
```
Alright, this should be okay.

niemeyer (approved):
Thank you!

PR #8, opened 2021-02-14
Support $PEBBLE_SOCKET and add $PEBBLE default.

niemeyer:
$PEBBLE now defaults to /var/lib/pebble/default.

### DIFF
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -337,7 +337,7 @@ func run() error {
 	}
 
 	pebbleDir := os.Getenv("PEBBLE")
-	if pebbleDir == "" || !osutil.IsDir(pebbleDir) {
+	if pebbleDir != "" && !osutil.IsDir(pebbleDir) {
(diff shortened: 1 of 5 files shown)


cmd/pebble/main.go
```diff
@@ -340,7 +340,10 @@ func run() error {
 	if pebbleDir == "" || !osutil.IsDir(pebbleDir) {
 		return fmt.Errorf("$PEBBLE must point to a pebble directory")
```
hpidcock, line 341: This here needs to allow pebbleDir to be unspecified (otherwise a default is pointless).

hpidcock (commented):
This is good thank-you but I think we need to allow /var/lib/pebble to not exist and be treated the same as an empty dir.

But for that to make sense we will need the ability to dynamically load layers via the API.

cmd/pebble/main.go
niemeyer: Good point, thanks.

PR #9, opened 2021-02-15
 Define and enforce layer filename format

niemeyer:
This also splits layer keys into independent order and label
fields. The new file format looks like 001-some-label.yaml,
where both order and label are unique. This drops the benefit of
timestamps, originally described as it being easy to blindly
create, in favor of it being better to read and manipulate.
Seems like a good tradeoff to make.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,19 +11,16 @@ designed with unique features that help with more specific use cases.
 Pebble is organized as a single binary that works as a daemon and also as a
 client to itself. When the daemon runs it loads its own configuration from the
 _$PEBBLE_ directory, as defined in the environment, and also writes down in
-that same directory its state and unix sockets for communication.
+that same directory its state and unix sockets for communication. If that variable
+is not defined, Pebble will attempt to look for its configuration from a default
+system-level setup at _/var/lib/pebble/default_. Using that directory is encouraged
+for whole-system setup such as when using Pebble to control services in a container.
 
-The _$PEBBLE_ directory must also contain a _layers/_ subdirectory that holds a
-list of yaml files conventionally named as `2020-12-01T15:00:00.yaml`.  The reason
-for the timestamp in the filename is that these configuration files are layered,
-as the directory name implies. That is, each layer sits above the former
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -11,19 +11,16 @@ designed with unique features that help with more specific use cases.
 Pebble is organized as a single binary that works as a daemon and also as a
 client to itself. When the daemon runs it loads its own configuration from the
 _$PEBBLE_ directory, as defined in the environment, and also writes down in
-that same directory its state and unix sockets for communication.
+that same directory its state and unix sockets for communication. If that variable
+is not defined, Pebble will attempt to look for its configuration from a default
+system-level setup at _/var/lib/pebble/default_. Using that directory is encouraged
```
benhoyt, line 16: Nit: For path names, would `<code>` spans (via backticks) be more typical / readable than italics (via underscores)? For env vars like `$PEBBLE`, seems like you've mixed underscores and backticks -- would be good to be consistent.

internal/setup/setup.go
```diff
@@ -250,29 +254,59 @@ func ParseLayer(key string, data []byte) (*Layer, error) {
 	return &layer, err
 }
 
+var fnameExp = regexp.MustCompile("^([0-9]{3})-([a-z](?:-?[a-z0-9]){2,}).yaml$")
```
benhoyt, line 257: This regex seems to prohibit a label less than 3 chars -- is that restriction necessary? For example, `123-go.yaml` or `001-a.yaml` are disallowed.

benhoyt (approved):

hpidcock (approved):

README.md
niemeyer: Thanks for raising that. I've moved all cases (hopefully) of code, paths, and environment variables to backticks.

internal/setup/setup.go
niemeyer: This is intentional, in part precisely to avoid things such as `123-a.yaml`. We can always open up further if we learn this was a bad choice, but we cannot restrict further once this is released into stable, so better to err on the side of caution.

PR #10, opened 2021-02-22
Dynamic add-layer and get-plan APIs

benhoyt:
An API to add in a layer dynamically (`POST /v1/layers`), and one to fetch the plan / combined layers view (`GET /v1/plan`).

There are two different modes for adding:

* `pebble add label layer.yaml`: this appends a new layer to the top of the layers
* `pebble add --combine label layer.yaml`: this combined the new layer with an existing layer of the given label (or appends if a layer with the label doesn't yet exist)

This also adds support for these two calls to the Go client and the CLI (`pebble add` and `pebble plan`), and tests for all of the above.


internal/daemon/api_setup.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2014-2020 Canonical Ltd
```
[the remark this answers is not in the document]
benhoyt: Good call, fixed.

internal/daemon/api_setup.go
```diff
@@ -0,0 +1,48 @@
...
+	layer, err := servmgr.FlattenedSetup()
+	if err != nil {
+		return statusInternalError("cannot fetch flattened layer: %v", err)
+	}
+	return SyncResponse(string(layer))
+}
+
+func v1PostLayer(c *Command, r *http.Request, x *userState) Response {
+	var payload struct {
+		Layer string `json:"layer"`
```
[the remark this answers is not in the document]
benhoyt: Good question. We don't currently accept or output the layer in JSON format anywhere else, just YAML (in the layer.yaml files). But it's an interesting idea. To me it seems like accepting the JSON struct directly would be more pure, but accepting the YAML is a bit more user-friendly. Worth further discussion tonight.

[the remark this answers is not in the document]
benhoyt: Per discussion, we're going to keep using YAML for now, but make this explicit using `"format": "yaml"` in the request (leaving it open for other formats like "dict" later).

benhoyt:
@niemeyer I've done both of these in `POST /v1/layers` as discussed. However, I'm still not at all sure it's consistent with either existing Pebble endpoints of Snapd endpoints -- they always seem to use GET for fetching information / views on config and POST for changes to the server. For example, here are all the Snapd endpoints:

```
 GET: appIconGet,
 GET: assertsFindMany,
 GET: checkThemes,
 GET: getAliases,
 GET: getAppsInfo,
 GET: getAssertTypeNames,
 GET: getChange,
 GET: getChanges,
 GET: getConnections,
 GET: getDebug,
 GET: getLogs,
 GET: getModel,
 GET: getPprof,
 GET: getSections,
 GET: getSerial,
 GET: getSnapConf,
 GET: getSnapFile,
 GET: getSnapInfo,
 GET: getSnapshotExport,
 GET: getSnapsInfo,
 GET: getSystemRecoveryKeys,
 GET: getSystems,
 GET: getUsers,
 GET: getValidationSet,
 GET: getWarnings,
 GET: interfacesConnectionsMultiplexer,
 GET: listSnapshots,
 GET: listValidationSets,
 GET: readyToBuy,
 GET: searchStore,
 GET: sysInfo,
 GET: tbd,
POST: abortChange,
POST: ackWarnings,
POST: applyValidationSet,
POST: changeAliases,
POST: changeInterfaces,
POST: changeSnapshots,
POST: consoleConfStartRoutine,
POST: doAssert,
POST: installThemes,
POST: loginUser,
POST: logoutUser,
POST: postApps,
POST: postBuy,
POST: postCohorts,
POST: postCreateUser,
POST: postDebug,
POST: postModel,
POST: postSnap,
POST: postSnapDownload,
POST: postSnaps,
POST: postSystemsAction,
POST: postSystemsAction,
POST: postUsers,
POST: runSnapctl,
```

As you can see, GETs are all "return a view on information", and POSTs are "make some change to server state". The "flatten" actually is really a (flattened) view on the Pebble configuration, rather than performing a mutating flatten operation on the server.

At the end of the day it doesn't matter that much, but given the above, I think `GET /v1/layers?view=flattened` or something would be more consistent with existing endpoints in both Pebble and Snap, as well as more typical for an HTTP API (GETs are informational and cacheable).

internal/overlord/servstate/manager.go
```diff
@@ -53,6 +53,76 @@ func (m *ServiceManager) reloadSetup() error {
...
+			return 0, err
+		}
+		dynamicFlattened.Order = last.Order
+		newOrder = dynamicFlattened.Order
+		newSetup = &setup.Setup{Layers: append(layers[:len(layers)-1], dynamicFlattened)}
+		newFlattened, err = newSetup.Flatten()
+		if err != nil {
+			return 0, err
+		}
+	} else if last != nil {
```
hpidcock, line 104: Let's just simplify this to
```
if last != nil && last.IsDynamic() {
		// Last layer is dynamic, merge new layer into existing dynamic layer
		dynamicSetup := &setup.Setup{Layers: []*setup.Layer{last, layer}}
		dynamicFlattened, err := dynamicSetup.Flatten()
		if err != nil {
			return 0, err
		}
		dynamicFlattened.Order = last.Order
		newOrder = dynamicFlattened.Order
		newSetup = &setup.Setup{Layers: append(layers[:len(layers)-1], dynamicFlattened)}
		newFlattened, err = newSetup.Flatten()
		if err != nil {
			return 0, err
		}
	} else {
		// Last layer is not dynamic, add new dynamic layer
		layer.Order = 1
		if last != nil {
			layer.Order = last.Order + 1
		}
		newOrder = layer.Order
		newSetup = &setup.Setup{Layers: append(layers, layer)}
		newFlattened, err = newSetup.Flatten()
		if err != nil {
			return 0, err
		}
	}
```

internal/overlord/servstate/manager.go
```diff
@@ -53,6 +53,76 @@ func (m *ServiceManager) reloadSetup() error {
...
+	if len(layers) > 0 {
+		last = layers[len(layers)-1]
+	}
+
+	var newOrder int
+	var newSetup *setup.Setup
+	var newFlattened *setup.Layer
+	if last != nil && last.IsDynamic() {
+		// Last layer is dynamic, merge new layer into existing dynamic layer
+		dynamicSetup := &setup.Setup{Layers: []*setup.Layer{last, layer}}
```
hpidcock, line 92: ptr to Setup instance not necessary, also I would prefer to see Flatten become a package level function so you can just do setup.FlattenLayers(last, layer) or setup.FlattenLayers(layers...)

internal/overlord/servstate/manager.go
benhoyt: Heh, I was on the fence about this and actually had it the other way at first, but then I changed it because "flat is better than nested". But I think you're right, this way is slightly simple/shorter, so changed it back. :-)

internal/overlord/servstate/manager.go
benhoyt: You don't need a pointer here, but it's stored as a `*Setup` on the `ServiceManager` instance, so it's a bit nicer to do so (and it's also consistent with `*Layer` and the fact that the `Setup` methods are on `*Setup`). So I'm going to leave this as is.

I agree with you about `setup.FlattenLayers`, thanks -- done.

benhoyt:
Gustavo's terminology plan (so to speak): https://www.notion.so/Pebble-terminology-and-API-0673ef8b95db4982a4fb5537f4964717

hpidcock (approved):

client/layers.go
```diff
@@ -0,0 +1,63 @@
...
+
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+type MergeLayerOptions struct {
+	Layer string
```
niemeyer: This is related to the question you've asked during our call: Go is statically typed, so we need to think well which parameters we have here to make it future proof. We've done the "format" trick in the payload to enable us to send it as structured data, so we need to ponder for a moment that might be here eventually. I suspect "Layer" would be the most natural name for the structured type, which means we need to move this out onto a separate field. Maybe LayerData as a []byte, and document it as being yaml?  Then, in the future we may introduce "Format", if necessary, or just "Layer" as something strongly typed.

client/plan.go
```diff
@@ -0,0 +1,56 @@
...
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/url"
+)
+
+type MergeLayerOptions struct {
+	Layer string
```
niemeyer, line 24: This is related to the question you've asked during our call: Go is statically typed, so we need to think well which parameters we have here to make it future proof. We've done the "format" trick in the payload to enable us to send it as structured data, so we need to ponder for a moment that might be here eventually. I suspect "Layer" would be the most natural name for the structured type, which means we need to move this out onto a separate field. Maybe LayerData as a []byte, and document it as being yaml? Then, in the future we may introduce "Format", if necessary, or just "Layer" as something strongly typed.

client/plan.go
```diff
@@ -0,0 +1,56 @@
...
+	if err := json.NewEncoder(&body).Encode(&payload); err != nil {
+		return err
+	}
+	_, err := client.doSync("POST", "/v1/layers", nil, nil, &body, nil)
+	return err
+}
+
+type GetPlanOptions struct{}
+
+func (client *Client) GetPlan(_ *GetPlanOptions) (layerYAML string, err error) {
```
niemeyer, line 47: In Go it's often conventional to avoid the Get prefix for those cases. We already have a few similar examples here in the client package itself. Also, similar to the above, we might have this as PlanData, and return it as a []byte which is most conventional for raw text file data. This way we save the Plan name for the structured version of it. PlanOptions might remain as such since other than the format, any options we may have will be common for both.

cmd/pebble/cmd_help.go
```diff
@@ -181,7 +181,7 @@ var helpCategories = []helpCategory{{
 
 var (
 	longPebbleDescription = strings.TrimSpace(`
-Pebble lets you control services and perform managemnet actions on
+Pebble lets you control services and perform management actions on
```
niemeyer, line 184: Thanks!

cmd/pebble/cmd_plan.go
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2021 Canonical Ltd
```
niemeyer, line 1: The file is misnamed (cmd_plan vs. `cmd_<cmd>`).

cmd/pebble/cmd_plan.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type cmdMerge struct {
+	clientMixin
+	Positional struct {
+		LayerPath string `positional-arg-name:"<layer-path>" required:"1"`
+	} `positional-args:"yes"`
+}
+
+var shortMergeHelp = "Dynamically merge a layer on top of the plan's layers"
```
niemeyer, line 33: I've spent some time here thinking about how we organize things in a way that, as usual, is fast for us to support right now, while leaving the road open for us to implement all the things we want in the near future. I've struggled a bit for two reasons:

1) "merge" seems to mean something else here vs. the internal "override" directive that exists in services. The best way to see that conflict is to think about the fact we have "merge" vs. "replace", in that directive. But if we have "merge" vs. "replace", we then have "merge" with "merge", or "merge" with "replace". Ugh. Terminology is hard.

2) Second reason is that this seems like a more fundamental operation: add a layer. There will likely be several different ways of adding layers: combining them, not combining them, dynamically, statically, persisting, etc. We don't want independent commands for all of those cases. Instead, it'd be much nicer to have one _solid_ command that has the most obvious default behavior, and then flags to customize the remaining options.

So, here is an idea: we already have a word for the idea of getting two different layers and merging them together into a single layer: combining. So, what if we changed both the API and the CLI here so that we spell out what's going on here a bit more precisely?  In other words, this operation would become:
```
$ pebble add --combine --dynamic <layer path>
```
Which is translated into an operation with `action: "add", "combine": true, "dynamic": true`.

In the near future, we'll fill the gaps and introduce the opportunity to not have these options set. For now, we can simply error out on the server side if they aren't set (`error: layer must be combined and dynamic - other options coming soon`), so that this PR can land sooner rather than later. The CLI and client can already support these options properly by forwarding them over, since that's trivial.

How does that sound?

internal/daemon/api_plan.go
```diff
@@ -0,0 +1,65 @@
...
+func v1GetPlan(c *Command, r *http.Request, _ *userState) Response {
+	format := r.URL.Query().Get("format")
+	if format != "yaml" {
+		return statusBadRequest("invalid format %q", format)
+	}
+
+	servmgr := c.d.overlord.ServiceManager()
+	plan, err := servmgr.Plan()
+	if err != nil {
+		return statusInternalError("cannot get plan: %v", err)
```
niemeyer, line 33: I think we can just forward the underlying error message here, since that operation ("getting the plan") is already what the method is supposed to do, so if we need to decorate an underlying naked error that should have happened inside it. It's of course not a big deal, but that kind of practice avoids repetitive error messages such as "cannot get plan: cannot generate plan: ..." or similar.

internal/daemon/api_plan.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	planYAML, err := yaml.Marshal(plan)
+	if err != nil {
+		return statusInternalError("cannot serialize plan: %v", err)
```
niemeyer, line 37: This is the counter example: the error message here is relevant as it's adding lacking context without which the message would be meaningless in the Pebble context.

internal/overlord/servstate/handlers.go
```diff
@@ -64,9 +64,9 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 		return err
 	}
 
-	service, ok := m.flattened.Services[req.Name]
+	service, ok := m.plan.Services[req.Name]
```
niemeyer, line 67: Nice, that's indeed more pleasant to read.

internal/overlord/servstate/manager.go
```diff
@@ -39,34 +38,97 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string) (*Se
...
-	m.setup = setup
-	m.flattened = flattened
-	return nil
+	defer releasePlan()
+	return m.plan, nil // TODO: copy?
+}
+
+// MergeLayer merges the given layer YAML into the dynamic layers, returning
+// the new layer's "order" (won't increase if adding another dynamic layer).
+func (m *ServiceManager) MergeLayer(layerYAML []byte) (int, error) {
```
niemeyer, line 62: This probably needs terminology tuning depending on previous points. The implementation below looks great, though.

internal/plan/plan.go
```diff
@@ -27,33 +27,36 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-type Setup struct {
-	Layers []*Layer
+type Plan struct {
+	Layers      []*Layer            `yaml:"-"`
+	Summary     string              `yaml:"summary,omitempty"`
+	Description string              `yaml:"description,omitempty"`
```
niemeyer, line 33: Per conversation online, we need to decide what to do with these before we have fields here.

internal/plan/plan.go
```diff
@@ -27,33 +27,36 @@ import (
...
 }
 
-type Layer struct {
-	Order       int
-	Label       string
-	Summary     string
-	Description string
-	Services    map[string]*Service
+func (l *Layer) IsDynamic() bool {
+	return l.Label == ""
```
niemeyer, line 46: Will it ever make sense to have more than one dynamic layer?  Perhaps one atop and one in the middle?  I guess we might eventually have dynamic layers with labels and orders. Not a big deal either way because we can change this method in a compatible way, but food for thought.

internal/plan/plan.go
```diff
@@ -256,7 +263,7 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 
 var fnameExp = regexp.MustCompile("^([0-9]{3})-([a-z](?:-?[a-z0-9]){2,}).yaml$")
 
-func ReadLayersDir(dirname string) ([]*Layer, error) {
+func ReadLayersDir(dirname string) (*Plan, error) {
```
niemeyer, line 266: Shouldn't this still return []*Layers, after all it's literally "read layers dir"?  That's the original idea at least. The ReadDir below does return a *Plan, though.

internal/plan/plan.go
```diff
@@ -312,16 +319,21 @@ func ReadLayersDir(dirname string) ([]*Layer, error) {
 		}
 		layers = append(layers, layer)
 	}
-	return layers, nil
+	return CombineLayers(layers...)
```
niemeyer, line 322: Similarly, it feels like this should return a *Layer, not a *Plan. This is the same operation we'll get when combining the top dynamic layer with the new dynamic layer provided over the wire. So combining feels like a general operation that results a new, combined, layer.

niemeyer (changes requested):
Thanks a lot, Ben. A lot of good code here. I have a few comments, but they are mainly aligning terminology and preparing for next steps. There's nothing significant to change I believe.

client/plan.go
```diff
@@ -0,0 +1,75 @@
...
+
+type AddLayerOptions struct {
+	// LayerData is the new layer in YAML format.
+	LayerData []byte
+
+	// Action is the action performed when adding the new layer. Only
+	// "combine" is supported right now, which means combine the new layer
+	// into the existing dynamic layer, or add a new dynamic layer if none
+	// exists. If not set, default to "combine".
+	Action AddLayerAction
```
benhoyt, line 37: Arguably this should be called "method" or similar (the method to use when adding).

cmd/pebble/cmd_add.go
```diff
@@ -0,0 +1,71 @@
...
+	"io/ioutil"
+
+	"github.com/jessevdk/go-flags"
+
+	"github.com/canonical/pebble/client"
+)
+
+type cmdAdd struct {
+	clientMixin
+	Action     string `long:"action"`
```
benhoyt, line 28: Again, I went back and forth between this name and "method" for this.

client/plan.go
benhoyt: Yep, agreed -- changed to `LayerData []byte`.

client/plan.go
benhoyt: Good call -- changed to `PlanData() ([]byte, error)`.

cmd/pebble/cmd_plan.go
benhoyt: Thanks. I like the idea of "add" with different modes / sub-actions. However, two things I'm not sure about:

1) I think it's a bit annoying to have to specify `--combine --dynamic` when at least for now that's all we have, and I think those are sensible defaults. So I think we should just make that the default, `pebble add <layer_path>`. Can always add `--append` or `--action=append` or something later.
2) For the add algorithm/action, I've gone with `--action=foo` rather than a boolean as there are more than two possibilities, and you don't want multiple booleans with only certain combinations that work. Right now only the default of `--action=combine` is supported, but we could `--action=append` or `--action=insert --index=3` or similar later.
3) I'm not sure we'll need to add static layers, but if we do, I'd prefer `--static` rather than always having to specify `--dynamic` now.

internal/daemon/api_plan.go
benhoyt: Good call -- fixed.

internal/overlord/servstate/manager.go
benhoyt: Fixed: `CombineLayer`.

internal/plan/plan.go
benhoyt: Fixed. Plan no longer has summary/description.

internal/plan/plan.go
benhoyt: Good idea, fixed.

internal/plan/plan.go
benhoyt: Ditto -- fixed.

cmd/pebble/cmd_plan.go
niemeyer: Totally agree on point 1. What we need to figure out is what's the best experience we can design for people interacting with the command line. Then, the default should be based on that understanding, and not on what we're doing just now, because we cannot change defaults once this is out. So what I've been trying to imagine is what will someone running a command such as this implicitly imagine it would do:
```
$ pebble add mylayer.yaml
```
The behavior we've defined to start with is a bit quirky to have as a default, unless we polish it a bit further. Maybe we can do that then?

While reading your comments I got another idea that maybe will polish it enough that we can make it common across API, the CLI, and the framework. How about this: let's introduce the label field right now and force people to provide it. Then, the behavior in the document is tuned just slightly to take into account the label. In other words, the possibilies are:

a. Add new layer at the top (order = max order+1), fail if it exists:
    `$ pebble add mylabel mylabel.yaml`

b. Same as before, but combine if it exists:
    `$ pebble add --combine mylabel mylabel.yaml`

c. Same as _a,_ but persist if it exists:
    `$ pebble add --persist mylabel mylabel.yaml`

d. Same as _b_ + _c_:
    `$ pebble add --combine --persist mylabel mylabel.yaml`

For this PR seems closer to _b_, I think?  So maybe we can just support that for now and then walk on with smaller improvements. If that sounds reasonable to you, I will update the document as well.

The --action proposal seems a bit too close to the wire, which is probably comfortable from a development standpoint, but I'm not sure it'd lead us into a nice experience for people using the CLI itself. In fact, even using the the operator framework it might be a bit unfriendly, in the sense of having to figure indexes, for example.

benhoyt: Per Mattermost discussion, we're going to go with requiring a label as per above. I'll update the PR to support `pebble add label layer.yaml` and `pebble add --combine label layer.yaml` (but leave out `--persist` and other options for now).

benhoyt: This is done now.

client/plan.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	var body bytes.Buffer
+	if err := json.NewEncoder(&body).Encode(&payload); err != nil {
+		return err
+	}
+	_, err := client.doSync("POST", "/v1/layers", nil, nil, &body, nil)
+	return err
+}
+
+type PlanDataOptions struct{}
```
niemeyer, line 59: Per previous review, I think we can name this as PlanOptions. If we ever have non-data plan adding, the options will almost certainly be exactly the same, just with the different type. Otherwise we'll end up with one type per method.

I'm also wondering if we should call the method below as PlanBytes instead of PlanData, as "data" as a bit more general and "bytes" implies some sort of encoding. What do you think?

client/plan_test.go
```diff
@@ -0,0 +1,76 @@
...
+	data, err := cs.cli.PlanData(&client.PlanDataOptions{})
+	c.Assert(err, check.IsNil)
+	c.Check(cs.req.Method, check.Equals, "GET")
+	c.Check(cs.req.URL.Path, check.Equals, "/v1/plan")
+	c.Check(cs.req.URL.Query(), check.DeepEquals, url.Values{"format": []string{"yaml"}})
+	c.Assert(string(data), check.Equals, `
+services:
+ foo:
+  override: replace
+  command: cmd
```
niemeyer, line 74: Just FYI, there's a function inside the setup (now plan) package that reformats yaml so that we can have it properly aligned in cases like this. If it's generally useful, we might move it to the strutil package.

cmd/pebble/cmd_plan.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type cmdPlan struct {
+	clientMixin
+}
+
+var shortPlanHelp = "Show the plan with layers combined"
+var longPlanHelp = `
+The plan command reads the plan (configuration) and displays it as YAML. The
+plan's layers are combined according to Pebble's layer override rules.
```
niemeyer, line 30: It feels like there's just a bit of tension between the concepts described here.

I've tried to take that same idea and create something similar: 
```
The plan command prints out the complete configuration of pebble as it stands
in memory. Layers are combined according to the override rules defined in them.
``` 
How do you like it?

internal/overlord/servstate/manager.go
```diff
@@ -39,34 +49,132 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string) (*Se
...
+	newLayers := append(m.plan.Layers, layer)
+	err := m.updateLayers(newLayers)
+	if err != nil {
+		return err
+	}
+	layer.Order = newOrder
+	return nil
+}
+
+func (m *ServiceManager) updateLayers(layers []*plan.Layer) error {
```
niemeyer, line 105: I wonder if this might be a bit more clear if named updatePlan. On the first pass I was a bit confused when reading appendLayer calling updateLayers. Not a big deal, of course.

internal/overlord/servstate/manager_test.go
```diff
@@ -218,3 +225,242 @@ func (s *S) TestStartFastExitCommand(c *C) {
...
+	return layer
+}
+
+func (s *S) planLayersHasLen(c *C, manager *servstate.ServiceManager, expectedLen int) {
+	plan, err := manager.Plan()
+	c.Assert(err, IsNil)
+	c.Assert(plan.Layers, HasLen, expectedLen)
+}
+
+func (s *S) TestAppendLayer(c *C) {
```
niemeyer, line 273: Nice selection of tests here and below. Thanks!

internal/plan/plan.go
```diff
@@ -225,12 +225,12 @@ func (l *Layer) order(names []string, stop bool) ([]string, error) {
 	return order, nil
 }
 
-func (l *Layer) CheckCycles() error {
+func (l *Layer) checkCycles() error {
```
niemeyer, line 228: One of the reasons this was public is that we need to make sure our effective in-memory configuration never gets corrupted to a point where running it is impossible. I believe we have introduced that possibility now that we can mutate the layer stack without this being called. We can either make this public and call it, or we might make CombineLayers run the check itself.

Not sure about which way is best. Can you please think a bit about it and implement the best approach? As long as we don't allow the issue, we can always tune later.

niemeyer (changes requested):
Thanks for keeping it up for all that time, Ben. I'm pretty sure this is the last review.

client/plan.go
benhoyt: Agreed on both -- done: `PlanDataOptions` -> `PlanOptions` and `PlanData` -> `PlanBytes`

client/plan_test.go
benhoyt: Noted. In this case I just "properly aligned" it in the test (4 spaces like `yaml.Marshal` does) -- I think it's clearer in this case  just write explicitly in the test the bytes you want to see.

cmd/pebble/cmd_plan.go
benhoyt: I can see what you're saying. However, I don't like "as it stands in memory" -- that's not very intuitive not correct, because in memory (taken literally) it's laid out however the Go struct/map internals are laid out. It's a YAML encoding, so I'd rather explicitly say that. I've tweaked to a (hopefully good enough) combination, including the word "effective" which might capture what you mean.

internal/overlord/servstate/manager.go
benhoyt: Agreed, done.

internal/plan/plan.go
benhoyt: Yeah, good call. I still don't think it should be exported, as callers shouldn't have to remember to do that -- the parse/combine functions should just check for it and return error. So I've kept it unexported but made `CombineLayers` check (and added a test for this case).

So now both the functions that create layers (parse and combine) check this. A caller could still bypass it by instantiating a `&plan.Layer{...}` with cycles directly, but I think that's an okay tradeoff, and kind of unavoidable with how the type is defined with all exported fields.

niemeyer: Sounds good, thanks!

cmd/pebble/cmd_plan.go
niemeyer: That's great as well, thanks.

niemeyer (approved):

PR #11, opened 2021-02-23
Run go fmt and tests on push and PR

benhoyt:
Currently `go fmt` is not being run on this codebase, and the tests aren't being run on every PR. This adds a GitHub Actions config file to make that happen. (Also run `go fmt` over the existing code.)


niemeyer:
Thanks!

PR #12, opened 2021-02-26
Create pebble dir on startup if it doesn't exist

benhoyt:
Also a bit of light refactoring to put the environment variable fetching and override logic in one place (and move duplication out of internal/daemon so it's all in cmd/pebble).

This change precedes a change to the Juju K8s code which will use the default Pebble directory `/var/lib/pebble/default` in the container, rather than overriding `PEBBLE` and `PEBBLE_SOCKET`.


cmd/pebble/cmd_run.go
```diff
@@ -181,3 +185,23 @@ out:
 
 	return d.Stop(ch)
 }
+
+// ensurePebbleDir ensures that the pebble directory exists and is a directory.
+// If the directory doesn't exist, create it.
+func ensurePebbleDir(pebbleDir string) error {
```
hpidcock, line 191: Wondering if this only makes sense for the default directory, regardless I think its fine.

cmd/pebble/main.go
```diff
@@ -336,14 +340,8 @@ func run() error {
...
 
-	pebbleDir := os.Getenv("PEBBLE")
-	if pebbleDir != "" && !osutil.IsDir(pebbleDir) {
-		return fmt.Errorf("$PEBBLE must point to a pebble directory")
-	}
-	clientConfig.Socket = os.Getenv("PEBBLE_SOCKET")
-	if clientConfig.Socket == "" {
-		clientConfig.Socket = filepath.Join(pebbleDir, ".pebble.socket")
-	}
+	_, socketPath := getEnvPaths()
```
hpidcock, line 343: This would be fine:
_, clientConfig.Socket = getEnvPaths()

hpidcock (approved):
Yeah pretty happy with this, thank-you.

cmd/pebble/cmd_run.go
```diff
@@ -181,3 +185,23 @@ out:
...
+		}
+		err = os.MkdirAll(pebbleDir, 0755)
+		if err != nil {
+			return fmt.Errorf("cannot create pebble directory: %w", err)
+		}
+		return nil
+	}
+	if !st.IsDir() {
+		return fmt.Errorf("%q is not a directory", pebbleDir)
+	}
```
niemeyer, line 205: The content of this function looks nice, but it should live inside the osutil package next to similar functions we have there. It's also worth checking out to see if we don't have this there already.

cmd/pebble/cmd_run.go
```diff
@@ -108,12 +108,16 @@ func sanityCheck() error {
 }
 
 func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
-
 	t0 := time.Now().Truncate(time.Millisecond)
 
+	pebbleDir, socketPath := getEnvPaths()
+	err := ensurePebbleDir(pebbleDir)
```
niemeyer, line 114: This sounds like a good idea for the juju use case, but it doesn't sound like a good general feature for pebble. Imagine someone makes a typo on the directory name in the CLI, and suddenly there's a second empty directory, which does nothing while pebble is running, while the real one is sitting there still with the actual content. Confusing and unfortunate bug.

We can fix that by putting the feature behind a flag such as `--create-dirs`.

niemeyer (changes requested):
Looks reasonable. Just some minor considerations:

cmd/pebble/cmd_run.go
benhoyt: Done. I added `--create-dirs`. I also realized I can just use `os.MkdirAll` directly, as that has no effect if the directory already exists.

cmd/pebble/cmd_run.go
benhoyt: I realized I can just use `os.MkdirAll` directly, as that has no effect if the directory already exists. Updated to that.

Side node: the `osutil.IsX` functions are a bit weird to me, because they don't return an error if they can't stat. It seems like there's a difference between successfully stat-ed (directory or otherwise) and error stat-ing due to permission denied or whatever.

cmd/pebble/cmd_run.go
benhoyt: This code has gone now, I'm just using `os.MkdirAll`.

cmd/pebble/main.go
benhoyt: Thanks, updated.

cmd/pebble/cmd_run.go
niemeyer: Interesting.. we should check to see if that causes actual issues in call sites, or if they're mainly using the positive side of it to mean "Yeah, I'm sure that's a directory.".

cmd/pebble/cmd_run.go
niemeyer: Thanks!

niemeyer (approved):

PR #13, opened 2021-03-04
Avoid race due to concurrent cmd.Wait() call in doStart/doStop

benhoyt:
doStart creates the exec.Command and starts a goroutine that waits for
it using cmd.Wait(), closing the active.done channel to signal that it
exited. However, nothing looks at that channel. It was ignored, and
doStop created its own channel and called cmd.Wait() on the same
exec.Command instance. This is a no-no, as cmd.Wait(), like most
stdlib functions unless explicitly noted is not concurrency-safe.

This changes simplifies (we already have an active.done channel) and
removes the race (which "go test -race" found). It also runs the tests
in "-race" mode in the GitHub Actions configuration.


niemeyer (approved):
Great catch, thanks!

PR #14, opened 2021-03-09
Add "pebble services" command and API

benhoyt:
This adds a new `GET /v1/services?name=foo,bar` API to fetch service status of all configured services, as well as the Go client and `pebble services` CLI command for the same. It's modeled after the `snap services` command.

See also the [design doc](https://docs.google.com/document/d/1OxaG9WUp2IR_S6qpeh9J5LgfmgW794KO9FJmGqPO_Ng/edit#).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ cmd/VERSION
 *.swp
 *.swo
 .idea
+/pebble
--- a/README.md
+++ b/README.md
@@ -128,11 +128,12 @@ Here are some of the things coming soon:
 
   - [x] Support `$PEBBLE_SOCKET` and default `$PEBBLE` to `/var/lib/pebble/default`
   - [x] Define and enforce convention for layer names
-  - [ ] Dynamic layer support over the API
+  - [x] Dynamic layer support over the API
+  - [x] Configuration retrieval commands to investigate current settings
+  - [x] Status command that displays active services and their current status
+  - [ ] General system modification commands (writing configuration files, etc)
+  - [ ] Improve signal handling, e.g., sending SIGHUP to a service
   - [ ] Terminate all services before exiting run command
-  - [ ] Status command that displays active services and their current status
-  - [ ] Configuration retrieval commands to investigate current settings
-  - [ ] General system modification commands (writing files, etc)
   - [ ] More tests for existing CLI commands
   - [ ] Better log caching and retrieval support
   - [ ] Consider showing unified log as output of `pebble run`
--- a/client/services.go
+++ b/client/services.go
@@ -18,6 +18,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"strings"
 )
 
 type ServiceOptions struct {
@@ -57,4 +59,33 @@ func (client *Client) doMultiServiceAction(actionName string, services []string)
 		"Content-Type": "application/json",
 	}
 	return client.doAsyncFull("POST", "/v1/services", nil, headers, bytes.NewBuffer(data))
+}
+
+// ServicesOptions are the filtering options for querying services.
+type ServicesOptions struct {
+	// Names is the list of service names to query for. If slice is nil or
+	// empty, fetch information for all services.
+	Names []string
+}
+
+// ServiceInfo holds status information for a single service.
+type ServiceInfo struct {
+	Name    string `json:"name"`    // service name
+	Default string `json:"default"` // default state: "start" or "stop"
+	Status  string `json:"status"`  // "active", "inactive", or "error"
+	Message string `json:"message"` // message (e.g., error) or empty string
+}
+
+// Services fetches information about specific services (or all of them),
+// ordered by service name.
+func (client *Client) Services(opts *ServicesOptions) ([]*ServiceInfo, error) {
+	query := url.Values{
+		"names": []string{strings.Join(opts.Names, ",")},
+	}
+	var services []*ServiceInfo
+	_, err := client.doSync("GET", "/v1/services", query, nil, nil, &services)
+	if err != nil {
+		return nil, err
+	}
+	return services, nil
 }
--- a/client/services.go
+++ b/client/services.go
@@ -70,11 +70,27 @@ type ServicesOptions struct {
 
 // ServiceInfo holds status information for a single service.
 type ServiceInfo struct {
-	Name    string `json:"name"`    // service name
-	Default string `json:"default"` // default state: "start" or "stop"
-	Status  string `json:"status"`  // "active", "inactive", or "error"
-	Message string `json:"message"` // message (e.g., error) or empty string
+	Name    string         `json:"name"` // service name
+	Startup ServiceStartup `json:"startup"`
+	Current ServiceStatus  `json:"current"`
 }
+
+// ServiceStartup defines the different startup modes for a service.
+type ServiceStartup string
+
+const (
+	StartupEnabled  ServiceStartup = "enabled"
+	StartupDisabled ServiceStartup = "disabled"
+)
+
+// ServiceStatus defines the current states for a service.
+type ServiceStatus string
+
+const (
+	StatusActive   ServiceStatus = "active"
+	StatusInactive ServiceStatus = "inactive"
+	StatusError    ServiceStatus = "error"
+)
 
 // Services fetches information about specific services (or all of them),
 // ordered by service name.
--- a/cmd/pebble/cmd_help.go
+++ b/cmd/pebble/cmd_help.go
@@ -167,16 +167,20 @@ type helpCategory struct {
 // helpCategories helps us by grouping commands
 var helpCategories = []helpCategory{{
 	Label:       "Basics",
-	Description: "basic management",
-	Commands:    []string{"start", "stop", "autostart"},
+	Description: "basic service management",
+	Commands:    []string{"autostart", "run", "start", "stop"},
+}, {
+	Label:       "Configuration",
+	Description: "manage configuration",
+	Commands:    []string{"add", "plan"},
 }, {
 	Label:       "Changes",
-	Description: "basic management",
-	Commands:    []string{"history", "tasks", "abort"},
+	Description: "view changes, tasks, and warnings",
+	Commands:    []string{"changes", "okay", "tasks", "warnings"},
 }, {
-	Label:       "...more",
-	Description: "slightly more advanced management",
-	Commands:    []string{"help"},
+	Label:       "Other",
+	Description: "miscellaneous commands",
+	Commands:    []string{"help", "version"},
 }}
 
 var (
@@ -196,7 +200,6 @@ func printHelpHeader() {
 	fmt.Fprintln(Stdout, pebbleUsage)
 	fmt.Fprintln(Stdout)
 	fmt.Fprintln(Stdout, pebbleHelpCategoriesIntro)
-	fmt.Fprintln(Stdout)
 }
 
 func printHelpAllFooter() {
@@ -212,6 +215,7 @@ func printHelpFooter() {
 // this is called when the Execute returns a flags.Error with ErrCommandRequired
 func printShortHelp() {
 	printHelpHeader()
+	fmt.Fprintln(Stdout)
 	maxLen := 0
 	for _, categ := range helpCategories {
 		if l := utf8.RuneCountInString(categ.Label); l > maxLen {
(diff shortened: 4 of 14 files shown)


README.md
```diff
@@ -128,11 +128,12 @@ Here are some of the things coming soon:
 
   - [x] Support `$PEBBLE_SOCKET` and default `$PEBBLE` to `/var/lib/pebble/default`
   - [x] Define and enforce convention for layer names
-  - [ ] Dynamic layer support over the API
+  - [x] Dynamic layer support over the API
+  - [x] Configuration retrieval commands to investigate current settings
```
benhoyt, line 132: Note: dynamic layer support and the configuration/plan API was added in a previous PR.

cmd/pebble/cmd_help.go
```diff
@@ -167,16 +167,20 @@ type helpCategory struct {
 // helpCategories helps us by grouping commands
 var helpCategories = []helpCategory{{
 	Label:       "Basics",
-	Description: "basic management",
-	Commands:    []string{"start", "stop", "autostart"},
+	Description: "basic service management",
+	Commands:    []string{"autostart", "run", "services", "start", "stop"},
```
benhoyt, line 171: Some of this CLI help stuff had gotten out of date.

internal/daemon/api_services.go
```diff
@@ -22,10 +22,36 @@ import (
...
-	return statusBadRequest("not implemented")
+	names := strutil.CommaSeparatedList(r.URL.Query().Get("names"))
+
+	servmgr := c.d.overlord.ServiceManager()
+	services, err := servmgr.Services(names)
+	if err != nil {
+		return statusInternalError(err.Error())
+	}
+
+	var infos []serviceInfo
```
hpidcock, line 44: probably could be `infos := make([]serviceInfo, 0, len(services))` but that's a bit of premature optimization.

internal/overlord/servstate/manager.go
```diff
@@ -185,22 +186,62 @@ func (m *ServiceManager) Ensure() error {
...
+	defer releasePlan()
+
+	requested := make(map[string]bool)
+	for _, name := range names {
+		requested[name] = true
+	}
+
+	var services []*ServiceInfo
 	for name, service := range m.plan.Services {
+		if len(names) > 0 && !requested[name] {
```
hpidcock, line 220: This is convoluted logic to express pretty important functionality for this API method.

I'd prefer:

```
onlyMatchRequested := len(names) > 0
for name, service := range m.plan.Services {
    if onlyMatchRequested && !requested[name] {
        continue
    }
```

hpidcock (approved):
Yeah looks good, thanks.

.gitignore
```diff
@@ -5,3 +5,4 @@ cmd/VERSION
 *.swp
 *.swo
 .idea
+/pebble
```
niemeyer, line 8: Curious.. what's this about?

README.md
```diff
@@ -128,11 +128,12 @@ Here are some of the things coming soon:
...
+  - [x] Status command that displays active services and their current status
+  - [ ] General system modification commands (writing configuration files, etc)
+  - [ ] Improve signal handling, e.g., sending SIGHUP to a service
   - [ ] Terminate all services before exiting run command
```
niemeyer, line 136: We should do this one soon and before we tackle the custom sending of signals, since it affects every single user of Pebble.

client/services.go
```diff
@@ -58,3 +60,32 @@ func (client *Client) doMultiServiceAction(actionName string, services []string)
...
+	// Names is the list of service names to query for. If slice is nil or
+	// empty, fetch information for all services.
+	Names []string
+}
+
+// ServiceInfo holds status information for a single service.
+type ServiceInfo struct {
+	Name    string `json:"name"`    // service name
+	Default string `json:"default"` // default state: "start" or "stop"
+	Status  string `json:"status"`  // "active", "inactive", or "error"
```
niemeyer, line 75: Can we keep up with the "Current" vs. "Default" terminology?  It feels a bit more inherently clear.

I also wonder if we should have some consts for the known values these fields may have, and then make them typed (still as strings, as that's easier to work with).

client/services.go
```diff
@@ -58,3 +60,32 @@ func (client *Client) doMultiServiceAction(actionName string, services []string)
...
+	Message string `json:"message"` // message (e.g., error) or empty string
```
niemeyer, line 76: What kind of messages would we have here?  In particular, when will we display something here that is not an error, and when it is an error how do we tell what the error is without listing the entire log?  The concern is two fold: the listing command should remain a listing command. Even a short message, with 50 chars, will turn the columns into a mess, while often not giving enough context for the error. So we need to find out how we both make it easily introspectable, while providing a good experience for people navigating through it.

As usual in those cases, we should probably do less before we know what we want. Happy to discuss this further elsewhere for more bandwidth.

cmd/pebble/cmd_help.go
niemeyer: We need to either remove "run" from the list, or remove "service" from the description. If we keep it in the list, we need to reorder it first as well, as that's how to get Pebble running at all.

cmd/pebble/cmd_help.go
```diff
@@ -167,16 +167,20 @@ type helpCategory struct {
...
+}, {
+	Label:       "Configuration",
+	Description: "manage configuration",
+	Commands:    []string{"add", "plan"},
 }, {
 	Label:       "Changes",
-	Description: "basic management",
-	Commands:    []string{"history", "tasks", "abort"},
+	Description: "view changes, tasks, and warnings",
+	Commands:    []string{"changes", "okay", "tasks", "warnings"},
```
niemeyer, line 179: Warnings (warnings, okay) are global concepts, and not just about changes/tasks. It doesn't feel like a coherent group. The order is also strange as it mixes the two ideas (changes, warnings, changes, warnings), but I think we should separate those two for the groups to make more sense together.

cmd/pebble/cmd_run.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/canonical/pebble/internal/systemd"
 )
 
-var shortRunHelp = "Run the pebble environment"
+var shortRunHelp = "Run the pebble server"
```
niemeyer, line 34: The reason it was worded as "pebble environment" is to give a feeling that you can have many of those. I was trying to avoid the somewhat usual feeling that there's that one server that you run for your whole machine, and instead give it a more user-level oriented taste: "You can have N environments, and this command runs one of them.

Does that make sense? Is there a way we can improve on the original wording while preserving that feeling? Maybe "plan" something something?

cmd/pebble/cmd_services.go
```diff
@@ -0,0 +1,84 @@
...
+		} else {
+			fmt.Fprintln(Stderr, "No matching services")
+		}
+		return nil
+	}
+
+	w := tabWriter()
+	defer w.Flush()
+
+	fmt.Fprintln(w, "Service\tStartup\tCurrent\tNotes")
```
niemeyer, line 62: It would be nice to unify terminology for this one. Do we keep "default" or "startup"?  Default may be nicer as it avoids the cacophony of "startup: start", and it is also slightly more open ended (we can have more values for "default", probably). On the other hand, "startup" is more well defined, probably. What do you think?

cmd/pebble/cmd_services.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for _, svc := range services {
+		startup := svc.Default
+		switch svc.Default {
+		case "start":
+			startup = "enabled"
```
niemeyer, line 68: In either case, we'd benefit from one term being used everywhere.

internal/daemon/api_services.go
```diff
@@ -46,7 +72,7 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 		if len(payload.Services) != 0 {
 			return statusBadRequest("%s accepts no service names", payload.Action)
 		}
-		services, err = servmgr.DefaultServices()
+		services, err = servmgr.DefaultServiceNames()
```
niemeyer, line 75: :ok_hand:

internal/overlord/servstate/manager.go
```diff
@@ -185,22 +186,62 @@ func (m *ServiceManager) Ensure() error {
...
+	Status  ServiceStatus
+	Message string
+}
+
+type ServiceStatus string
+
+const (
+	ServiceActive   ServiceStatus = "active"
+	ServiceInactive ServiceStatus = "inactive"
+	ServiceError    ServiceStatus = "error"
```
niemeyer, line 201: Nice, that's what I had in mind in one of the first comments above.

internal/overlord/servstate/manager.go
niemeyer: Agreed, except I'd invert the logic and say simply "all := true/false" to make it more obvious.

niemeyer (changes requested):
Thanks, Ben!  A few nitpicks, but nothing fundamental here. Nice implementation overall.

.gitignore
benhoyt: It's because I do `go build ./cmd/pebble` to build locally, and that produces a `pebble` binary in the current directory. It was odd to me that you need to prefix with `/` (looks like an absolute path), but that's `.gitignore` syntax for "only the file `pebble` in the root repository directory, not `pebble/foo` or `bar/pebble`".

README.md
benhoyt: It's unclear to me that the workload processes *should* stop if the service manager process (Pebble) dies. Isn't it good that they keep going? Or is what you're suggesting just good cleanup behavior and/or how other service managers work?

cmd/pebble/cmd_help.go
benhoyt: I removed "service" from the description and moved "run" to the first item (I had organized them alphabetically, but that's probably not the most useful way in a short list like this.

internal/daemon/api_services.go
benhoyt: Agreed it's probably premature optimization, but doesn't hurt - done.

internal/overlord/servstate/manager.go
benhoyt: "all" actually led to more complex/inverted boolean logic (`!all && !requested[name]` or `!(all || requested[name])`, so I kept the logic the same but called it `matchNames` so it reads better but isn't too long.

client/services.go
benhoyt: Yeah, I wondered that too. I think Startup and Current like the CLI is most clear, so I've renamed to that across the board (manager and client). And I've added a type and consts in the Go client too (I had them only on the manager end).

My only request is that, for consistency, we change the YAML layer config to use `startup: enabled/disabled` instead of `default: start/stop`. (start/stop seem more like actions or commands than the startup state in any case). @niemeyer  Thoughts?

client/services.go
benhoyt: Yeah, I think you're right. I've removed for now as I wasn't currently populating this field anyway. We can always add it later when we've defined it / got more experience with Pebble.

cmd/pebble/cmd_help.go
benhoyt: Sounds good - I've moved to the "Other" category (same as `snap help` does).

cmd/pebble/cmd_run.go
benhoyt: Fair enough. We could say "a pebble server" rather than "the", but that doesn't sound quite right either. Reverted to "environment" for now.

cmd/pebble/cmd_services.go
benhoyt: As above, unified to Startup: enabled/disabled and Current: active/inactive everywhere.

cmd/pebble/cmd_services.go
benhoyt: See above.

.gitignore
niemeyer: I see.. I generally have cmd/foo/foo in .gitignore, since that's what we get if we run "go build" there. Otherwise it'll drop wherever $CWD happens to be.

Either way, if that helps you somehow, that's certainly okay.

README.md
niemeyer: In the best scenario, we get a SIGTERM in k8s, and politely terminate all the services we're running. In the worst case, the container is abruptly terminated. In both of these, though, the underlying services are going away.

Another perspective: right now we have no way to keep track of services unless we're running. If Pebble gets restarted and we run "pebble services", what do we get?

Yet another perspective: ideally we'd run services in a way that if Pebble dies abruptly, services die as well. Otherwise we'll lose control of these services for the reason above. Not an issue for now since k8s will collect for us regardless, but once Pebble is running standalone more often we should fix that too.

client/services.go
niemeyer: Sounds sane. Thanks for making it all consistent.

niemeyer (approved):
Thank you!

PR #18, opened 2021-03-21
Pass service description's environment variables to child process

### LINKED ISSUE #17: Pebble not passing configured environment variables to services
When testing out the config of a new Pebble-based charm, it seems to not be passing env variables to child processes.

A simple test (these files are uploaded to this issue - including the built pebble binary I'm using) :

Dockerfile:

```Dockerfile
FROM ubuntu:focal

ENV PEBBLE /config

COPY pebble /usr/bin/pebble
COPY pebble-test /usr/bin/pebble-test
COPY 000-test-layer.yaml /config/layers/000-test-layer.yaml

CMD ["/usr/bin/pebble", "run"]
```

pebble-test:

```bash
#!/bin/bash
if [[ -z "${MYVAR:-}" ]]; then
  echo "Variable MYVAR is unset" > /tmp/out.log
else
  echo "The value of MYVAR is: ${MYVAR}" > /tmp/out.log
fi
```

000-test-layer.yaml

```yaml
---
summary: Test Layer

services:
  pebbletest:
    override: replace
    summary: Simple test layer
    command: /usr/bin/pebble-test
    default: start
    environment:
      - MYVAR: my-very-important-value
```

The following commands then illustrate the issue:

```bash
$ docker build -t jnsgruk/pebble-test:latest .
$ docker run -d --name pbl --rm -it jnsgruk/pebble-test:latest
$ docker exec -it pbl cat /tmp/out.log
Variable MYVAR is unset

# Cleanup: docker stop pbl
```

Thanks :)

[pebble-env-issue-test.zip](https://github.com/canonical/pebble/files/6170049/pebble-env-issue-test.zip)

benhoyt:
Pretty self-explanatory. Fixes #17

### DIFF
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -2,6 +2,7 @@ package servstate
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"syscall"
 	"time"
@@ -82,6 +83,12 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 	}
 	cmd := exec.Command(args[0], args[1:]...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
+	// Pass service description's environment variables to child process
+	cmd.Env = os.Environ()
+	for _, v := range service.Environment {
+		cmd.Env = append(cmd.Env, v.Name+"="+v.Value)
+	}
 
 	buffer := strutil.NewLimitedBuffer(160, 10*1024)
 	cmd.Stdout = buffer
--- a/internal/overlord/servstate/manager_test.go
+++ b/internal/overlord/servstate/manager_test.go
@@ -505,3 +505,57 @@ func (s *S) TestServices(c *C) {
 		{Name: "test4", Current: servstate.StatusInactive, Startup: servstate.StartupDisabled},
 	})
 }
+
+var planLayerEnv = `
+services:
+    envtest:
+        override: replace
+        command: /bin/sh -c "env | grep PEBBLE_ENV_TEST | sort > %s; sleep 300"
+        environment:
+            - PEBBLE_ENV_TEST_1: foo
+            - PEBBLE_ENV_TEST_2: bar bazz
+`
+
+func (s *S) TestEnvironment(c *C) {
+	// Setup new state and add "envtest" layer
+	st := state.New(nil)
+	dir := c.MkDir()
+	runner := state.NewTaskRunner(st)
+	manager, err := servstate.NewManager(st, runner, dir)
+	c.Assert(err, IsNil)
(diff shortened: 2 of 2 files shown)


hpidcock (approved):
LGTM thank-you

hpidcock:
@benhoyt @niemeyer would also be good to do env var expansion here so someone could do

```
environment:
   - PATH: $PATH:/my/bin/dir
```

benhoyt:
> @benhoyt @niemeyer would also be good to do env var expansion here so someone could do `PATH: $PATH:/my/bin/dir`

Yeah, I agree, though I think we can leave that for a subsequent PR (this is strictly the bug fix).

niemeyer:
@hpidcock @benhoyt Agreed, on both counts. We should do it on a follow up, but we shouldn't wait too much as that may be seen as breaking compatibility. Note that this is the key reason why we have that as a list and note as a map: ordering matters if we want to expand.

internal/overlord/servstate/handlers.go
```diff
@@ -83,6 +84,12 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 	cmd := exec.Command(args[0], args[1:]...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 
+	// Pass service description's environment variables to child process
+	cmd.Env = os.Environ()
+	for _, v := range service.Environment {
+		cmd.Env = append(cmd.Env, v.Name+"="+v.Value)
+	}
```
niemeyer, line 91: Thanks for fixing this, Ben.  We indeed need a good test for this sort of change.

niemeyer (changes requested):

internal/overlord/servstate/handlers.go
niemeyer: Ah, and it's also worth checking (and testing) what happens if there's a variable in the environment process which needs to be replaced by something explicit on the service.

benhoyt: Added a test for this: it tests brand new env vars, env vars only set in the parent, and env vars in the parent but overridden in the layer config. I'll do any `$PATH` type of interpolation in a subsequent commit.

internal/overlord/servstate/manager_test.go
```diff
@@ -505,3 +505,60 @@ func (s *S) TestServices(c *C) {
...
 }
+
+var planLayerEnv = `
+services:
+    envtest:
+        override: replace
+        command: /bin/sh -c "env | grep PEBBLE_ENV_TEST | sort > %s; sleep 300"
+        environment:
+            - PEBBLE_ENV_TEST_1: foo
+            - PEBBLE_ENV_TEST_2: bar bazz
```
niemeyer, line 516: Nice test.

niemeyer (approved):
Thanks!

PR #19, opened 2021-03-21
Fix marshaling of StringVariable to match unmarshaling

benhoyt:
When I added "pebble plan" to fetch the plan, I didn't add a custom
MarshalYAML function on plan.StringVariable, so it marshaled like so:

```
environment:
    - name: FOO
      value: bar
```

This changes it to matching the unmarshaling, like so:

```
environment:
    - FOO: bar
```

Corresponding Python Operator Framework issue:
https://github.com/canonical/operator/issues/486

### DIFF
--- a/internal/plan/plan_test.go
+++ b/internal/plan/plan_test.go
@@ -23,9 +23,10 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/canonical/pebble/internal/plan"
-
 	. "gopkg.in/check.v1"
+	"gopkg.in/yaml.v3"
+
+	"github.com/canonical/pebble/internal/plan"
 )
 
 // TODOs:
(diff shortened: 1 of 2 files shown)


hpidcock (approved):
Thanks, LGTM

internal/plan/plan_test.go
```diff
@@ -478,3 +479,38 @@ func (s *S) TestReadDirDupNames(c *C) {
...
+            - var1: val1
+            - var0: val0
+            - var2: val2
+    srv2:
+        override: replace
+        command: srv2cmd
+    srv3:
+        override: replace
+        command: srv3cmd
+`[1:]
```
niemeyer, line 510: Can we please use the same function we use in the setup (now plan) package tests to indent this YAML properly and not need the [1:] at the end?  It's much nicer to read the code when the strings are aligned properly with the code instead of de-dented out of its context.

niemeyer (approved):
LGTM. Just one suggestion:

internal/plan/plan_test.go
benhoyt: Fixed.

PR #20, opened 2021-03-23
Change default: start/stop to startup: enabled/disabled

benhoyt:
I missed this in the previous PR when adding "pebble services": https://github.com/canonical/pebble/pull/14

This is a breaking change, but given the rc/early status of this, we are okay with that. (There are only a few internal charms that use Pebble right now.)


niemeyer (approved):
Thanks. Reading this makes me feel like it's going in the right direction.

PR #21, opened 2021-03-25
README updates to reflect default being renamed to startup

mthaddon:
Some README updates to reflect `default` being renamed to `startup`.


benhoyt:
Oooh, thanks @mthaddon -- good catch!

benhoyt (approved):

PR #25, opened 2021-04-01
Files API

benhoyt:
This implements the Pebble file I/O API as per the [spec](https://docs.google.com/document/d/1NPPZZB_qCwtTru-YsKkRSAb6mXJzBiE0fG3GTZnx4fM/edit#heading=h.tqx73o2f3f0q).

The code is done, and tests for all features (including error handling) are in place. The one thing I'm not yet testing is setting of user/group -- will try to add tests for that tomorrow.

This PR doesn't incldue the Pebble Go client or Pebble CLI commands in this PR. To keep things moving, I plan to do that in a subsequent PR.

Note that per discussion with Gustavo, it'd be nice if, for consistency, all endpoints could optionally be sent as `multipart/form-data`. However, this is not necessary right now, and can be done later if we want.

### DIFF
--- a/internal/daemon/api_files.go
+++ b/internal/daemon/api_files.go
@@ -26,7 +26,7 @@ import (
 	"net/textproto"
 	"os"
 	"os/user"
-	"path"
+	pathpkg "path"
 	"strconv"
 	"time"
 
@@ -50,8 +50,8 @@ func v1GetFiles(_ *Command, req *http.Request, _ *userState) Response {
 		}
 		return readFilesResponse{paths: paths}
 	case "list":
-		p := query.Get("path")
-		if p == "" {
+		path := query.Get("path")
+		if path == "" {
 			return statusBadRequest("must specify path")
 		}
 		pattern := query.Get("pattern")
@@ -59,7 +59,7 @@ func v1GetFiles(_ *Command, req *http.Request, _ *userState) Response {
 		if directory != "true" && directory != "false" && directory != "" {
 			return statusBadRequest(`directory parameter must be "true" or "false"`)
 		}
-		return listFilesResponse(p, pattern, directory == "true")
+		return listFilesResponse(path, pattern, directory == "true")
 	default:
 		return statusBadRequest("invalid action %q", action)
 	}
@@ -88,13 +88,13 @@ func (r readFilesResponse) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	// Read each file's contents to multipart response.
 	result := make([]fileResult, len(r.paths))
 	status := http.StatusOK
-	for i, p := range r.paths {
-		err := readFile(p, mw)
+	for i, path := range r.paths {
+		err := readFile(path, mw)
 		if err != nil {
 			status = http.StatusBadRequest
 		}
 		result[i] = fileResult{
-			Path:  p,
+			Path:  path,
 			Error: fileErrorToResult(err),
 		}
 	}
@@ -131,24 +131,24 @@ func (r readFilesResponse) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
-func nonAbsolutePathError(p string) error {
-	return fmt.Errorf("paths must be absolute, got %q", p)
+func nonAbsolutePathError(path string) error {
+	return fmt.Errorf("paths must be absolute, got %q", path)
 }
 
-func readFile(p string, mw *multipart.Writer) error {
-	if !path.IsAbs(p) {
-		return nonAbsolutePathError(p)
+func readFile(path string, mw *multipart.Writer) error {
+	if !pathpkg.IsAbs(path) {
+		return nonAbsolutePathError(path)
 	}
-	if osutil.IsDir(p) {
-		return fmt.Errorf("cannot read a directory: %q", p)
+	if osutil.IsDir(path) {
+		return fmt.Errorf("cannot read a directory: %q", path)
 	}
-	f, err := os.Open(p)
+	f, err := os.Open(path)
 	if err != nil {
 		return err
 	}
 	defer f.Close()
 
-	fw, err := mw.CreateFormFile("files", p)
+	fw, err := mw.CreateFormFile("files", path)
 	if err != nil {
 		return err
 	}
@@ -250,7 +250,7 @@ func fileInfoToResult(fullPath string, info os.FileInfo) fileInfoResult {
 	}
 	result := fileInfoResult{
 		Path:         fullPath,
-		Name:         path.Base(fullPath),
+		Name:         pathpkg.Base(fullPath),
 		Type:         fileModeToType(mode),
 		Size:         psize,
 		Permissions:  fmt.Sprintf("%03o", mode.Perm()),
@@ -259,19 +259,19 @@ func fileInfoToResult(fullPath string, info os.FileInfo) fileInfoResult {
 	return result
 }
 
-func listFilesResponse(p, pattern string, directoryItself bool) Response {
-	if !path.IsAbs(p) {
-		return statusBadRequest("path must be absolute, got %q", p)
+func listFilesResponse(path, pattern string, directoryItself bool) Response {
+	if !pathpkg.IsAbs(path) {
+		return statusBadRequest("path must be absolute, got %q", path)
 	}
-	result, err := listFiles(p, pattern, directoryItself)
+	result, err := listFiles(path, pattern, directoryItself)
 	if err != nil {
 		return listErrorResponse(err)
 	}
 	return SyncResponse(result)
 }
 
-func listFiles(p, pattern string, directoryItself bool) ([]fileInfoResult, error) {
-	info, err := os.Stat(p)
+func listFiles(path, pattern string, directoryItself bool) ([]fileInfoResult, error) {
+	info, err := os.Stat(path)
 	if err != nil {
 		return nil, err
 	}
@@ -281,14 +281,14 @@ func listFiles(p, pattern string, directoryItself bool) ([]fileInfoResult, error
 	if !info.IsDir() || directoryItself {
 		// Info about a single file (or directory entry itself).
 		infos = []os.FileInfo{info}
-		dir = path.Dir(p)
+		dir = pathpkg.Dir(path)
 	} else {
 		// List an entire directory.
-		infos, err = ioutil.ReadDir(p)
+		infos, err = ioutil.ReadDir(path)
 		if err != nil {
 			return nil, err
 		}
-		dir = p
+		dir = path
 	}
 
 	var result []fileInfoResult
@@ -296,10 +296,10 @@ func listFiles(p, pattern string, directoryItself bool) ([]fileInfoResult, error
 		name := info.Name()
 		matched := true
 		if pattern != "" {
-			matched, err = path.Match(pattern, name)
+			matched, err = pathpkg.Match(pattern, name)
 		}
 		if matched {
-			fullPath := path.Join(dir, name)
+			fullPath := pathpkg.Join(dir, name)
 			result = append(result, fileInfoToResult(fullPath, info))
 		}
 	}
@@ -400,12 +400,12 @@ func writeFiles(body io.Reader, boundary string) Response {
 		if part.FormName() != "files" {
 			return statusBadRequest(`field name must be "files", got %q`, part.FormName())
 		}
-		p := part.FileName()
-		info, ok := infos[p]
+		path := part.FileName()
+		info, ok := infos[path]
 		if !ok {
-			return statusBadRequest("no metadata for path %q", p)
+			return statusBadRequest("no metadata for path %q", path)
 		}
-		errors[p] = writeFile(info, part)
+		errors[path] = writeFile(info, part)
 		part.Close()
 	}
 
@@ -430,13 +430,13 @@ func writeFiles(body io.Reader, boundary string) Response {
 }
 
 func writeFile(item writeFilesItem, source io.Reader) error {
-	if !path.IsAbs(item.Path) {
+	if !pathpkg.IsAbs(item.Path) {
 		return nonAbsolutePathError(item.Path)
 	}
 
 	// Create parent directory if needed
 	if item.MakeDirs {
-		err := os.MkdirAll(path.Dir(item.Path), 0o755)
+		err := os.MkdirAll(pathpkg.Dir(item.Path), 0o755)
 		if err != nil {
 			return fmt.Errorf("cannot create directory: %w", err)
 		}
@@ -502,7 +502,7 @@ func makeDirs(dirs []makeDirsItem) Response {
 }
 
 func makeDir(dir makeDirsItem) error {
-	if !path.IsAbs(dir.Path) {
+	if !pathpkg.IsAbs(dir.Path) {
 		return nonAbsolutePathError(dir.Path)
 	}
 	perm, err := parsePermissions(dir.Permissions, 0o755)
@@ -572,25 +572,25 @@ type removePathsItem struct {
 func removePaths(paths []removePathsItem) Response {
 	result := make([]fileResult, len(paths))
 	var firstErr error
-	for i, p := range paths {
-		err := removePath(p.Path, p.Recursive)
+	for i, path := range paths {
+		err := removePath(path.Path, path.Recursive)
 		if err != nil && firstErr == nil {
 			firstErr = err
 		}
 		result[i] = fileResult{
-			Path:  p.Path,
+			Path:  path.Path,
 			Error: fileErrorToResult(err),
 		}
 	}
 	return syncResponseWithError(result, firstErr)
 }
 
-func removePath(p string, recursive bool) error {
-	if !path.IsAbs(p) {
-		return nonAbsolutePathError(p)
+func removePath(path string, recursive bool) error {
+	if !pathpkg.IsAbs(path) {
+		return nonAbsolutePathError(path)
 	}
 	if recursive {
-		return os.RemoveAll(p)
+		return os.RemoveAll(path)
 	}
-	return os.Remove(p)
+	return os.Remove(path)
 }
--- a/internal/daemon/api_files.go
+++ b/internal/daemon/api_files.go
@@ -140,7 +140,7 @@ func (r readFilesResponse) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		if file.Error != nil {
 			continue
 		}
-		fw, err := mw.CreateFormFile("p:"+file.Path, path.Base(file.Path))
+		fw, err := mw.CreateFormFile("path:"+file.Path, path.Base(file.Path))
 		if err != nil {
 			http.Error(w, "\n"+err.Error(), http.StatusInternalServerError)
 			return
--- a/internal/daemon/api_files.go
+++ b/internal/daemon/api_files.go
@@ -1,0 +1,121 @@
+// Copyright (c) 2021 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package daemon
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"path"
+	"path/filepath"
+	"time"
+)
+
+func v1GetFiles(c *Command, r *http.Request, _ *userState) Response {
+	query := r.URL.Query()
+	action := query.Get("action")
+	switch action {
+	case "read":
+		paths := query["path"]
+		if len(paths) == 0 {
+			return statusBadRequest("must specify one or more paths")
+		}
+		return getFiles(paths)
+	case "list":
+		return listFiles(query.Get("pattern"))
+	default:
+		return statusBadRequest("invalid action %q", action)
+	}
+}
+
+func getFiles(paths []string) Response {
+	return nil
+}
+
+func errorResponse(kind errorKind, message string) Response {
+	status := 400
+	switch kind {
+	case errorKindNotFound:
+		status = 404
+	case errorKindPermissionDenied:
+		status = 403
+	}
+	return &resp{
+		Type: ResponseTypeError,
+		Result: &errorResult{
+			Message: message,
+			Kind:    kind,
+		},
+		Status: status,
+	}
+}
+
+// TODO: should we include "." and ".."?
+func listFiles(pattern string) Response {
+	st, err := os.Stat(pattern)
+	if err != nil {
+		if errors.Is(err, os.ErrPermission) {
+			return errorResponse(errorKindPermissionDenied, err.Error())
+		}
+		if !errors.Is(err, os.ErrNotExist) {
+			return statusBadRequest("fetching path information: %v", err)
+		}
+	} else if st.IsDir() {
+		pattern = path.Join(pattern, "*")
+	}
+	matches, err := filepath.Glob(pattern)
+	if err != nil {
+		return statusBadRequest("listing files: %v", err)
+	}
+	type listFilesResult struct {
+		Path         string `json:"path"`
+		Name         string `json:"name"`
+		Type         string `json:"type"`
+		Size         *int64 `json:"size,omitempty"`
+		Permissions  string `json:"permissions"`
+		LastModified string `json:"last-modified"`
+	}
+	var result []listFilesResult
+	for _, match := range matches {
+		st, err := os.Lstat(match)
+		if err != nil {
+			return statusBadRequest("fetching file information: %v", err)
+		}
+		fileType := "file"
+		var psize *int64
+		if st.IsDir() {
+			fileType = "directory"
+			psize = nil
+		} else {
+			size := st.Size()
+			psize = &size
+		}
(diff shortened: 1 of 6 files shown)


internal/daemon/api_files.go
```diff
@@ -0,0 +1,595 @@
...
+	case "read":
+		paths := query["path"]
+		if len(paths) == 0 {
+			return statusBadRequest("must specify one or more paths")
+		}
+		return readFilesResponse{paths: paths}
+	case "list":
+		pattern := query.Get("pattern")
+		if pattern == "" {
+			return statusBadRequest("must specify pattern")
```
wallyworld, line 54: I guess we're not defaulting to returning everything if pattern is ""?

benhoyt: Well, you have to specify the directory, e.g., `pattern=/etc`

internal/daemon/api_files.go
```diff
@@ -0,0 +1,594 @@
...
+
+func removePath(p string, recursive bool) error {
+	if !path.IsAbs(p) {
+		return fmt.Errorf("paths must be absolute (%q is not)", p)
+	}
+	if recursive {
+		return os.RemoveAll(p)
+	} else {
+		return os.Remove(p)
+	}
```
wallyworld, line 617: Don't really need the `else`

internal/daemon/api_files.go
```diff
@@ -0,0 +1,594 @@
...
+			Path:  dir.Path,
+			Error: fileErrorToResult(err),
+		}
+	}
+	return syncResponseWithError(result, firstErr)
+}
+
+func makeDir(dir makeDirsItem) error {
+	if !path.IsAbs(dir.Path) {
+		return fmt.Errorf("paths must be absolute (%q is not)", dir.Path)
```
wallyworld, line 539: not sure if we want a msg template const or helper func for this common error

internal/daemon/api_files.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+	if dir.MakeParents {
+		err = os.MkdirAll(dir.Path, perm)
+	} else {
+		err = os.Mkdir(dir.Path, perm)
+	}
+	if err != nil {
+		return err
+	}
+	err = updateUserGroup(dir.Path, dir.UserID, dir.User, dir.GroupID, dir.Group)
```
wallyworld, line 553: maybe a comment as why why we need to do this, umask and all that

internal/daemon/api_files.go
```diff
@@ -0,0 +1,594 @@
...
+	if err != nil {
+		_ = f.Close()
+		return fmt.Errorf("error writing file: %w", err)
+	}
+	err = f.Close()
+	if err != nil {
+		return fmt.Errorf("error closing file: %w", err)
+	}
+
+	// Update user and group if necessary.
```
wallyworld, line 475: See other comment - maybe explain why

wallyworld (commented):
A few small suggestions.
Not sure if api_files.go could have a few functions grouped together with block comments like "These are for writing files" , "These are for listing files" etc to segment things a little

internal/daemon/api_files.go
```diff
@@ -0,0 +1,594 @@
...
+		if len(paths) == 0 {
+			return statusBadRequest("must specify one or more paths")
+		}
+		return readFilesResponse{paths: paths}
+	case "list":
+		pattern := query.Get("pattern")
+		if pattern == "" {
+			return statusBadRequest("must specify pattern")
+		}
+		directoryItself := query.Get("directory") == "true"
```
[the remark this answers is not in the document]
benhoyt: Good call, done.

internal/daemon/api_files.go
benhoyt: Removed.

internal/daemon/api_files.go
benhoyt: Good call, done.

internal/daemon/api_files.go
benhoyt: The umask issue is actually to do with permissions, not user and group stuff. But I actually decided against fiddling with umask or manually doing `os.Chmod` (the latter didn't seem to work anyway): my reasoning is that the process umask should probably be respected, not overridden. So at least for now I've just made the tests work with standard umasks (022 and 002).

internal/daemon/api_files.go
benhoyt: See above.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	contentType := req.Header.Get("Content-Type")
+	mediaType, params, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return statusBadRequest("invalid Content-Type %q", contentType)
+	}
+
+	switch mediaType {
+	case "multipart/form-data":
+		boundary := params["boundary"]
+		if boundary == "" {
```
niemeyer, line 336: We should probably do a basic check here as well to make sure this is not too short, otherwise a cooked result might end up creating a disproportional amount of work on the server end.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	errors := make(map[string]error)
+	for i := 0; ; i++ {
+		part, err = mr.NextPart()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return statusBadRequest("cannot read file part %d: %v", i, err)
+		}
+		if !strings.HasPrefix(part.FormName(), "file:") {
```
niemeyer, line 417: Is there an established convention for such prefixes?  Reading this I was wondering between "file:" or "file=", and whether there is already a most used pattern.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+			return statusBadRequest(`field name must be in format "file:/path", not %q`, part.FormName())
+		}
+		p := part.FormName()[len("file:"):]
```
niemeyer, line 420: strings.TrimPrefix comes to mind. Sometimes I do something like
```
formName := part.FormName()
actualPath := strings.TrimPrefix(formName, "file:")
if actualPath == formName {
     ... error
}
```
So we don't repeat the prefix, and do both the check and the stripping at once. This is nitpicking and not important, though.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+		}
+	}
+
+	// Create file and write contents to temporary file.
+	perm, err := parsePermissions(item.Permissions, 0o644)
+	if err != nil {
+		return err
+	}
+	tempPath := item.Path + ".pebble-temp"
+	f, err := os.OpenFile(tempPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, perm)
```
niemeyer, line 468: O_WRONLY?  Not sure if it matters...

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	if err != nil {
+		return fmt.Errorf("creating file: %w", err)
```
niemeyer, line 470: Depending on where this is presented and what happens in "err", that message might be unclear as it sounds like an action is taking place right now.  The traditional `"cannot create file: %w"`, would be safer. Same for the cases below.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	}
+	_, err = io.Copy(f, source)
+	if err != nil {
+		_ = f.Close()
+		_ = os.Remove(tempPath)
```
niemeyer, line 475: Instead of removing at every error, this could be deferred a single time right after creation inside a function that checks if err != nil.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+			err = fmt.Errorf("no file content for path %q", file.Path)
+		}
+		if err != nil && firstErr == nil {
+			firstErr = err
+		}
+		result[i] = fileResult{
+			Path:  file.Path,
+			Error: fileErrorToResult(err),
+		}
+	}
```
niemeyer, line 445: Should we sync the filesystem here?  Please have a look at the atomic write commit we have somewhere in osutil, as this is surprisingly boring.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+		err = os.MkdirAll(dir.Path, perm)
+	} else {
+		err = os.Mkdir(dir.Path, perm)
+	}
+	if err != nil {
+		return err
+	}
+	err = updateUserAndGroup(dir.Path, dir.UserID, dir.User, dir.GroupID, dir.Group)
+	if err != nil {
+		return fmt.Errorf("setting user and group: %w", err)
```
niemeyer, line 556: Same as earlier note. Not sure if I missed one of these so would be worth double checking.

niemeyer (changes requested):
Here is a first pass. Overall looks great, thanks Ben!

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+		}
+		return readFilesResponse{paths: paths}
+	case "list":
+		pattern := query.Get("pattern")
+		if pattern == "" {
+			return statusBadRequest("must specify pattern")
+		}
+		directory := query.Get("directory")
+		if directory != "true" && directory != "false" && directory != "" {
+			return statusBadRequest(`directory parameter must be "true" or "false"`)
```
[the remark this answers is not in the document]
benhoyt: It's equivalent to `ls -d`, which shows the directory itself rather than what's inside the directory. Compare:

```
$ ls -al -d /etc
drwxr-xr-x 148 root root 12288 Apr  9 16:48 /etc
$ ls -al /etc | head -n5
total 1268
drwxr-xr-x 148 root root   12288 Apr  9 16:48 .
drwxr-xr-x  21 root root    4096 Mar 17 02:59 ..
drwxr-xr-x   3 root root    4096 Aug  1  2020 acpi
-rw-r--r--   1 root root    3028 Aug  1  2020 adduser.conf
```

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	}
+	w.WriteHeader(status)
+
+	// Write first part: response metadata in JSON format.
+	mh := textproto.MIMEHeader{}
+	mh.Set("Content-Type", "application/json")
+	mh.Set("Content-Disposition", `form-data; name="response"`)
+	part, err := mw.CreatePart(mh)
+	if err != nil {
+		http.Error(w, "\n"+err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
benhoyt: Yeah, this is tricky. By the time you know you have an error, you've already written the headers and some bytes in the body. So with a streamed multipart response like this there's no good way to include all the errors up front (unless you do something nasty like read the entire files' contents into memory).

Your idea with a list of errors at the end is good ... I started implementing that, but then there were two pieces of metadata, the first one with just the paths, and the last with any additional errors. Then I realized, why not just put the entire "response" JSON metadata at the end *instead* of the start? This actually works out really nicely, and allows you to include all errors. The HTTP status is always 200 OK, and the metadata (really just paths and errors) are always in the `response` field at the end. It's simpler code and more consistent. Let me know what you think.

However, there are still a couple of cases where you have to stop early -- if I get an error outputting the next multipart part, I still return early with a broken response. I don't see a real workaround for this (it's kind of like getting an error when you're logging an error).

The "\n" prefix in the early returns are just to put the error message on a new line, so that if a human is examining the output, the error message will be sure to be on a line of its own.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+		}
+		fw, err := mw.CreateFormFile("path:"+file.Path, path.Base(file.Path))
+		if err != nil {
+			http.Error(w, "\n"+err.Error(), http.StatusInternalServerError)
+			return
+		}
+		_, err = io.Copy(fw, file.f)
+		if err != nil {
+			http.Error(w, "\n"+err.Error(), http.StatusInternalServerError)
+			return
```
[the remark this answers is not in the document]
benhoyt: I've done this, at least the for the `io.Copy`, which now goes through `osutil.AtomicWriteChown` -- if that functions returns any error, it records the error to include in the response and keeps going.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	// Write the multipart trailer (last boundary marker).
+	err = mw.Close()
+	if err != nil {
+		http.Error(w, "\n"+err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func nonAbsolutePathError(p string) error {
+	return fmt.Errorf("paths must be absolute (%q is not)", p)
```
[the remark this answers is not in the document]
benhoyt: Done.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+		return statusBadRequest("pattern must be absolute (%q is not)", pattern)
+	}
+	result, err := listFilesErr(pattern, directoryItself)
+	if err != nil {
+		return listErrorResponse(err)
+	}
+	return SyncResponse(result)
+}
+
+func listFilesErr(pattern string, directoryItself bool) ([]fileInfoResult, error) {
```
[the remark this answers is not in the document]
benhoyt: Good call, done.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+	info, err := os.Stat(pattern)
+	if errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
benhoyt: Yeah, but the [docs for `os.IsNotExist`](https://golang.org/pkg/os/#IsNotExist) say the following:

> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, os.ErrNotExist).

internal/daemon/api_files.go
```diff
@@ -0,0 +1,627 @@
...
+		dir, base := filepath.Split(pattern)
+		result, err := readDirFiltered(dir, base)
```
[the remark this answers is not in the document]
benhoyt: I believe this is the same behavior as `ls`, though it's a little harder to see because of shell filename expansion:

```
$ touch fooa
$ touch foob
$ touch 'foo?'
$ ls foo?
'foo?'   fooa   foob
$ ls 'foo?'
'foo?'
```

This is only going to be an issue if you have filenames with `*` or `?` in them, right? In the spirit of being opinionated (cf our discussion about logging the other day), I'm okay just saying, "that's a really bad idea, don't do that".

Alternatively, if you feel strongly about this, we could make the list-files endpoint *always* take a pattern, or take either a `path` or `pattern` (or even `file`, `dir`, or `pattern` to make it even more explicit). If we go this route, my vote would be to only take `pattern` to keep it simple. Then you'd have to say `GET /v1/files?pattern=/etc/*` explicitly, which isn't necessarily a bad thing.

internal/daemon/api_files.go
benhoyt: Done. I've added `const minBoundaryLength 8`.

internal/daemon/api_files.go
benhoyt: Not that I could find. There is a thing in PHP (and maybe other web libraries, I don't know) where `user[name]=benhoyt` means build a nested object that you can access as `query["user"]["name"]`, but that doesn't really apply here. Between `:` and `=`, I prefer `:` as `=` seems too much like query string parameters, which this is not. Fairly bike-sheddy though, as no one will actually see this prefix in practice.

internal/daemon/api_files.go
benhoyt: Nice. Done.

internal/daemon/api_files.go
benhoyt: Done, thanks. (Though now this uses osutil.AtomcFile, which does use `O_WRONLY`.)

internal/daemon/api_files.go
benhoyt: Good call, updated.

internal/daemon/api_files.go
benhoyt: Good call, done.

internal/daemon/api_files.go
benhoyt: Ha, nice -- osutil.AtomicFile takes care of a lot of this, including setting user and group. I've updated it.

internal/daemon/api_files.go
benhoyt: Done.

internal/daemon/api_files.go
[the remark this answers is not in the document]
niemeyer: The StatusInternalServerError there is not actually a 200, but it won't be sent as the actual HTTP status as is generally the intention of the http.Errror function, because the header has been sent already. Then, we prefix the error message with \n to dump an error message in the middle of some bulk data in the hope someone will introspect the output to see why it was actually wrong, instead of planning for an error to be reported properly to the other side. So we're abusing a function that clearly is not meant for this, and the result is not even so good.

I think we need to have a convention that makes building the server and clients straightforward, while following a nice convention that is easy to explain, and hopefully compatible with alternative clients that are not our own.

After going back and forth on this a few times, I think we should go with your suggestion of having "response" last and including both any reading errors and the digest computed on-the-fly (so we only read files once). On the write side we then store all files temporary in the local filesystem, evaluate the "response" to make sure we're happy with the entire operation, and only then commit the modifications to the filesystem.

What made me tilt towards this is the realization that today the mime parser in Go will do that itself when we attempt to parse a message that is over 10MB, which means any streaming optimizations we do are pointless, at least for the time being and without someone doing work to improve this. So we may as well just do that.

As part of reading about this, I also found the answer to the other question: the way browsers send the filename is indeed via an independent filename="..." field, with a form-data such as name="files" that is repeated for all the files present in the given form field (for our case, all of them). The header entry for each file looks like:
```
Content-Disposition: form-data; name="files"; filename="dir/name/somefile.txt"
```

From [section 4.2 of RFC7578](https://tools.ietf.org/html/rfc7578#section-4.2):

>    For form data that represents the content of a file, a name for the file SHOULD be supplied as well, by using a "filename" parameter of the Content-Disposition header field.  The file name isn't mandatory for cases where the file name isn't available or is meaningless or private; this might result, for example, when selection or drag-and- drop is used or when the form data content is streamed directly from a device.
> 
> If a "filename" parameter is supplied, the requirements of Section 2.3 of [RFC2183] for the "receiving MUA" (i.e., the receiving Mail User Agent) apply to receivers of multipart/form-data as well: do not use the file name blindly, check and possibly change to match local file system conventions if applicable, and do not use directory path information that may be present.

From [section 4.3 of RFC7578](https://tools.ietf.org/html/rfc7578#section-4.3):

> To match widely deployed implementations, multiple files MUST be sent by supplying each file in a separate part but all with the same "name" parameter.

From [section 2.3 of RFC2183](https://tools.ietf.org/html/rfc2183#section-2.3):
> The receiving MUA SHOULD NOT respect any directory path information that may seem to be present in the filename parameter.  The filename should be treated as a terminal component only.  Portable specification of directory paths might possibly be done in the future via a separate Content-Disposition parameter, but no provision is made for it in this draft.

We're clearly not following the RFC purely as intended, since we use an extra field with data, and we plan to respect the directory to define the target location, but I can't see a reason for us to not approach the standard design as much as possible and use the filename field somewhat as planned. The Go package supports it already as well, and given the widespread use I assume Python does too.

internal/daemon/api_files.go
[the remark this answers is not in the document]
niemeyer: I see, makes sense. The name of that parameter is a bit awkward, but I guess consistency wins.

internal/daemon/api_files.go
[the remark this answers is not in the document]
niemeyer: Thanks. I misremembered which one was first, but with the whole change in error nesting this makes sense.

internal/daemon/api_files.go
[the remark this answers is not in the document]
niemeyer: This is not the same behavior as `ls`, since `ls` has no idea of wildcard expansion at all. The expansion is always done by the shell, and when you escape the string you disable the shell behavior altogether:
```
$ touch fooa foob 'foo?'
$ ls foo?
'foo?'   fooa   foob
$ ls 'foo?'
'foo?'
```
What I'm suggesting is that we do exactly that: we need two different modes, one in which wildcards are considered, and one where they are not, and we should not guess which is which based on the parameter alone, because that creates behavior people will be surprised and frustrated about.

niemeyer (changes requested):

internal/daemon/api_files.go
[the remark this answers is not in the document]
benhoyt: Good call, renamed to `itself` for clarity.

internal/daemon/api_files.go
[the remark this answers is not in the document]
benhoyt: You're right that my use of `http.Error` with `StatusInternalServerError` was misleading here. I've changed it to just `fmt.Fprint` to the output writer in those cases. Note that for these few cases (writing and encoding the metadata part) we can't do much better than printing to the output and hoping it succeeds. In fact, `CreatePart` basically only returns an error if there was an error writing to the output, so it may not do much good to write more anyway. What will happen in practice in the rare cases that this does occur is the client will not find the multipart boundary and will fail due to invalid multipart / unexpected EOF when reading the multipart data.

Thanks for the links to the multipart RFC. I had thought about using the filename field earlier, but originally decided against it because I know some clients strip off the path part of it (per the "SHOULD NOT" warning in RFC section 2.3 that you've quoted). However, Go does not, and we'll have a lot of control over the Python client (there's no multipart parser in the stdlib) so I think it's okay. I've updated the code to use "filename" -- it is simpler and cleaner.

internal/daemon/api_files.go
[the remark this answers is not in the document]
benhoyt: Fair enough. I've updated it to always have two parameters: `path` and `pattern`. Path is always a real file or directory path. Pattern is optional, and if specified, filters the list down to only the names which match the pattern. So the different ways you would call it are as follows:

```
path=/my/file               # get info about a single file (will still return list, but length 1)
path=/my/dir                # list info about all entries in directory
path=/my/dir&itself=true    # list info about directory itself
path=/my/dir&pattern=*.txt  # list info about names in directory that match *.txt
```

(I've also updated the name of the "directory" field to "itself" per your earlier comment.)

internal/daemon/api_files.go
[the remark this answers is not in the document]
benhoyt: Oh, I forgot to mention: I did see your note about calculating the digest as we read and including it in the metadata at the end, and I think that's a good idea, however, we'd really love to get this first PR included in Juju for tomorrow's 2.9rc10 release (we're trying to do it on a weekly cadence). I believe that's something that can easily be added later, and clients can check it if it's present.

benhoyt:
@niemeyer Note the late-in-the-game tweak to the response status code. This makes responses more consistent, and easier to handle on the client (I'm half way through writing the Python client). I think it's more in line with what HTTP 400 Bad Request means in any case. Now:

* If the HTTP status is 200, it means the entire response was processed successfully (though there may be one or more errors with certain paths). In this case the response `type` is `sync`, and the "shape" of the result is always a list of path info objects with an `error` key if there's an error for that path.
* If the status is 400, it means there was something malformed about the entire request (not just a single path). In this case the response `type` is `error`, and the "shape" of the result is always an error object.

A client can still signal an error if there are one or more path errors if it wants to. But while writing the Python client it seemed weird to have to sometimes parse the result as an error, and sometimes as a list of path information.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,576 @@
...
+	"os/user"
+	pathpkg "path"
+	"strconv"
+	"time"
+
+	"github.com/canonical/pebble/internal/osutil"
+	"github.com/canonical/pebble/internal/osutil/sys"
+)
+
+const minBoundaryLength = 8
```
niemeyer, line 37: Aren't boundaries pretty long strings in general?  Can we do at least 32 here, considering Python and Go conventions?

internal/daemon/api_files.go
```diff
@@ -0,0 +1,576 @@
...
+
+func nonAbsolutePathError(path string) error {
+	return fmt.Errorf("paths must be absolute, got %q", path)
+}
+
+func readFile(path string, mw *multipart.Writer) error {
+	if !pathpkg.IsAbs(path) {
+		return nonAbsolutePathError(path)
+	}
+	if osutil.IsDir(path) {
```
niemeyer, line 138: This should probably whitelist instead of blacklist (we don't want devices, sockets, etc). We should probably allow plain files only, but accept symlinks to plain files as well.

internal/daemon/api_files.go
```diff
@@ -0,0 +1,576 @@
...
+	status := http.StatusBadRequest
+	kind := errorKindGenericFileError
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		status = http.StatusNotFound
+		kind = errorKindNotFound
+	case errors.Is(err, os.ErrPermission):
+		status = http.StatusForbidden
+		kind = errorKindPermissionDenied
+	}
```
niemeyer, line 189: The function above is entirely contained inside this function. We can just `return &resp{..., fileErrorToResult(...)}` I think.

niemeyer (approved):
Nice, this API is looking pretty sweet. Thanks for it all, Ben!

Some trivial bits, and +1 when you're happy.

internal/daemon/api_files.go
benhoyt: Good call. I'd originally used 8 so I didn't have to go crazy in the tests, but now I'm just using `01234567890123456789012345678901` is the tests, no big deal.

internal/daemon/api_files.go
benhoyt: Good call. Switched to use `os.Stat` with `info.Mode().IsRegular()`.

internal/daemon/api_files.go
benhoyt: Refactored/simplified - thanks. (We still need the status here.)

PR #28, opened 2021-04-16
Files API: add user/group to list-files, and a few minor fixes

benhoyt:
This PR includes the following:

* Add user-id, user, group-id, group fields to the list-files response (with user/group lookup cached per API call for efficiency).
* Minor fix to the list-files response to give `result: []` in the JSON instead of `result: null` when there are no results, e.g., the directory is empty.
* Change UserID/GroupID fields from int to *int as 0 is actually a valid value ("root"), so we couldn't distinguish between unset and someone trying to set both to "root".
* Add a couple more tests.


niemeyer (approved):
Sensible tweaking, thanks!

PR #29, opened 2021-04-22
Foundation work for service log buffers with `pebble run -v`

hpidcock:
This PR adds the management fundamentals for for in-memory service logging.

Including:
- Concurrency safe fixed sized byte ring buffer for storing stream output from services.
- Concurrency safe write buffer that preserves write information (timestamp, stdout/stderr separation and ordering). The write buffer stores the payload in the ring buffer described above, while storing write metadata in its own ring buffer.
- Runtime allocations currently limited to notification channels (one per write, closing the previous channel to notify readers)
- Multi-reader through iterator objects that are concurrency safe, provide notifications for more data, prevent recycling of ring buffer slots by holding reference to unread portions.
- `pebble run -v` flag to aggregate and dump all service logs to stdout.

`pebble logs` command, plus API machinery to support that command will be included in the next PR (https://github.com/canonical/pebble/pull/31)

There are two tunable limits we can play with:
1. Ring buffer size (for storing the payload) which currently is 1MiB per service.
2. Write buffer size (for storing write metadata) which currently is ~10500 writes (using an avg. log line of 100 bytes).

Design decisions:
- Opted for a truncation-free design (except when the write exceeds the buffer size, this truncates the write into multiple writes)
- Opted to share a common ring buffer for variable length payloads, this maximizes the memory to allow it be used by every write. Allows each write to be variable in length.
- Opted to track writes rather than lines. Simplifies state keeping, simplifies streaming (immediate visibility for incomplete lines) and works for applications that behave poorly (i.e. doesn't write newlines).

Known limitations:
- Slow consumer problem, need the means to drop slow readers when the ring buffer recycling stalls + implement eager recycling to maintain a suitably sized free region.
- Long writes that exceed the buffer size have to be split up over multiple write calls (reports io.ErrShortWrite)

See also the [Pebble log management spec](https://docs.google.com/document/d/1GeM4V7FH-4JeMy_bXpBxaxWnlqrqXcuQxWNz5eJFMt0/edit#heading=h.yij8xaij9lcy)


internal/servicelog/write_buffer.go
```diff
@@ -0,0 +1,354 @@
...
+func (wb *WriteBuffer) advanceHead() {
+	atomic.AddInt32(&wb.numWrites, 1)
+	atomic.AddInt32(&wb.headIndex, 1)
+}
+
+func (wb *WriteBuffer) signalWaiters() {
+	wb.signalMut.Lock()
+	defer wb.signalMut.Unlock()
+	close(wb.signalChan)
+	wb.signalChan = make(chan struct{})
```
benhoyt, line 258: I am concerned that this is a heap allocation every log -- and in our quest to alleviate pressure on GC using ring buffers etc, we're reintroducing the allocation here anyway. I don't know how large the channel struct is (looks like it might be around 12 words or 96 bytes: https://golang.org/src/runtime/chan.go). Maybe we should do some measurements?

In any case, I don't think it's pressing -- I think as long as we have the API right, this can be considered an implementation detail.

Also, we should probably do some tests where we leave Pebble running overnight with a couple of services that log a lot, to see if there are any memory or goroutine leaks or other long-term resource issues. I can do that tomorrow.

internal/servicelog/iterator.go
```diff
@@ -0,0 +1,152 @@
...
+		return 0, io.EOF
+	}
+	start := it.write.start + RingPos(it.read)
+	end := it.write.end
+	read, err := it.wb.ringBuffer.WriteTo(writer, start, end)
+	it.read += int(read)
+	return read, err
+}
+
+func (it *iterator) Buffers() [2][]byte {
```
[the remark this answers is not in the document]
hpidcock: Heh, well since we had all those discussions, we kept changing our minds, so I kept each iteration we did. I think its fine as we don't know where this work will go at this stage, since it's all tested we can probably keep it and retire it later.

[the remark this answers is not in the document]
benhoyt: Hah, yep, fair enough.

internal/servicelog/output.go
```diff
@@ -0,0 +1,50 @@
...
+// Sink blocks until done is closed.
+func Sink(it Iterator, out Output, serviceName string, done <-chan struct{}) error {
+	defer it.Close()
+	last := false
+	for {
+		more := it.More()
+		for it.Next() {
+			err := out.WriteLog(it.Timestamp(), serviceName, it.StreamID(), it)
+			if err != nil {
+				return err
```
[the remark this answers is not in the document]
hpidcock: For now I'd rather bubble up errors so we can see them

internal/servicelog/write_buffer.go
hpidcock: I'm not too concerned with this.

Next best thing would be buffered channels per iterator and a registration list. It means sending on multiple channels, but still preserves the ability of the receiver to cancel waiting (unlike with waitgroups/mutex).

Signaling becomes more expensive, but less churn on the gc.

benhoyt:
I ran a simple "load test" against this code for most of a day, with Pebble running 3 services, each of which output a log every 50sm on average, to exercise it and check for thread or memory leaks. Definitely not a rigorous proof that it's bug free, but passes the sniff test. It stabilized on 18 OS threads and 20MB RSS (columns are time-of-day: pid num-threads rss-mb cmdline):

```
 8:50: 1071129    17 15896 ./pebble run -v
 9:15: 1071129    18 18372 ./pebble run -v
 9:39: 1071129    18 19304 ./pebble run -v
 9:47: 1071129    18 19672 ./pebble run -v
10:22: 1071129    18 20396 ./pebble run -v
11:33: 1071129    18 20264 ./pebble run -v
11:45: 1071129    18 19952 ./pebble run -v
12:54: 1071129    18 20440 ./pebble run -v
14:52: 1071129    18 20556 ./pebble run -v
16:20: 1071129    18 20560 ./pebble run -v
16:47: 1071129    18 20092 ./pebble run -v
```

hpidcock:
Running the few benchmarks we have so far on my local machine to see if there is any significant slowdown. Still need to do some more benchmarks, but I believe these results are acceptable.

```
goos: linux
goarch: amd64
pkg: github.com/canonical/pebble/internal/servicelog
cpu: AMD Ryzen 9 3900X 12-Core Processor
BenchmarkRingBufferWriteSmall             	46361967	        24.67 ns/op
BenchmarkRingBufferWriteSmall-2           	53676553	        22.44 ns/op
BenchmarkRingBufferWriteSmall-4           	45628158	        22.15 ns/op
BenchmarkRingBufferWriteSmall-8           	44999754	        22.35 ns/op
BenchmarkRingBufferWriteSmall-24          	43869945	        22.94 ns/op
BenchmarkRingBufferWrite                  	39913106	        25.68 ns/op
BenchmarkRingBufferWrite-2                	50517768	        26.07 ns/op
BenchmarkRingBufferWrite-4                	41453491	        24.98 ns/op
BenchmarkRingBufferWrite-8                	41033641	        24.63 ns/op
BenchmarkRingBufferWrite-24               	41153439	        24.49 ns/op
BenchmarkRingBufferCopy                   	80187750	        15.10 ns/op
BenchmarkRingBufferCopy-2                 	79855620	        15.00 ns/op
BenchmarkRingBufferCopy-4                 	79999392	        15.05 ns/op
BenchmarkRingBufferCopy-8                 	78182415	        14.69 ns/op
BenchmarkRingBufferCopy-24                	83575770	        14.74 ns/op
BenchmarkWriteBufferConcurrentSmall       	 3020390	       475.4 ns/op
BenchmarkWriteBufferConcurrentSmall-2     	 2301459	       519.5 ns/op
BenchmarkWriteBufferConcurrentSmall-4     	 2719784	       398.7 ns/op
BenchmarkWriteBufferConcurrentSmall-8     	 2771666	       403.9 ns/op
BenchmarkWriteBufferConcurrentSmall-24    	 2486174	       545.5 ns/op
BenchmarkWriteBufferConcurrent            	 3261088	       367.1 ns/op
BenchmarkWriteBufferConcurrent-2          	 2158870	       521.1 ns/op
BenchmarkWriteBufferConcurrent-4          	 3040135	       414.5 ns/op
BenchmarkWriteBufferConcurrent-8          	 2919121	       428.6 ns/op
BenchmarkWriteBufferConcurrent-24         	 2157354	       532.8 ns/op
PASS
ok  	github.com/canonical/pebble/internal/servicelog	40.813s
```

internal/logger/logger.go
```diff
@@ -46,94 +41,78 @@ var NullLogger = nullLogger{}
 
 var (
 	logger Logger = NullLogger
-	lock   sync.Mutex
```
hpidcock, line 49: @benhoyt I still don't believe we can remove this, its important to protect the `logger` variable's memory.

benhoyt:
@hpidcock Nice! For reference, I re-ran the benchmarks with memory profiling enables. For most it shows no allocs, which is good. Oddly, for the 1-CPU run of the write buffer benchmarks it shows 1 alloc per op, but for the other runs it doesn't. Odd:

```
$ go test -run='^$' -benchmem -bench=. -v -cpu=1,2,4,8,24 ./internal/servicelog/
goos: linux
goarch: amd64
pkg: github.com/canonical/pebble/internal/servicelog
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkRingBufferWriteSmall
BenchmarkRingBufferWriteSmall             	22724634	        49.64 ns/op	       1 B/op	       0 allocs/op
BenchmarkRingBufferWriteSmall-2           	24565872	        48.52 ns/op	       1 B/op	       0 allocs/op
BenchmarkRingBufferWriteSmall-4           	24271093	        48.33 ns/op	       1 B/op	       0 allocs/op
BenchmarkRingBufferWriteSmall-8           	22878421	        49.01 ns/op	       1 B/op	       0 allocs/op
BenchmarkRingBufferWriteSmall-24          	23681080	        48.33 ns/op	       1 B/op	       0 allocs/op
BenchmarkRingBufferWrite
BenchmarkRingBufferWrite                  	21643294	        56.61 ns/op	      24 B/op	       0 allocs/op
BenchmarkRingBufferWrite-2                	20682942	        53.26 ns/op	      24 B/op	       0 allocs/op
BenchmarkRingBufferWrite-4                	20364487	        53.94 ns/op	      24 B/op	       0 allocs/op
BenchmarkRingBufferWrite-8                	20256765	        53.06 ns/op	      24 B/op	       0 allocs/op
BenchmarkRingBufferWrite-24               	22761544	        53.62 ns/op	      24 B/op	       0 allocs/op
BenchmarkRingBufferCopy
BenchmarkRingBufferCopy                   	34290349	        36.94 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferCopy-2                 	30353947	        38.03 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferCopy-4                 	33008377	        37.72 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferCopy-8                 	31941633	        37.78 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferCopy-24                	30861390	        39.09 ns/op	       0 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrentSmall
BenchmarkWriteBufferConcurrentSmall       	 1879486	       619.6 ns/op	     153 B/op	       1 allocs/op
BenchmarkWriteBufferConcurrentSmall-2     	 2236633	       571.1 ns/op	     105 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrentSmall-4     	 3007882	       415.3 ns/op	      81 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrentSmall-8     	 3495286	       342.9 ns/op	      69 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrentSmall-24    	 3307507	       349.8 ns/op	      61 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrent
BenchmarkWriteBufferConcurrent            	 2100082	       583.9 ns/op	     176 B/op	       1 allocs/op
BenchmarkWriteBufferConcurrent-2          	 2048551	       590.3 ns/op	     128 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrent-4          	 2952591	       480.5 ns/op	     104 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrent-8          	 3553885	       332.6 ns/op	      92 B/op	       0 allocs/op
BenchmarkWriteBufferConcurrent-24         	 3195494	       346.0 ns/op	      84 B/op	       0 allocs/op
PASS
ok  	github.com/canonical/pebble/internal/servicelog	36.240s
```

benhoyt:
Further to that, it looks like my suspicion about signalWaiters closing the channel and re-making it for each write is correct: `signalWaiters` uses about 60% of the total CPU and is responsible for 50% of the memory allocation (during the whole benchmark run, including the overall `NewWriteBuffer` allocation which only happens once per test). I'm not saying we need to fix this now (it's probably fast enough unless there are a huge volume of logs), but wanted to post here while it's fresh.

These profiling results were written with the following command (and then I used `go tool pprof` to view):

```
$ go test -run='^$' -benchmem -bench=WriteBuffer -v -cpu=1,2 -memprofile=mem.out -cpuprofile=cpu.out ./internal/servicelog/ -test.memprofilerate=1
```

cpu

mem

internal/servicelog/write_buffer.go
benhoyt: See benchmark results about the cost of this here: https://github.com/canonical/pebble/pull/29#issuecomment-829716115 (again, not saying we need to fix it now, just wanted to copy results while I have them handy).

internal/logger/logger.go
benhoyt: I'm not sure I follow what the concern is: multiple goroutines *reading* from a variable concurrently is fine, it's only when there's something writing as well that it's unsafe. `SetLogger` is the only thing that writes it, but that's only called once at startup from the main goroutine, which I've noted in `SetLogger`'s doc comment (`MockLogger` also calls `SetLogger`, but we not running tests concurrently within the package, so that's fine).

I removed this because it seems a shame to be locking something that never needs it, as we're never calling `SetLogger` except the once on startup. And now that the Logger type locks itself (the stdlib's logger did as well), it would have been two lock/unlock sequences for every write.

hpidcock:
> Further to that, it looks like my suspicion about signalWaiters closing the channel and re-making it for each write is correct: `signalWaiters` uses about 60% of the total CPU and is responsible for 50% of the memory allocation (during the whole benchmark run, including the overall `NewWriteBuffer` allocation which only happens once per test). I'm not saying we need to fix this now (it's probably fast enough unless there are a huge volume of logs), but wanted to post here while it's fresh.
> 
> These profiling results were written with the following command (and then I used `go tool pprof` to view):
> 
> ```
> $ go test -run='^$' -benchmem -bench=WriteBuffer -v -cpu=1,2 -memprofile=mem.out -cpuprofile=cpu.out ./internal/servicelog/ -test.memprofilerate=1
> ```
> 
> cpu
> 
> mem

Can you rerun the cpu profile with memprofile turned off, thats the biggest chunk under `runtime.profilealloc`

benhoyt:
@hpidcock Good call, done and updated my comment and the graphs.

internal/servicelog/write_buffer.go
```diff
@@ -0,0 +1,354 @@
...
+	write.index = index
+	write.time = writeTime
+	write.streamID = streamID
+	write.start = wb.ringBuffer.Pos()
+	n, err := wb.ringBuffer.Write(p[:writeLength])
+	if err != nil && err != io.ErrShortWrite {
+		return 0, err
+	}
+	write.end = wb.ringBuffer.Pos()
+	written = n
```
niemeyer, line 168: This file is where I got a bit lost. I don't understand what this piece does in the puzzle, since I was assuming all we needed was a ring buffer, and something similar to an iterator that consumes data from it. To have multiple consumers I imagine we need a list of earliest reads. So in comparison this all looks surprisingly complex.

There are even sleeps in the middle of that logic, which is a strong smell.

I'm hoping we can significantly simplify this.

niemeyer (changes requested):
Most of it looks fantastic. Many nitpicks, and a couple of more relevant questions.

internal/servicelog/iterator.go
```diff
@@ -0,0 +1,166 @@
...
+type BufferedWrite interface {
+	// Reset the read offset used by io.Reader and io.WriterTo.
+	Reset()
+	// Length of the buffered write.
+	Length() int
+	// Timestamp of when the buffered write was received by the write
+	// buffer.
+	Timestamp() time.Time
+	// StreamID of which stream the buffered write was received on.
+	// This can be either Stdout or Stderr.
```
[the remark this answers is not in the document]
hpidcock: @niemeyer the stream ID is just a label on the log write. For a single process we have a single buffer, both stdout and stderr are recorded into the same buffer to maintain the order and to share the same buffer limits.

cmd/pebble/main.go
```diff
@@ -335,10 +335,7 @@ func (e *exitStatus) Error() string {
 }
 
 func run() error {
-	err := logger.SimpleSetup()
-	if err != nil {
-		fmt.Fprintf(Stderr, "WARNING: Cannot activate logging: %v\n", err)
-	}
+	logger.SetLogger(logger.New(os.Stderr, "[pebble] stderr: "))
```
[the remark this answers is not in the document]
benhoyt: The `stderr:` part was to make it consistent with service log output, which could be either stdout or stderr. However, it's not really necessary. I've simplified to your suggestion `<timestamp> [pebble] <msg>` and added code to forbid `pebble` as a service name (in `ParseLayer`).

internal/daemon/daemon.go
```diff
@@ -57,6 +58,9 @@ type Options struct {
 	// to communicate with the daemon. Defaults to a hidden (dotted) name inside
 	// the pebble directory.
 	SocketPath string
+
+	// VerboseOutput is set when service logs should be aggregated via stdout.
```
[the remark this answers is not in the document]
benhoyt: I've updated this comment -- hopefully it's clearer now.

cmd/pebble/cmd_run.go
```diff
@@ -189,3 +202,80 @@ out:
...
+	msg    []byte
+	mutex  sync.Mutex
+}
+
+func (w *logWriter) WriteLog(timestamp time.Time, serviceName string, stream servicelog.StreamID, message io.Reader) error {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	// Use a buffer for the prefix to minimize the number of writes. Use the
+	// format "2021-08-04T12:34:45Z00:33 serviceName stdout/stderr: message".
```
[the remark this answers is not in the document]
benhoyt: Done -- see below. Note that the times are already all UTC and suffixed with `Z` -- it was just the comment that was wrong.

internal/logger/logger_test.go
```diff
@@ -36,50 +34,17 @@ type LogSuite struct {
 	restoreLogger func()
 }
 
-func (s *LogSuite) SetUpTest(c *C) {
-	s.logbuf, s.restoreLogger = logger.MockLogger()
+func (s *LogSuite) SetUpTest(_ *C) {
```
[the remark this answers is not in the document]
benhoyt: Done.

internal/overlord/servstate/handlers.go
```diff
@@ -48,6 +49,10 @@ var (
 	okayWait = 1 * time.Second
 	killWait = 5 * time.Second
 	failWait = 10 * time.Second
+
+	maxLogBytes  = 1024 * 1024
+	avgLogLine   = 100
+	maxLogWrites = maxLogBytes / avgLogLine
```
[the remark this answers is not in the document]
benhoyt: This does actually limit the number of writes, and old writes will simply be overwritten when it hits this maximum (around 10k). Is that correct, @hpidcock?

internal/overlord/servstate/handlers.go
```diff
@@ -90,23 +95,43 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
 		active.err = cmd.Wait()
+		_ = active.logBuffer.Close()
 		close(active.done)
 	}()
+	if m.verboseOutput != nil {
+		go func() {
+			defer outputIterator.Close()
+			err := servicelog.Sink(outputIterator, m.verboseOutput, req.Name, active.done)
+			if err != nil {
+				logger.Noticef("service %s log sink closed with %v", req.Name, err)
```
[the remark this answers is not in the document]
benhoyt: Done.

internal/overlord/servstate/manager.go
```diff
@@ -288,3 +293,31 @@ func (m *ServiceManager) StopOrder(services []string) ([]string, error) {
...
+
+// ServiceLogs returns iterators to the provided services. Each iterator
+// must be closed via the Close method.
+func (m *ServiceManager) ServiceLogs(services []string) (map[string]servicelog.Iterator, error) {
+	releasePlan, err := m.acquirePlan()
+	if err != nil {
+		return nil, err
+	}
+
+	defer releasePlan()
```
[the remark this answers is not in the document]
benhoyt: Done.

internal/overlord/servstate/manager_test.go
```diff
@@ -89,8 +95,18 @@ func (s *S) SetUpTest(c *C) {
 	err = ioutil.WriteFile(filepath.Join(s.dir, "layers", "002-two.yaml"), []byte(planLayer2), 0644)
 	c.Assert(err, IsNil)
 
+	s.logBufferMut.Lock()
+	s.logBuffer.Reset()
+	s.logBufferMut.Unlock()
+	logOutput := servicelog.OutputFunc(func(_ time.Time, _ string, _ servicelog.StreamID, message io.Reader) error {
```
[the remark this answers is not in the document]
benhoyt: Good call -- fixed.

internal/overlord/servstate/handlers.go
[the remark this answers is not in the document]
hpidcock: Correct, this is the maximum number of writes that the buffer stores, not the max number of writes all time.

internal/servicelog/iterator.go
```diff
@@ -0,0 +1,166 @@
...
+package servicelog
+
+import (
+	"io"
+	"time"
+)
+
+type Iterator interface {
+	// Close must be called after the consumer is finished to release
+	// backing buffers.
```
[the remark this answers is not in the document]
hpidcock: I'm working through this, I've got a bunch of them done, but I'll do a full review soon.

internal/servicelog/output.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+		}
+		if last {
+			break
+		}
+		select {
+		case <-more:
+		case <-done:
+			last = true
```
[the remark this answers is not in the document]
hpidcock: Yeah this is a much better approach, I've replicated what you've described here.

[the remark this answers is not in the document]
hpidcock: If done channel is nil, we only returned whats buffered and don't block. Closed channel has the same effect. If the channel closes while waiting, it also returns early with false.

internal/servicelog/iterator.go
```diff
@@ -0,0 +1,166 @@
...
+	// writes to advance to.
+	More() <-chan struct{}
+	BufferedWrite
+}
+
+type BufferedWrite interface {
+	// Reset the read offset used by io.Reader and io.WriterTo.
+	Reset()
+	// Length of the buffered write.
+	Length() int
```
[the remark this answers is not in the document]
hpidcock: I think the semantics of Buffered make the most here, so I'll update it to match. Thanks.

internal/servicelog/ring_buffer.go
```diff
@@ -0,0 +1,231 @@
...
+
+import (
+	"errors"
+	"io"
+	"sync"
+)
+
+var (
+	ErrFreeOutOfOrder = errors.New("free out of order")
+	ErrOutOfRange     = errors.New("index out of range")
```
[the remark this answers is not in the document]
hpidcock: Agreed, thanks for the suggestion.

internal/servicelog/write_buffer.go
hpidcock: @niemeyer WriteBuffer tracks each write that comes out of stdout/stderr for a process, keeping track of the metadata around the write (ts, which stream, where it is in the ring buffer). It also handles multiple writer/multiple readers and the concurrency issues around that.

My original implementation was much more complicated as you saw. I've made an attempt to simplify it with a list of the iterators following the buffer, so that we can both see what are the earliest reads and keep a notification channel for each iterator.

niemeyer: I see what it does. I don't understand why it does that, though. Why isn't this just the ring buffer itself with just tracking of read position?  We do need to track the position of readers, but that's just a RingBufferReader with an int (or RingPos) of its current position, that once Closed gets off the list.

For timestamps, we can add them when we write the data into the buffer, as logging normally does.

For finding out which stream, I don't think we want that per earlier comment. Separating stdout and stderr creates bad logging because they get out of sync. To fix that we need to indeed pass a single file descriptor into the process (in Go, cmd.Stdout and cmd.Stderr need to be the same buffer). So tracking the stream ID is impossible if we don't even have two streams.

Tracking individual writes: why?   I could understand tracking individual lines and having all of our API be line-based, but we are neither benefiting from a simple API that manipulates a list of strings, nor a simple API that manipulates a buffer.

Happy to have a conversation about this with more bandwidth.

cmd/pebble/cmd_run.go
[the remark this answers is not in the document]
jameinel: just as a reference, I believe Juju stores logs as UTC timestamps, and there is a flag on "juju debug-log" as to whether you want to see them in local time. (eg, it is the client rendering the timestamp that chooses what timezone offset to display), which has generally been helpful for users. We might consider that for the 'pebble logs' CLI.

internal/servicelog/iterator.go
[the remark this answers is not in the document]
jameinel: Is that just a documentation issue, and it should be "This is the combined output of Stdout and Stderr" ?

internal/servicelog/write_buffer.go
jameinel: FWIW, I believe in many processes stdout is buffered and stderr is not, which means lines get out of sync anyway.
I would certainly do a best effort to try and keep them in sync (don't make the problem worse), but also recognize that they are fundamentally 2 streams that do get out of sync.

niemeyer: I don't understand what that means, given that when you create a process you're the one defining what the stdout and stderr file descriptors are. If it is the same buffer, they cannot get out of sync because there is only one stream and one buffer. If they are out of sync, the only way to get that is if some child processes spawned further processes and messed that up, in which case Pebble has no saying and it's not its problem anymore.

jameinel:
So inside the child process, they may wrap the raw file handle in a
buffered writer or not. I have definitely seen applications which do buffer
their writes to stdout but not to stderr. I do believe there isn't much
reason for Pebble to pass in 2 different file handles if we are going to be
putting the content into a single stream/buffer anyway.

internal/servicelog/write_buffer.go
niemeyer: @jameinel Your email reply got off thread in GitHub, so I'm bringing it back:

> So inside the child process, they may wrap the raw file handle in a buffered writer or not. I have definitely seen applications which do buffer their writes to stdout but not to stderr. I do believe there isn't much reason for Pebble to pass in 2 different file handles if we are going to be putting the content into a single stream/buffer anyway.

This is the case I described above: if an application misbehaves internally, of course there's nothing we can do. The discussion here is about getting our part right.

hpidcock: ### Single stream
Interesting. So by having a single fd we can guarantee write order on both stdout/stderr because they **are** the same. In doing so we lose the ability to distinguish between stdout/stderr. I don't think that is too much of an issue, it does limit what level of filtering the end user can do, but I don't think that is too much of a concern. `kubectl logs` for example doesn't let you separate the two, they use a single file for both stdout/stderr.

Separate FDs does make more sense in a `ssh`/`exec` scenario, when we add the ability to run commands via pebble, but in that case we don't care about timestamps, we can just use two ring buffers.

### Embedding timestamps
If we embed the timestamps in the ring byte buffer, it will make multiple log formats a little bit more difficult. The log format I personally care about the most is a **raw** output, so I can see the exact output of the program (sans timestamps/service name), so if the encoding in the buffer is strict (as to be able to recreate the output as it was received), then I don't see any problem with this.

There is a tiny amount of state keeping for insertion of timestamp, but it can be as simple as `on the next byte after \n insert the timestamp`. The state is required because a line could span multiple write on the fd.

Streamed output from a program with long lines (spanned over multiple `write`s) makes it difficult to represent the API in the form of lines, I think a streaming style API will better suit this. Within go this would be an io.Reader and over HTTP its just as simple. A line by line API would have to omit incomplete lines when streaming/following, this is probably fine.

With embedded timestamps, mixing multiple streams (multiple services' output) will be a little bit more difficult, will need to read ahead on all the io.Reader for each service, then select the most recent by parsing the timestamps and comparing.
I think we can use the same iterator interface we currently have, the iterator can handle all this.

benhoyt: I'm on board with having a single stream - I don't think that's too useful. However, I do prefer the timestamps being "metadata" rather than string data in the ring buffer that clients may have to parse out if they want to reformat. This is a tricky discussion to have async. I suggest we discuss and finalize this in a video chat -- I'll send a calendar invite / Mattermost message to see if we can touch base today.

internal/servicelog/iterator.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	n, err := it.rb.WriteTo(writer, it.index)
+	if err == ErrRange {
+		it.truncated()
+		err = nil
+	}
+	if n > 0 {
+		it.truncWritten = false
+	}
+	it.index += RingPos(n)
```
niemeyer, line 145: On truncation, won't this just keep writing the truncation message forever if the buffers are off? Should probably be set to the start of the current position, but even that sounds a bit suspect because there's a chance the buffer will be off again next time we try. Ideal situation is probably to make the index float on truncation, and then only redefine it once we're effectively reading some real data, so that we'll immediately move the marker forward and make room for the ring buffer write to overwrite some data without another immediate truncation.

internal/servicelog/ringbuffer.go
```diff
@@ -0,0 +1,369 @@
...
+	rb.rwlock.RLock()
+	defer rb.rwlock.RUnlock()
+	return rb.writeClosed
+}
+
+// Write writes p to the backing buffer, allocating the number of bytes in p.
+// If p is larger than the size of the buffer then io.ErrShortWrite is returned and the
+// number of bytes written. If the p is larger than the number of bytes available,
+// then the tail is discarded to make room.
+func (rb *RingBuffer) Write(p []byte) (written int, _ error) {
```
niemeyer, line 78: Can we please have `err` as the second variable name?  This is reading incorrect to me right now as I generally associate the `_` with "unused", but that error may actually be set to a valid value below and should be checked by the call site.

internal/servicelog/ringbuffer.go
```diff
@@ -0,0 +1,369 @@
...
+// Positions returns the start and end positions of readable data in the RingBuffer.
+func (rb *RingBuffer) Positions() (start RingPos, end RingPos) {
+	rb.rwlock.RLock()
+	defer rb.rwlock.RUnlock()
+	return rb.readIndex, rb.writeIndex
+}
+
+// Copy copies bytes from the range into the supplied dest buffer. If dest is not large enough
+// to fill the bytes from start to end, then start to start+len(dest) is copied and
+// the error io.ErrShortBuffer is returned.
```
niemeyer, line 161: The error condition may be unexpected eventually, as in general filling the space available in dest is the success case of copy functions in Go. Probably not a big deal given that this is an internal API though. Just in case, though: are we making use of this error, or would using the returned `n` as usual with `copy` make the logic just as simple for call sites?

internal/servicelog/ringbuffer.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	if rb.Closed() {
+		close(iter.closeChan)
+	}
+	rb.iteratorList = append(rb.iteratorList, iter)
+	return iter
+}
+
+// HeadIterator returns an iterator from the head of the buffer.
+func (rb *RingBuffer) HeadIterator(lines int) Iterator {
```
niemeyer, line 247:
```
// If lines is greater than zero, the iterator will start that many lines
// backwards from the head.
```
... I suppose.

internal/servicelog/ringbuffer.go
```diff
@@ -0,0 +1,369 @@
...
+			continue
+		}
+		close(iter.closeChan)
+		rb.iteratorList[i] = rb.iteratorList[len(rb.iteratorList)-1]
+		rb.iteratorList = rb.iteratorList[:len(rb.iteratorList)-1]
+		return
+	}
+}
+
+func (rb *RingBuffer) buffers(start, end RingPos) [2][]byte {
```
niemeyer, line 346: Would be nice to have a few words documenting this function, as it's not typical nor obvious.

internal/servicelog/ringbuffer.go
```diff
@@ -0,0 +1,369 @@
...
+	return nil
+}
+
+func (rb *RingBuffer) signalIterators() {
+	rb.iteratorMutex.RLock()
+	defer rb.iteratorMutex.RUnlock()
+	for _, iter := range rb.iteratorList {
+		select {
+		case iter.notifyChan <- struct{}{}:
+		default:
```
niemeyer, line 318: Given that these channels are unbuffered, the `default:` here will be used rule whenever the other side is busy, which means if it is busy and we try to notify it that there's more content being written one instruction before it decides to wait, it's dead and won't ever recognize the fact that there's more data pending before the next notification. For this logic to be correct, we need to have at least 1 item of buffer in the channel.

On a separate note, I also tend to have my channels as `bool` instead of `struct{}`, because code looks nicer and on the consuming side the bool sometimes make logic neater copying the byte back and forth tends to be irrelevant in terms of space or time. Whatever you prefer in this case, though.

niemeyer (changes requested):
Thanks, this looks a lot simpler.

Probably the last few details:

internal/servicelog/formatter.go
```diff
@@ -0,0 +1,95 @@
...
+			f.timestampBuffer = append(f.timestampBuffer, "] "...)
+			f.timestamp = f.timestampBuffer
+		}
+
+		for len(f.timestamp) > 0 {
+			// Timestamp bytes don't count towards the returned count.
+			n, err := f.dest.Write(f.timestamp)
+			f.timestamp = f.timestamp[n:]
+			if err != nil {
+				return 0, err
```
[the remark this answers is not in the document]
hpidcock: I'm not totally sure I understand what you mean. Not sure if you are talking about the written byte count or the error itself.

Consider [`gzip.Writer`](https://golang.org/src/compress/gzip/gzip.go#L139), it doesn't report back the `encoding` size, it reports back the payload written. In cases of error relating to encoding, it reports back the error received from the underlying writer. In cases of success it only report backs how much payload was written.

internal/servicelog/ringbuffer.go
hpidcock: ahhh this is the old comment, I need to fix this sorry. Copy no longer takes an end position, therefore ErrShortBuffer is no longer a possible error state.

internal/servicelog/ringbuffer.go
hpidcock: That is a much better comment, thank-you.

internal/servicelog/ringbuffer.go
hpidcock: Good catch, these were supposed to be buffered channels with a size of 1.

Happy to change to bool, I think struct{} channels only make true sense for cancellation.

internal/servicelog/formatter.go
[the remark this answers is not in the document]
hpidcock: Ahh I think I see it now, we should return `written` for previous calls to write.

internal/servicelog/iterator.go
hpidcock: Good idea, I've moved this to a floating index on truncation, so that the next read will always be the tail.

internal/servicelog/formatter.go
```diff
@@ -30,7 +30,8 @@ type formatter struct {
 }
 
 const (
-	TimeFormat = time.RFC3339
+	// TimeFormat is RFC3339 with millisecond precision.
+	TimeFormat = "2006-01-02T15:04:05.999Z07:00"
```
hpidcock, line 34: @niemeyer it was suggested by @benhoyt we might need slightly more precise timestamps for interleaving logs for multiple services. This would help when debugging multiple services. Regardless I think we can change this at any point in time because RFC3339 allows for any length fraction on the second.

jameinel:
Note that while RFC3339 allows for anything, I believe there is a
restriction on what normal libraries allow.
IIRC Python limits it to 6 digits (microseconds), while Go likes to default
to 9 (nanoseconds).

```
>>> dt.fromisoformat('2021-05-18T10:07:26.123')
datetime.datetime(2021, 5, 18, 10, 7, 26, 123000)
>>> dt.fromisoformat('2021-05-18T10:07:26.123456')
datetime.datetime(2021, 5, 18, 10, 7, 26, 123456)
>>> dt.fromisoformat('2021-05-18T10:07:26.123456789')
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '2021-05-18T10:07:26.123456789'
```

Now I believe it does that because the object it uses to track time only
has microsecond resolution (so it can't hold the extra digits of precision).
But I would just caution that while the spec allows arbitrary, common
libraries don't.

internal/servicelog/iterator.go
```diff
@@ -0,0 +1,175 @@
...
+	trunc        []byte
+	truncWritten bool
+	notifyChan   chan bool
+	closeChan    chan struct{}
+}
+
+var _ Iterator = (*iterator)(nil)
+
+var (
+	truncBytes = []byte("(... output truncated ...)\n")
```
niemeyer, line 50: I suppose this might end up at the end of a long line and be completely missed.

If so, let's add a "\n" as the first char as well to make sure the bump is noticeable on the output.

niemeyer (approved):
LGTM, thanks for all the tuning.

Just a final minor suggestion:

benhoyt (approved):

internal/servicelog/iterator.go
benhoyt: Fixed.

benhoyt:
@hpidcock is out today so I'm going to merge this so we can get this change into Juju.

PR #31, opened 2021-04-28
API for fetching logs

benhoyt:
This change provides an HTTP API to access the logs (the ring buffer that Harry added in https://github.com/canonical/pebble/pull/29). The endpoint is `GET /v1/logs` with these query parameters:

* `n`: number of logs to go back - defaults to 10, and negative (e.g., `n=-1`) means show all logs for the given services
* `services`: name of service to list logs for - you can specify this multiple times for multiple services, and the default is to show logs for all services
* `follow`: set to `true` to follow (tail) the given services logs until the client connection is closed (defaults to `false`). Before following the new logs, `n` latest logs from the ring buffer are output first (you can disable this by setting `n=0`).

The format of the output is [JSON Lines](https://jsonlines.org/) (denoted with mime type `application/x-ndjson`). This is widely used for logging (for example [by Docker](https://docs.docker.com/config/containers/logging/json-file/)), supported by log ingestion tools as well as CLI tools such as `jq`. It looks like so:

```
{"time":"2021-04-23T01:28:52.660Z","service":"redis","message":"redis started up"}
{"time":"2021-04-23T01:28:52.798Z","service":"thing","message":"did something"}
```

We also considered using syslog format (RFC5424), but [decided against that](https://github.com/canonical/pebble/pull/31#issuecomment-845671341) as JSON Lines is more typical, and easier to produce and parse.


jnsgruk:
This looks great! I agree with your observation about the quirky format though. While the format decided isn't *that hard* to parse (thinking about people who might be ingesting these logs using something like the EFK stack), it would be nicer to have something that's either completely human readable, or something that is *just* JSON.

It seems there might be a case to give the user some options;

- Default to some standard, plain-text and human readable format - something like:

```
2021-04-30 18:13:56 (stdout) myservice | This is a log message that would be easy to read....
2021-04-30 18:17:52 (stderr) another   | This is a log message that would be easy to read....
2021-04-30 18:21:52 (stdout) other     | This is a log message that would be easy to read, this is a log 
message that would be easy to read, this is a log message 
that would be easy to read, this is a log message that would be easy to read
```

- Provide an option to use JSON logging, where newlines are just inserted as `\n` in a `message` field, designed more for non-human parsing?

```
{"time":"2021-04-30T18:13:56.243533673Z", "service":"myservice", "stream":"stdout", "message":"This is a log message that would be easy to read...."}
{"time":"2021-04-30T18:17:52.987654321Z", "service":"other", "stream":"stderr", "message":"This is a log message that would be easy to read...."}
{"time":"2021-04-30T18:21:52.123456789Z", "service":"other", "stream":"stdout", "message":"This is a log message that would be easy to read, this is a log\nmessage that would be easy to read, this is a log message\nthat would be easy to read, this is a log message that would be easy to read"}
```

benhoyt:
Thanks @jnsgruk. Note that normally this API would be accessed by the (upcoming) `pebble logs` command, and the client would do the formatting. So I think the API response should be easily machine-readable (human consumption is nice, but definitely secondary).

hpidcock:
Most languages supports buffered stream reading with `ReadLine() string` and `Read(n) []byte` like semantics. The benefit of separating the message from the metadata is that it requires no escaping, no additional buffering and is still simple to implement.

While I agree APIs should be human readable if possible, I think the compromises here are acceptable.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,326 @@
...
+			_, err := io.Copy(&buf, it)
+			if err != nil {
+				fmt.Fprintf(w, "\ncannot write log from %q: %v", name, err)
+				return
+			}
+			entries = append(entries, entry{
+				timestamp: it.Timestamp(),
+				service:   name,
+				stream:    it.StreamID(),
+				message:   buf.String(),
```
hpidcock, line 164: we should return to this shortly with reverse iterators on the write buffer.

Imagine 10 (let m=10) services, 1mb logs each, using `pebble logs -n 10000` (let n=10000). It's going to allocate another ~10mb just to then sort through `n*m=100000` logs. Worst case that's 10mb and `O(m*n^2)=10b` iterations, best case 100kb and `O(m*n*log(m*n))=500k` iterations.

With reverse iterators we could do this in no extra memory, `O(2n)=200k` iterations.

benhoyt: Yeah, agreed it's quite expensive at present. Per offline discussion, I'm totally fine if you do this later as a separate PR.

benhoyt:
> The underlying PR is approved. Once it's merged, can you please update this so we get a clean PR to review? This underlying piece is a bit too large for us to review comfortably with it mixed up.

@niemeyer Yep, definitely. I merged Harry's branch and mostly updated this yesterday. Just one test to finish today (and of course merge from master when the underlying PR is merged).

benhoyt:
For reference, we also considered using syslog format here. Copying my research/decision for not choosing that here:

Here's what I've found out about the syslog protocol by skimming the RFCs (RFC5424 is the main one, with RFC6587 covering TCP transport and framing of messages). It's simple enough to write a syslog message. The RFC5424 format is:

```
<pri>version timestamp hostname appname proc_id msg_id structured_data log_message
```

Version is always "1". Most other fields are optional and replaced with "-" if not needed. In our case we only need timestamp, appname (Pebble service name), and log message, for example:

```
<13>1 2003-10-11T22:14:15.003Z - nginx - - - something happened
```

The structured data is a `[key1="val1" key2="val2"]` type of thing, but we wouldn't need that (and, according to some Loggly docs I was reading, most people that want structured logs just stuff JSON in the log message anyway). One weird thing is that if the log message is UTF-8, you're supposed to include a UTF-8 byte order mark prefix, otherwise it's to be interpreted as raw bytes.

However, because the log message isn't escaped you have to deal with framing. Common approaches are what's called "non-transparent framing" with just newlines to terminate the log message (but then what if the message has a newline in it?) and the more modern "octet-count" framing, where each message is prefixed with a count of bytes, like so (though I've added newlines to make it more readable):

```
48 <1>1 2003-10-11T22:14:15.003Z host.local - - - -
25 <3>1 - host.local - - - -
63 <13>1 2003-10-11T22:14:15.003Z - nginx - - - something happened
```

Generating these messages is trivial. Parsing them is a bit messy. There are a couple of Go and Python syslog parsing libraries out there, but none of them has particularly high usage from what I can tell (and we'd have to pull the Python one in as a dependency of Operator Framework when we add the logs API there, or write our own).

Overall I don't think we gain any practical benefit from using the syslog format. JSON Lines is much easier to parse in both Go and Python, and more external tools support it. There's also a fair amount of precedent for JSON Lines logging, including Docker, for example https://docs.docker.com/config/containers/logging/json-file/, where the format is very similar to what I'm proposing:

```
{"log":"Log line is here\n","stream":"stdout","time":"2019-01-01T11:11:11.111111111Z"}
```

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+	encoder := json.NewEncoder(w)
+	encoder.SetEscapeHTML(false)
+	err := iterateLogs(it, reader, nil, now, func(log logEntry) error {
+		return encoder.Encode(log)
+	})
+	return err
+}
+
+// Output last numLogs logs from multiple iterators (not particularly
+// efficient, as it writes them all to a slice first, then sorts by time).
```
niemeyer, line 148: When we have N producers generating items that are already ordered inside each of the producers, a more memory conscious implementation and often more efficient as well is to just order the consumption itself.

For example: if we have A, B, and C, we consume one item from each, and keep track of two of them: the earliest one, and the second-to-earliest one. Then, we produce the earliest one and loop in place for as long as that specific producer has timestamps earlier than the second-to-earliest one. Since logs will tend to be clustered and sparse, besides simple and memory efficient, that should also be much faster than doing the entire operation in memory before producing anything.

Another relevant advantage is that with that one single algorithm implementation, we can cover the single case, the multi case, and the follow case, so `outputLogsMulti`, `outputLogsSingle`, and `followLogs` (and maybe `iterateLogs`) can become one single loop.

Without that, the idea of having to allocate and sort dozens or hundreds of thousands of log entries in memory for a trivial operation like this feels disproportionate to the operation being done.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+	for service, it := range itsByName {
+		wg.Add(1)
+		go func(service string, it servicelog.Iterator) {
+			defer wg.Done()
+			reader := bufio.NewReaderSize(it, bufioReaderSize)
+			fw := &followWriter{w, writeMutex}
+			encoder := json.NewEncoder(fw)
+			encoder.SetEscapeHTML(false)
+			err := iterateLogs(it, reader, done, time.Time{}, func(log logEntry) error {
+				return encoder.Encode(log)
```
niemeyer, line 204: This concurrency here feels a bit brittle. I'll comment more on why below, but as a suggested fix, it would be better if we managed to integrate this into the approach I described above, so that we're still sorting individual lines instead of entire buffers. Our ring buffer iterator is already a middle-man tracking positions between an in-memory buffer which is a moving target and the logic we have above. Maybe we can take that described algorithm and make it work even if the buffer reaches the current EOF.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+			})
+			if err != nil {
+				fmt.Fprintf(w, "\nerror writing logs: %v", err)
```
niemeyer, line 207: Again, we can't write freely to w.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+type followWriter struct {
+	w     io.Writer
+	mutex *sync.Mutex
+}
+
+func (w *followWriter) Write(p []byte) (int, error) {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+	n, err := w.w.Write(p)
+	flushWriter(w.w) // Flush HTTP response after each line in follow mode
```
niemeyer, line 309: Per earlier note, this function is not safe. Even if we ignore error/retry cases, Imagine what happens if any part of the stack writes half a JSON object because that's a perfectly fine way to use Write.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	// .999 allows any number of fractional seconds (including none at all)
+	timestamp, err := time.Parse("2006-01-02T15:04:05.999Z07:00", string(fields[0]))
+	if err != nil {
+		return logEntry{}, false
+	}
+	if len(fields[1]) < 3 {
+		return logEntry{}, false
+	}
+	service := string(fields[1][1 : len(fields[1])-1]) // Trim [ and ] from "[service]"
```
niemeyer, line 293: We need to check and complain loudly if it does not conform.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+				// EOF reached, break after processing line
+				break
+			}
+		}
+	}
+	return nil
+}
+
+// Parse a log entry of the form "2021-05-20T15:39:12.345Z [service] log message"
+func parseLog(line []byte) (logEntry, bool) {
```
niemeyer, line 280: Would be great to have this moved to the services log package where the logs / format is defined, and also move the timing format to a common variable there, and then have a test that straps producing and consuming together to make sure we don't get very ugly bugs where we don't respect our own internal inter-dependencies.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+
+			log, ok := parseLog(line)
+			if ok {
+				// Output normal log
+				err := writeLog(log)
+				if err != nil {
+					return err
+				}
+				if !now.IsZero() && log.Time.After(now) {
+					// Stop if we're past current time
```
niemeyer, line 257: I get the intention here, but this feels a bit confusing. We cannot possibly get beyond the current time when reading logs from the past. The parameter should probably be called `stop` or similar, and the comment can be dropped.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+					return nil
+				}
+				prevTime = log.Time
+				prevService = log.Service
+			} else if !prevTime.IsZero() {
+				// Output partial log with previous timestamp and service name
```
niemeyer, line 263: When will that come into action?  Would be nice to add a note here with some details.

niemeyer (changes requested):
Thanks, Ben. Here are some initial observations.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+		if len(itsByName) == 1 {
+			// Handle single-service case more efficiently
+			for _, it := range itsByName {
+				err = outputLogsSingle(w, it)
+			}
+		} else {
+			err = outputLogsMulti(w, itsByName, numLogs)
+		}
+		if err != nil {
+			fmt.Fprintf(w, "\nerror writing logs: %v", err)
```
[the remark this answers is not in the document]
benhoyt: Yeah, fair enough -- I've changed this to log an error (local only) and return instead of breaking the output.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	// If requested, follow and output logs real time until request closed
+	if follow {
+		flushWriter(w)
+		followLogs(w, itsByName, req.Context().Done())
+	}
+}
+
+// Output a single iterator's logs without reading them all into memory.
```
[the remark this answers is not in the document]
benhoyt: I presume you're referring to starting the comment with the name of the function? Assuming so, fixed (here and other places).

internal/daemon/api_logs.go
benhoyt: Yeah, I think this is fair. It was fairly inefficient. I've reworked the code to do something very close to the above algorithm, merging N streams assuming they're in order (it picks the earliest one each time around). Two things:

1) You can't quite use the same algorithm for the "output last N logs (default)" case and the "output all logs in the ring buffer" case, because we only have a `HeadIterator(N)` which iterates forward. There's currently no way to iterate backward until you have enough (and even if we did have that, you'd still have to buffer them and reverse them). So what I've done is in the "output last N logs" case (the default is 10) I iterate through them as described, and only keep the last 10, using a FIFO queue.
2) For the "follow" case, it's simpler to have a different algorithm, partly because there's a done/cancel channel involved. So I've left `followLogs` largely as it was, though I've changed the outputting to send to a channel to avoid the `Write` concurrency issues you mentioned below.

internal/daemon/api_logs.go
benhoyt: Concurrency: good call. I've updated this to send to a channel and write to the output from only a single goroutine. Safe now, and more Go-like in any case.

internal/daemon/api_logs.go
benhoyt: Fixed.

internal/daemon/api_logs.go
benhoyt: Good call, thanks for pointing that out! I've done this by the sub-goroutines sending to a channel now, so there's only the main goroutine doing the writing.

internal/daemon/api_logs.go
benhoyt: Fixed. It returns an error if the `[` or `]` are missing now.

internal/daemon/api_logs.go
benhoyt: Good call. I've moved `parseLog` to `servicelog.Parse` and added an iterating `servicelog.Parser`. I've added tests for those, as well as a round-trip test like you're describing that tests the log formatter and log parser together.

internal/daemon/api_logs.go
benhoyt: Good call. Renamed to "stop" and updated the comment.

internal/daemon/api_logs.go
benhoyt: Thanks. It will happen when a log line is too big for the read buffer or when `(... output truncated ...)` has been written by the ring buffer. I've updated the comment.

internal/daemon/api_logs.go
benhoyt: Reassuringly, all of the tests in `api_logs_test.go` still pass after this significant rework.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,275 @@
...
+			// HeadIterator(n)) to a FIFO queue of size n and drop oldest. This
+			// avoids reading numServices*numLogs into memory and sorting.
+			fifo := make(chan servicelog.Entry, numLogs)
+			err = writeLogs(itsByName, func(entry servicelog.Entry) error {
+				// If FIFO channel is full, discard oldest log entry before
+				// writing new one so it doesn't block.
+				if len(fifo) == cap(fifo) {
+					<-fifo
+				}
+				fifo <- entry
```
benhoyt, line 131: Note: I'm actually using this channel as a deque / ringbuffer, which has some overhead due to the channel synchronization code. It's about 6x as slow as a hand-rolled deque implementation (see results at https://play.golang.org/p/l09jd5ovdZG), but we're still only talking about ~60ns per append operation (compared to ~10ns) so I prefer the simplicity of just using a channel here.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,275 @@
...
+	stop := time.Now().UTC()
+
+	// Service names, ordered alphabetically for consistent output.
+	names := make([]string, 0, len(itsByName))
+	for name := range itsByName {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	parsersByName := make(map[string]*servicelog.Parser, len(itsByName))
```
hpidcock, line 178: parser setup should happen outside of writeLogs so that writeLogs and followLogs can share the same parser so to not skip data already buffered.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,275 @@
...
+// channel as they're written. It stops when the done channel is closed, and
+// closes the output channel when everything is finished.
+func followLogs(itsByName map[string]servicelog.Iterator, done <-chan struct{}, output chan<- servicelog.Entry) {
+	// Start one goroutine per service to listen for and output new logs.
+	var wg sync.WaitGroup
+	for _, it := range itsByName {
+		wg.Add(1)
+		go func(it servicelog.Iterator) {
+			defer wg.Done()
+			parser := servicelog.NewParser(it, logReaderSize)
```
hpidcock, line 239: parser setup should happen outside of followLogs so that writeLogs and followLogs can share the same parser so to not skip data already buffered.

hpidcock (changes requested):

internal/servicelog/parser.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Parse parses a log entry of the form
+// "2021-05-20T15:39:12.345Z [service] log message".
+func Parse(line []byte) (Entry, error) {
+	fields := bytes.SplitN(line, []byte(" "), 3)
+	if len(fields) != 3 {
+		return Entry{}, errParseFields
+	}
+	// .999 allows any number of fractional seconds (including none at all)
+	timestamp, err := time.Parse("2006-01-02T15:04:05.999Z07:00", string(fields[0]))
```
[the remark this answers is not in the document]
benhoyt: Unfortunately they're slightly different: this uses `.999` instead of `.000`, because that means allow any number of fractional seconds. I wanted to do that in case we decide to use microseconds (or just seconds), the parser will still work fine. `.000` means exactly 3 digits.

internal/servicelog/parser.go
```diff
@@ -0,0 +1,127 @@
...
+	if errors.Is(p.err, io.EOF) {
+		return nil
+	}
+	return p.err
+}
+
+// Parse parses a log entry of the form
+// "2021-05-20T15:39:12.345Z [service] log message".
+func Parse(line []byte) (Entry, error) {
+	fields := bytes.SplitN(line, []byte(" "), 3)
```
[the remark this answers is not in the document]
benhoyt: Yeah, the Go compiler is definitely smart enough to not escape these kinds of literal conversions. This is very common. You can ask the Go compiler whether memory allocations escape with `-gcflags -m`:

```
$ cat t.go
package main

import "os"

func main() {
	os.Stdout.Write([]byte("Hello world\n"))
}
$ go build -gcflags "-m" t.go
# command-line-arguments
./t.go:5:6: can inline main
./t.go:6:24: ([]byte)("Hello world\n") does not escape
```

internal/servicelog/parser.go
```diff
@@ -0,0 +1,127 @@
...
+	errParseFields  = errors.New("line has too few fields")
+	errParseTime    = errors.New("invalid log timestamp")
+	errParseService = errors.New("invalid log service name")
+)
+
+// Entry is a parsed log entry.
+type Entry struct {
+	Time    time.Time
+	Service string
+	Message string
```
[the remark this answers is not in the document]
benhoyt: Yeah, it's a shame. Do you see any way we could? I don't think we can, because the data in the ring buffer may be overwritten.

internal/servicelog/parser.go
```diff
@@ -0,0 +1,127 @@
...
+	p.err = nil
+}
+
+// Next parses the next log from the reader and reports whether another log
+// is available (false is returned on EOF or other read error).
+func (p *Parser) Next() bool {
+	for p.err == nil {
+		line, err := p.br.ReadSlice('\n')
+		if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, bufio.ErrBufferFull) {
+			// Non EOF (and non-buffer full) error, stop now
```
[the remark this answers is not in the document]
benhoyt: Yes, that's correct. With the second entry having the same timestamp as the first (see below in the code for this case).

internal/servicelog/parser.go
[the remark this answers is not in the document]
hpidcock: Fair enough, maybe make this a constant and place it next to the other one.

internal/servicelog/parser.go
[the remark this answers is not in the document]
hpidcock: Ahh cool, I like it.

internal/servicelog/parser.go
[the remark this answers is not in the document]
hpidcock: The only way I can see this happen is with io.WriterTo and making the parser a writer that blocks the Write call until the parsed Entry is no longer required. So no, not a nice way.

internal/servicelog/parser.go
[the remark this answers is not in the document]
benhoyt: I've made this a constant now (however, I've left them in parser.go and formatter.go next to the relevant code). Per discussion, I also un-exported `TimeFormat` and `TimeFormatRegex` as those are implementation details, especially now that the formatter and parser are all in the same package.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,275 @@
...
+			for it.Next(done) {
+				parser.Reset()
+				for parser.Next() {
+					output <- parser.Entry()
```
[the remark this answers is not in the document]
benhoyt: Thanks - done.

internal/daemon/api_logs.go
benhoyt: Done.

internal/daemon/api_logs.go
benhoyt: Done.

benhoyt:
Per voice discussion with @niemeyer, we want to try to unify the `outputLogs` and `followLogs` cases. Plus, there's a potential timing issue with `followLogs` having each goroutine output logs as soon as it receives them ... with the layers of buffering, if lots of logs from different services come at a similar time, they may end up out of order in the output, so we should still run the follow portion through ordering logic.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,262 @@
...
+	// Get log iterators by service (and close them when we're done).
+	itsByName, err := r.svcMgr.ServiceLogs(services, numLogs)
+	if err != nil {
+		response := statusInternalError("cannot fetch log iterators: %v", err)
+		response.ServeHTTP(w, req)
+		return
+	}
+	defer func() {
+		for _, it := range itsByName {
+			_ = it.Close()
```
hpidcock, line 100: excellent

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,262 @@
...
+			if !nexts[i].Time.IsZero() {
+				continue
+			}
+			if parser.Next() {
+				nexts[i] = parser.Entry()
+			} else if parser.Err() != nil {
+				log.Printf("error parsing logs: %v", parser.Err())
+				return
+			} else if iterators[i].Next(nil) {
+				parser.Reset()
```
niemeyer, line 165: At this point we may have an actual entry inside the parser which ideally should be considered for this round since this was just a buffer size behavior and not a timing one, so the time for the upcoming entry might still be earlier than the others. Just adding something like this here should solve it:
```
if parser.Next() {
    nexts[i] = parser.Entry()
}
```
No need to check errors even. We can pick that up on the next round.

Btw, there's a `Reset()` above which I don't quite understand. Why do we need to explicitly reset the parser, instead of just calling `Next` once more? Isn't it just picking up the next item out of the iterator?

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+			if earliest < 0 || next.Time.Before(nexts[earliest].Time) {
+				earliest = i
+			}
+		}
+
+		// No more logs: flush and exit, or wait for another if in follow mode.
+		if earliest < 0 {
+			// If they requested "n" logs, flush logs in the FIFO first.
+			if numLogs > 0 && len(fifo) > 0 {
```
niemeyer, line 183: This new loop is already looking pretty nice in the sense we have the logic almost entirely agnostic to what form of iteration is being requested. The final bits are about that fifo, which needs to wait until the end of the currently available logs because it's the last N and not the first N, and the follow flag, which becomes a breaking point to decide whether to keep going or abort. I think we can make the loop 100% agnostic and simplify it on the way by moving it to a go routine on its own function, and accepting a single cancellation channel as a parameter.

Something along those lines:

```
func ServeHTTP(...) {

    (...)

    logs := make(chan servicelog.Entry)
    go r.steamLogs(logs, requestDone)

    for {
        select {
        case log := <-logs:
            if log.Time.IsZero() {
                    if follow {
                        continue
                    }
                    break
            }
            if numLogs > 0 {
                 // ... add to fifo
            } else {
                // ... encode and write
            }
    }
    close(requestDone)
    // ... flush fifo
}

```

In other words, the entire logic inside the loop is now unaware of fifo or follow. It just sends logs out for as long as it can, until the stop channel is closed. These features are instead implemented at the call site. My theory at the moment is that the code inside the loop will then be simpler and more linear.

What do you think?

niemeyer (changes requested):
Thank you, that's looking nice. Final suggestion probably:

internal/daemon/api_logs.go
benhoyt: Good catch. I've added this and a test for the case in `api_logs_test.go`.

`Parser.Reset` was needed to reset the stored error, which was EOF from the previous call to `Next`. However, this is not really an error case, so I refactored slightly to avoid storing EOF in p.err. This also avoids the need for `Reset`, and we can just call `Next` again as you suggested -- thanks.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,312 @@
...
+	// or we return from this handler.
+	returning := make(chan struct{})
+	defer close(returning)
+	done := make(chan struct{}, 1)
+	go func() {
+		select {
+		case <-req.Context().Done():
+		case <-returning:
+		}
+		done <- struct{}{}
```
hpidcock, line 138: this should be a close

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,312 @@
...
+			logger.Noticef("error writing logs: %v", err)
+			return err
+		}
+		flushWriter(w)
+		return nil
+	}
+
+	// Close the streamLogs done channel when either the request is cancelled
+	// or we return from this handler.
+	returning := make(chan struct{})
```
hpidcock, line 130: I'd replace this whole mess with
```
ctx, cancel := context.WithCancel(req.Context())
defer cancel()
```
Then you can use ctx.Done()

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,312 @@
...
+		case <-returning:
+		}
+		done <- struct{}{}
+	}()
+
+	// Background goroutine to stream ordered logs: it sends parsed logs on
+	// logs channel, any error on errors channel.
+	logs := make(chan servicelog.Entry)
+	errors := make(chan error)
+	go streamLogs(itsByName, logs, errors, done)
```
hpidcock, line 145: make streamLogs return an error then

```
go func() {
   errors <- streamLogs(itsByName, logs, ctx.Done())
}()
```

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,312 @@
...
+		case <-req.Context().Done():
+		case <-returning:
+		}
+		done <- struct{}{}
+	}()
+
+	// Background goroutine to stream ordered logs: it sends parsed logs on
+	// logs channel, any error on errors channel.
+	logs := make(chan servicelog.Entry)
+	errors := make(chan error)
```
hpidcock, line 144: maybe rename to streamErrChan and make it buffered 1

internal/daemon/api_logs.go
benhoyt: Small update: I did add the test in `api_logs_test.go` but now with this code being refactored into a background goroutine and a main loop it's hard to test this case. For now I've just removed the test (I know it's working from the test I did add). Not 100% coverage I know, but I think it's a reasonable trade-off.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,312 @@
...
+	for {
+		// Try to fetch next log from each service (parser/iterator combo).
+		for i, parser := range parsers {
+			if !nexts[i].Time.IsZero() {
+				continue
+			}
+			if parser.Next() {
+				nexts[i] = parser.Entry()
+			} else if parser.Err() != nil {
+				errors <- fmt.Errorf("error parsing logs: %w", parser.Err())
```
hpidcock, line 259: unless errors is buffered and guarantee only one send on that channel, then this needs to be protected in a select with done

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+			if earliest < 0 || next.Time.Before(nexts[earliest].Time) {
+				earliest = i
+			}
+		}
+
+		// No more logs: send empty log to caller, then wait for more logs
+		// or done signal.
+		if earliest < 0 {
+			logs <- servicelog.Entry{}
```
hpidcock, line 283: needs to be in a select with done

internal/daemon/api_logs.go
benhoyt: That's a good suggestion. Strictly speaking we wouldn't need a goroutine, we could use some kind of `logStreamer` type with yet another `next()` method ... I tried this initially but IMO it made the code worse. So I switched to your goroutine/channel based approach, and I think it's pretty clean now. The `streamLogs` goroutine handles just the merging of log streams by timestamp, and the main loop handles the request-specific logic (`follow`, `numLog`).

internal/daemon/api_logs.go
benhoyt: Ah, good call -- I knew there was a better way to do this! Done.

internal/daemon/api_logs.go
benhoyt: Yeah, that's a bit nicer -- thanks. Done.

internal/daemon/api_logs.go
benhoyt: Added buffer of 1 and renamed to `errorChan` as IMO "stream" didn't need to be in it.

internal/daemon/api_logs.go
benhoyt: This is gone now that we're returning the error (and `errorChan` now is now 1-buferred in any case).

internal/daemon/api_logs.go
benhoyt: Done.

niemeyer (approved):
Thank you. Last round with minors only.

LGTM when you're happy with it.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,307 @@
...
+	defer func() {
+		for _, it := range itsByName {
+			_ = it.Close()
+		}
+	}()
+
+	// Make a channel and register it with each of the iterators to be
+	// notified when new data comes in. We don't strictly need this when not
+	// following, but it doesn't hurt either.
+	notification := make(chan bool, 1)
```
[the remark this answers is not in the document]
benhoyt: The notification here is to "wake up" the main processing loop -- it's only for knowing if there's more data on any one of the iterators (and we check them all). The producer here is `RingBuffer.signalIterators`, but the send to `notifyChan` doesn't block (it has a default case). So if we had no buffer it'd miss notifications, but a buffer size of 1 is all we need. `Iterator.Notify` is currently commented as follows:

```go
	// Notify sets the notification channel. When more data is available, the channel
	// passed in to Notify will have true sent on it. If the channel is not receiving (unbuffered)
	// or full (buffered), the notification will be dropped.
```

Because the buffer size is 1 some notifications will be dropped, but that's okay, the main loop will at least be notified once and pick up any more when it checks the iterators. Hope that helps.

internal/daemon/api_logs.go
```diff
@@ -0,0 +1,307 @@
...
+					continue
+				}
+				// Not following, we're done
+				return
+			}
+
+			// Logs are coming faster than we can send them (probably a slow
+			// client), so stop now.
+			if !follow && log.Time.After(requestStarted) {
+				_ = flushFifo()
```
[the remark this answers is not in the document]
benhoyt: Done.

PR #35, opened 2021-05-04
Allow starting services as another user

### LINKED ISSUE #33: Start a service with a custom user
While migrating [mysql-operator](https://github.com/canonical/mysql-operator/pull/44#issuecomment-830166733) to pebble I discover that some services (mysql for instance) refuses to start with root user.

Would be nice if we have the chance of setting a custom user to run the service, like for example:

```python
        layer = {
            "summary": "MySQL layer",
            "description": "Pebble layer configuration for MySQL",
            "services": {
                "mysql": {
                    "override": "merge",
                    "summary": "mysql daemon",
                    "command": "docker-entrypoint.sh mysqld",
                    "command_user": "mysql", 
                    "startup": "enabled",
                    "environment": {
                        "MYSQL_ROOT_PASSWORD": "SuperSecretPassword",
                    },
                },
            },
        }
```

benhoyt:
This is my minimalist approach to allowing us to start services under a different user (a non-`root` user in the case of Juju sidecar charms). I'll add tests if we agree the plan syntax and approach is right.

Fixes #33


internal/overlord/servstate/handlers.go
```diff
@@ -84,6 +86,20 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 	cmd := exec.Command(args[0], args[1:]...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 
+	// Start as another user if "user" specified in plan
+	if service.User != "" {
```
hpidcock, line 90: I think user should be able to take a uid as well. Not all containers images will have a /etc/passwd

I'd follow https://docs.docker.com/engine/reference/run/#user

benhoyt: Good call -- done. I've allowed numeric UID and GID as well. However, I've done them as two separate fields -- seeing this is structured YAML already, I don't think it makes sense to have to parse the `user[:group]` mini-format within a single field; might as well use separate fields (it also fits the files API better where `user` and `group` are two separate fields).

hpidcock:
Also this PR should update README

internal/overlord/servstate/handlers.go
jameinel: I agree that in an API user and group should be separate. I can understand why they weren't for the CLI, but it makes more sense to be precise here.

README.md
```diff
@@ -73,6 +75,10 @@ Some details worth highlighting:
   - There is the `override` field (for now required) which defines whether this 
 entry _overrides_ the previous service of the same name (if any - missing is 
 okay), or merges with it.
+  - The optional `user` field allows starting a service with a different user
```
jameinel, line 78: This text is no longer true, as you have to pass a numeric user -id to 'user-id' not 'user'.

README.md
```diff
@@ -183,6 +189,20 @@ services:
...
+
+        # (Optional) Username for starting service as a different user.
+        user: <username>
+
+        # (Optional) Group ID for starting service as a different user. Takes
+        # precedence over "group" field.
+        group-id: <gid>
+
+        # (Optional) Group name or GID for starting service as a different user.
+        group: <group name>
```
jameinel, line 205: I wonder if we just treat it as an error to set both group-id and group and have them not match in the target container.

internal/daemon/api_files_test.go
```diff
@@ -854,31 +851,28 @@ func (s *filesSuite) TestWriteUserGroupMocked(c *C) {
...
+		c.Check(username, Equals, "USER")
+		c.Check(group, Equals, "GROUP")
+		u, g := 56, 78
+		return &u, &g, nil
 	}
 	defer func() {
 		atomicWriteChown = osutil.AtomicWriteChown
-		lookupUser = user.Lookup
-		lookupGroup = user.LookupGroup
+		normalizeUidGid = osutil.NormalizeUidGid
```
jameinel, line 868: FWIW, I feel like this has actually turned into a bit of an anti-pattern in the Juju code base. These become brittle, and messy with race conditions. It is global state that you are mutating, etc. I much prefer having Objects that you can inject new dependencies into.

internal/osutil/user.go
```diff
@@ -75,3 +79,40 @@ func UidGid(u *user.User) (sys.UserID, sys.GroupID, error) {
...
+func NormalizeUidGid(uid, gid *int, username, group string) (*int, *int, error) {
+	if uid == nil && username == "" && gid == nil && group == "" {
+		return nil, nil, nil
+	}
+	if uid == nil && username != "" {
+		u, err := userLookup(username)
+		if err != nil {
+			return nil, nil, err
+		}
+		n, _ := strconv.Atoi(u.Uid)
```
jameinel, line 95: from my earlier comment, this could become:
```
if username != "" {
  u, err := userLookup(username)
  if err != nil {
...
  }
  n, _ := strconv.Atoi(u.Uid)
  if uid != nil && *uid != n {
    return nil, nil, Error("user and user-id supplied, but they don't map to the same user %s => %d != %d")
  }

```

Thoughts?

internal/osutil/user.go
```diff
@@ -75,3 +79,40 @@ func UidGid(u *user.User) (sys.UserID, sys.GroupID, error) {
...
+			return nil, nil, err
+		}
+		n, _ := strconv.Atoi(g.Gid)
+		gid = &n
+	}
+	if uid == nil && gid != nil {
+		return nil, nil, fmt.Errorf("must specify user, not just group")
+	}
+	if uid != nil && gid == nil {
+		return nil, nil, fmt.Errorf("must specify group, not just UID")
```
jameinel, line 115: It seems surprising that you could specify just 'user' and we will find the group id, but if you specify uid, we don't find the group.
Can you look up the user information by UID?

internal/overlord/servstate/manager_test.go
```diff
@@ -203,6 +212,45 @@ func (s *S) TestStartBadCommand(c *C) {
...
+	c.Check(err, IsNil)
+	chg := s.st.NewChange("test", "Start test")
+	chg.AddAll(ts)
+	s.st.Unlock()
+
+	s.ensure(c, 1)
+
+	s.st.Lock()
+	c.Check(chg.Status(), Equals, state.ErrorStatus)
+	c.Check(chg.Err(), ErrorMatches, `.*\n.*cannot start service: .* operation not permitted.*`)
```
jameinel, line 237: Does this test fail if run as root? We should probably at least skip it then?

jameinel (approved):

README.md
```diff
@@ -38,47 +38,47 @@ description: |
...
-    override: replace
-    startup: enabled
-    command: cmd
-    before:
-      - srv3
-
-  srv3:
-    override: replace
-    command: cmd
+    srv1:
```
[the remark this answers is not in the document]
benhoyt: I didn't actually change anything here - this was a merge from master. See https://github.com/canonical/pebble/pull/39

internal/daemon/api_files_test.go
benhoyt: I think I'm going to leave this as is. It just doesn't seem great for the caller to have to change a simple global function call like `osutil.NormalizeUidGid()` into this:

```go
utils := osutil.OSUtil{LookupUser: user.Lookup, LookupGroup: user.LookupGroup}
utils.NormalizeUidGid()

# or, if a nil default means "use the real OS function" you could do:
utils := osutil.OSUtil{}
utils.NormalizeUidGid()
```

Seeing these are global / singleton functions I'm okay with having globals for their dependencies. Go runs package-level tests serially by default, so race conditions aren't an issue here.

We can always revisit this later.

README.md
benhoyt: I think that behavior is probably a better choice -- however, this function is already used in the files API to normalize the user / user-id combination (which is in the wild already) so strictly speaking we can't change this behavior. We could make this usage of user/user-id behave differently, but that seems weird. So it's not perfect, but precedence does seem like a fairly common and easy-to-understand concept, so not too bad either.

README.md
benhoyt: Updated.

internal/osutil/user.go
benhoyt: See above

internal/osutil/user.go
benhoyt: Yeah, it's slightly odd. The rationale is I don't want to do the lookup by UID (which you can do using `user.LookupId`) because these lookups require `/etc/passwd` to be present, which may not always be the case in a container. So if you specify IDs I want to avoid the lookup.

internal/overlord/servstate/manager_test.go
benhoyt: Good call. I added code to skip it if root.

README.md
```diff
@@ -183,6 +188,20 @@ services:
         # that should be set in the context of the process
         environment:
             <env var name>: <env var value>
+
+        # (Optional) User ID for starting service as a different user. Takes
+        # precedence over "user" field.
+        user-id: <uid>
+
+        # (Optional) Username for starting service as a different user.
+        user: <username>
```
jnsgruk, line 197: Can we add some text in the comment that describes the behaviour if the specified user doesn't exist?

jnsgruk (approved):
Looks great, thanks @benhoyt :)

README.md
niemeyer: The behavior I remember discussing in the context of the operator was precisely to conflict if both are provided until we had a clear idea, but maybe my memory fails. It seems awkward for the ID to have precedence instead of the name, because the name is more indicative of what the user is. Let's just ponder for a moment: imagine we have user=postgres/uid=400, and 400 is serial-port. Does it make sense to make the postgres files owned by serial-port? Feels like it doesn't. If there is a precedence, it's probably the name that makes more sense than the ID. I'd prefer to fix that now than to make the weird behavior permanent.

internal/daemon/api_files_test.go
niemeyer: Agree with Ben here, and we also need to keep in mind that this code base is inherited from snapd, not juju. There's a pattern well established already, so we should either respect it or change it all at once at some point, if we ever find a reason to.

internal/osutil/user.go
niemeyer: Per earlier rationale, let's please revisit this and fix it so we change both this and the API to conform to username precedence, or to raise an error until we have a plan we're all happy with in place.

niemeyer (changes requested):
This looks good. Only concern we must solve is the precedence/conflict situation.

README.md
benhoyt: Added sentence about that under "user" and "group" field.

README.md
benhoyt: Fair enough - I've made this change now (and this function is both used here and for the files API). As far as backwards compatibility goes, it seems *very* unlikely with the number of charms we have so far -- and even fewer use the files API -- that someone would be specifying both in a conflicting way. So on reflection I do think it's fine to make this minor breaking change.

jameinel: I'm happy with a "these differ, therefore error"

niemeyer: Thanks, looks good.

internal/osutil/user.go
niemeyer: Interesting trade-off indeed. I can't think of a strong reason to go one way or the other right now, and I see Ben's motivation as reasonable. I suggest going forward with Ben's approach for now and observing how people use it. Once we depend on passwd, we cannot undo it.

internal/osutil/user.go
```diff
@@ -75,3 +79,48 @@ func UidGid(u *user.User) (sys.UserID, sys.GroupID, error) {
 
 	return sys.UserID(uid), sys.GroupID(gid), nil
 }
+
+// NormalizeUidGid returns the "normalized" UID and GID for the given IDs and
+// names. If both uid and username are specified, the username's UID must match
+// the given uid (similar for gid and group), otherwise an error is returned.
+func NormalizeUidGid(uid, gid *int, username, group string) (*int, *int, error) {
```
niemeyer, line 86: The API here is a bit atypical/awkward. I'd be worried if this was public, but since it's an "internal/" package, sounds okay and we can tune it later if we ever find the need.

niemeyer (approved):
LGTM, thanks!

A minor point FYI only, no need to do anything about it.

PR #36, opened 2021-05-05
Return custom error for plan format issue; have API return 400, not 500

benhoyt:
This addresses part of https://github.com/canonical/operator/issues/514 -- currently `POST /v1/layers` returns a generic HTTP 500 error if the layer format is incorrect, whereas it should be 400 bad request for better error handling on the client side. So add a custom error type for this case in the plan package, and handle it and return 400 in the API handler.


hpidcock (approved):
LGTM.

PR #38, opened 2021-05-07
Add layer spec

jnsgruk:
Have had a few requests for a comprehensive spec of Pebble layers. This PR attempts to add that.

Some minor re-organisation of the README, and the inclusion of a simple table of contents, as it's getting a little long.


README.md
```diff
@@ -121,31 +127,68 @@ And start or stop a specific service with:
...
+In that folder you can also find a nice looking html file reflecting the contents of the file above.
+The html file can be generated with: `npx redoc-cli bundle doc/api.yaml --output doc/index.html`
+
+## Layer specification
+
+### Type: Layer
+
+| Field         |         Type         | Default | Required? | Description                                                                                     |
+| :------------ | :------------------: | :-----: | :-------: | :---------------------------------------------------------------------------------------------- |
+| `summary`     |       `string`       |  `nil`  |    no     | Short one line summary of the layer                                                             |
```
benhoyt, line 142: Note that for most of these strings the default is not actually `nil`, but empty string, but here has the same meaning: "not set".

README.md
```diff
@@ -121,31 +127,68 @@ And start or stop a specific service with:
     $ pebble start <name1> [<name2> ...]
     $ pebble stop  <name1> [<name2> ...]
 
+## API
+
+The documentation for the currently exposed API endpoints can be found in `doc/api.yaml`
```
benhoyt, line 132: Hmm, I haven't been updating this, so it's already out of date, and it requires Node/npx to update. I wonder if we should just drop it for now. Maybe we can add again later once most of the Pebble API and features are in place. In any case, it probably won't be used by anyone except Pebble devs, in which case we can just say "refer to the code". :-) Everyone else will use the Pebble Python client in Operator Framework, or the Pebble Go client or CLI.

benhoyt (approved):

README.md
jnsgruk: I don't disagree, will remove for now.

README.md
jnsgruk: Yeh I debated how to represent this - perhaps I'll go to putting `""` in for correctness.

jnsgruk:
Addressed comments from @benhoyt, removed the `redoc` API docs for now.

benhoyt (approved):

doc/api.yaml
```diff
@@ -1,261 +0,0 @@
-openapi: 3.0.2
```
niemeyer, line 1: Thanks for cleaning this up.

niemeyer (commented):
Thanks for the additional documentation, @jnsgruk.

Since Ben already merged, any change will need to be on a follow up.

A few comments:

README.md
```diff
@@ -18,57 +23,58 @@ for whole-system setup such as when using Pebble to control services in a contai
...
-the order of the layer and the following label uniquely identifies it.  Each
+the order of the layer and the following label uniquely identifies it. Each
 layer in the stack sits above the former one, and has the chance to improve or
 redefine the service configuration as desired.
 
-## Layer configuration
+## Layer configuration examples
 
-This is a complete example of the current configuration format:
+This is a complete example of the current configuration format. You can also see the [layer specification](#layer-specification):
```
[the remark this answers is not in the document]
jnsgruk: Good call, will address in a new PR

README.md
```diff
@@ -18,57 +23,58 @@ for whole-system setup such as when using Pebble to control services in a contai
...
-        command: cmd
-        before:
-            - srv3
-
-    srv3:
-        override: replace
-        command: cmd
+  srv1:
+    override: replace
+    summary: Service summary
```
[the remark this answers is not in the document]
niemeyer: Thank you, in that case the other best practice is to self-review before submitting, to ensure the diff is intended. Frequently we leave behind things that shouldn't be there. That's more frequent with code, but it seems to have applied here too.

PR #51, opened 2021-06-28
Avoid panics in layer parsing/combining

### LINKED ISSUE #50: Pebble's not robust to wrong input
If pebble is fed a layer which has a service with no `command` specified, it will panic upon starting that service. The daemon will instead crash immediately upon start if the service's dictionary is empty.

### Steps to reproduce

```shell
git clone https://github.com/canonical/pebble.git
cd pebble
make build
export PEBBLE=`pwd`/pebble
mkdir $PEBBLE/layers
cat >$PEBBLE/layers/001-config.yaml <<EOF
services:
    my-service:
        override: replace
EOF
# NOTE no command on the my-service service
_build/_/bin/pebble run -v --create-dirs
```

In another terminal:

```shell
_build/_/bin/pebble start my-service
```

That command produces the following error in the terminal were the pebble daemon is running:

```
panic: runtime error: slice bounds out of range [1:0]

goroutine 52 [running]:
github.com/canonical/pebble/internal/overlord/servstate.(*ServiceManager).doStart(0xc00014c1e0, 0xc00014b320, 0xc00021c0f0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/overlord/servstate/handlers.go:89 +0x177d
github.com/canonical/pebble/internal/overlord/state.(*TaskRunner).run.func1(0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/overlord/state/taskrunner.go:195 +0xca
gopkg.in/tomb%2ev2.(*Tomb).run(0xc00021c0f0, 0xc000212300)
	/home/facu/go/pkg/mod/gopkg.in/tomb.v2@v2.0.0-20161208151619-d5d1b5820637/tomb.go:163 +0x3c
created by gopkg.in/tomb%2ev2.(*Tomb).Go
	/home/facu/go/pkg/mod/gopkg.in/tomb.v2@v2.0.0-20161208151619-d5d1b5820637/tomb.go:159 +0x112
```

If the service's dictionary is empty:

```
cat >$PEBBLE/layers/001-config.yaml <<EOF
services:
    my-service: ~
EOF
_build/_/bin/pebble run -v --create-dirs
```

That'll produce:

```
2021-06-23T10:21:00.044Z [pebble] Started daemon.
2021/06/23 12:21:00 http: panic serving pid=2083067;uid=1000;socket=/home/facu/canonical/sidecar-charm/pebble/pebble/.pebble.socket;: runtime error: invalid memory address or nil pointer dereference
goroutine 50 [running]:
net/http.(*conn).serve.func1(0xc000208000)
	/usr/lib/go-1.13/src/net/http/server.go:1767 +0x149
panic(0xa032a0, 0xe21550)
	/usr/lib/go-1.13/src/runtime/panic.go:679 +0x1e0
github.com/canonical/pebble/internal/plan.ParseLayer(0x1, 0xc00020208c, 0x6, 0xc000246000, 0x1c, 0x21c, 0x0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/plan/plan.go:279 +0x697
github.com/canonical/pebble/internal/plan.ReadLayersDir(0xc0001fe100, 0x37, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/plan/plan.go:340 +0xe5d
github.com/canonical/pebble/internal/plan.ReadDir(0xc000028087, 0x30, 0x0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/plan/plan.go:362 +0x27d
github.com/canonical/pebble/internal/overlord/servstate.(*ServiceManager).reloadPlan(0xc000122140, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/overlord/servstate/manager.go:60 +0x64
github.com/canonical/pebble/internal/overlord/servstate.(*ServiceManager).acquirePlan(0xc000122140, 0x0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/overlord/servstate/manager.go:174 +0x11e
github.com/canonical/pebble/internal/overlord/servstate.(*ServiceManager).DefaultServiceNames(0xc000122140, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/overlord/servstate/manager.go:258 +0x7c
github.com/canonical/pebble/internal/daemon.v1PostServices(0xe28520, 0xc00021c200, 0x0, 0x0, 0x0)
	/home/facu/canonical/sidecar-charm/pebble/internal/daemon/api_services.go:73 +0x5f3
github.com/canonical/pebble/internal/daemon.(*Command).ServeHTTP(0xe28520, 0xb2a740, 0xc0002000c0, 0xc00021c200)
	/home/facu/canonical/sidecar-charm/pebble/internal/daemon/daemon.go:248 +0x547
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0001a2000, 0xb2a740, 0xc0002000c0, 0xc00021c200)
	/home/facu/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210 +0x1e6
github.com/canonical/pebble/internal/daemon.logit.func1(0xb2ab00, 0xc000222000, 0xc00021c000)
	/home/facu/canonical/sidecar-charm/pebble/internal/daemon/daemon.go:308 +0x143
net/http.HandlerFunc.ServeHTTP(0xc000127200, 0xb2ab00, 0xc000222000, 0xc00021c000)
	/usr/lib/go-1.13/src/net/http/server.go:2007 +0x44
net/http.serverHandler.ServeHTTP(0xc0001b4000, 0xb2ab00, 0xc000222000, 0xc00021c000)
	/usr/lib/go-1.13/src/net/http/server.go:2802 +0x20f
net/http.(*conn).serve(0xc000208000, 0xb2bec0, 0xc0001f4040)
	/usr/lib/go-1.13/src/net/http/server.go:1890 +0x1716
created by net/http.(*Server).Serve
	/usr/lib/go-1.13/src/net/http/server.go:2928 +0x931
```

### Expected behavior

Pebble daemon shouldn't crash, and ideally should return to the client/user some indication that the missing `command` is an issue or that the service's dictionary is invalid.

Also, pebble's README.md lists `command` as an optional field, which apparently it's not.

benhoyt:
* Fix panic when command is not set in combined layer
* Fix panic when service object/dict is null
* Also disallow use of empty string as service name

Fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,9 +154,9 @@ services:
         # override the existing service spec in the plan with the same name.
         override: merge | replace
 
-        # (Optional) The command to run the service. This is optional in an
-        # individual layer, but required in the combined layer. The command is
-        # executed directly; use "/bin/sh -c '...'" to run via the shell.
+        # (Required in combined layer) The command to run the service. The
+        # command is executed directly; use "/bin/sh -c '...'" to run via the
+        # shell.
         #
         # Example: /usr/bin/somecommand -b -t 30
         command: <commmand>
--- a/README.md
+++ b/README.md
@@ -151,8 +151,15 @@ services:
         #
         # The value 'merge' will ensure that values in this layer specification
         # are merged over existing definitions, whereas 'replace' will entirely
-        # override the existing service spec in the plan with the same name
+        # override the existing service spec in the plan with the same name.
         override: merge | replace
+
+        # (Optional) The command to run the service. This is optional in an
+        # individual layer, but required in the combined layer. The command is
+        # executed directly; use "/bin/sh -c '...'" to run via the shell.
+        #
+        # Example: /usr/bin/somecommand -b -t 30
+        command: <commmand>
 
         # (Optional) A short summary of the service
         summary: <summary>
@@ -160,10 +167,6 @@ services:
         # (Optional) A detailed description of the service
         description: |
             <description>
-
-        # (Optional) The command to run the service.
-        # Example: /usr/bin/somecommand -b -t 30
-        command: <commmand>
 
         # (Optional) Control whether the service is started automatically when
         # Pebble starts. 
--- a/internal/plan/plan.go
+++ b/internal/plan/plan.go
@@ -83,7 +83,7 @@ func (e *FormatError) Error() string {
 	return e.Message
 }
 
-// CombineLayers combines the given layers into a plan, with the later layers
+// CombineLayers combines the given layers into a single layer, with the later
 // layers overriding earlier ones.
 func CombineLayers(layers ...*Layer) (*Layer, error) {
 	combined := &Layer{
--- a/internal/plan/plan.go
+++ b/internal/plan/plan.go
@@ -150,13 +150,16 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 	}
 
 	// Ensure fields in combined layers validate correctly.
-	err := combined.validateFields()
-	if err != nil {
-		return nil, err
+	for name, service := range combined.Services {
+		if service.Command == "" {
+			return nil, &FormatError{
+				Message: fmt.Sprintf(`plan must define "command" for service %q`, name),
+			}
+		}
 	}
 
 	// Ensure combined layers don't have cycles.
-	err = combined.checkCycles()
+	err := combined.checkCycles()
 	if err != nil {
 		return nil, err
 	}
@@ -254,17 +257,6 @@ func order(services map[string]*Service, names []string, stop bool) ([]string, e
 	return order, nil
 }
 
-func (l *Layer) validateFields() error {
-	for name, service := range l.Services {
-		if service.Command == "" {
-			return &FormatError{
-				Message: fmt.Sprintf(`plan must define "command" for service %q`, name),
-			}
-		}
-	}
-	return nil
-}
-
 func (l *Layer) checkCycles() error {
 	var names []string
 	for name := range l.Services {
@@ -287,6 +279,11 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 	layer.Order = order
 	layer.Label = label
 	for name, service := range layer.Services {
+		if name == "" {
+			return nil, &FormatError{
+				Message: fmt.Sprintf("cannot use empty string as service name"),
+			}
+		}
 		if name == "pebble" {
 			// Disallow service name "pebble" to avoid ambiguity (for example,
 			// in log output).
--- a/internal/plan/plan.go
+++ b/internal/plan/plan.go
@@ -83,7 +83,7 @@ func (e *FormatError) Error() string {
 	return e.Message
 }
 
-// CombineLayers combines the given layers into a Plan, with the later layers
+// CombineLayers combines the given layers into a plan, with the later layers
 // layers overriding earlier ones.
 func CombineLayers(layers ...*Layer) (*Layer, error) {
 	combined := &Layer{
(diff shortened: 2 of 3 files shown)


benhoyt:
One thing I noticed while testing this is that if there are format errors in the "static" layers from the file system, this is caught only after startup, and `pebble run` starts successfully. This is not a problem for dynamically-added layers, just for the layers loaded right after startup from the layers directory. I think this should probably be hard fail (i.e., it shouldn't start at all).

README.md
```diff
@@ -151,20 +151,23 @@ services:
...
-        # override the existing service spec in the plan with the same name
+        # override the existing service spec in the plan with the same name.
         override: merge | replace
 
+        # (Optional) The command to run the service. This is optional in an
+        # individual layer, but required in the combined layer. The command is
+        # executed directly; use "/bin/sh -c '...'" to run via the shell.
+        #
+        # Example: /usr/bin/somecommand -b -t 30
+        command: <commmand>
```
hpidcock, line 162: Why did this move?

internal/plan/plan.go
```diff
@@ -83,7 +83,7 @@ func (e *FormatError) Error() string {
 	return e.Message
 }
 
-// CombineLayers combines the given layers into a Plan, with the later layers
+// CombineLayers combines the given layers into a plan, with the later layers
```
hpidcock, line 86: `with the later layers layers`

Maybe consider revising?

internal/plan/plan.go
```diff
@@ -247,6 +254,17 @@ func order(services map[string]*Service, names []string, stop bool) ([]string, e
 	return order, nil
 }
 
+func (l *Layer) validateFields() error {
+	for name, service := range l.Services {
+		if service.Command == "" {
```
hpidcock, line 259: Do we need to add anything else to this list?

hpidcock (approved):

README.md
benhoyt: *It* didn't move, I moved it. ;-) I thought it made more sense near the top, right after "override", as it's really the most important field after "override". (More so than summary/description, which are totally optional.)

internal/plan/plan.go
benhoyt: Thanks, updated.

internal/plan/plan.go
benhoyt: No, I don't think so -- I just double-checked, and everything else is optional. "override" was required and already checked (in `CombineLayers`), and command is the only other required field (well, optional in individual layers, required in combined).

internal/plan/plan.go
```diff
@@ -276,6 +294,11 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 				Message: fmt.Sprintf("cannot use reserved service name %q", name),
 			}
 		}
+		if service == nil {
+			return nil, &FormatError{
+				Message: fmt.Sprintf("service object cannot be null for service %q", name),
```
jameinel, line 299: This would also be true if 'service == ""' wouldn't it?
Would it be better to do something like:

`if service == nil || service == "" `
?

jameinel (approved):
I think clarifying the language a bit and then also checking for "" as well as 'nil' is useful.

internal/plan/plan.go
benhoyt: `service` is a `Service` struct, so comparing it to `""` doesn't make sense. But I've added a check for the service name being empty string above as well.

benhoyt:
> So I think the documentation saying that command is (Optional) is more confusing than helpful. I realize that if you are adding a layer to another layer, it wouldn't have to change the command, but surely you should have a command in the final rendered layer.
>
> Is there a clearer syntax for 'required in the final plan, but optional for a given layer'?

Yeah, you're not wrong, it was weird. I've clarified to the more verbose but more helpful/accurate "(Required in combined layer)".

> And in your Validate call, you do, in fact, make it required to have a command.

Ah true -- `validateFields` was a layer method, but called from `CombineLayers`. That's kinda weird, so I just moved the logic into `CombineLayers`.

niemeyer (commented):
A bit late, but nice fix, thanks!

PR #54, opened 2021-07-09
add building section to readme

stulluk:
this commit adds a building section to readme.
todo: we need to explain cross compiling pebble.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,10 @@ designed with unique features that help with more specific use cases.
 
   - [General model](#general-model)
   - [Layer configuration examples](#layer-configuration-examples)
+  - [Building Pebble](#building-pebble)
   - [Running pebble](#running-pebble)
   - [Layer specification](#layer-specification)
+
   - [TODO/Contributing](#todo-contributing)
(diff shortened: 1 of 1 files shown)


niemeyer:
Hi Sertac,

Thanks for the PR.  I think we may need something like this eventually, but this feels somewhat lightweight for now, in the sense that if people don't know that they need to use "make" with a "Makefile", there will be many other reasons why this won't be very useful.  I suggest holding it off just a bit and introducing a building section when we are solving more non-trivial questions there.

I'm closing this for now, just so we can get off the queue, but happy to chat next week if you'd like to bring this back alive.

benhoyt:
Huh, I was just looking at this, @niemeyer, and chatting with Burak last night about the Makefile -- he said there were a couple of things about it that didn't work. It hasn't been maintained, and I suggest we get rid of it and just show in the README or wherever how to use `go build` / `go run` for local development. I'm going to put up a PR that updates the README and adds a HACKING.md to this effect (and probably deletes the Makefile at the same time).

PR #58, opened 2021-07-14
Add restart cmd + auto/start, stop idempotency.

### LINKED ISSUE #47: autostart fails if service has already started
This happens when pebble_ready is deferred, but another event (re)starts the service. When pebble_ready re-enters, I get this error:
```
  File "./src/charm.py", line 443, in <module>
    main(AlertmanagerCharm)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/main.py", line 406, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/main.py", line 140, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/framework.py", line 278, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/framework.py", line 722, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/framework.py", line 767, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 67, in wrapped
    func(self, event, *args, **kwargs)
  File "./src/charm.py", line 213, in _on_alertmanager_pebble_ready
    container.autostart()
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/model.py", line 1042, in autostart
    self._pebble.autostart_services()
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/pebble.py", line 791, in autostart_services
    return self._services_action('autostart', [], timeout, delay)
  File "/var/lib/juju/agents/unit-alertmanager-0/charm/venv/ops/pebble.py", line 831, in _services_action
    raise ChangeError(change.err, change)
ops.pebble.ChangeError: cannot perform the following tasks:
- Start service "alertmanager" (service "alertmanager" was previously started)
```
This is expected when trying to start the same service twice, but with autostart I expected this to work. Is this a bug or by design?

hpidcock:
This PR implements the `restart` command for (re)starting services.

Additionally
- autostart, start and stop are now idempotent.
- variable expansion on the command for a service now takes place

Resolves #47

### DIFF
--- a/cmd/pebble/cmd_restart.go
+++ b/cmd/pebble/cmd_restart.go
@@ -1,0 +1,59 @@
+// Copyright (c) 2014-2021 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"github.com/jessevdk/go-flags"
+
+	"github.com/canonical/pebble/client"
+)
+
+var shortRestartHelp = "Restart a service"
+var longRestartHelp = `
+The restart command restarts the named service(s) in the correct order.
+`
+
+type cmdRestart struct {
+	waitMixin
+	Positional struct {
+		Services []string `positional-arg-name:"<service>" required:"1"`
+	} `positional-args:"yes"`
+}
+
+func init() {
+	addCommand("restart", shortRestartHelp, longRestartHelp, func() flags.Commander { return &cmdRestart{} }, nil, nil)
+}
+
+func (cmd cmdRestart) Execute(args []string) error {
+	if len(args) > 1 {
+		return ErrExtraArgs
+	}
+
+	servopts := client.ServiceOptions{
+		Names: cmd.Positional.Services,
+	}
+	changeID, err := cmd.client.Restart(&servopts)
+	if err != nil {
+		return err
+	}
+
+	if _, err := cmd.wait(changeID); err != nil {
+		if err == noWait {
+			return nil
+		}
+		return err
+	}
+	return nil
+}
--- a/internal/daemon/api_services.go
+++ b/internal/daemon/api_services.go
@@ -101,16 +101,17 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 		}
 		taskSet, err = servstate.Start(st, services)
 	case "stop":
-		services, err = servmgr.StopOrder(payload.Services, true)
+		services, err = servmgr.StopOrder(payload.Services)
 		if err != nil {
 			break
 		}
 		taskSet, err = servstate.Stop(st, services)
 	case "restart":
-		services, err = servmgr.StopOrder(payload.Services, false)
+		services, err = servmgr.StopOrder(payload.Services)
 		if err != nil {
 			break
 		}
+		services = intersectOrdered(payload.Services, services)
 		var stopTasks *state.TaskSet
 		stopTasks, err = servstate.Stop(st, services)
 		if err != nil {
@@ -157,4 +158,20 @@ func v1GetService(c *Command, r *http.Request, _ *userState) Response {
 
 func v1PostService(c *Command, r *http.Request, _ *userState) Response {
 	return statusBadRequest("not implemented")
+}
+
+// intersectOrdered returns the intersection of left and right where
+// the right's ordering is persisted in the resulting set.
+func intersectOrdered(left []string, orderedRight []string) []string {
+	m := map[string]bool{}
+	for _, v := range left {
+		m[v] = true
+	}
+	var out []string
+	for _, v := range orderedRight {
+		if m[v] {
+			out = append(out, v)
+		}
+	}
+	return out
 }
--- a/internal/daemon/api_services.go
+++ b/internal/daemon/api_services.go
@@ -64,13 +64,12 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 	}
 
 	var err error
-	var services []string
 	servmgr := c.d.overlord.ServiceManager()
 	if payload.Action == "autostart" {
 		if len(payload.Services) != 0 {
 			return statusBadRequest("%s accepts no service names", payload.Action)
 		}
-		services, err = servmgr.DefaultServiceNames()
+		services, err := servmgr.DefaultServiceNames()
 		if err != nil {
 			return statusInternalError("%v", err)
 		}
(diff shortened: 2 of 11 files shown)


cmd/pebble/cmd_restart.go
```diff
@@ -0,0 +1,59 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"github.com/canonical/pebble/client"
+	"github.com/jessevdk/go-flags"
```
benhoyt, line 19: Nit: like Juju, we've been separating imports into three groups: stdlib, 3rd party (jessevdk/go-flags), and Pebble (canonical/pebble/client).

internal/overlord/servstate/handlers.go
```diff
@@ -77,17 +78,38 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+	envMap := make(map[string]string)
+	for _, envLine := range env {
+		parts := strings.SplitN(envLine, "=", 2)
+		if len(parts) == 2 {
+			envMap[parts[0]] = parts[1]
+		}
 	}
 
-	args, err := shlex.Split(service.Command)
+	commandString := os.Expand(service.Command, func(k string) string {
```
benhoyt, line 100: I think this is a good change, but it doesn't look like it's documented (in the `README`) or tested anywhere. Can we add those?

Nit: what about calling this `commandExpanded`?

internal/overlord/servstate/handlers.go
```diff
@@ -134,6 +150,13 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 	go func() {
 		active.err = cmd.Wait()
 		_ = active.logBuffer.Close()
+		releasePlan, err := m.acquirePlan()
+		if err == nil {
+			if m.services[req.Name].cmd == cmd {
```
benhoyt, line 155: What does this `if` statement check for? (I realize it was there already below.)

benhoyt (approved):
Approve (with the addition of README/tests for the `$var` expanding in command).

internal/overlord/servstate/handlers.go
hpidcock: Oh yeah I forgot I had made this change, I'll make sure to add some tests and documentation.

internal/overlord/servstate/handlers.go
hpidcock: acquirePlan wouldn't actually fail here, since the plan is loaded already. I guess I can make this more future proof by relying on the mutex directly incase acquirePlan implementation changes.

README.md
```diff
@@ -156,9 +156,9 @@ services:
 
         # (Required in combined layer) The command to run the service. The
         # command is executed directly; use "/bin/sh -c '...'" to run via the
-        # shell.
+        # shell. All environment variables in the $X or ${X} form are expanded.
         #
-        # Example: /usr/bin/somecommand -b -t 30
+        # Example: /usr/bin/somecommand -b -t 30 $ENV_VAR
```
niemeyer, line 161: Left overs from the other PR?

cmd/pebble/cmd_restart.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/jessevdk/go-flags"
+
+	"github.com/canonical/pebble/client"
+)
+
+var shortRestartHelp = "Restart a service and its dependencies"
+var longRestartHelp = `
+The restart command restarts the service with the provided name and
+any other services it depends on, in the correct order.
```
niemeyer, line 26: There's an open point on the other PR about this.  I seem to have reviewed this one together with that other one for not realizing this was independently available. Other than this, the PR looks simple and nice. :+1:

niemeyer (changes requested):

internal/daemon/api_services.go
```diff
@@ -86,29 +85,50 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
...
+		}
+		var startTasks *state.TaskSet
+		startTasks, err = servstate.Start(st, services)
+		if err != nil {
+			break
+		}
+		startTasks.WaitAll(stopTasks)
+		taskSet = state.NewTaskSet()
+		taskSet.AddAll(stopTasks)
+		taskSet.AddAll(startTasks)
```
niemeyer, line 131: It's nice that both the stop order and the start order is correct here.

internal/overlord/servstate/handlers.go
```diff
@@ -183,7 +187,8 @@ func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
 
 	active, ok := m.services[req.Name]
 	if !ok {
-		return fmt.Errorf("service %q is not active", req.Name)
+		// Already stopped
```
niemeyer, line 190: Should we log something in the task itself (goes into the console), so it is at least known that this was a no-op?

internal/plan/tarjan.go
```diff
@@ -38,8 +38,15 @@ func tarjanSort(successors map[string][]string) [][]string {
 		index:      make(map[string]int, len(successors)),
 	}
 
+	// Stabilize iteration through successors map to prevent
+	// disjointed components producing unstable output due to
+	// golang map randomized iteration.
+	stableIDs := make([]string, 0, len(successors))
 	for id := range successors {
-		id := string(string(id))
```
niemeyer, line 42: That's a funny left over from an automatic replacement. The id used to be an ObjectId type.

niemeyer (approved):
Thank you. Just a few trivial remarks, and LGTM when you're happy.

niemeyer:
Btw, please keep an eye on the commit message. The description of this PR looks out of date, referring to the env change which was split out.

internal/plan/plan.go
```diff
@@ -172,21 +172,21 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+func (p *Plan) StopOrder(names []string, includeDependants bool) ([]string, error) {
+	return order(p.Services, names, true, includeDependants)
 }
 
-func order(services map[string]*Service, names []string, stop bool) ([]string, error) {
+func order(services map[string]*Service, names []string, stop bool, includeRelated bool) ([]string, error) {
 	// For stop, create a list of reversed dependencies.
 	predecessors := map[string][]string(nil)
-	if stop {
+	if stop && includeRelated {
```
niemeyer, line 189: While reviewing the follow up PR, I started to suspect that this logic can't be right, as it's blocking ordering knowledge from the algorithm altogether. We need to feed the hierarchy into the dependency algorithm even if we're not going to be using all of it, because we don't know which paths are present in the graph. In other words, even if we only care about A and D, we need to have A, B, C, and D into the dependency logic, so that we can find the path betweem A and D and order them. As such, I suggest leaving StartOrder and StopOrder alone, and just having a helper function that when given two lists L1 and L2, will just reorder L1 according to the order of these elements L2.

niemeyer (changes requested):
Oops, sorry, one additional comment that probably needs looking into:

internal/overlord/servstate/handlers.go
```diff
@@ -75,9 +75,17 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
-	if previous {
-		return fmt.Errorf("service %q was previously started", req.Name)
+	if service, previous := m.services[req.Name]; previous {
+		select {
+		case <-service.done:
+			// Service exited but not yet removed, safe to override.
+		default:
+			// Already started
+			m.state.Lock()
+			task.Logf("service %q already started", req.Name)
```
niemeyer, line 85: The error message became a log, so we need to capitalize and add punctuation so it fits in the context:
```
"Service %q already started."
```

internal/overlord/servstate/handlers.go
```diff
@@ -183,7 +190,11 @@ func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
 
 	active, ok := m.services[req.Name]
 	if !ok {
-		return fmt.Errorf("service %q is not active", req.Name)
+		// Already stopped
+		m.state.Lock()
+		task.Logf("service %q already stopped", req.Name)
```
niemeyer, line 195: Same here.

niemeyer (approved):
Just a couple of trivials about error messages that became logs, and LGTM! Thank you!

PR #60, opened 2021-07-30
Add Replan operation to restart changed services.

hpidcock:
This change adds the **replan** command for restarting services that have changed or removed (we don't have remove AFAIK yet) since when they were started.

Replan acts much the same as **autostart** with the added restarting of changes.

Based on #58

Compare changes:
https://github.com/canonical/pebble/compare/hpidcock:restart...hpidcock:replan

### DIFF
--- a/client/services.go
+++ b/client/services.go
@@ -46,6 +46,11 @@ func (client *Client) Restart(opts *ServiceOptions) (changeID string, err error)
 	return changeID, err
 }
 
+func (client *Client) Replan(opts *ServiceOptions) (changeID string, err error) {
+	_, changeID, err = client.doMultiServiceAction("replan", opts.Names)
+	return changeID, err
+}
+
 type multiActionData struct {
 	Action   string   `json:"action"`
 	Services []string `json:"services"`
--- a/cmd/pebble/cmd_replan.go
+++ b/cmd/pebble/cmd_replan.go
@@ -1,0 +1,54 @@
+// Copyright (c) 2014-2020 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"github.com/canonical/pebble/client"
+	"github.com/jessevdk/go-flags"
+)
+
+var shortReplanHelp = "Ensure running services match the current plan"
+var longReplanHelp = `
+The replan command starts, stops, or restarts services that have changed,
+so that running services match exactly the desired configuration in the current plan.
+`
+
+type cmdReplan struct {
+	waitMixin
+}
+
+func init() {
+	addCommand("replan", shortReplanHelp, longReplanHelp, func() flags.Commander { return &cmdReplan{} }, nil, nil)
+}
+
+func (cmd cmdReplan) Execute(args []string) error {
+	if len(args) > 1 {
+		return ErrExtraArgs
+	}
+
+	servopts := client.ServiceOptions{}
+	changeID, err := cmd.client.Replan(&servopts)
+	if err != nil {
+		return err
+	}
+
+	if _, err := cmd.wait(changeID); err != nil {
+		if err == noWait {
+			return nil
+		}
+		return err
+	}
+	return nil
+}
--- a/internal/daemon/api_services.go
+++ b/internal/daemon/api_services.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"sort"
 	"strings"
 
 	"github.com/canonical/pebble/internal/overlord/servstate"
@@ -34,7 +35,7 @@ type serviceInfo struct {
 func v1GetServices(c *Command, r *http.Request, _ *userState) Response {
 	names := strutil.CommaSeparatedList(r.URL.Query().Get("names"))
 
-	servmgr := c.d.overlord.ServiceManager()
+	servmgr := overlordServiceManager(c.d.overlord)
 	services, err := servmgr.Services(names)
 	if err != nil {
 		return statusInternalError("%v", err)
@@ -64,8 +65,13 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 	}
 
 	var err error
-	servmgr := c.d.overlord.ServiceManager()
-	if payload.Action == "autostart" {
+	servmgr := overlordServiceManager(c.d.overlord)
+	switch payload.Action {
+	case "replan":
+		if len(payload.Services) != 0 {
+			return statusBadRequest("%s accepts no service names", payload.Action)
+		}
+	case "autostart":
 		if len(payload.Services) != 0 {
 			return statusBadRequest("%s accepts no service names", payload.Action)
 		}
@@ -81,7 +87,7 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 			})
 		}
 		payload.Services = services
-	} else {
+	default:
 		if len(payload.Services) == 0 {
 			return statusBadRequest("no services to %s provided", payload.Action)
 		}
@@ -130,6 +136,40 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 		taskSet = state.NewTaskSet()
 		taskSet.AddAll(stopTasks)
 		taskSet.AddAll(startTasks)
+	case "replan":
+		var stopNames, startNames []string
+		stopNames, startNames, err = servmgr.Replan()
+		if err != nil {
+			break
+		}
+		var stopTasks *state.TaskSet
+		stopTasks, err = servstate.Stop(st, stopNames)
+		if err != nil {
+			break
+		}
+		var startTasks *state.TaskSet
+		startTasks, err = servstate.Start(st, startNames)
+		if err != nil {
+			break
+		}
+		startTasks.WaitAll(stopTasks)
+		taskSet = state.NewTaskSet()
+		taskSet.AddAll(stopTasks)
+		taskSet.AddAll(startTasks)
+
+		// Populate a list of services affected by the replan for summary.
+		replanned := make(map[string]bool)
+		for _, v := range stopNames {
+			replanned[v] = true
+		}
+		for _, v := range startNames {
+			replanned[v] = true
+		}
+		for k := range replanned {
+			services = append(services, k)
+		}
+		sort.Strings(services)
+		payload.Services = services
 	default:
 		return statusBadRequest("action %q is unsupported", payload.Action)
 	}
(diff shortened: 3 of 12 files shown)


internal/overlord/servstate/handlers.go
```diff
@@ -77,17 +78,37 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+	}
+
+	// Pass service description's environment variables to child process
+	env := os.Environ()
+	for k, v := range service.Environment {
+		env = append(env, k+"="+v)
+	}
+
+	// Expand the command string with environment variables.
+	envMap := make(map[string]string)
```
niemeyer, line 92: This looks unrelated to this PR. I'm keen on supporting this ASAP as it introduces an incompatibility, but this is something worth talking about to make sure we get it right.

For example, how should "mycmd "$ASD"` expand?

I suggest taking that out of the PR so we can merge this, and let's discuss this by itself.

internal/overlord/servstate/handlers.go
```diff
@@ -126,14 +141,22 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
 	go func() {
 		active.err = cmd.Wait()
 		_ = active.logBuffer.Close()
+		releasePlan, err := m.acquirePlan()
+		if err == nil {
+			if m.services[req.Name].cmd == cmd {
+				delete(m.services, req.Name)
+			}
+			releasePlan()
+		}
```
niemeyer, line 159: But that change means we are now losing the metadata about the command termination. With the previous logic, we can comfortable look at the metadata for the running state at any point, while with this change it will be gone as soon as the service itself dies, so we in fact can never trust the cmd data to be there anymore because we don't control the process termination. Am I missing something?  If not and there wasn't a clear motivation for this change, I suggest reverting it for the time being.

internal/overlord/servstate/manager.go
```diff
@@ -327,3 +342,52 @@ func (m *ServiceManager) ServiceLogs(services []string, last int) (map[string]se
...
+	releasePlan, err := m.acquirePlan()
+	if err != nil {
+		return nil, nil, err
+	}
+	defer releasePlan()
+
+	needsRestart := make(map[string]bool)
+	stopOrder := []string(nil)
+	for name, service := range m.services {
+		if currentPlan, ok := m.plan.Services[name]; ok &&
```
niemeyer, line 358: currentService?

niemeyer (changes requested):
Looks very nice, thanks!

Some comments, but nothing fundamental.

client/services.go
```diff
@@ -41,6 +41,16 @@ func (client *Client) Stop(opts *ServiceOptions) (changeID string, err error) {
 	return changeID, err
 }
 
+func (client *Client) Restart(opts *ServiceOptions) (changeID string, err error) {
+	_, changeID, err = client.doMultiServiceAction("restart", opts.Names)
+	return changeID, err
+}
+
+func (client *Client) Replan(opts *ServiceOptions) (changeID string, err error) {
+	_, changeID, err = client.doMultiServiceAction("replan", opts.Names)
```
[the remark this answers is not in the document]
benhoyt: I definitely prefer "replan" (though it was weird to me at first!). "Comply" sounds a bit authoritarian :-), plus comply is vague (comply with what?) whereas "replan" links it so nicely to "plan".

internal/overlord/servstate/manager.go
```diff
@@ -12,7 +12,21 @@ import (
 	"github.com/canonical/pebble/internal/servicelog"
 )
 
-type ServiceManager struct {
+type ServiceManager interface {
```
benhoyt, line 15: Curious why you changed this to an interface? Normally in Go packages return concrete types, and interfaces are only created where needed (e.g., in `Overlord` itself, so the `ServiceManager` can be mocked out).

internal/overlord/servstate/manager.go
```diff
@@ -327,3 +342,52 @@ func (m *ServiceManager) ServiceLogs(services []string, last int) (map[string]se
...
+// their plans had changed between when they started and this call.
+func (m *serviceManager) Replan() ([]string, []string, error) {
+	releasePlan, err := m.acquirePlan()
+	if err != nil {
+		return nil, nil, err
+	}
+	defer releasePlan()
+
+	needsRestart := make(map[string]bool)
+	stopOrder := []string(nil)
```
benhoyt, line 356: Nit: in my experience `var stopOrder []string` is more common for the zero (nil) value. Not sure exactly why, but it is (maybe because there's less punctuation?). Same below.

For example, in Juju there are 1371 instances of `var foo []T`, but only 3 of `foo := []T(nil)`. Similar in the Go stdlib.

internal/plan/plan.go
```diff
@@ -57,6 +58,38 @@ type Service struct {
...
+	}
+	return &copy
+}
+
+// Equal returns true when the two services are equal in value.
+func (s *Service) Equal(other *Service) bool {
+	if s == other {
+		return true
+	}
+	// TODO: implement this efficiently.
```
benhoyt, line 89: Probably not a concern for this use case?

cmd/pebble/cmd_replan.go
```diff
@@ -0,0 +1,55 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"github.com/canonical/pebble/client"
+	"github.com/jessevdk/go-flags"
+)
+
+var shortReplanHelp = "Restart running services that have changed and start default services"
```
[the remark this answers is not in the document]
hpidcock: Much better description thank-you

internal/overlord/servstate/handlers.go
hpidcock: This is in #58 so I'll remove it from there and propose a separate PR we can drill into this more with.

cmd/pebble/cmd_restart.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/jessevdk/go-flags"
+
+	"github.com/canonical/pebble/client"
+)
+
+var shortRestartHelp = "Restart a service and its dependencies"
+var longRestartHelp = `
+The restart command restarts the service with the provided name and
+any other services it depends on, in the correct order.
```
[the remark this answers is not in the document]
hpidcock: StopOrder adds in dependents. So therefore the stop command stops dependents.
I think we should be consistent here, should dependents always stop/restart for the replan/stop/restart commands?

I don't mind either way, just is strange behavior. Starts make sense to happen in order with dependencies, to reduce failure on the critical path. Dependents should be fault tolerant and handle failure gracefully, if not the health-check work and auto-restarts should handle that.

internal/overlord/servstate/manager.go
```diff
@@ -25,10 +25,11 @@ type ServiceManager struct {
 }
 
 type activeService struct {
-	cmd       *exec.Cmd
-	err       error
-	done      chan struct{}
-	logBuffer *servicelog.RingBuffer
+	plannedService *plan.Service
```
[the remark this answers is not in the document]
hpidcock: I've gone with originalService, hopefully that is sufficient.

cmd/pebble/cmd_restart.go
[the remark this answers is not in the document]
hpidcock: StopOrder can optionally not add dependants to the returned slice. This fixes the strange behaviour in restart & replan

client/services.go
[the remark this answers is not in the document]
hpidcock: Replan is the cleanest option I think, happy to continue with it.

cmd/pebble/cmd_restart.go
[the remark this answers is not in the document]
niemeyer: This is being sorted on the underlying #58.

internal/overlord/servstate/manager.go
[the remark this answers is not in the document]
niemeyer: Thanks :+1:

internal/plan/plan.go
```diff
@@ -172,21 +205,21 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+func (p *Plan) StopOrder(names []string, includeDependants bool) ([]string, error) {
+	return order(p.Services, names, true, includeDependants)
 }
 
-func order(services map[string]*Service, names []string, stop bool) ([]string, error) {
+func order(services map[string]*Service, names []string, stop bool, includeRelated bool) ([]string, error) {
 	// For stop, create a list of reversed dependencies.
 	predecessors := map[string][]string(nil)
-	if stop {
+	if stop && includeRelated {
```
niemeyer, line 222: While reviewing the follow up PR, I started to suspect that this logic can't be right, as it's blocking ordering knowledge from the algorithm altogether.  We need to feed the hierarchy into the dependency algorithm even if we're not going to be using all of it, because we don't know which paths are present in the graph. In other words, even if we only care about A and D, we need to have A, B, C, and D into the dependency logic, so that we can find the path betweem A and D and order them.  As such, I suggest leaving StartOrder and StopOrder alone, and just having a helper function that when given two lists L1 and L2, will just reorder L1 according to the order of these elements L2.

niemeyer (changes requested):
Oops, sorry, one additional comment that probably needs looking into:

internal/overlord/servstate/handlers.go
```diff
@@ -75,9 +75,14 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 		return fmt.Errorf("cannot find service %q in plan", req.Name)
 	}
 
-	_, previous := m.services[req.Name]
-	if previous {
-		return fmt.Errorf("service %q was previously started", req.Name)
+	if service, previous := m.services[req.Name]; previous {
+		select {
+		case <-service.done:
+			// Service exited but not yet removed, safe to override.
```
niemeyer, line 81: This is only safe because we check for the cmd field before actually removing. We might probably give a hint here so future generations don't fall into the trap of thinking that this is always safe, and we can also be more explicit about what it is that it's safe to do. Something along the lines of:
```
// Service exited but not yet removed. It's safe to replace it
// because the removal will check if the cmd still matches.
```

internal/plan/plan.go
```diff
@@ -172,21 +205,21 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+	return order(p.Services, names, false, true)
 }
 
 // StopOrder returns the required services that must be stopped for the named
 // services to be properly stopped, in the order that they must be stopped.
 // An error is returned when a provided service name does not exist, or there
 // is an order cycle involving the provided service or its dependencies.
-func (p *Plan) StopOrder(names []string) ([]string, error) {
-	return order(p.Services, names, true)
+func (p *Plan) StopOrder(names []string, includeDependants bool) ([]string, error) {
```
niemeyer, line 215: This needs to be looked into per note on #58: I suggest leaving the original logic in the package alone, and just having that sorting function that applies sorting from a list into a different subset list.

niemeyer (changes requested):
Only blocker here is the underlying #58.O

Otherwise, a trivial and LGTM, thanks!

internal/daemon/api_services.go
```diff
@@ -82,33 +87,89 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
...
+			replanned[v] = true
+		}
+		for _, v := range startNames {
+			replanned[v] = true
+		}
+		for k := range replanned {
+			services = append(services, k)
+		}
+		sort.Strings(services)
+		payload.Services = services
```
niemeyer, line 172: This reads very cleanly, thank you.

niemeyer (approved):
LGTM, thanks!

jnsgruk:
Nice work! Can we just make sure we get this covered in the docs over at https://juju.is/docs/sdk/pebble please? :)

PR #61, opened 2021-08-11
One-shot commands (pebble exec)

benhoyt:
This implements the Pebble/Go side of the [one-shot commands spec](https://docs.google.com/document/d/1npbNQMokDUoGX1UT61CvnMq6lIJ6MnU6CwExkymHEcA/edit) for command execution using websocket communication, including the Go client and `pebble exec` CLI.

**Other notes for the code reviewer:**

* I've added a number of tests for the API with use-terminal=false. They're not exactly "unit" tests as they run real commands and use the Go client for connecting to the websockets, but this makes the code and tests significantly simpler, and they run fast.
* There is currently only one trivial test with use-terminal=true, as it's hard to test terminal-specific features in Go tests. Instead, I suggest we focus on manual testing for testing terminal interactivity (see "Terminal resizing" under manual tests below).
* I may add client tests later when we've figured out what shape the extracted LXD websocket proxying library takes. Ideally it would use a websocket-like interface rather than `*websocket.Conn` (which makes things kinda hard to test right now).
* The `ptyutil` or `wsutil` packages were pulled basically verbatim from LXD, so I'm not going to add tests for those. The LXD team will be extracting those to a shared library, at which point we can `go get` them instead of vendoring.

**Manual tests:**

```
# Build and run the Pebble server with something like this:
$ go build ./cmd/pebble
$ PEBBLE_DEBUG=1 PEBBLE=~/pebble ./pebble run --hold

# Then in another terminal you can run "pebble exec"

# Basic stdout
$ PEBBLE=~/pebble ./pebble exec -- echo foo
foo
$ PEBBLE=~/pebble ./pebble exec -- echo foo >out
$ cat out
foo

# Stdin and stdout
$ PEBBLE=~/pebble ./pebble exec -- cat
asdf
asdf
# press Ctrl-D to exit
$ PEBBLE=~/pebble ./pebble exec -- cat
foo
foo
# also try Ctrl-C to exit (signal handling)

# Executable not found
$ PEBBLE=~/pebble ./pebble exec -- foobar
error: exec: "foobar": executable file not found in $PATH

# Command error
$ PEBBLE=~/pebble ./pebble exec -- ls notfound
ls: cannot access 'notfound': No such file or directory
$ echo $?
2
$ PEBBLE=~/pebble ./pebble exec -T -- ls notfound 2>err
$ echo $?
2
$ cat err
ls: cannot access 'notfound': No such file or directory

# Terminal resizing
[...]

PR #63, opened 2021-08-16
Add wait-change API endpoint and client function

benhoyt:
This adds an endpoint to the Pebble API to wait for a change to be ready, with an optional timeout. It's intended for use with the upcoming "exec" feature ([spec](https://docs.google.com/document/d/1npbNQMokDUoGX1UT61CvnMq6lIJ6MnU6CwExkymHEcA/edit#) and [WIP PR](https://github.com/canonical/pebble/pull/61)). The LXD code has a similar feature, except changes are called "operations" in LXD, and there's a "wait operation" API endpoint.

This is particularly important for "pebble exec", so we can wait for the command to be finished without polling every short interval, and we can return a response immediately when the command finishes executing.

In future we could also use this new endpoint in the `pebble start` and `pebble stop` commands (and Python Operator Framework equivalents) to avoid polling there too.

### DIFF
--- a/client/changes.go
+++ b/client/changes.go
@@ -160,15 +160,14 @@ func (client *Client) Changes(opts *ChangesOptions) ([]*Change, error) {
 
 // WaitChangeOptions holds the options for the WaitChange call.
 type WaitChangeOptions struct {
-	// If non-zero, wait at most this long before returning the current change
-	// data. The timeout elapsing is not considered an error, so if a timeout
-	// is specified, the caller should check Change.Ready to determine whether
-	// the change is actually finished.
+	// If nonzero, wait at most this long before returning. If a timeout
+	// occurs, WaitChange will return an error.
 	Timeout time.Duration
 }
 
-// WaitChange waits for a given change to be finished (whether or not there
-// was an error, which is indicated by Change.Err being set).
+// WaitChange waits for the change to be finished. If the wait operation
+// succeeds, the returned Change.Err string will be non-empty if the change
+// itself had an error.
 func (client *Client) WaitChange(id string, opts *WaitChangeOptions) (*Change, error) {
 	var chgd changeAndData
 
--- a/client/changes.go
+++ b/client/changes.go
@@ -157,3 +157,27 @@ func (client *Client) Changes(opts *ChangesOptions) ([]*Change, error) {
 
 	return chgs, err
 }
+
+// WaitChangeOptions holds the options for the WaitChange call.
+type WaitChangeOptions struct {
+	Timeout time.Duration
+}
+
+// WaitChange waits for a given change to be finished (whether or not it had
+// an error). If a timeout is specified, wait at most that long.
+func (client *Client) WaitChange(id string, opts *WaitChangeOptions) (*Change, error) {
+	var chgd changeAndData
+
+	query := url.Values{}
+	if opts != nil && opts.Timeout != 0 {
+		query.Set("timeout", opts.Timeout.String())
+	}
+
+	_, err := client.doSync("GET", "/v1/changes/"+id+"/wait", query, nil, nil, &chgd)
+	if err != nil {
+		return nil, err
+	}
+
+	chgd.Change.data = chgd.Data
+	return &chgd.Change, nil
+}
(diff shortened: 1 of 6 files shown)


internal/daemon/api_changes.go
```diff
@@ -165,18 +165,52 @@ func v1GetChanges(c *Command, r *http.Request, _ *userState) Response {
...
+		}
+		timer := time.NewTimer(timeout)
+		select {
+		case <-change.Ready():
+			timer.Stop() // change ready, release timer resources
+		case <-timer.C:
+		}
+	} else {
+		// No timeout, wait indefinitely for change to be ready.
+		<-change.Ready()
```
hpidcock, line 206: select with wait on r.Context().Done()

internal/daemon/api_changes.go
```diff
@@ -165,18 +165,52 @@ func v1GetChanges(c *Command, r *http.Request, _ *userState) Response {
...
+		// whichever is first.
+		timeout, err := time.ParseDuration(timeoutStr)
+		if err != nil {
+			return statusBadRequest("invalid timeout %q: %v", timeoutStr, err)
+		}
+		timer := time.NewTimer(timeout)
+		select {
+		case <-change.Ready():
+			timer.Stop() // change ready, release timer resources
+		case <-timer.C:
```
hpidcock, line 202: add select clause for select on r.Context().Done()

hpidcock (approved):
Looks good, just two selects that need fixing.

internal/daemon/api_changes.go
benhoyt: Thanks - fixed!

internal/daemon/api_changes.go
benhoyt: Thanks - fixed!

client/changes.go
```diff
@@ -157,3 +157,31 @@ func (client *Client) Changes(opts *ChangesOptions) ([]*Change, error) {
...
+// was an error, which is indicated by Change.Err being set).
+func (client *Client) WaitChange(id string, opts *WaitChangeOptions) (*Change, error) {
+	var chgd changeAndData
+
+	query := url.Values{}
+	if opts != nil && opts.Timeout != 0 {
+		query.Set("timeout", opts.Timeout.String())
+	}
+
+	_, err := client.doSync("GET", "/v1/changes/"+id+"/wait", query, nil, nil, &chgd)
```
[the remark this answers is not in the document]
benhoyt: Good question. I had observed it to be a very long timeout / no timeout in my manual testing, but just confirmed that now in the code -- we always create the `http.Client` with no `Timeout` specified, so no timeout applies here. Arguably that's bad practice, but we can follow up in a subsequent PR if we want to change that.

client/changes.go
```diff
@@ -157,3 +157,31 @@ func (client *Client) Changes(opts *ChangesOptions) ([]*Change, error) {
 
 	return chgs, err
 }
+
+// WaitChangeOptions holds the options for the WaitChange call.
+type WaitChangeOptions struct {
+	// If non-zero, wait at most this long before returning the current change
+	// data. The timeout elapsing is not considered an error, so if a timeout
+	// is specified, the caller should check Change.Ready to determine whether
+	// the change is actually finished.
```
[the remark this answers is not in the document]
benhoyt: Yeah, Harry and I went back and forth on that and ended up that it shouldn't be an error, primarily because when Pebble returns an error it can't also return other data (the Change in this case), and I thought it might be useful to have the current state of the change at the time of the timeout.

However, on reflection it makes for a less error-prone API if it's just an error, and in the success case the change is always ready. If the caller really wants the current state of a change in the case of a timeout/error, they can always make an additional call to `client.Change()` to get the current status. However, it's unlikely you'll want to.

In any case, I've used HTTP status code `504 Gateway Timeout` for this (along with a reasonable error message). It's not quite a perfect semantic match, but it seems close enough.

client/changes.go
```diff
@@ -157,3 +157,31 @@ func (client *Client) Changes(opts *ChangesOptions) ([]*Change, error) {
...
+	Timeout time.Duration
+}
+
+// WaitChange waits for a given change to be finished (whether or not there
+// was an error, which is indicated by Change.Err being set).
```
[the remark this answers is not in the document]
benhoyt: I've updated this documentation.

benhoyt:
@niemeyer I've addressed your concerns -- mainly making the timeout an error and tweaking the doc comments. I'm going to go ahead and merge; we can always tweak later if necessary.

PR #64, opened 2021-08-17
Avoid use of multipart.Part.FileName() to fix files API on Go 1.17

benhoyt:
In Go 1.17 they changed `part.FileName()` to pass the result through `filepath.Base()` before returning, stripping off the directory component which is important for Pebble's files API. See https://golang.org/doc/go1.17#mime/multipart and https://github.com/golang/go/commit/784ef4c53135644d70f3476a4bd90010b9acff66.

Note that `ReadForm()` used in the tests also calls `part.FileName()` so we have to roll that ourselves too.

Incidentally, we'd had some discussion about this in the original PR: https://github.com/canonical/pebble/pull/25#discussion_r611891810

> I had thought about using the filename field earlier, but originally decided against it because I know some clients strip off the path part of it (per the "SHOULD NOT" warning in RFC section 2.3 that you've quoted). However, Go does not, and we'll have a lot of control over the Python client (there's no multipart parser in the stdlib) so I think it's okay. I've updated the code to use "filename" -- it is simpler and cleaner.

Oh well, it's a fairly easy fix. :-)

This commit also runs Go 1.17's version of "go fmt", which adds some `//go:build` lines.

### DIFF
--- a/internal/daemon/api_files_test.go
+++ b/internal/daemon/api_files_test.go
@@ -1054,21 +1054,33 @@ func readMultipart(c *C, response *http.Response, body io.Reader, result interfa
 	c.Assert(mediaType, Equals, "multipart/form-data")
 	c.Assert(params["boundary"], Not(Equals), "")
 	mr := multipart.NewReader(body, params["boundary"])
-	form, err := mr.ReadForm(4096)
-	c.Assert(err, IsNil)
 
-	err = json.Unmarshal([]byte(form.Value["response"][0]), result)
-	c.Assert(err, IsNil)
-
+	// First decode all the files
 	files := make(map[string]string)
-	for _, fh := range form.File["files"] {
-		f, err := fh.Open()
+	var part *multipart.Part
+	for {
+		part, err = mr.NextPart()
 		c.Assert(err, IsNil)
-		b, err := ioutil.ReadAll(f)
+		if part.FormName() != "files" {
+			break
+		}
+		b, err := ioutil.ReadAll(part)
 		c.Assert(err, IsNil)
-		_ = f.Close()
-		files[fh.Filename] = string(b)
+		files[multipartFilename(part)] = string(b)
+		part.Close()
 	}
+
+	// Then decode "response" JSON at the end
+	c.Assert(part.FormName(), Equals, "response")
+	decoder := json.NewDecoder(part)
+	err = decoder.Decode(result)
+	c.Assert(err, IsNil)
+	part.Close()
+
+	// Ensure that was the last part
+	_, err = mr.NextPart()
+	c.Assert(err, Equals, io.EOF)
+
 	return files
 }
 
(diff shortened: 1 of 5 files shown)


hpidcock (approved):

internal/osutil/sys/sysnum_16_linux.go
```diff
@@ -1,4 +1,5 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
+//go:build arm || 386
```
benhoyt, line 2: These are due to the new `//go:build` style build constraints being introduced (and `go fmt` adds them in Go 1.17). See https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

internal/daemon/api_files_test.go
```diff
@@ -1054,21 +1054,33 @@ func readMultipart(c *C, response *http.Response, body io.Reader, result interfa
...
+	// First decode all the files
 	files := make(map[string]string)
-	for _, fh := range form.File["files"] {
-		f, err := fh.Open()
+	var part *multipart.Part
+	for {
+		part, err = mr.NextPart()
 		c.Assert(err, IsNil)
-		b, err := ioutil.ReadAll(f)
+		if part.FormName() != "files" {
```
jameinel, line 1064: This seems a significant difference in the looping. In the former we were iterating over everything in 'files' now we seem to be iterating just over the multipart but then not iterating over files.

(The old form looked like the data was something like:
```
{'blah': ?,
'files': [content, content, content],
}
```

But the new way we are treating it seems more like:
```['blah', {'files': content}, {'files': content'}, {'files': content}]```

I'm assuming that the test suite is passing so I'm wrong about something (I seem to recall some Python HTTP classes that would let you treat repeated headers as a single list.)

Comment (benhoyt) [reply internal/daemon/api_files_test.go:1064]: Yeah, I think you're assuming something wrong. The raw read-files response (before and after) is like so:

```
--01234567890123456789012345678901
Content-Disposition: form-data; name="files"; filename="/root/hi.txt"

Hello
--01234567890123456789012345678901
Content-Disposition: form-data; name="files"; filename="/root/bye.txt"

Bye bye
--01234567890123456789012345678901
Content-Disposition: form-data; name="files"; filename="/root/foobar.txt"

Foo
Bar
--01234567890123456789012345678901
Content-Disposition: form-data; name="response"
{
    "action": "read",
    "files": [
        {"path": "/root/hi.txt"},
        {"path": "/root/bye.txt"},
        {"path": "/root/foobar.txt"}
    ]
}
--01234567890123456789012345678901--
```

In the old code, we called `multipart.Reader.ReadForm` to loop over the multipart sections (files and response). In the new code, we do that loop ourselves and don't call `ReadForm` at all -- either way it loops over the files first, and then grabs the response at the end. Hope that helps.

jameinel (approved):

benhoyt:
@niemeyer I merged this relatively trivial fix with two approvals, but let me know if you have any concerns.

PR #65, opened 2021-09-07
Make Pebble "go vet" clean

benhoyt:
Fix a couple of minor issues that `go vet` raises, just to make Pebble "go vet clean" so that we can see real vet issues easily. We could also add `go vet ./...` to the GitHub Actions to ensure this stays up to date, but I decided not to for now.

Before:

```
$ go vet ./...
# github.com/canonical/pebble/internal/overlord/servstate
internal/overlord/servstate/handlers.go:167:2: unreachable code
internal/overlord/servstate/handlers.go:217:2: unreachable code
# github.com/canonical/pebble/cmd/pebble
cmd/pebble/cmd_logs.go:111:2: misuse of unbuffered os.Signal channel as argument to signal.Notify
$
```

After:

```
$ go vet ./...
$
```

### DIFF
--- a/cmd/pebble/cmd_logs.go
+++ b/cmd/pebble/cmd_logs.go
@@ -107,7 +107,7 @@ func (cmd *cmdLogs) Execute(args []string) error {
 // Needed because signal.NotifyContext is Go 1.16+
 func notifyContext(parent context.Context, signals ...os.Signal) context.Context {
 	ctx, cancel := context.WithCancel(parent)
-	ch := make(chan os.Signal)
+	ch := make(chan os.Signal, 10)
 	signal.Notify(ch, signals...)
 	go func() {
 		// Wait for signal, then cancel the context.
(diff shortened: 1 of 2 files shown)


cmd/pebble/cmd_logs.go
```diff
@@ -107,7 +107,7 @@ func (cmd *cmdLogs) Execute(args []string) error {
 // Needed because signal.NotifyContext is Go 1.16+
 func notifyContext(parent context.Context, signals ...os.Signal) context.Context {
 	ctx, cancel := context.WithCancel(parent)
-	ch := make(chan os.Signal)
+	ch := make(chan os.Signal, 10)
```
hpidcock, line 110: what is the wisdom behind this magic number? const/comment maybe?

hpidcock (approved):

cmd/pebble/cmd_logs.go
benhoyt: Fair call. The 10 was just a magic value >0 that I saw used in LXD code. Anything above 0 shuts up go vet, and other places in Pebble use 2 for the os.Signal channel, so I used that and added a clarifying comment.

PR #66, opened 2021-09-07
Remove a bunch of snapd references

benhoyt:
Because of the history of pebble code being copied from snapd, there are still a bunch of snap/snapd references scattered throughout the code. This removes the most obvious ones in comments and the like. It also deletes a bunch of commented-out code in `cmd/pebble/export_test.go` (that's what version control is for :-).

I would also like to remove the systemd/standby stuff that I believe is specific to snapd, but that's a heavier change so I'd like to discuss that first.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -52,9 +52,9 @@ type doer interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
-// Config allows to customize client behavior.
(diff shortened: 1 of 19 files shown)


hpidcock (approved):
Seems benign. Nice catch on the os.Unsetenv

PR #68, opened 2021-09-15
Let pebble list files recursively with a flag

rbarry82:


rbarry82:
@benhoyt it's quick, but it's a start. If we don't worry about exclusions from walking the path and callback functions, "if this file is a directory, follow it"...

benhoyt:
@rbarry82 Thanks. As mentioned on Mattermost, I think there are some tricky design issues here when you dig in. That's fine -- this is as good a place as any to discuss. I think this is a reasonable feature to include, but we want to get the design in Pebble. There are a few discussion points here:

1) As it stands, `recurse` doesn't interact very well with `pattern`. You're doing the matching first, and you don't recurse into a directory if it doesn't match. But it seems to me someone who does `list_files(path, pattern='*.txt', recurse=True)` will want it to find all `*.txt` files under the path, whether or not the directory matches the pattern. Currently I believe it'll only find files like `dir.txt/file.txt`, which is rare. So I think we want to do the recursion before the matching.
    - It may be better to extend pattern itself to allow `**` style recursive globs, which is well-known and used in `.gitignore` and the like. If the path includes a `**`, recursion is enabled (a pattern of just `**` finds all files recursively). You can't do that with Go's standard library glob, but there are several packages that seem to do it nicely, so it wouldn't be hard to achieve (https://github.com/bmatcuk/doublestar, https://github.com/mattn/go-zglob, https://github.com/yargevad/filepathx).
    - I also think it'd be a bit nicer of an API to return the results in a flat list, instead of arbitrarily nested children sub-lists. So you'd get a flat list with paths `sub`, `sub/one`, `sub/one/three.txt`, etc. This avoids the caller having to iterate the response recursively -- it's flattened.

2) Pruning. It seems to me this would be very desirable in a large directory tree, particularly things like ignoring `.`-prefixed files and directories like `.git` (and all its hundreds of sub-folders). You *can* prune client-side, of course, but it's much less inefficient for large trees. I'm just not sure we want an "inefficient by design" kind of API. If we're going down this track, I think it might be good to at least have a simplistic pruning option, `ignore_dot=True`. Or perhaps an additional `ignore` pattern, so you could say `list_files(path, pattern='**', ignore='**/.*')` (we'd have to think about whether ignore matched directories or just files).

3) Handling of file system cycles (I think you mentioned this on Mattermost, and it's a good point). If someone has a symlink that points back up in the tree, a simplistic approach would recurse forever. I think the code you have doesn't follow symlinks at all, so can't have cycles? If so, that's probably a good default. Just want to make sure we've thought about it.

4) Error handling. What happens if iterating a sub or sub-sub directory has an error: do we fail hard and return no results (the current code), or skip it and keep going (maybe keeping a list of error directories to include in the response), or something else? Failing hard like you've done is the obvious answer, but it's a bit of a pain if you have a big tree with one deeply-nested sub-folder that has bad permissions and that causes the entire search to fail. For an ideal API, I'd probably have it fail hard by default (`errors='fail'`), but have an option like `errors='report'` to skip errors and report them in the response.

rbarry82:
I agree with basically all of these, but I wanted to get at least a trivial implementation up so we could start discussing the pain points somewhere other than Mattermost. In general, I'm less worried about the callback functions we discussed (since, from this point of view, we can more or less treat Pebble as an API rather than a "raw" filesystem.

1. You raise a good point about `pattern`. Pattern matching after recursion is definitely the way to go, and I missed it in the initial implementation.
- `**`-style recursive globs just as well for me. I'm not particular about it at all. https://github.com/bmatcuk/doublestar is well-regarded and maintained. I"ll go with that.
- Flattening the list gets a :+1: from me. The nested datastructure lends itself to non-Golang JSON parsing easily enough, but making the API "nice" would require reflecting back on it and marshaling it to JSON in a way which cuts `Children` out of the struct and flattens it out so `resp["result"]["foo"]["bar"]["quux.txt"]` comes back instead, and it's not likely to be worth the effort compared to flattening the list.
2. The response from the Pebble API is remarkably fast even with huge directory trees, and there are enough "hidden files matter" use cases for configuration files that I have mixed feelings about excluding them by default. Programatically, I'd also expect hidden files to be shown (as it is by default in go, python, `readdir()`, and virtually every other API). From my POV, this seems to be consistency with existing utilities, even if it's ineffecient.
3. Right -- the existing code does not follow symlinks. `filepath.Walk` and `os.path.walk` also do not follow links by default, so it seemed like a good pattern to stick with.
4. :+1: to returning the errors in the response. Failing hard by default is always reasonable in Go, The question then is going to be "what code do we return"? It's not really 200 if it's errored, but it also doesn't fit nicely into the paradigms for `4xx` or `5xx`

benhoyt:
Thanks. Yeah, I think I like the `**` recursive globs. They're well-known from shells and `.gitignore` and so on. That doublestar package looks reasonable at a quick scan.

I'm still concerned about pruning, as the size may get huge quickly. When I do a `sudo find / >files.txt` on my machine it took about 15 seconds to complete (the first time, ~5 seconds after), and returned some 2.8 million paths. The raw paths output by find was 200MB, and I estimated that would be about 700MB in list-files response format (with the additional fields and in JSON format). That's pretty huge... I think even if we don't implement `ignore` or some other kind of pruning we should have a `limit` which defaults to 10,000 items or so (that would be approx 2MB of JSON, and you could bump up the limit in the request if you knew you were really looking at a huge tree).

Note: currently the Pebble `Response` type is a bit dumb, in that it marshals the entire response to a `[]byte` array, effectively instantiating the data twice, so it would be 700MB times 2 in the above example. We could probably fix that doubling by improving `Response.ServeHTTP` to encode directly to the `ResponseWriter`.

So I'd be okay with the following for a first-pass API, and we can always add ignore patterns or more advanced pruning or error handling or other features later, incrementally.

* recursive `**` globbing for `pattern`
* flat response structure with a single list of file infos (with each `path` containing the full path relative to the base of `pattern`)
* a new `limit` parameter, with a default of maybe 10,000
* fail hard
* don't follow symlinks

However, I would suggest getting @jnsgruk and @niemeyer's approval on this design before too much more coding.

benhoyt:
Oh, I realized we probably can't add this to the existing `GET /v1/files?action=list` action, for compatibility reasons. For example, if a new client is talking to an old Pebble server without `**` support, it will silently ignore the `**` and return whatever it matches non-recursively. Whereas we really want it to return an error in that case, so the client knows it's talking to an old Pebble that doesn't support recursion. Otherwise the client will blindly use the (incorrect) results. The same would be true if we used a `recurse=true` query string parameter -- existing versions of Pebble would just ignore that query parameter.

So what I think we should do is add this functionality under a new `action=list-recursive` mode (same `GET /v1/files` endpoint). With existing versions of Pebble that will return an HTTP 400 Bad Request error, so the client knows it's not supported. (A smart client *could* even fall back to interpreting the `**`'s and doing the recursion locally, though that's probably a bit too clever.)

niemeyer (changes requested):
The PR has no description so I'm not quite sure about what's the backing story. From the summary and the initial patch, it sounds like we might not want this, as it's turning a long iterative process into a server side process that runs for potentially a long time, potentially caching a lot of data, and without any feedback at all. So bad effects on the daemon, and bad UX.

Again, not sure about the use case, but have you considered just iterating on the files on the client side?  That was the idea when we were talking about this API back then.

rbarry82:
I'm writing out a spec to flesh this out more and have an actual discussion about it.

It's a tradeoff between a very high number of API calls from the client to iterate over the files on the client side in order to perform a directory pull and a relatively fast/low-impact walk of a tree on the daemon side.

If we were measuring performance between "list the contents of a directory, issue an API call to pull every file individually, recurse if needed" and doing it all server-side to filter, the server side will win every time. This PR was intended to be a starting point for a discussion, but it's moving to the spec. Closing for now.

Issue #57, opened 2021-07-12
`pebble logs` does not see previous runs of a service


hpidcock:
Restarting a service `pebble stop <svc> && pebble start <svc>` causes logs from previous service processes to be lost.

benhoyt:
This was fixed with the refactoring done in https://github.com/canonical/pebble/pull/79

PR #69, opened 2021-09-21
Show process's last output in error message when service exits quickly

benhoyt:
The `pebble logs` command can now see the logs of previously stopped/terminated services (see https://github.com/canonical/pebble/issues/57). But it's also useful to include the last output (log lines) in the common "exited quickly" case, so this change adds that output to the task log of the change.

This output will appear in the exception string automatically in charms, via Python Operator Framework change https://github.com/canonical/operator/pull/653.

The output (from `pebble start` trying to start a service with command `sleep x`) appears as follows:

```
$ pebble start sleepx
- Start service "sleepx" (service exited too quickly with code 1)
```

And then if you view the task's log using `pebble tasks` (with the change ID from `pebble changes`):

```
$ pebble tasks 1385
Error   today at 14:08 NZDT  today at 14:08 NZDT  Start service "sleepx"

......................................................................
Start service "sleepx"

2021-10-20T14:08:07+13:00 INFO most recent service output:
    sleep: invalid time interval ‘x’
    Try 'sleep --help' for more information.
2021-10-20T14:08:07+13:00 ERROR service exited too quickly with code 1
```

Note that in the ring buffer the log lines are prefixed with `2021-09-21T00:33:20.605Z [sleepx] `, so we strip that off here to avoid double timestamps and make it look nicer for the user.


internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +177,23 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+		}
+		return errors.New(msg)
 	}
 	// unreachable
 }
 
+func readLastLogs(logBuffer *servicelog.RingBuffer, numLines int) ([]byte, error) {
+	it := logBuffer.HeadIterator(numLines)
+	defer it.Close()
+	b, err := ioutil.ReadAll(it)
```
hpidcock, line 193: `return ioutil.ReadAll(it)`?

internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +177,23 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 			}
 			releasePlan()
 		}
-		return fmt.Errorf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
+		msg := fmt.Sprintf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
```
hpidcock, line 180: we should be able to read the last log here and omit the `lastLog` variable

hpidcock (approved):
dropping `lastLog` variable as its not needed.

internal/overlord/servstate/handlers.go
benhoyt: Good call. Done.

internal/overlord/servstate/handlers.go
benhoyt: I actually had that at first, but there's a race condition between the goroutine that calls `cmd.Wait()`, because it's closing the log buffer. So if the goroutine's `logBuffer.Close()` occurs before or during this reads the last logs, we get an error reading the logs because the log buffer is already closed.

benhoyt: Per discussion, this was a bug in the ring buffer where it would double-close `iterator.closeChan` if you created the iterator after the ring buffer `Close()` method was called (which should only close it for writes, not reads). Fixed the bug per your suggestion and updated the test to catch this (it wasn't closing the iterators).

hpidcock (approved):

internal/overlord/servstate/handlers.go
```diff
@@ -162,7 +168,20 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+		it := logBuffer.HeadIterator(lastLogLines)
+		defer it.Close()
+		lastLogs, err := ioutil.ReadAll(it)
+		if err != nil {
+			logger.Noticef("could not read service %q logs: %v", req.Name, err)
+		}
+
+		msg := fmt.Sprintf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
+		if len(lastLogs) > 0 {
+			msg = fmt.Sprintf("%s - last %d lines of output:\n%s",
```
niemeyer, line 181: I'm hoping we can get something a bit simpler and less verbose.  How about something like this:
```
cannot start %q: exited quickly with code %d:
    (...)
    <line1>
    <line2>
```
Four spaces at the start, and (...) is only present if there was any output before it. We have logic to do almost exactly this in a function called OutputErr somewhere (likely strutil or osutil).

Also, the timestamp and extra service name on every line is not nice. It makes sense on an aggregated log, but here it seems alien and unpolished. We should be able to produce a cleaner log, either using the same logic used elsewhere (reusing, not copying), or maybe with the knowledge that we have above (the formatting is setup right here).

Happy to talk further.

niemeyer (changes requested):
Looks pretty good. I'm hoping we can just polish the way it's exposed a bit.

internal/overlord/servstate/handlers.go
```diff
@@ -162,7 +168,20 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 			}
 			releasePlan()
 		}
-		return fmt.Errorf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
+
+		it := logBuffer.HeadIterator(lastLogLines)
+		defer it.Close()
+		lastLogs, err := ioutil.ReadAll(it)
+		if err != nil {
+			logger.Noticef("could not read service %q logs: %v", req.Name, err)
```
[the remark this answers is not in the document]
benhoyt: Done

internal/overlord/servstate/handlers.go
```diff
@@ -162,7 +168,20 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 			}
 			releasePlan()
 		}
-		return fmt.Errorf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
+
+		it := logBuffer.HeadIterator(lastLogLines)
```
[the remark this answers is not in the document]
benhoyt: That's confused me several times as well, because usually we think of "tail" as tailing the end of the log as in the `tail` command. But in the ring buffer the tail is the first logs that were written (it follows the head) and the head is the most recent. Interestingly, I just typed "ring buffer head and tail" into Google and the first hit was a [thread](https://groups.google.com/g/comp.os.msdos.programmer/c/H75qEeKMGVo) with people debating which should be named which!

internal/overlord/servstate/handlers.go
benhoyt: I've fixed this now to simplify the message (we don't actually need the `%q` service name as it's already in the outer message (see updated PR description), but I've made the other changes: indentation by 4 spaces, and the `(...)` truncation marker if it exceeds 20 lines.

I'm not planning to remove the timestamps for now, as we have a separate task on our [TODO list](https://github.com/canonical/pebble#roadmap--todo) to parse/remove timestamps, and we can clean this up then as well.

benhoyt: Per discussion, I'm stripping off the timestamp + service name now.

internal/overlord/servstate/handlers.go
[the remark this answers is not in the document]
jameinel: You could avoid calling it head and tail and call it "oldest" and "newest". Might be clearer.

internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +169,48 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+			msg += ":\n" + logs
+		}
+		return errors.New(msg)
 	}
 	// unreachable
 }
 
+// Used to strip the Pebble log prefix, for example: "2006-01-02T15:04:05.000Z [service] "
+// Timestamp must match format in logger.timestampFormat.
+var timestampServiceRegexp = regexp.MustCompile(`\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z \[.*?\] `)
```
niemeyer, line 189: Probably irrelevant here, but instead of `.*?` this is better spelled as `[^]]+`, which means "at least one character, but not ]". This prevents backtracking, as it can deterministically tell where to stop without depending on the rest of the string.

It also need to be anchored. More on this below.

internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +169,48 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
 			releasePlan()
 		}
-		return fmt.Errorf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
+
+		logs, err := getLastLogs(logBuffer)
+		if err != nil {
+			logger.Noticef("Cannot read service %q logs: %v", req.Name, err)
+		}
+
+		msg := fmt.Sprintf("service exited too quickly with code %d", cmd.ProcessState.ExitCode())
```
niemeyer, line 178: The old message looks slightly more clear. There's no context here about the fact we were attempting to start that service.

internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +169,48 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 			}
 			releasePlan()
 		}
-		return fmt.Errorf("cannot start service: exited quickly with code %d", cmd.ProcessState.ExitCode())
+
+		logs, err := getLastLogs(logBuffer)
+		if err != nil {
+			logger.Noticef("Cannot read service %q logs: %v", req.Name, err)
```
niemeyer, line 175: Instead of logging locally, this should be bundled on the error message itself, as we're already raising an error, and this almost certainly is part of that error issue and should be observed and understood together.

internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +169,48 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+		}
+
+		msg := fmt.Sprintf("service exited too quickly with code %d", cmd.ProcessState.ExitCode())
+		if logs != "" {
+			msg += ":\n" + logs
```
niemeyer, line 180: Per note on the last review, can we please use OutputErr here?  It is oriented exactly for this use case: merging some output and a related error value. It handles a few different cases, has formatting to isolate the text, and it works well when there is no text as well, so you can just blindly pass the text in without worrying about it. Also, if it happens to not be great for a certain case, we can fix a single location and have it improved everywhere consistently.

niemeyer (changes requested):
I'm sorry as this looks very nitpicky. The whole thing is good and should go in, it's just trivial details that I'd like to talk about to make sure we're not by mistake breaking what is working.

internal/overlord/servstate/handlers.go
```diff
@@ -162,11 +169,48 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+	// Indent lines
+	trimmed := strings.TrimSpace(string(logBytes))
+	lines := strings.Split(trimmed, "\n")
+	if len(lines) > lastLogLines {
+		// Prefix with truncation marker if too many lines
+		lines[0] = "(...)"
+	}
+	for i, line := range lines {
+		// Strip Pebble timestamp and "[service]" prefix
+		line = timestampServiceRegexp.ReplaceAllString(line, "")
```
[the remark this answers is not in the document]
benhoyt: Good call on the anchoring, done. I've kept it line-by-line as we're doing line-by-line processing in any case to do the indentation, so this just keeps it simple (and should still be relatively efficient with the `^`).

internal/overlord/servstate/handlers.go
benhoyt: `OutputErr` is a bit of a strange one, as it *doesn't* show the `err` at all if there is output. But in this case the `err.Error()` message is useful for context, and includes the exit code. For example, with `OutputErr`:

```
cannot perform the following tasks:
- Start service "sleepx":
-----
sleep: invalid time interval ‘x’
Try 'sleep --help' for more information.
-----
```

Rather than the more helpful:

```
cannot perform the following tasks:
- Start service "sleepx": service exited too quickly with code 1:
    sleep: invalid time interval ‘x’
    Try 'sleep --help' for more information.
```

I'm fine with the additional `-----` formatting here if you want, but the lack of "service exited too quickly with code 1" is not an improvement. We could change `OutputErr` to include the err as well, and maybe that'd be an improvement for everyone? But it seems it was intentionally designed that way, though I'm not quite sure why -- error information is lost.

internal/servicelog/ringbuffer.go
```diff
@@ -342,7 +342,12 @@ func (rb *RingBuffer) releaseIterators() {
 	rb.iteratorMutex.Lock()
 	defer rb.iteratorMutex.Unlock()
 	for _, iter := range rb.iteratorList {
-		close(iter.closeChan)
+		// Close closeChan if not already closed
```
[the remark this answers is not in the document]
benhoyt: True. This change was because after you close the ring buffer for writes (`RingBuffer.Close`), you should still be able to create read iterators, and this wasn't working due to the double close, hence this fix.

However, I've discussed with @hpidcock, and there is a better way to structure the RingBuffer to avoid the double close, as well as combine the individual `Iterator.closeChan` into one (semantically it should probably be one channel on the ring buffer). This change isn't hurting, so I'll leave it here, but we plan to refactor that in a subsequent commit.

internal/overlord/servstate/handlers.go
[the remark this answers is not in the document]
benhoyt: Probably clearer, yeah. However, I'm not going to make that change in this PR. @hpidcock can open a separate PR if he wants, though IMO it's not worth it.

internal/overlord/servstate/handlers.go
benhoyt: This is moot now: `OutputErr` is no longer relevant now that we're putting the service output logs in the task's log rather than returning it in the change error.

internal/overlord/servstate/handlers.go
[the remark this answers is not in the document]
hpidcock: Heh, I think of this as if it's a lizard "if you cut the head off, then it's dead, but you can cut off the tail as much as you want, it grows back". Anyway, happy to change the naming to be less ambiguous.

internal/overlord/servstate/handlers.go
```diff
@@ -437,6 +444,51 @@ func (s *serviceData) exited(waitErr error) error {
...
+	st.Lock()
+	defer st.Unlock()
+
+	logs, err := getLastLogs(logBuffer)
+	if err != nil {
+		task.Errorf("cannot read service logs: %v", err)
+	}
+	if logs != "" {
+		// Add lastLogLines last lines of service output to the task's log.
+		task.Logf("most recent service output:\n%s", logs)
```
niemeyer, line 459: s/most/Most/, and s/cannot/Cannot/ above.

niemeyer (approved):
Just the logs, otherwise thanks and sorry for being very slow on this one.

internal/overlord/servstate/handlers.go
benhoyt: Done, thanks

PR #73, opened 2021-10-04
Exec client tests

benhoyt:
Add client tests for exec (one-shot commands). To make this testable in a straight-forward way, I needed to refactor `wsutil` a little bit to accept interfaces rather than concrete `*websocket.Conn` instances, and make the function which creates client websockets settable (`Client.getWebsocket`).


client/client.go
```diff
@@ -87,28 +106,47 @@ func New(config *Config) *Client {
...
-			userAgent: config.UserAgent,
+		// By default talk over a UNIX socket.
+		transport = &http.Transport{Dial: unixDialer(config.Socket), DisableKeepAlives: config.DisableKeepAlive}
+		baseURL := url.URL{Scheme: "http", Host: "localhost"}
+		client = &Client{baseURL: baseURL}
+	} else {
+		// Otherwise talk regular HTTP-over-TCP.
+		baseURL, err := url.Parse(config.BaseURL)
+		if err != nil {
+			panic(fmt.Sprintf("cannot parse server base URL: %q (%v)", config.BaseURL, err))
```
hpidcock, line 121: We should just return an error vs panicing.

hpidcock (approved):

client/client.go
benhoyt: This panic was there already, but I agree -- fixed.

client/client.go
```diff
@@ -79,36 +83,70 @@ type Client struct {
...
+	jsonWriter
+}
+
+type jsonWriter interface {
+	WriteJSON(v interface{}) error
 }
 
 // New returns a new instance of Client
-func New(config *Config) *Client {
+func New(config *Config) (*Client, error) {
```
niemeyer, line 104: The change looks good, and okay to do right now. But soon we won't be able to tolerate that kind of change in the client package, as this is exactly the part of Pebble that is supposed to be used as a library by third-parties. The exposed API needs to be rock-solid.

PR #74, opened 2021-10-04
Only test/build Linux-specific parts of wsutil and ptyutil on Linux

benhoyt:
This avoids running the tests for the Linux-specific parts of this on non-Linux OSes, so that `github.com/canonical/pebble` is (for example) importable in Juju and `go test ./...` runs on Windows and macOS.

The `_linux.go` suffix is also how the LXD codebase does it.


benhoyt:
This should be a no-op change (just moving code from one file into another), so I'm going to merge directly to make faster progress incorporating into Juju.

PR #77, opened 2021-10-19
Exec: Change EOF designator from empty TEXT frame to {"command":"end"}

benhoyt:
Per discussion at https://github.com/canonical/operator/pull/593#discussion_r730471325.

The corresponding Python Operator Framework commit is https://github.com/canonical/operator/pull/593/commits/1d72211f4910c26545547cb91fc5f6336bcde4d5.

### DIFF
--- a/internal/wsutil/wsutil.go
+++ b/internal/wsutil/wsutil.go
@@ -90,6 +90,8 @@ func WebsocketRecvStream(w io.Writer, conn MessageReader) chan bool {
 }
 
 func recvLoop(w io.Writer, conn MessageReader) {
+	buf := make([]byte, 32*1024) // only allocate once per websocket, not once per loop
+
 	for {
 		mt, r, err := conn.NextReader()
 		if err != nil {
@@ -97,16 +99,22 @@ func recvLoop(w io.Writer, conn MessageReader) {
 			return
 		}
 
-		if mt == websocket.CloseMessage {
+		switch mt {
+		case websocket.CloseMessage:
 			logger.Debugf("Got close message for reader")
 			return
-		}
 
-		if mt == websocket.TextMessage {
+		case websocket.TextMessage:
+			// A TEXT message is an out-of-band "command".
+			payload, err := ioutil.ReadAll(r)
+			if err != nil {
+				logger.Debugf("Cannot read from message reader: %v", err)
+				return
+			}
 			var command struct {
 				Command string `json:"command"`
 			}
-			err := json.NewDecoder(r).Decode(&command)
+			err = json.Unmarshal(payload, &command)
 			if err != nil {
 				logger.Noticef("Cannot decode I/O command: %v", err)
 				continue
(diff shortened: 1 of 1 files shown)


internal/wsutil/wsutil.go
```diff
@@ -109,48 +81,59 @@ func WebsocketSendStream(conn MessageWriter, r io.Reader, bufferSize int) chan b
...
+				Command string `json:"command"`
 			}
+			err := json.NewDecoder(r).Decode(&command)
 			if err != nil {
-				logger.Debugf("Error writing buf %s", err)
-				break
+				logger.Noticef("Cannot decode I/O command: %v", err)
+				continue
+			}
+			if command.Command != "end" {
```
hpidcock, line 114: This should just be a switch statement.

internal/wsutil/wsutil.go
```diff
@@ -109,48 +81,59 @@ func WebsocketSendStream(conn MessageWriter, r io.Reader, bufferSize int) chan b
 func WebsocketRecvStream(w io.Writer, conn MessageReader) chan bool {
 	ch := make(chan bool)
```
hpidcock, line 110: Probably this should just be buffered with a size of 1 so the goroutine can complete earlier.

hpidcock (approved):

internal/wsutil/wsutil.go
benhoyt: True. I changed it to a `close(ch)` instead to match the close in `WebsocketSendStream`.

internal/wsutil/wsutil.go
```diff
@@ -109,48 +81,61 @@ func WebsocketSendStream(conn MessageWriter, r io.Reader, bufferSize int) chan b
...
 
-			i, err := w.Write(buf)
-			if i != len(buf) {
-				logger.Debugf("Didn't write all of buf")
-				break
+		if mt == websocket.TextMessage {
+			var command struct {
+				Command string `json:"command"`
 			}
+			err := json.NewDecoder(r).Decode(&command)
```
niemeyer, line 109: What's the intended handling of the remaining data in the reader here?  I suppose right now it's ignoring anything after the first JSON object.  We probably don't care much for now while we really only have one message and at the very end, but I'm curious if there's candidate bug here or not. If so, there should at least be a note here to remind ourselves that this is an issue.

internal/wsutil/wsutil.go
```diff
@@ -109,48 +81,61 @@ func WebsocketSendStream(conn MessageWriter, r io.Reader, bufferSize int) chan b
...
+			default:
+				logger.Noticef("Invalid I/O command %q", command.Command)
+				continue
 			}
 		}
-		ch <- true
-	}(w, conn)
 
-	return ch
+		buf, err := ioutil.ReadAll(r)
```
niemeyer, line 124: Seems worthwhile checking mt here, and only acting on BinaryMessage otherwise this will break down in the future when the type is unknown.

internal/wsutil/wsutil.go
```diff
@@ -109,48 +81,61 @@ func WebsocketSendStream(conn MessageWriter, r io.Reader, bufferSize int) chan b
...
+		if err != nil {
+			logger.Debugf("Cannot read from message reader: %v", err)
+			return
```
niemeyer, line 127: Should these be more visible errors somehow?  What happens elsewhere if we return silently here and below?

internal/wsutil/wsutil.go
```diff
@@ -109,48 +81,64 @@ func WebsocketSendStream(conn MessageWriter, r io.Reader, bufferSize int) chan b
...
-				logger.Debugf("Got error getting next reader %s", err)
-				break
-			}
+func recvLoop(w io.Writer, conn MessageReader) {
+	buf := make([]byte, 32*1024) // only allocate once per websocket, not once per loop
+
+	for {
+		mt, r, err := conn.NextReader()
+		if err != nil {
+			logger.Debugf("Cannot get next reader: %v", err)
```
hpidcock, line 98: Not really required for now but in the future I'd probably make recvLoop return an error, that way all this error logging can be handled in the caller.

internal/wsutil/wsutil.go
benhoyt: Yeah, fair point. I've changed to use `json.Unmarshal` which only supports a single value. `json.Decoder` is more for streams in any case.

internal/wsutil/wsutil.go
benhoyt: Good call. I've updated, and also refactored a bit to only allocate a buffer once instead of every loop as `ReadAll` will.

internal/wsutil/wsutil.go
benhoyt: I started doing that and a couple of them do actually occur under normal conditions, for example error getting the reader if one side closes the connection first. So I decided to just leave for now. Can always go into `PEBBLE_DEBUG=1` mode if/when we need to debug.

Issue #72, opened 2021-09-30
pebble exec: sending stdin to "use-terminal" process doesn't work


benhoyt:
Per [discussion](https://github.com/canonical/pebble/pull/61#discussion_r719482813) on PR #61, there's a bug in the one-shot commands handling where you can't send stdin to an exec process with "use-terminal" true.

We were aware of this issue before merging #61, but wanted to get that PR merged and fix this later, as it isn't a critical issue.

It's easy to reproduce:

```
$ echo foo | PEBBLE=~/pebble go run ./cmd/pebble exec -t -- cat  # -t is the default
exit status 129

# You either get that error, or sometimes this one:
$ echo foo | PEBBLE=~/pebble go run ./cmd/pebble exec -t -- cat
error: cannot perform the following tasks:
- exec command "cat" (fork/exec /usr/bin/cat: input/output error)
exit status 1

# In non-terminal mode ("use-terminal" false), it works fine:
$ echo foo | PEBBLE=~/pebble go run ./cmd/pebble exec -T -- cat
foo
```

On the server end, it's receiving some kind of `POLLNVAL` signal, and prints that in the logs:

```
2021-09-30T20:55:14.332Z [pebble] POST /v1/exec 14.740358ms 202
2021-09-30T20:55:14.345Z [pebble] GET /v1/tasks/1153/websocket/control 12.786948ms 200
2021-09-30T20:55:14.346Z [pebble] GET /v1/tasks/1153/websocket/stdio 54.774µs 200
2021-09-30T20:55:14.346Z [pebble] Detected poll(POLLNVAL) event: exiting.
2021-09-30T20:55:14.375Z [pebble] GET /v1/changes/1153/wait 29.210628ms 200
```

benhoyt:
This was fixed with https://github.com/canonical/pebble/pull/81

PR #78, opened 2021-10-21
Revert to lxc exec / ssh behavior for detecting terminal mode

benhoyt:
Per discussion at https://github.com/canonical/pebble/pull/61#discussion_r719482813,
we wanted to make the default use-terminal mode based only on stdout,
but that's different to what both "lxc exec" and "ssh" do, so seems
wise to copy them. Additionally, it's problematic to implement,
because when allocating a PTY on the remote, you get a "master" file
descriptor that handles both stdin and stdout. You have to call
close() on the master FD to close stdin so that a command that reads
stdin (eg: "cat") stops, but you're closing the entire master FD so
that also closes stdout, causing the POLLNVAL errors we were seeing on
the stdout side, as well as the command output sometimes being cut
short.

We probably could address that by having a separate pipe for stdin and
specifying that in the API, or allocating two PTYs on the remote, or
similar, but it's complex and I don't think deviating from ssh or
"lxc exec" is a good idea.

In practice you very rarely want interactivity if you're piping stdin,
which is I guess why ssh (and lxc exec) do it the way they do.

When you try to pipe stdin and specify -t, here is what "lxc exec" and
ssh do:

* lxc exec: "echo foo | lxc exec <instance> -t -- cat" ... this fails
  with the error we were seeing here, exit code 129 and intermittent
  output for the same reason.
* ssh: "echo foo | ssh giftyweddings.com -t -- cat" ... because it's
  not going to work anyway, ssh explicitly prevents this, uses
  non-terminal mode anyway, and prints the following message on stderr:

$ echo foo | ssh giftyweddings.com -t -- cat
Pseudo-terminal will not be allocated because stdin is not a terminal.
foo

I think the stderr message is a good idea, because it tells you you've
done something wrong, so I've copied that here in "pebble exec".

This reverts commit d7ca5289e9d23b8f5d04c0bcc970f7939bce905c plus
adding the stderr msg if -t is specified and stdin is not a terminal.

Fixes https://github.com/canonical/pebble/issues/72

### DIFF
--- a/cmd/pebble/cmd_exec.go
+++ b/cmd/pebble/cmd_exec.go
@@ -54,7 +54,7 @@ var execDescs = map[string]string{
 	"gid":     "Group ID to run command as",
 	"group":   "Group name to run command as (group's GID must match gid if both present)",
 	"timeout": "Timeout after which to terminate command",
-	"t":       "Allocate remote pseudo-terminal (default if stdout is a TTY)",
+	"t":       "Allocate remote pseudo-terminal (default if stdin and stdout are TTYs)",
 	"T":       "Disable remote pseudo-terminal allocation",
 }
 
@@ -94,19 +94,24 @@ func (cmd *cmdExec) Execute(args []string) error {
 		env[key] = value
 	}
 
-	// Specify UseTerminal if -t/--terminal is given, or if stdout is a TTY.
+	// Specify UseTerminal if -t is given, or if both stdin and stdout are TTYs.
+	stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
 	stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
 	var useTerminal bool
 	if cmd.Terminal {
 		useTerminal = true
+		if !stdinIsTerminal {
+			fmt.Fprintln(os.Stderr, "Pseudo-terminal will not be allocated because stdin is not a terminal")
+			useTerminal = false
+		}
 	} else if cmd.NoTerminal {
 		useTerminal = false
 	} else {
-		useTerminal = stdoutIsTerminal
+		useTerminal = stdinIsTerminal && stdoutIsTerminal
 	}
 
 	// Record terminal state (and restore it before we exit).
-	if useTerminal && ptyutil.IsTerminal(unix.Stdin) {
+	if useTerminal && stdinIsTerminal {
 		oldState, err := ptyutil.MakeRaw(unix.Stdin)
 		if err != nil {
 			return fmt.Errorf("cannot change terminal to raw mode: %v", err)


hpidcock (approved):
Thanks for the detailed reasoning and research on this. LGTM

benhoyt:
Okay, I just did a quick proof-of-concept for making this happen, and my quick hack at least seems to work. For `echo foo | pebble exec -- cat`, on the remote/server side, we still open a PTY for stdout/stderr, but we also open a separate pipe for stdin, and connect the read side of that pipe to `cmd.Stdin`.

However, we'd need to add a field to our API so the server knows to use a pipe for stdin instead of the PTY. Something like a `stdin-is-pipe` boolean (which you would only set if `use-terminal` is true, because if `use-terminal` is false, stdin is already a pipe). This wouldn't affect Python Operator Framework (which doesn't support `use-terminal` true anyway) and is backwards compatible.

What I suggest we do is merge this change so that at least `echo foo | pebble exec -- cat` works at a basic level instead of being broken (`use-terminal` will be false like lxc/ssh), but update issue https://github.com/canonical/pebble/issues/72 and keep that open to track the work to actually get the `stdin-is-pipe` approach working. I'm still not convinced the effort is worth it ... but in any case, it's not a trivial tweak and requires a small amount of restructuring, so I'd rather focus on other features for a while (specifically auto-restart and health checks).

benhoyt:
I think it's a reasonable idea to add a new API parameter to allow creation of a PTY but stdin to be a pipe. However, I'm less convinced it's particularly important, given how long `lxc exec` has had its current behavior. I use `lxc exec` for single commands quite regularly (because I don't want to enter a shell only to run a single command, then exit it), so I'm not sure I buy the argument that it's "optimized for remote shell use".

I'm also not convinced it's worth making a backwards-incompatible change. I realize it's early and we probably could change it without much pain, but at some point we have to draw a line in the sand ... what's too long to make a backwards-incompatible change? a week, a month, a year?

In any case, what about the following as a (backwards-compatible) compromise:

* Keep the existing `terminal=true/false` field to mean "open a terminal/PTY". (Adding a "use-" prefix doesn't add much IMO.)
* Add a new field `stdin-mode` which has values of "auto" (the default), "pipe" (connect stdin to a pipe), and "terminal" (connect stdin to a terminal/PTY). The default would be "auto", meaning "terminal" if `terminal=true` and "pipe" if `terminal=false`, i.e., the current behavior stays the same. (The combination terminal=false, stdin-mode=terminal would be invalid, at least for now.)

Given that this would be backwards compatible, we could simply merge this PR to fix `echo foo | pebble exec -- cat`. Later when we add `stdin-mode`, it'd default to `terminal=true`, `stdin-mode=pipe`. But I don't have strong feelings ... we could leave this as is and not merge this PR to signal "CLI fix / change is coming".

Regarding your suggested name, I don't think the name `interactive=true/false` for the new field describes what it does very well. What's interactive? Does it refer to stdin or stdout? It doesn't seem particularly clear.

niemeyer:
You're right that we need to draw a line, and you're also right that we cannot wait too long. So let me try to help with both of these:

Let's please change both the API and the Go API to have the following fields: terminal/Terminal and interactive/Interactive, with the semantics defined above.

The rationale is this: "terminal" forces the allocation of a terminal, but does not assign stdin to the terminal unless "interactive" is also true. What's interactive is the command being run, and I hope we can at least agree on what an interactive process is general.

You're right that it's not a perfectly transparent, but none of the other options are either, unless one reads the documentation, so I'm more worried about it being nice and consistent. "terminal" also means nothing, for instance. And "terminal" and "stdin-mode" aren't a good pair.

I'm on holiday this week, but I'm keen on unblocking this, so hopefully this helps a bit.

benhoyt:
Okay, working on this now. Closing this PR and will open another shortly to make those changes.

PR #79, opened 2021-10-27
Service auto-restart and backoffs (refactor as state machine)

benhoyt:
This PR implements auto-restart and backoff functionality for Pebble services ([spec link](https://docs.google.com/document/d/1d6-h3UAt2VPUSvlkVF30l8iuDW8raRNkHbp5M6NUo1A/edit)). To achieve this, I've refactored the service handling into a state machine to make it consistent and easy to add the new states like `backoff` -- most of `internal/overlord/servstate/handlers.go` has been rewritten.

The PR adds the following new fields to the layer configuration under services > name (documented in README.md and defaults shown below):

```
on-success: restart
on-failure: restart
backoff-delay: 500ms
backoff-factor: 2.0
backoff-limit: 30s
```

The PR also refactors how/when we track "active services", and means the logs are saved across service stop/starts (and backoffs/restarts). So we're fixing a few long-standing issues at once with the refactor.

The list of states is:

```
const (
	stateInitial serviceState = iota
	stateStarting
	stateRunning
	stateTerminating
	stateKilling
	stateStopped
	stateBackoff
)
```

When you start a service it starts at `initial`, then moves to `starting` for the 1s okay-wait period, then transitions to `running`. When you stop a service it sends SIGTERM then goes to `terminating` and then `stopped` (via `killing` if SIGKILL is necessary). 

If a service is configured for auto-restart (the new default), it goes through the `backoff` state, and there's a `backoffTime` to keep track of the current backoff time. The new `Backoff*` fields in the service configuration specify how the backoff delay will advance.

This PR also adds a new status for the `Current` field in the API/CLI: `backoff`. Example CLI output:

```
$ pebble services
Service  Startup  Current
svc1     enabled  inactive
svc2     enabled  active
svc3     enabled  backoff
```

The PR also adds a `ServiceManager.SendSignal` function to send a signal to a running service. This is only used for tests right now, but we're planning to expose it via the API in another PR -- that will be easy to add.

The tests of restart and backoff currently use short timeouts, with some `time.Sleep()` calls. This seems to work very reliably in my tests, but if it becomes a problem we change change those to a wait/poll loop to wait for the logs the service writes, and check the timestamp deltas are what we expect (but it's a bit more work).


cmd/pebble/wait.go
```diff
@@ -110,16 +110,16 @@ func (wmx waitMixin) wait(id string) (*client.Change, error) {
 		}
 
 		for _, t := range chg.Tasks {
+			nowLog := lastLogStr(t.Log)
```
benhoyt, line 113: This fixes a bug where it was missing logs when task.Status != Doing (eg: Done).

internal/daemon/daemon.go
```diff
@@ -701,7 +701,17 @@ func New(opts *Options) (*Daemon, error) {
 		normalSocketPath:    opts.SocketPath,
 		untrustedSocketPath: opts.SocketPath + ".untrusted",
 	}
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput)
+
+	exitPebble := make(chan struct{})
```
benhoyt, line 705: This channel is wired up so that the servstate manager can close the channel to gracefully stop the Pebble server when "on-success: exit-pebble" or similar is specified in the service config.

internal/plan/plan.go
```diff
@@ -89,6 +95,44 @@ func (s *Service) Equal(other *Service) bool {
...
+func (s *Service) ParseBackoff() ([]time.Duration, error) {
+	if s.Backoff == nil {
+		return []time.Duration{
+			0,
+			10 * time.Second,
+			20 * time.Second,
+			40 * time.Second,
+			80 * time.Second,
+			160 * time.Second,
+			320 * time.Second,
```
benhoyt, line 109: This is the default backoff list.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+	err = s.cmd.Start()
 	if err != nil {
 		if outputIterator != nil {
 			_ = outputIterator.Close()
 		}
-		_ = logBuffer.Close()
+		_ = s.logs.Close()
 		return fmt.Errorf("cannot start service: %v", err)
 	}
+	startTime, _ := s.config.ParseStartTime() // ignore error; it's already been validated
```
niemeyer, line 320: The comment explaining why it's okay to ignore the error is longer than the code to check the error. :)

Instead of trusting some remote logic that must remain true for this to not be an awkward issue, we can just move this logic up and return err, simplifying assumptions and reading.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+	done := make(chan struct{})
 	go func() {
-		active.err = cmd.Wait()
-		_ = active.logBuffer.Close()
-		close(active.done)
+		waitErr := s.cmd.Wait()
+		close(done)
+		err := s.exited(waitErr)
+		if err != nil {
+			logger.Noticef("Cannot execute exited action: %v", err)
```
niemeyer, line 330: This message needs to be a bit more user friendly so people at the terminal can tell what it means.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
 
-	releasePlan()
+	return nil
+}
 
-	okay := time.After(okayWait)
-	select {
-	case <-okay:
+// okayWaitElapsed is called when the okay-wait timer has elapsed (and the
+// service is considered running successfully).
```
niemeyer, line 351: In contrast, that's a nice code doc, thanks.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+		logger.Noticef("Service %q stopped unexpectedly with code %d", s.config.Name, exitCode(s.cmd))
+		action, onType := getAction(s.config, waitErr == nil)
+		switch action {
+		case plan.ActionLog:
+			// Log has already been written above, no further log is necessary.
+			logger.Debugf("Service %q %s action is %q, transitioning to stopped state", s.config.Name, onType, action)
+			s.transition(stateStopped)
+
+		case plan.ActionExitPebble:
+			logger.Noticef("Service %q %s action is %q, triggering server exit", s.config.Name, onType, action)
```
niemeyer, line 392: We need to review this name before it's final. I'm hoping we can find a term that implies termination clearly (and in contrast with the service termination) without having the self-reference to the project name.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+
+	case stateTerminating, stateKilling:
+		s.stopped <- nil
+		s.transition(stateStopped)
+
+	default:
+		return fmt.Errorf("exited invalid in state %q", s.state)
 	}
-	// unreachable
+	return nil
```
niemeyer, line 422: Nice logic in this function. Makes the whole thing feel well organized, thank you for the change as a whole.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+
+	default:
+		return fmt.Errorf("sendSignal invalid in state %q", s.state)
 	}
-	cmd := active.cmd
+	return nil
+}
 
-	releasePlan()
+// stop is called to stop a running (and backing off) service.
```
niemeyer, line 476: I assume the comment actually means "running or backing off". Would be good to replace if so.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+	defer s.lock.Unlock()
 
-	syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM)
-	//cmd.Process.Signal(syscall.SIGTERM)
+	switch s.state {
+	case stateRunning:
+		// First send SIGTERM to try to terminate it gracefully.
+		err := syscall.Kill(-s.cmd.Process.Pid, syscall.SIGTERM)
+		if err != nil {
+			logger.Noticef("cannot send SIGTERM to process: %v", err)
```
niemeyer, line 486: s/cannot/Cannot/

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+		err := syscall.Kill(-s.cmd.Process.Pid, unix.SignalNum(signal))
+		if err != nil {
+			return err
+		}
+
+	case stateBackoffWait, stateTerminating, stateKilling, stateStopped:
+		return fmt.Errorf("cannot send signal while service is stopped or stopping")
+
+	default:
+		return fmt.Errorf("sendSignal invalid in state %q", s.state)
```
niemeyer, line 471: Missing "internal error: " prefix? Would also be nice to avoid referring to the method name, as code gets refactored and this message would be left behind. Maybe: `"internal error: attempted to signal service in state %q"

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
-	syscall.Kill(-cmd.Process.Pid, syscall.SIGTERM)
-	//cmd.Process.Signal(syscall.SIGTERM)
+	switch s.state {
+	case stateRunning:
+		// First send SIGTERM to try to terminate it gracefully.
+		err := syscall.Kill(-s.cmd.Process.Pid, syscall.SIGTERM)
+		if err != nil {
+			logger.Noticef("cannot send SIGTERM to process: %v", err)
+		}
+		s.stopped = make(chan error, 1)
```
niemeyer, line 488: I haven't looked into it, but given the complexity involved here, we need to make sure that this buffering is enough, and there's no risk of attempting to send twice on this channel. Doing so would mean a blocked goroutine and a frozen system.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+
+// backoffWaitElapsed is called when the current backoff's timer has elapsed,
+// to restart the service.
+func (s *service) backoffWaitElapsed() error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	switch s.state {
+	case stateBackoffWait:
+		logger.Debugf("Restarting service %q", s.config.Name)
```
niemeyer, line 509: Shouldn't that be a nice and more general log message (non-debug)? Or do we have that elsewhere already?

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
-					delete(m.services, req.Name)
-				}
-				releasePlan()
-			}
-			return nil
+	case stateBackoffWait:
+		s.transition(stateStopped)
+
+	default:
+		return fmt.Errorf("stop invalid in state %q", s.state)
```
niemeyer, line 496: Same note as above. Maybe: `"internal error: attempted to stop service in state %q"

internal/overlord/servstate/manager.go
```diff
@@ -338,20 +352,28 @@ func (m *ServiceManager) Replan() ([]string, []string, error) {
...
+	defer m.servicesLock.Unlock()
+
 	needsRestart := make(map[string]bool)
 	var stop []string
-	for name, activeService := range m.services {
-		if currentService, ok := m.plan.Services[name]; ok &&
-			currentService.Equal(activeService.originalPlan) {
-			continue
+	for name, s := range m.services {
+		s.lock.Lock()
```
niemeyer, line 361: Related to earlier comment: what's the benefit of having individual locks on every service?  Wouldn't it be enough to have a more encompassing lock similar to what we had before (servicesLock), since these are just management control tasks and not fast-paced or necessarily concurrent ones?  Easier to get right if so.

internal/plan/plan.go
niemeyer: factor = 2 :)

This seems too spaced out, btw, but we can tune later. People won't like waiting 10+ seconds until a process restarts after its first few exits.

internal/plan/plan.go
```diff
@@ -89,6 +95,44 @@ func (s *Service) Equal(other *Service) bool {
...
+			return nil, err
+		}
+		durations[i] = d
+	}
+	return durations, nil
+}
+
+// ParseStartTime returns the parsed version of the start time duration, or
+// the default if it's not specified.
+func (s *Service) ParseStartTime() (time.Duration, error) {
```
niemeyer, line 129: Seems better to have this as a ParseTime function of the plan package, instead of expecting every time attribute to come with an analogous function.

That said, why are we holding those values as strings instead of parsing them together with the rest of the service, and having the right values (including the default) in place at all times?

internal/plan/plan.go
```diff
@@ -105,6 +149,15 @@ const (
 	ReplaceOverride ServiceOverride = "replace"
 )
 
+type ServiceAction string
+
+const (
+	ActionUnset      ServiceAction = ""
+	ActionRestart    ServiceAction = "restart"
+	ActionExitPebble ServiceAction = "exit-pebble"
```
niemeyer, line 157: Per earlier note, let's please review this before it's final.

internal/plan/plan.go
```diff
@@ -105,6 +149,15 @@ const (
...
+	ActionLog        ServiceAction = "log"
```
niemeyer, line 158: I find this one a bit awkward as well. I mean, good logging should be in place regardless and is not analogous to the other options. But this is easy for us to tune before we're done.

niemeyer (changes requested):
Looks great overall, thanks Ben!

Lots of comments, but that's because the PR is long. Nothing major here.

client/services.go
```diff
@@ -80,9 +80,11 @@ type ServicesOptions struct {
...
 // ServiceInfo holds status information for a single service.
 type ServiceInfo struct {
-	Name    string         `json:"name"` // service name
-	Startup ServiceStartup `json:"startup"`
-	Current ServiceStatus  `json:"current"`
+	Name        string         `json:"name"` // service name
+	Startup     ServiceStartup `json:"startup"`
+	Current     ServiceStatus  `json:"current"`
+	BackoffNum  int            `json:"backoff-num"`
+	NumBackoffs int            `json:"num-backoffs"`
```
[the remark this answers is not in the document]
benhoyt: Done. Two things:

1) These specific fields have gone away. This was sending the backoff number and number of backoffs in the API for the CLI, which we've decided not to do for now.
2) As discussed, we've changed the backoff list of durations to a calculation-based approach, with new service fields backoff-delay, backoff-factor, and backoff-limit to customize the parameters.

cmd/pebble/cmd_services.go
```diff
@@ -59,10 +59,16 @@ func (cmd *cmdServices) Execute(args []string) error {
 	w := tabWriter()
 	defer w.Flush()
 
-	fmt.Fprintln(w, "Service\tStartup\tCurrent")
+	fmt.Fprintln(w, "Service\tStartup\tCurrent\tBackoff")
```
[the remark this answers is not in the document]
benhoyt: Done. I've removed the backoff numbers and just added a status "backoff", which is shown in the "Current" column.

internal/daemon/api_services.go
```diff
@@ -27,9 +27,11 @@ import (
...
 
 type serviceInfo struct {
-	Name    string `json:"name"`
-	Startup string `json:"startup"`
-	Current string `json:"current"`
+	Name        string `json:"name"`
+	Startup     string `json:"startup"`
+	Current     string `json:"current"`
+	BackoffNum  int    `json:"backoff-num"`
+	NumBackoffs int    `json:"num-backoffs"`
```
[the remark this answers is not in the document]
benhoyt: As above, these are gone now.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
 
-	maxLogBytes = 100 * 1024
+const (
+	stateInitial serviceState = iota
+	stateStarting
+	stateRunning
+	stateTerminating
+	stateKilling
+	stateStopped
+	stateBackoffWait
```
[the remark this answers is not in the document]
benhoyt: Done.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+func (s serviceState) String() string {
+	switch s {
+	case stateInitial:
+		return "initial"
+	case stateStarting:
+		return "starting"
+	case stateRunning:
+		return "running"
+	case stateBackoffWait:
+		return "backoff-wait"
```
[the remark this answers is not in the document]
benhoyt: Done.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+		return "killing"
+	case stateStopped:
+		return "stopped"
+	default:
+		return "unknown"
+	}
+}
+
+// service holds the information for a service under our control.
+type service struct {
```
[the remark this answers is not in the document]
benhoyt: Done, and renamed some of the `s` variables back to `service`. However, I've used `serviceData` for the type name, as I really like `serviceState` for the state machine state (and `ServiceStatus` is already a type, returned as part of the struct that  `Services()` returns).

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+	lock         sync.Mutex
+	manager      *ServiceManager
+	state        serviceState
+	config       *plan.Service
+	logs         *servicelog.RingBuffer
+	started      chan *int
```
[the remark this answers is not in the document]
benhoyt: Done, good call to have them symmetrical.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+	m.state.Unlock()
+	if err != nil {
+		return fmt.Errorf("cannot get service request: %w", err)
+	}
+
 	releasePlan, err := m.acquirePlan()
 	if err != nil {
+		return fmt.Errorf("cannot acquire plan lock: %w", err)
+	}
+	config, ok := m.plan.Services[request.Name]
```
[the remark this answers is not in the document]
benhoyt: Thanks, done.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+	}
+	releasePlan()
+
+	m.servicesLock.Lock()
+	s := m.services[config.Name]
+	if s != nil && s.getStateLocked() != stateStopped {
+		m.servicesLock.Unlock()
+		m.state.Lock()
+		task.Logf("Service %q already started.", config.Name)
+		m.state.Unlock()
```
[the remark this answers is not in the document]
benhoyt: Yeah, you weren't wrong, it was a bit crazy. I've made most of these changes, and it really simplifies doStart/doStop. In particular:

1) Unified: gotten rid of the individual `serviceData.lock` and just use the "global" `servicesLock` everywhere. As you say, this locking/unlocking shouldn't be happening a ton, so it's not perf-critical.
2) Added helper methods to get the service, moving the `servicesLock` lock/unlock out of doStart/doStop and into those methods, where it can be a simple lock/defer unlock at the top.
3) Moved the lock-task.Logf-unlock dance to a helper function to clean that up.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
 
+	// Wait for a small amount of time, and if the service hasn't exited,
+	// consider it a success.
+	timeout := time.After(okayWait + 100*time.Millisecond)
+	for {
+		select {
+		case exitCode := <-s.started:
+			if exitCode != nil {
+				m.removeService(config.Name)
+				return fmt.Errorf("cannot start service: exited quickly with code %d", *exitCode)
```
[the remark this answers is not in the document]
benhoyt: Done. Now the error formatting matches the stop case too: `return fmt.Errorf("cannot start service: %w", err)`

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+			}
+			// Started successfully (ran for small amount of time without exiting).
+			return nil
+		case <-tomb.Dying():
+			// Start cancelled (shouldn't really happen, so disallow for simplicity).
+			logger.Noticef("Cannot cancel start (service %q)", config.Name)
```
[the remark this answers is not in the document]
benhoyt: Previously `tomb.Dying` was ignored / not referenced at all. I've added it here so at least there's a log, but I don't actually change the behavior (ignoring it). I've clarified the comment and log message to make this explicit. If we want to handle more gracefully later we can, but realistically I don't think it makes much sense to cancel the start within the 1s okay-wait period, so I'm punting on that for now. Again, what I have now is a slight (logging) improvement over what we had for this case.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+		case <-timeout:
+			// Should never happen, but don't block just in case we get something wrong.
+			m.removeService(config.Name)
+			return fmt.Errorf("internal error: timed out waiting for start")
```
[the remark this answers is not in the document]
benhoyt: Yeah, it's because the state machine should always send to the `started` channel, either in `okayWaitElapsed` after it's been running fine for 1s, or when it exits within 1s. So this is only to catch a coding/logic bug. I've clarified in the comments (for the stop case as well).

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +54,528 @@ var (
...
+		s.cmd.Env = append(s.cmd.Env, k+"="+v)
 	}
-	logBuffer := servicelog.NewRingBuffer(maxLogBytes)
+
+	// Set up stdout and stderr to write to log ring buffer.
 	var outputIterator servicelog.Iterator
-	if m.serviceOutput != nil {
-		outputIterator = logBuffer.TailIterator()
+	if s.manager.serviceOutput != nil {
+		outputIterator = s.logs.HeadIterator(0)
```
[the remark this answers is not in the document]
benhoyt: Heh, not really either. `TailIterator` was fine before, because we weren't keeping logs after the service died or between runs. Now we are keeping logs from previous runs, but we don't want to copy from the beginning of the ring buffer to `serviceOutput` again (that would duplicate the logs), so start iterating from the head when we restart the service.

internal/overlord/servstate/handlers.go
benhoyt: True. This code is actually gone now, as we're parsing it to a duration directly in the plan code now.

internal/overlord/servstate/handlers.go
benhoyt: Note that this is not actually something a user should ever see, as the errors `exited()` returns are both internal "shouldn't happen" error. I've tweaked the wording slightly, though. So if we did have a bug they'd see something like:

```
Cannot transition state after service exited: internal error: exited invalid in state "backoff"
```

internal/overlord/servstate/handlers.go
benhoyt: Done. Per discussion the other night, I've updated this to just "exit" (and "log" to "ignore").

internal/overlord/servstate/handlers.go
benhoyt: Yep, updated.

internal/overlord/servstate/handlers.go
benhoyt: Done.

internal/overlord/servstate/handlers.go
benhoyt: This isn't actually an internal error, but is bubbled up to the user in `ServiceManager.SendSignal()`. However, good point about the method name/refactoring, so I've removed the `sendSignal` name and now have: `"sending signal invalid in state %q"`.

internal/overlord/servstate/handlers.go
benhoyt: Yes, good point. This channel is sent to in state "terminating" or "killing" (after SIGTERM or SIGKILL has been sent), or in state "killing" after the kill wait has elapsed. However, if the kill wait elapses it sends a stop error on the channel (but remains in "killing" state for lack of something better to do), so if the process exits after that (unlikely but possible) it'll mean a blocked goroutine. So a size of 2 is sufficient -- I've updated it.

internal/overlord/servstate/handlers.go
benhoyt: Yeah, you're right. I've made this a `logger.Noticef` that happens whenever the service starts. That's an important enough event!

internal/overlord/servstate/handlers.go
benhoyt: Similar to the `start()` one, this may actually bubble up to the user.

internal/overlord/servstate/manager.go
benhoyt: Per above, I've merged everything into one `servicesLock`.

internal/plan/plan.go
benhoyt: As mentioned above, this is gone now as it's parsed directly into a duration in `plan.Service`.

internal/plan/plan.go
benhoyt: Per discussion, updated to just "exit".

internal/plan/plan.go
benhoyt: Per discussion, updated to just "ignore".

internal/plan/plan.go
benhoyt: Yep, with the new values it's much faster: 0.5s, 1s, 2s, ... up to 30s.

internal/daemon/daemon.go
benhoyt: Thanks. It does need to be wired up somehow: at the end of the day `daemon.Stop()` needs to be called, and servstate is a dependency of overlord (which is a dependency of daemon), not the other way around. But I've simplified this a bit by wiring up `daemon.Stop` a bit more directly.

cmd/pebble/wait.go
benhoyt: This is the one thing I still need to look into -- will do tomorrow.

benhoyt: For now, I've reverted this change until I understand it better. We can always fix in a subsequent PR if there's fixing that needs to happen.

niemeyer (changes requested):
Thanks for these changes, Ben.  Here is a second pass:

README.md
```diff
@@ -212,6 +212,43 @@ services:
...
+
+        # (Optional) If the service exits, perform this action: either restart
+        # the service after a backoff wait, exit the Pebble server, or simply
+        # ignore it (log the failure and transition the service to the
+        # "stopped" status). Default is "restart".
+        #
+        # If on-failure is set and the service's exit code is nonzero,
+        # on-failure takes precedence; if on-success is set and the service's
+        # exit code is zero, on-success takes precedence.
+        on-exit: restart | exit | ignore
```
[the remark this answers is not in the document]
benhoyt: Thanks. Yeah, it was a bit messy. I agree with the simplification (removal of `on-exit`) and change to `halt` -- both done.

README.md
```diff
@@ -212,6 +212,43 @@ services:
...
+        # (Optional) If the service exits with a zero exit code, override
+        # on-exit and perform this action (see on-exit for details).
+        on-success: restart | exit | ignore
+
+        # (Optional) Initial backoff delay for the "restart" exit action.
+        # Default is half a second ("500ms").
+        backoff-delay: <duration>
+
+        # (Optional) After each backoff, the backoff delay is multiplied by
+        # this factor to get the next backoff time. Must be greater than or
```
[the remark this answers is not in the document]
benhoyt: Done.

README.md
```diff
@@ -212,6 +212,43 @@ services:
...
+        # equal to one. Default is 2.0.
+        backoff-factor: <factor>
+
+        # (Optional) Limit for the backoff delay: when multiplying by
+        # backoff-factor to get the next backoff time, if the result is
```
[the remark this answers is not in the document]
benhoyt: Done.

client/services.go
```diff
@@ -80,11 +80,9 @@ type ServicesOptions struct {
 
 // ServiceInfo holds status information for a single service.
 type ServiceInfo struct {
-	Name        string         `json:"name"` // service name
-	Startup     ServiceStartup `json:"startup"`
-	Current     ServiceStatus  `json:"current"`
-	BackoffNum  int            `json:"backoff-num"`
-	NumBackoffs int            `json:"num-backoffs"`
+	Name    string         `json:"name"` // service name
```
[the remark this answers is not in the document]
benhoyt: Yeah, I don't know why I'd put that in there in the first place. :-) Removed.

README.md
```diff
@@ -212,6 +212,43 @@ services:
...
+        # greater than this value, it is capped to this value. Default is
+        # half a minute ("30s").
+        backoff-limit: <duration>
+
+        # (Optional) After the service has been running successfully for this
+        # time, reset the backoff time. Default is ten seconds ("10s").
+        backoff-reset: <duration>
```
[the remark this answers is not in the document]
benhoyt: Good call, I like this simplification / removal of `backoff-reset`.

internal/daemon/daemon.go
```diff
@@ -702,16 +702,10 @@ func New(opts *Options) (*Daemon, error) {
...
-		if err != nil {
-			logger.Noticef("Cannot stop daemon: %v", err)
-		}
-	}()
-
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, exitPebble)
+	stopDaemon := func() error {
+		return d.Stop(nil)
+	}
+	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, stopDaemon)
```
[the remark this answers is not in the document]
benhoyt: Yeah, I'd seen the restart functionality and `Daemon.Stop`, it was just a matter of wiring it up. However, I see now that the overlord already has the restart functionality exposed (via `HandleRestart`). Thanks - fixed to use that existing wiring.

internal/overlord/servstate/handlers.go
```diff
@@ -187,65 +194,80 @@ func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
...
 			if err != nil {
 				return fmt.Errorf("cannot stop service: %w", err)
 			}
 			// Stopped successfully.
 			return nil
 		case <-tomb.Dying():
-			// Stop cancelled (shouldn't really happen, so disallow for simplicity).
-			logger.Noticef("Cannot cancel stop (service %q)", request.Name)
+			// Stop operation cancelled (ignore for simplicity).
+			logger.Noticef("Ignoring cancellation of stop for service %q", request.Name)
```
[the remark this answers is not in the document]
benhoyt: It's hard to tell whether you realize or not, but aborting (cancelling) the `start` or `stop` only aborts the starting and stopping "changes" -- it's not referring to aborting the process itself. So here's what I've done:

In `doStop`, for aborting a stop operation, I'm not sure it makes sense to say "Waiting for service %q to stop...", because you've just tried to abort/cancel it from stopping. Aborting a stop would be trying to resuscitate / restart the service, but who knows what state it's in. As your comment mentions, we've already sent the terminate/kill signal at this point, so we can't do much except log it. In any case, I've changed it to the following:

```go
	case <-tomb.Dying():
			// User tried to abort the stop, but SIGTERM and/or SIGKILL have
			// already been sent to the process, so there's not much more we
			// can do than log it.
			logger.Noticef("Cannot abort stop for service %q, kill signals already sent", request.Name)
```

In `doStart`, for aborting a start operation, we can actually abort/cancel the start by killing the process. So I've updated this to send SIGKILL and report the `start` change as having failed. Alternatively, we could just disallow aborting a stop operation (you've only got 1 second to do so!) and log "Cannot abort start for service %q, starting up" or similar.

internal/overlord/servstate/handlers.go
```diff
@@ -187,65 +194,80 @@ func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
...
 		case <-timeout:
-			// Should never happen, but don't block just in case we get something wrong.
+			// Should never happen, because killTimeElapsed and exited both send to the
+			// "stopped" channel, but don't block in case we got something wrong.
 			return fmt.Errorf("internal error: timed out waiting for stop")
```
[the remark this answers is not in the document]
benhoyt: Yeah, you're right, there is a timing race here (could happen, even if unlikely). What I meant by should never happen is that the `stopped` channel should always be sent to: either due to the process exiting or the kill timeout elapsing (for start, the `started` channel should always be sent to as well: either due to the process exiting or the 1s wait elapsing).

But I think it's simpler and better to just avoid this, and if the code isn't sending to the `started` / `stopped` channel correctly, the tests should catch that and time out.

internal/overlord/servstate/handlers.go
```diff
@@ -161,24 +120,72 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+			return nil
+		case stateBackoff, stateStopped:
+			// Start allowed in "stopped" and "backoff" states, handle below.
+		default:
+			// Cannot start service while terminating or killing, handle in start().
+			return service
+		}
+	}
+
+	if service == nil {
```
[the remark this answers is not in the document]
benhoyt: Good call -- simplified.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +53,584 @@ var (
...
+			config:  config.Copy(),
+			logs:    servicelog.NewRingBuffer(maxLogBytes),
+		}
+		m.services[config.Name] = service
+	} else {
+		service.backoffNum = 0
+		service.backoffTime = 0
+		service.transition(stateInitial)
+	}
+	service.started = make(chan error, 1)
```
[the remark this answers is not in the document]
benhoyt: Good call, done.

internal/overlord/servstate/handlers.go
```diff
@@ -51,183 +53,584 @@ var (
...
+}
+
+func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
 	m.state.Lock()
-	req, err := TaskServiceRequest(task)
+	request, err := TaskServiceRequest(task)
 	m.state.Unlock()
 	if err != nil {
-		return err
+		return fmt.Errorf("cannot get service request: %w", err)
```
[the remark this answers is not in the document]
benhoyt: Yeah, I had added this error decoration but you're right, it doesn't add anything, and the underlying function already prepends "internal error" as appropriate. Removed.

README.md
```diff
@@ -213,42 +213,31 @@ services:
...
-        # exit code is zero, on-success takes precedence.
-        on-exit: restart | exit | ignore
-
-        # (Optional) If the service exits with a nonzero exit code, override
-        # on-exit and perform this action (see on-exit for details).
-        on-failure: restart | exit | ignore
+        # (Optional) If the service exits with a zero exit code, perform
+        # this action: either restart the service after a backoff wait, halt
+        # (terminate the Pebble server), or simply ignore it (log the failure
+        # and transition the service to the "stopped" status). Default is
```
niemeyer, line 219: Sorry for missing this earlier: I think we're not yet exposing (and probably should not) the "stopped" status that is used by the internal state machine, but rather showing this up as "inactive".  We need to be a bit careful in terms of consistency between the internal logic and what we have in the user interface. The sentence also seems slightly broken up, when perhaps we could make it more linear. Giving it a try:
```
# (Optional) The on-success option defines what happens when the service exits with
# a zero exit code. Possible values are: 'restart' (default) which restarts the service after
# the backoff time, 'halt' which stops Pebble and exits, and 'ignore' which does nothing
# in that situation.

File: README.md [hunk 44]
```diff
@@ -213,42 +213,31 @@ services:
...
+        # "restart".
+        on-success: restart | halt | ignore
 
-        # (Optional) If the service exits with a zero exit code, override
-        # on-exit and perform this action (see on-exit for details).
-        on-success: restart | exit | ignore
+        # (Optional) If the service exits with a nonzero exit code, perform
+        # this action (see on-success for details). Default is "restart".
+        on-failure: restart | halt | ignore
```
niemeyer, line 225: Assuming the above looks good, we might spell that out completely in a similar amount of text:
```
# (Optional) The on-failure option defines what happens when the service exits with
# a non-zero exit code. Possible values are: 'restart' (default) which restarts the service after
# the backoff time, 'halt' which stops Pebble and exits, and 'ignore' which does nothing
# in that situation.
```

internal/overlord/servstate/handlers.go
```diff
@@ -117,25 +117,25 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+		// User tried to abort the start, sending SIGKILL to process is about
+		// the best we can do.
+		m.removeService(config.Name)
+		m.servicesLock.Lock()
+		defer m.servicesLock.Unlock()
+		err := syscall.Kill(-service.cmd.Process.Pid, syscall.SIGKILL)
+		if err != nil {
+			return fmt.Errorf("start aborted, but cannot send SIGKILL to process: %v", err)
 		}
+		return fmt.Errorf("start aborted, sent SIGKILL to process")
```
niemeyer, line 138: Do we have context for which service is dying in these messages above, or do we need to add some sort of `service %q`?

internal/overlord/servstate/handlers.go
```diff
@@ -148,34 +148,34 @@ func (m *ServiceManager) serviceForStart(task *state.Task, config *plan.Service)
...
 		service.backoffNum = 0
 		service.backoffTime = 0
 		service.transition(stateInitial)
+		return service
+	default:
+		// Cannot start service while terminating or killing, handle in start().
+		return service
 	}
-	service.started = make(chan error, 1)
-	return service
```
niemeyer: Thanks for the changes. :+1:

internal/overlord/servstate/handlers.go
```diff
@@ -216,12 +215,10 @@ func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
...
-			// Stop operation cancelled (ignore for simplicity).
-			logger.Noticef("Ignoring cancellation of stop for service %q", request.Name)
-		case <-timeout:
-			// Should never happen, because killTimeElapsed and exited both send to the
-			// "stopped" channel, but don't block in case we got something wrong.
-			return fmt.Errorf("internal error: timed out waiting for stop")
+			// User tried to abort the stop, but SIGTERM and/or SIGKILL have
+			// already been sent to the process, so there's not much more we
+			// can do than log it.
+			logger.Noticef("Cannot abort stop for service %q, kill signals already sent", request.Name)
```
niemeyer, line 221: Either s/signals/signal/ or or s/kill//, I guess?

niemeyer (approved):
Final notes below. LGTM when you're happy. It's a bit dense to be sure it's all lined up exactly right, but I think we got the interface right and we can nail down any issues as we see them.

Thanks for this whole refactoring!

internal/overlord/servstate/handlers.go
benhoyt: Yes. For example, if you press Ctrl-C in the 1s okay-wait period while running the `pebble start` CLI command, it calls "abort" on the change, and the error is wrapped with the service name. You get `Start service "test" (start aborted, sent SIGKILL to process)`.

README.md
benhoyt: Thanks. Updated to basically your wording with a couple of tweaks.

README.md
benhoyt: Thanks. Updated to basically your wording with a couple of tweaks.

internal/overlord/servstate/handlers.go
benhoyt: Thanks - I updated to the latter.

cmd/pebble/wait.go
benhoyt: To close the loop on this, I opened https://github.com/snapcore/snapd/pull/11121 to discuss with the Snap folks. We should probably not be using `wait()` here. We could use our new Pebble "long polling" wait variant instead anyway, instead of polling every 100ms. It's not like Snap where we really need lots of progress updates.

PR #81, opened 2021-11-02
Rename "use-terminal" to "terminal"; add separate "interactive" parameter

benhoyt:
Per discussion on https://github.com/canonical/pebble/pull/78, we're going to switch the `use-terminal` parameter to be called just `terminal`, and add a new `interactive` bool parameter to mean "also assign stdin to this PTY. If input and output are from/to the terminal, interactive=true and terminal=true. If input is a pipe, but output is the terminal, terminal=true. Otherwise, both false or not provided.

Note that this backwards-incompatible change doesn't actually affect the Python Operator Framework, as that only uses exec in terminal=false mode. It only affects the Pebble CLI (and Go client). So shouldn't affect any real users at this point.


internal/wsutil/wsutil_linux.go
```diff
@@ -26,42 +26,31 @@ import (
...
+	for {
+		buf, ok := <-in
+		if !ok {
+			r.Close()
+			logger.Debugf("Sending write barrier")
+			err := conn.WriteMessage(websocket.TextMessage, endCommandJSON)
 			if err != nil {
-				logger.Debugf("Got err writing %s", err)
-				break
+				logger.Debugf("Got err writing barrier %s", err)
```
niemeyer, line 39: This was already an issue before in this file, and we are probably inheriting this code from elsewhere, but for the record these log messages aren't great as they have no context to them, and even knowing the context they feel somewhat unclear (e.g. "Got err writing" below.. writing what, to what, when, etc).

Just a note. It was already like this before and not worth going around fixing everything in this PR, but would be nice to clean it up eventually.

internal/wsutil/wsutil_linux.go
```diff
@@ -26,42 +26,31 @@ import (
...
+		if err != nil {
+			logger.Debugf("Got err writing %s", err)
+			break
+		}
+	}
 
-	return readDone, writeDone
+	closeMsg := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+	conn.WriteMessage(websocket.CloseMessage, closeMsg)
+	r.Close()
```
niemeyer, line 53: That looks a lot more readable now, thanks!

niemeyer (approved):
Thank you!

hpidcock (approved):

PR #87, opened 2021-11-29
Crossport restart package from snapd.

hpidcock:
This PR crossports the restart package and related changes from snapd.

It brings pebble inline with the snapd upstream to simplify state by removing the restart concerns into its own package with good separation.

Relevant commits pulled across https://github.com/snapcore/snapd/commits/master/overlord/restart

### DIFF
--- a/internal/daemon/api.go
+++ b/internal/daemon/api.go
@@ -20,6 +20,7 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/canonical/pebble/internal/overlord"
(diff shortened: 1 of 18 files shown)


benhoyt (approved):
Looks good to me. I scanned the ported code and reviewed the servstate and other Pebble-specific changes.

niemeyer (approved):
Assuming most of this comes from snapd and was previously reviewed, did a shallow pass and it does look nice. Thanks for porting over, Harry.

PR #88, opened 2021-11-29
Fix various missing error checks and missing wait mixin descriptions.

hpidcock:
Used staticcheck to detect a few missing error checks and unused code of concern.

### DIFF
--- a/internal/daemon/api_files.go
+++ b/internal/daemon/api_files.go
@@ -325,6 +325,9 @@ func listFiles(path, pattern string, itself bool) ([]fileInfoResult, error) {
 		matched := true
 		if pattern != "" {
 			matched, err = pathpkg.Match(pattern, name)
+			if err != nil {
+				return nil, err
+			}
 		}
 		if matched {
 			fullPath := pathpkg.Join(dir, name)
--- a/internal/overlord/servstate/manager_test.go
+++ b/internal/overlord/servstate/manager_test.go
@@ -768,6 +768,7 @@ services:
 	svc := s.waitUntilService(c, "test2", func(svc *servstate.ServiceInfo) bool {
 		return svc.Current == servstate.StatusActive
 	})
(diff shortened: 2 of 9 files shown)


hpidcock:
@benhoyt @niemeyer might be wise to add in a staticcheck github action so we catch this stuff earlier.

internal/daemon/api_files.go
```diff
@@ -325,6 +325,9 @@ func listFiles(path, pattern string, itself bool) ([]fileInfoResult, error) {
 		matched := true
 		if pattern != "" {
 			matched, err = pathpkg.Match(pattern, name)
+			if err != nil {
```
benhoyt, line 328: Good catch, thanks!

internal/overlord/servstate/manager_test.go
```diff
@@ -765,7 +765,7 @@ services:
 
 	// Start service and wait till it starts up the first time.
 	chg := s.startServices(c, []string{"test2"}, 1)
-	svc := s.waitUntilService(c, "test2", func(svc *servstate.ServiceInfo) bool {
+	s.waitUntilService(c, "test2", func(svc *servstate.ServiceInfo) bool {
```
benhoyt, line 768: We weren't actually using the return value from this helper function, so I just got rid of it (it was always non-nil anyway, otherwise it fatal'd the test).

benhoyt (approved):
LGTM, thanks! I'd be in favour of a staticcheck type of thing (assuming it didn't find false positives or we could easily tune it not to).

It wouldn't hurt to add `go vet ./...` into the "Test and build" step of our GitHub actions CI. I'm sure that wouldn't catch as much as staticcheck, but still seems worth doing now that Pebble is `go vet` clean.

niemeyer (approved):
LGTM, thanks!

PR #89, opened 2021-11-29
Add finalizer support to block shutdown for a remote finalizer

hpidcock:
When juju is using pebble in Kubernetes pods, each pebble instance is running
in its own container. When Kubernetes kills a pod, a juju charm might need to
first run some commands, change the file system etc before pebble disappears.

This change gives juju charms time to run the required steps by blocking shutdown
and then allows juju to notify pebble that it has performed all finalization operations
using the /v1/finalize api.

It is noted that there is a limited time period for both the finalizer to run and pebble
shutdown (including service process termination). In future these timings will be
tuneable from the run cli and also from juju/charm side so that in the case of
Kubernetes the termination grace period can be extending for charms/services that
need more time.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -63,6 +63,9 @@ type Options struct {
 	// ServiceOuput is an optional io.Writer for the service log output, if set, all services
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
+
+	// Finalizer is true when the daemon requires an external finalizer to run before it can stop.
+	Finalizer bool
 }
 
 // A Daemon listens for requests and routes them to the right command
@@ -81,6 +84,7 @@ type Daemon struct {
 	tomb                tomb.Tomb
 	router              *mux.Router
 	standbyOpinions     *standby.StandbyOpinions
+	finalizer           chan bool
 
 	// set to remember we need to restart the system
 	restartSystem bool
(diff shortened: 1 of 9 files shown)


internal/daemon/daemon.go
```diff
@@ -695,13 +730,27 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 	return state.ErrExpectedReboot
 }
 
+// finalize is used by v1Finalize to allow the Daemon to stop when
+// launched with Finalizer set to true.
+func (d *Daemon) finalize() {
+	if d.finalizer == nil {
+		return
+	}
+	select {
```
benhoyt, line 739: Why is the select...default needed when `d.finalizer` is a 1-buffered channel? Why not `close()` an unbuffered channel instead of trying to send?

benhoyt (approved):
Approach and code looks good to me. Just one nit comment.

internal/daemon/daemon.go
hpidcock: can't close a closed channel, this means that the /v1/finalize api can be called as many times as it wants and only works once without needing more complextity

niemeyer (changes requested):
Thanks for this, Harry. I've skimmed through the patch, and overall it seems nice and simple. The points that would be nice to catch up on are mainly the API and terms used.

Enumerating so it's easier to discuss:

1) The overall change seems centered around the idea that we're "running an external finalizer", but from the perspective of Pebble, it has no visibility of such a finalizer. The only thing it seems to be doing is waiting for confirmation, which feels quite simple in contrast with "finalization", so maybe it would be more clear here to organize things around that idea. For example, the flag might be --confirm-shutdown instead (or --confirm-halt, pending decision on #85).

2) Related to (1), I wonder how we should design the API to prepare for upcoming features. It sounds like shutdown is a larger theme that we could set ourselves up for looking into, and having an independent endpoint just to confirm the shutdown will render us unable to expand it further.  For example, just to get us on the right mindset: do we have an API for restarting Pebble today, or for shutting it down?  Where would such an API live?  Do we have something analogous on snapd that we should look into?

3) Once we start waiting for such a confirmation, what happens when the container is supposed to die for being restarted because something went wrong and the check (or whatever) instructed Pebble to shutdown?   Do we have any monitoring in juju that will be ready to confirm the operation, so that we don't end up with containers hanging waiting for a confirmation that will never come, or will at least not come timely?

We'll probably benefit from chatting a bit about those details interactively.

benhoyt:
@hpidcock Harry, going to close this for now at least, seeing this is going to require significant rework -- feel free to reopen when it's ready for review again (or post a new PR, whatever you think is best).

Issue #83, opened 2021-11-12
Structured logs


mmanciop:
Pebble logs entries as plain text with a specified pattern. While that is functional for human consumption, it is far more error probe to process plain text log entries than structured ones, e.g., JSON. One of the use-cases to process Pebble logs in the observability domain would be to correlate downtime with issues reported by Pebble, e.g., services that won’t start.

benhoyt:
My general thinking is that logs are for human consumption (while debugging or diagnosing issues), and that if we wanted to support automated observability use cases we should push metrics for those. But maybe structured logs are a simple way to kill two birds with one stone ([so to speak](https://www.youtube.com/watch?v=71gilEP4aJY) :-).

@jnsgruk @niemeyer What are your thoughts?

mmanciop:
Metrics and logs fulfill different needs in O11y: logs are fundamentally events*; metrics are mostly about counting the recurrence of events that are planned ahead, and generally work poorly to document specifics of which failures occur, even when the possible failures are strictly enumerated ahead of time (which happens so seldom in nowadays systems), as you lose the specific timestamp for single occurrences of the failure (and thus the opportunity for correlation with other events). What often happens is that metrics like “how many errors occur” and calculated by counting how many events of a certain nature occur in a given timeframe.

* What is the difference between logs and events is matter of much debate, usually preceded by the ingestion of intoxicating substances of choice)

jnsgruk:
I'm not against the idea of adding more structured logging -- I think it's likely a good thing as people take sidecar charms to production.

That said, I'm with Ben that we need to find a way to keep Pebble logs human readable too, so perhaps the Pebble client could provide that functionality through a command line option or similar?

mmanciop:
I think we should have a look at how cloud providers to structured logging for K8s control plane. I agree that what one sees in `kubectl logs` should be human readable, but it seems to me https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/ could be of help here.

petevg:
Hrm. I'm skeptical that structured logs are a good idea here. A charm will often be reporting on the behavior of an arbitrary service, which may or may not present its own logs in a way that can be consumed safely and reliably as a json object.

Wrapping things in JSON additionally makes this much less friendly to humans troubleshooting a service, as well as requiring extra complexity in the suite of tools that are passing these logs back and forth.

[...]

PR #90, opened 2021-11-29
Simplify client.Logs "N" parameter and avoid pointer

benhoyt:
For rationale for this change, see discussion at https://github.com/canonical/pebble/issues/83#issuecomment-973980643.

This removes the pointer on the `N *int` (number of logs) parameter to be just a plain `int`. When calling `Logs` (non-follow), the default is the server-specified default (currently 30). When calling `FollowLogs`, the default is 0 (don't show any buffered logs).

This PR also changes the server API to have the same semantics (except that the default is "unset n" rather than "zero N"). This shouldn't affect any real users -- the `pebble logs` CLI command already sends through an explicit "n" value anyway. Any negative "n" values other than -1 are now an error (but this won't affect any clients).

I've also changed the default number of logs from 10 to 30 (see discussion on this PR).

Note: this is a breaking change in the Go client. However, it's a trivial fix, and at this point we probably have exactly one internal user of the Go client (the Pebble CLI itself), so I think it's reasonable.

### DIFF
--- a/client/logs.go
+++ b/client/logs.go
@@ -27,22 +27,21 @@ import (
 )
 
 const (
-	AllLogs = -1
-
 	logReaderSize = 4 * 1024
 )
 
 // LogsOptions holds the options for a call to Logs or FollowLogs.
 type LogsOptions struct {
-	// Function called to write a single log to the output (required).
+	// WriteLog is called to write a single log to the output (required).
 	WriteLog func(entry LogEntry) error
 
-	// The list of service names to fetch logs for (nil or empty slice means
-	// all services).
+	// Services is the list of service names to fetch logs for (nil or empty
+	// slice means all services).
 	Services []string
 
-	// Number of logs to fetch (before following if calling FollowLogs). To
-	// fetch all buffered logs, set N to AllLogs.
+	// N defines the number of log lines to return from the buffer. In follow
+	// mode, the default is zero, in non-follow mode it's server-defined
+	// (currently 30). Set to -1 to return the entire buffer.
 	N int
 }
 
@@ -69,7 +68,9 @@ func (client *Client) logs(ctx context.Context, opts *LogsOptions, follow bool) 
 	for _, service := range opts.Services {
 		query.Add("services", service)
 	}
-	query.Set("n", strconv.Itoa(opts.N))
+	if opts.N != 0 {
+		query.Set("n", strconv.Itoa(opts.N))
+	}
 	if follow {
 		query.Set("follow", "true")
 	}
--- a/client/logs.go
+++ b/client/logs.go
@@ -27,9 +27,12 @@ import (
 )
 
 const (
+	AllLogs = -1
+
 	logReaderSize = 4 * 1024
 )
 
+// LogsOptions holds the options for a call to Logs or FollowLogs.
 type LogsOptions struct {
 	// Function called to write a single log to the output (required).
 	WriteLog func(entry LogEntry) error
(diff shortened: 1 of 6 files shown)


client/logs.go
```diff
@@ -38,10 +41,9 @@ type LogsOptions struct {
 	// all services).
 	Services []string
 
-	// Total number of logs to fetch (before following if calling FollowLogs).
-	// If nil, use Pebble's default (10). If negative, fetch all buffered logs.
-	// If set to zero when calling FollowLogs, write no logs before following.
-	N *int
+	// Number of logs to fetch (before following if calling FollowLogs). To
+	// fetch all buffered logs, set N to AllLogs.
+	N int
```
niemeyer, line 46: This is solving part of the problem by removing the pointer usage and making it a bit more clear, but it's still missing part of the problem: the current _server_ API is confusing. The default value is being set by the command, when it should really be in the server when N is unset. Otherwise, what sense it makes to request logs, and then get neither what's buffered nor what is coming after (follow)? It's an awkward reply, and we're addressing that not in the server nor in the client, but way down in the command. I think we should really fix the server.

My suggestion in the issue was to have the default behavior when N is zero being a nice outcome in both cases: if following, just follow and ignore the buffer. If not following, show a default number server-defined. To show all logs, we don't need any special logic or constant, since the client can trivially provide a large number as desired

To keep compatibility, we can allow the server to support -1 specifically, but it needs to be clear about it. Right now that whole side of the logic looks somewhat accidental instead of intentional when looking at the API code

niemeyer (changes requested):
Thanks for working on this, Ben. Some notes:

client/logs.go
benhoyt: Ah, I'd originally misinterpreted your comment [here](https://github.com/canonical/pebble/issues/83#issuecomment-973980643) as saying we shouldn't change the server API, but I see now I was misreading it. Well, there are a few aspects to this:

1) Do we need a special case for "all"? I think we do. For one, it's a more direct API than specifying n=1000000 or whatever, and the user of the API won't have to think "wait, *how* big do I have to make it?" or worse, choose something like n=10000 that's not necessarily big enough. Also, the `RingBuffer` implementation and API needs to know, because it calls `TailIterator()` if n<0, or `HeadIterator(n)` if n>=0. In the latter case, `HeadIterator` scans back the given number of lines byte-by-byte, so it'd be quite inefficient to scan through the whole buffer if you just want it all.

2) Should we change the server API? I don't think it's particularly awkward -- it doesn't involve pointers as it's just a query parameter. Not specifying the `n` query parameter means "use the server default" in both follow and non-follow cases, which makes sense to me. The only special case is negative `n`, which means "all". We could change that to literally `?n=all` (we use `-n=all` in the CLI already).

3) I think it's confusing to have the Go client have a default of 10 in one case (non-follow) but a default of 0 in the other (follow). In the `pebble logs` CLI command, we want `pebble logs` and `pebble logs -f` to have the same default, like `tail` and `tail -f` both show the last 10 buffered logs. But with your suggested Go client change, we'd have to have a client-side default for `pebble logs -f` anyway.

4) Partly because of the above, it seems to me it was a bad idea to have a server-specified default at all. That should be something that's client-dependent. For example, on the CLI a default of 10 is good, but in a web UI, maybe 100 would make more sense. That's why I got rid of the default in the Go client in this PR -- I think any default should be client-specified (essentially making `LogsOptions.N` a required field).

My concrete suggestion is we keep the server API as is (or maybe tweak it to allow `n=all` as well as `n=-1`), and make the Go client change suggested in this PR. Alternatively for the Go client, we could have

```go
type LogsOptions struct {
   ...
   Fetch LogsFetch // one of FetchN (zero value), FetchDefault, FetchAll
   N     int       // not used if Fetch is FetchDefault or FetchAll 
}
```

If I'm not explaining myself clearly in the above, let's chat on video.

benhoyt: Okay, I think that's reasonable, thanks. API design is hard! :-) I've made this change to the Go client and the same tweak to the server API and updated the PR -- ready for re-review.

niemeyer (approved):
Thanks!

PR #92, opened 2021-12-02
Add signals API to send a signal to one or more services

benhoyt:
This implements spec [JU029 Pebble: sending signals to services](https://docs.google.com/document/d/1R9gdQU7q1CdH3raqLeB9d86pICbOSpm0P04P_CNF2g0/edit#). Specifically:

* An API to send a signal to one or more running services:
  - `POST /v1/signals {"signal": "SIGHUP", "services": ["s1", "s2"]}`
* A Go client to send a signal:
  - `c.SendSignal(&client.SendSignalOptions{Signal: "SIGHUP", Services: []string{"s1, "s2"})`
* A CLI command to send a signal:
  - `pebble signal HUP s1 s2 ...`

One question I had was whether the API should return an error if the service being signaled is not running. After discussion with Harry we decided to mimic the behavior of the Unix `kill` command, which gives an error if the pid is not active. But that's open for debate -- it will mean if you call `send_signal()` in a charm it'll raise an exception if the service hasn't been started yet (maybe that's a good thing).

### DIFF
--- a/client/signals.go
+++ b/client/signals.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 )
 
-// SendSignalOptions holds the options for SendSignal.
 type SendSignalOptions struct {
 	Signal   string
 	Services []string
--- a/client/signals.go
+++ b/client/signals.go
@@ -1,0 +1,47 @@
+// Copyright (c) 2021 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// SignalsOptions describe the options for SendSignal.
+type SignalsOptions struct {
+	Signal   string
+	Services []string
+}
+
+// SendSignal sends a signal to each of the specified services.
+func (client *Client) SendSignal(opts *SignalsOptions) error {
+	var body bytes.Buffer
+	payload := signalsPayload{
+		Signal:   opts.Signal,
+		Services: opts.Services,
+	}
+	err := json.NewEncoder(&body).Encode(&payload)
+	if err != nil {
+		return fmt.Errorf("cannot encode JSON payload: %v", err)
+	}
+	_, err = client.doSync("POST", "/v1/signals", nil, nil, &body, nil)
+	return err
+}
+
+type signalsPayload struct {
+	Signal   string   `json:"signal"`
+	Services []string `json:"services"`
+}
--- a/client/signals.go
+++ b/client/signals.go
@@ -20,14 +20,14 @@ import (
 	"fmt"
 )
 
-// SignalsOptions describe the options for SendSignal.
-type SignalsOptions struct {
+// SendSignalOptions holds the options for SendSignal.
+type SendSignalOptions struct {
 	Signal   string
 	Services []string
 }
 
 // SendSignal sends a signal to each of the specified services.
-func (client *Client) SendSignal(opts *SignalsOptions) error {
+func (client *Client) SendSignal(opts *SendSignalOptions) error {
 	var body bytes.Buffer
 	payload := signalsPayload{
 		Signal:   opts.Signal,
--- a/cmd/pebble/cmd_help.go
+++ b/cmd/pebble/cmd_help.go
@@ -168,11 +168,15 @@ type helpCategory struct {
 var helpCategories = []helpCategory{{
 	Label:       "Basics",
 	Description: "basic management",
-	Commands:    []string{"run", "autostart", "start", "stop", "exec", "services", "logs", "signal"},
+	Commands:    []string{"run", "exec"},
+}, {
+	Label:       "Services",
+	Description: "service management",
+	Commands:    []string{"autostart", "start", "stop", "restart", "services", "logs", "signal"},
 }, {
 	Label:       "Configuration",
 	Description: "manage configuration",
-	Commands:    []string{"add", "plan"},
+	Commands:    []string{"add", "plan", "replan"},
 }, {
 	Label:       "Changes",
 	Description: "view changes and tasks",
--- a/cmd/pebble/cmd_help.go
+++ b/cmd/pebble/cmd_help.go
@@ -178,9 +178,9 @@ var helpCategories = []helpCategory{{
 	Description: "manage services",
 	Commands:    []string{"services", "logs", "start", "restart", "signal", "stop", "replan"},
 }, {
-	Label:       "Operations",
-	Description: "other operations",
-	Commands:    []string{"exec"}, // TODO: "checks" would go here
+	Label:       "Files",
+	Description: "work with files and execute commands",
+	Commands:    []string{"exec"},
 }, {
 	Label:       "Changes",
 	Description: "manage changes and their tasks",
--- a/cmd/pebble/cmd_help.go
+++ b/cmd/pebble/cmd_help.go
@@ -168,7 +168,7 @@ type helpCategory struct {
 var helpCategories = []helpCategory{{
 	Label:       "Basics",
 	Description: "basic management",
-	Commands:    []string{"run", "autostart", "start", "stop", "exec", "services", "logs"},
+	Commands:    []string{"run", "autostart", "start", "stop", "exec", "services", "logs", "signal"},
 }, {
 	Label:       "Configuration",
 	Description: "manage configuration",
--- a/cmd/pebble/cmd_help.go
+++ b/cmd/pebble/cmd_help.go
@@ -166,25 +166,29 @@ type helpCategory struct {
 
 // helpCategories helps us by grouping commands
 var helpCategories = []helpCategory{{
-	Label:       "Basics",
-	Description: "basic management",
-	Commands:    []string{"run", "exec"},
+	Label:       "Run",
+	Description: "run pebble",
+	Commands:    []string{"run", "help", "version"},
+}, {
+	Label:       "Plan",
+	Description: "view and change configuration",
+	Commands:    []string{"add", "plan"},
 }, {
 	Label:       "Services",
-	Description: "service management",
-	Commands:    []string{"autostart", "start", "stop", "restart", "services", "logs", "signal"},
+	Description: "manage services",
+	Commands:    []string{"services", "logs", "start", "restart", "signal", "stop", "replan"},
 }, {
-	Label:       "Configuration",
-	Description: "manage configuration",
-	Commands:    []string{"add", "plan", "replan"},
+	Label:       "Operations",
+	Description: "other operations",
+	Commands:    []string{"exec"}, // TODO: "checks" would go here
 }, {
 	Label:       "Changes",
-	Description: "view changes and tasks",
+	Description: "manage changes and their tasks",
 	Commands:    []string{"changes", "tasks"},
 }, {
-	Label:       "Other",
-	Description: "other commands",
-	Commands:    []string{"warnings", "okay", "help", "version"},
+	Label:       "Warnings",
+	Description: "manage warnings",
+	Commands:    []string{"warnings", "okay"},
 }}
 
 var (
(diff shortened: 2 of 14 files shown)


cmd/pebble/cmd_help.go
```diff
@@ -168,11 +168,15 @@ type helpCategory struct {
 var helpCategories = []helpCategory{{
 	Label:       "Basics",
 	Description: "basic management",
-	Commands:    []string{"run", "autostart", "start", "stop", "exec", "services", "logs"},
+	Commands:    []string{"run", "exec"},
```
benhoyt, line 171: A couple of things here:

* I added the new `signal` command for this commit.
* I'm adding in `restart` and `replan`, missed here when those were added.
* "basic management" is getting too long, especially with the "checks" command that will be added in another PR. I think the category "service management" is good, though I'm not sure about the "basic management" name anymore -- suggestions for a better name/grouping?

internal/daemon/api_services_test.go
```diff
@@ -321,15 +320,3 @@ func (s *apiSuite) TestServicesReplan(c *C) {
 	c.Check(tasks[0].Summary(), Equals, `Start service "test1"`)
 	c.Check(tasks[1].Summary(), Equals, `Start service "test2"`)
 }
-
-type wrappedServiceManager struct {
```
benhoyt, line 325: This wasn't being used anywhere, deleting.

internal/overlord/servstate/handlers.go
```diff
@@ -499,14 +499,18 @@ func getAction(config *plan.Service, success bool) (action plan.ServiceAction, o
 func (s *serviceData) sendSignal(signal string) error {
 	switch s.state {
 	case stateStarting, stateRunning:
+		sig := unix.SignalNum(signal)
```
benhoyt, line 502: The send-signal functionality already existed via the restart/backoff work; I'm just tweaking the error behavior here.

cmd/pebble/cmd_help.go
hpidcock: Thanks for fixing my mistake with restart/replan

internal/daemon/api_signals.go
```diff
@@ -0,0 +1,43 @@
...
+	Services []string `json:"services"`
+}
+
+func v1PostSignals(c *Command, req *http.Request, _ *userState) Response {
+	var payload signalsPayload
+	decoder := json.NewDecoder(req.Body)
+	if err := decoder.Decode(&payload); err != nil {
+		return statusBadRequest("cannot decode request body: %v", err)
+	}
+	if len(payload.Services) <= 0 {
```
hpidcock, line 33: In other APIs do we not deal with the wildcard case (might just be autostart?)? Here it probably doesn't make sense just wondering if we should seek consistency.

internal/daemon/api_signals.go
```diff
@@ -0,0 +1,43 @@
...
+package daemon
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+type signalsPayload struct {
+	Signal   string   `json:"signal"`
+	Services []string `json:"services"`
```
hpidcock, line 24: I think we should look at adding a ProcessGroup/Group/SignalGroup bool, where if true the signal is sent to the process group of the service rather than just the pid of the process.

internal/overlord/servstate/handlers.go
```diff
@@ -499,14 +499,18 @@ func getAction(config *plan.Service, success bool) (action plan.ServiceAction, o
...
+		if sig == 0 {
+			return fmt.Errorf("invalid signal name %q", signal)
+		}
 		logger.Noticef("Sending %s to service %q", signal, s.config.Name)
-		err := syscall.Kill(-s.cmd.Process.Pid, unix.SignalNum(signal))
+		err := syscall.Kill(-s.cmd.Process.Pid, sig)
```
hpidcock, line 507: This is incorrect for your usage, this signals the process group (-pid is a process group), which I think only makes sense in service stop.

I think we need to review this anyway for service stop:
- Stop should first signal SIGTERM to the service process (not the group, the process should SIGTERM its child processes)
- Wait the grace time
- Always send SIGKILL to the process group (even if the process exits, we need to signal the whole process group to ensure all child processes are killed).

@niemeyer does this sound right to you?

internal/daemon/api_signals.go
hpidcock: That way a SIGUSR1 can just be sent to the service process and not children in the same process group.

cmd/pebble/cmd_help.go
hpidcock: Also I like this grouping

internal/daemon/api_signals.go
benhoyt: Yeah, `replan` and `autostart` error if you *do* provide service names, and `start`, `stop`, and `restart` error if you don't (which makes sense for each of those, I think). I don't think it's a good idea (or makes a lot of sense) to send a signal to all services. If someone really wants to they can do something like this:

```python
service_names = [s.name for s in container.get_services()]
container.send_signal('SIGHUP', service_names)
```

internal/daemon/api_signals.go
benhoyt: That's a good idea. I was wondering whether it should send to children or not. I guess there are two questions:

1) What should the default be, send to just the service, or send to group? I think your approach of sending to just the service is a reasonable default (least potentially destructive).
2) What should the field be named? I think `group` is not great, as that indicates a group name (and is used as a group name in other Pebble APIs and the plan config). Maybe `SendToGroup bool` -- longer but nice and clear?

internal/overlord/servstate/handlers.go
benhoyt: This is a good call -- I fairly new to this signal handling and hadn't considered that. What you propose sounds reasonable. I'll try to do a bit more reading on it, and probably add this to the agenda for tomorrow night's spec review meeting.

cmd/pebble/cmd_help.go
```diff
@@ -168,11 +168,15 @@ type helpCategory struct {
...
+}, {
+	Label:       "Services",
+	Description: "service management",
+	Commands:    []string{"autostart", "start", "stop", "restart", "services", "logs", "signal"},
```
niemeyer, line 175: It feels worth reorganizing indeed, but the groups feels a bit blurred, in the sense that it's not clear what goes inside which given their descriptions. Also, as a previous problem we already had, the descriptions are currently not worded evenly: some of them are actions ("view changes") while others are passive ("basic/service management"). Finally, "basic" now has "run" and "exec", which are completely unlike each other, and the proximity of their names doesn't help either.

So maybe we should dedicate a further moment to try to help the situation here and have more distinctive "buckets".

Here is an initial attempt:
```
"Run", "run pebble"
    {"run", "help", "version"}
"Plan", "view and change configuration"
    {"add", "plan", "replan"}
"Services", "manage services"
    {"services", "logs", "start", "restart", "signal", "stop", "replan"}
"Changes", "manage changes and their tasks"
    {"changes", "tasks"}
"Warnings", "manage warnings"
    {"warnings", "okay"}
```
Some notes:
- Isn't "autostart" the old version of "replan"?  That means we can drop the old one.
- If "autostart" and "replan" are indeed friends, it's curious we had them in different groups. What's the right one?
- I've tried to order services inside the groups using some subjective logic while taking into account their meaning, relevance, and proximity

cmd/pebble/cmd_signal.go
```diff
@@ -0,0 +1,63 @@
...
+	Positional struct {
+		Signal   string   `positional-arg-name:"<SIGNAME>"`
+		Services []string `positional-arg-name:"<service>"`
+	} `positional-args:"yes" required:"yes"`
+}
+
+var shortSignalHelp = "Send a signal to one or more running services"
+var longSignalHelp = `
+The signal command sends a signal to one or more running services. The signal
+name must be uppercase to distinguish it from the service names, for example:
```
niemeyer, line 37: Seems worth dropping the "to distinguish it from service names", as this is more of a UX detail, and it's not strictly true as we could just take the position into account. It just makes the CLI nicer and less ambiguous for a person to read and write.

cmd/pebble/cmd_signal.go
```diff
@@ -0,0 +1,63 @@
...
+
+pebble signal HUP mysql nginx
+pebble signal SIGHUP mysql nginx    # full signal name is fine too
```
niemeyer, line 40: I'm not strongly against the optional SIG prefix, but can we please wait a bit before we support it?  It feels like consistency generally wins, and from the two lines above the first one reads the best. This way we also don't even need to document the exception.

internal/daemon/api_services_test.go
niemeyer: Thanks. I don't even know why this was here.

internal/daemon/api_signals.go
niemeyer: I'm not entirely sure if we need to signal children or not, but I'm pretty sure we should not have that as the default behavior.  We have no idea about what sub-processes the given service will have, and signaling the entire group seems somewhat of a bad practice.  The one exception being SIGKILL, probably, which we don't want to be sent this way in any case.

As such, I suggest just limiting to the process itself for now. As we learn why people want to signal the group, we can design the API and implement it accordingly.

internal/overlord/servstate/handlers.go
niemeyer: Just to double check, what's the API we intend to offer here?   "HUP", or "SIGHUP"?  I suggest enforcing one of them only, and perhaps only "SIGHUP" here as it's the more formal name, which we're opting not to use in the CLI as we have "signal HUP" already.

internal/overlord/servstate/handlers.go
```diff
@@ -499,14 +499,18 @@ func getAction(config *plan.Service, success bool) (action plan.ServiceAction, o
...
 		if err != nil {
 			return err
 		}
 
 	case stateBackoff, stateTerminating, stateKilling, stateStopped:
-		return fmt.Errorf("cannot send signal while service is stopped or stopping")
+		return fmt.Errorf("service is not running")
```
niemeyer, line 513: This seems to lack the context ("cannot send signal ...").

The message below seems to lack a verb as well ("is invalid" or similar).

internal/overlord/servstate/manager.go
```diff
@@ -410,11 +410,13 @@ func (m *ServiceManager) SendSignal(services []string, signal string) error {
 	for _, name := range services {
 		s := m.services[name]
 		if s == nil {
+			errors = append(errors, fmt.Sprintf("cannot send signal to %q: service is not running", name))
```
niemeyer, line 413: Oh, hey, or maybe not? :-)  Maybe that default clause is the one poorly worded indeed?

niemeyer (approved):
Thanks for this!

Assuming these comments make sense, LGTM once they are sorted.  Otherwise, happy to chat further.

internal/daemon/api_signals.go
benhoyt: Okay, I've limited it to just the process for now, not the group.

client/signals.go
```diff
@@ -0,0 +1,47 @@
...
+
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// SendSignalOptions holds the options for SendSignal.
```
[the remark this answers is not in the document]
benhoyt: Fair -- done.

cmd/pebble/cmd_help.go
benhoyt: The above is missing "exec", which doesn't really fit anywhere. I've updated suggesting adding "Operations" which will have "exec" and (soon) "checks". Let's discuss tonight.

cmd/pebble/cmd_signal.go
benhoyt: Done.

cmd/pebble/cmd_signal.go
benhoyt: I've removed it from the docs. However, it's still supported as we have to add the SIG prefix for the API anyway, so it would be explicit code to prevent it from working. I think consistently using `HUP` (for example) in the docs but allowing the alternative is fine.

internal/overlord/servstate/handlers.go
benhoyt: Yeah, given the context added below this is okay.

internal/overlord/servstate/manager.go
benhoyt: Fixed.

benhoyt:
> Just to double check, what's the API we intend to offer here? "HUP", or "SIGHUP"? I suggest enforcing one of them only, and perhaps only "SIGHUP" here as it's the more formal name, which we're opting not to use in the CLI as we have "signal HUP" already.

Yep, it's always the full `SIGHUP` in the API (also in the existing exec API).

internal/overlord/servstate/handlers.go
benhoyt: We discussed this on video and have a solution -- will follow up in a separate PR as it's not related to the send-signal API.

cmd/pebble/cmd_help.go
benhoyt: For now I've added a "Files" group for files and exec.

PR #96, opened 2021-12-15
Fix bug with replan/start/restart not updating internal config

benhoyt:
This fixes a bug in Replan not updating the stored `service.config` value due to bad logic I introduced in https://github.com/canonical/pebble/pull/79/files#diff-c2607d6e10ee84455ecce459b40a302a3a9f44197dc7a2205c52c0cc76c8bfaaR370. I've also added a test for this case.

It also fixes a similar bug where a manual stop/start (or restart) sequence didn't update the stored service config -- I introduced this bug in the same commit.

### DIFF
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -163,7 +163,7 @@ func (m *ServiceManager) serviceForStart(task *state.Task, config *plan.Service)
 		return service
 	}
 
-	// Ensure config is up-to-date from the plan
+	// Ensure config is up-to-date from the plan whenever the user starts a service
 	service.config = config.Copy()
 
 	switch service.state {
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -163,7 +163,7 @@ func (m *ServiceManager) serviceForStart(task *state.Task, config *plan.Service)
 		return service
 	}
 
-	// Ensure config is up-to-date from the plan whenever the user starts a service
+	// Ensure config is up-to-date from the plan whenever the user starts a service.
 	service.config = config.Copy()
 
 	switch service.state {
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -163,6 +163,9 @@ func (m *ServiceManager) serviceForStart(task *state.Task, config *plan.Service)
 		return service
 	}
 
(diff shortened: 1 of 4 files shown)


internal/overlord/servstate/handlers.go
```diff
@@ -406,7 +406,7 @@ func (s *serviceData) exited(waitErr error) error {
 		action, onType := getAction(s.config, waitErr == nil)
 		switch action {
 		case plan.ActionIgnore:
-			logger.Debugf("Service %q %s action is %q, transitioning to stopped state", s.config.Name, onType, action)
+			logger.Noticef("Service %q %s action is %q, transitioning to stopped state", s.config.Name, onType, action)
```
benhoyt, line 409: This is so all three actions (restart, halt, ignore) have a similar log message with these details. This shouldn't happen often (only on service exit) and seems important.

internal/overlord/servstate/manager.go
```diff
@@ -368,7 +368,7 @@ func (m *ServiceManager) Replan() ([]string, []string, error) {
 	needsRestart := make(map[string]bool)
 	var stop []string
 	for name, s := range m.services {
-		if config, ok := m.plan.Services[name]; ok && config.Equal(s.config) {
+		if config, ok := m.plan.Services[name]; ok {
```
benhoyt, line 371: This is the actual bug fix.

internal/overlord/servstate/manager_test.go
```diff
@@ -316,6 +316,62 @@ func (s *S) TestReplanServices(c *C) {
...
+	defer s.stopTestServices(c)
+
+	// Ensure the ServiceManager's config reflects the plan config
+	config := s.manager.Config("test2")
+	c.Assert(config, NotNil)
+	c.Assert(config.OnSuccess, Equals, plan.ActionUnset)
+	c.Assert(config.Summary, Equals, "")
+	command := config.Command
+
+	// Add a layer and override a couple of values
```
wallyworld, line 330: Nit - comments should end with a "." right?

wallyworld (approved):

internal/overlord/servstate/manager_test.go
benhoyt: Good call, updated for consistency.

benhoyt:
Going to merge this bug fix without further review as well, so we can include it in the next version of Juju. The fix is two lines (the rest is tests and a logging tweak).

PR #97, opened 2022-01-04
Generate Copy and Merge for plan structs

benhoyt:
NOTE: this branch is based off the `checks-api-cli` branch, but this PR shows the diff against master. Until #85 and #86 are merged, **see https://github.com/benhoyt/pebble/pull/5 for a reviewable diff of just these changes.**

Context: previously I hacked up a [proof-of-concept](https://github.com/benhoyt/pebble/pull/4) using `reflect`-based libraries to avoid the error-prone boilerplate in the `Copy` and `Merge` methods -- however, there wasn't appetite for that approach as it's too hard to change the behavior later, for example if we wanted to skip copying fields or copy certain pointer fields by pointer instead of value.

Generating the code has the same advantages (avoiding the tedious and error-prone hand-written code) but not the disadvantages. Specifically:

1) It's our code generator for this very specific use case, so it's small and easy to hack and change as needed. For example, skipping certain fields, copying pointer fields by pointer vs value, and so on.
2) If at some point we do decide this approach is not working anymore, we can just keep the generated code and drop the code generator, as the generated code is basically how you'd write it by hand ... without the bugs. :-)

**Speaking of bugs, this approach definitely pays off.** I already caught a couple of copy-and-paste bugs when hand-writing the new `Copy` and `Merge` methods for the health checks code, and writing this code generator found a few more bugs, shown in the following commits:

* A bug in `Check.Copy` due to a typo `copied` which should have been `c` ([diff](https://github.com/canonical/pebble/commit/c187a0e5b0d6e8b52102aa77a28cb5219eb99b38#diff-d05354a90aeb11eee6d8957c6de750ba70ae45bb7febc530025af25b099d54a8R94)). This would cause a subtle but visible bug with the `Check.Exec` field.
* `Service.Merge` missing the `Required` field ([diff](https://github.com/canonical/pebble/commit/2cfa7e9760325c97e90c1561bf717bd8cc3dc51d#diff-aca0ca7bb76b9d9562b227561d9dd1c065a59ced11b30b5ee506a25e1d547e5fR27)). This would cause a subtle but visible bug with the `Required` field not being merged when you added a layer.
* `Service.Merge` merging the `UserID` and `GroupID` fields by pointer instead of value, which may have caused issues if the caller changed the value by pointer underfoot ([diff](https://github.com/canonical/pebble/commit/2cfa7e9760325c97e90c1561bf717bd8cc3dc51d#diff-aca0ca7bb76b9d9562b227561d9dd1c065a59ced11b30b5ee506a25e1d547e5fR35)). This one probably wouldn't have caused an issue in practice.


benhoyt:
After further discussion via video, we've decided to use the reflect-based approach instead (unless or until we need special-cases for the various fields, in which case we'll probably revert to hand-rolled logic). The reflect-based approach from https://github.com/benhoyt/pebble/pull/4 has now been merged into https://github.com/canonical/pebble/pull/85.

PR #98, opened 2022-01-05
Fix subtle plan.Service merging issues

benhoyt:
* Add missing Requires field
* Copy values (not just pointers) for UserID and GroupID

These issues were found by #97.


wallyworld (approved):

benhoyt:
Merged without further review as this is a trivial bug-fix.

Issue #93, opened 2021-12-09
Wait for child processes to terminate when stopping a service


benhoyt:
Per discussion on Tuesday, we should update the stop-service API to wait for all child processes in the process group to terminate before returning (or sending `SIGKILL`).

Currently when doing "stop" we send `SIGTERM` to the process group, which won't change. But then we only wait for the immediate PID to exit. We should change that to use https://linux.die.net/man/2/waitpid and wait for all processes in the process group to finish. That will do the right thing for services that start other commands, for example a shell script `/bin/sh -c 'foo bar'`.

benhoyt:
This is fixed by https://github.com/canonical/pebble/pull/101. Note that there's also https://github.com/canonical/pebble/issues/6, which will be fixed by https://github.com/canonical/pebble/pull/114.

Issue #6, opened 2021-02-02
Forward signals to child processes


amouat:
Assuming pebble runs a PID 1 in a container, it needs to take care of some tasks, most notably forwarding signals to child processes and dealing with zombie & orphan processes. Handling signals is especially key, as this is required to allow processes to shutdown cleanly e.g. databases to finish writes and log, web apps to handle final requests. At the moment, signal forwarding doesn't seem to be happening by default at least.

To test this I used the following Dockerfile:

```
FROM debian:latest

COPY pebble /pebble
COPY test.sh /test.sh
RUN mkdir --parents /.pebble/layers
COPY 01.yaml /.pebble/layers/
ENV PEBBLE /.pebble

CMD ["/pebble", "run"]
```

With the pebble config:
```
summary: Test Layer

services:

    srv1:
        override: replace
        summary: Service summary
        command: /test.sh
        default: start
```

And test.sh:

```
#!/bin/bash

function shutdown()
{
    echo "Graceful shutdown" >> /log
    exit
}

trap shutdown SIGTERM

echo "Started" >> /log
while true
do
    echo Sleeping >> /log
    sleep 10
done
```

Build this image and run it using Docker:

```
$ docker build -t test .
...
$ docker run -d --name test testdocker run 
```

Now run `docker stop` which will send a SIGTERM to the container and then look at the logfile:

```
$ docker stop test
test
$ docker cp test:/log ./log
$ cat log
Started
Sleeping
Sleeping
```

Note that "Graceful shutdown" does not appear, indicating that the signal is not being forwarded to the child process. If we test the same script with [tini](https://github.com/krallin/tini/) we can observe the correct behaviour: 

```
$ cat Dockerfile.tini
FROM debian:latest

ENV TINI_VERSION v0.19.0
ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
RUN chmod +x /tini
COPY test.sh /test.sh

CMD ["/tini", "--", "/test.sh"]

$ docker build -f Dockerfile.tini -t test-tini .
...
$ docker rm test-tini                     
test-tini
$ docker run -d --name test-tini test-tini 
853db290cfeeb797d21203ccfa07221b7e2f6ef34b0f755a6e97bd4471ad8d16
$ docker stop test-tini
test-tini

$ docker cp test-tini:/log ./log
$ cat log
Started
Sleeping
Graceful shutdown
```

[...]

PR #99, opened 2022-01-16
feat(#93): Wait for child processes

merkata:
This feature implements waiting for child processes when stopping a
service via SIGTERM as described in #93. This is done via waiting for all child processes
that are part of the process group of the main (service) process, i.e.
the leader process. The syscall used is Wait4 with a waitpid behaviour.
Note that waiting for grandchildren is not supported with this approach.

Wait4 system call is used as wait and waitpid are not supported in Golang
(see https://github.com/golang/go/issues/9176). It loops over until there are
no more children to wait on (returns syscall.ECHILD as err, though this is not
checked here explicitly), ignoring hung processes (syscall.WNOHANG) and
retrying on a interrupted system call or library function call (syscall.EINTR,
see https://man7.org/linux/man-pages/man7/signal.7.html).

The test simulates a child process doing clean up work that takes longer than the timer
set to invoke the SIGKILL. Before the implementation, it fails (as expected), inserting
the wait4 functionality makes the test pass.

### DIFF
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -23,6 +23,13 @@ import (
 	"github.com/canonical/pebble/internal/strutil/shlex"
 )
 
+func init() {
+	// Reap sub-children via setting the child subreaper
+	// that way we can reap "grandchildren" from our main process,
+	// effectively we can wait for the children of the terminating process
+	unix.Prctl(unix.PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0)
+}
+
 // TaskServiceRequest extracts the *ServiceRequest that was associated
 // with the provided task when it was created, reflecting details of
 // the operation requested.
@@ -356,11 +363,10 @@ func (s *serviceData) startInternal() error {
 	done := make(chan struct{})
 	go func() {
 		waitErr := s.cmd.Wait()
-		err := s.waitProcessGroup()
-		if err != nil {
-			logger.Noticef("Cannot wait for process group: %v", err)
+		childErr := s.waitProcessGroup()
+		if childErr != nil {
+			logger.Noticef("Cannot wait for children processes after main process exit: %v", childErr)
 		}
-		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
 		close(done)
 		err = s.exited(waitErr)
 		if err != nil {
@@ -593,6 +599,7 @@ func (s *serviceData) stop() error {
 			logger.Noticef("Cannot send SIGTERM to process: %v", err)
 		}
 		s.transition(stateTerminating)
+		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
 
 	case stateBackoff:
 		logger.Noticef("Service %q stopped while waiting for backoff", s.config.Name)
@@ -606,15 +613,28 @@ func (s *serviceData) stop() error {
 }
 
 func (s *serviceData) waitProcessGroup() error {
+	var (
+		status syscall.WaitStatus
+		rusage syscall.Rusage
+		pid1   int
+		err    error
+	)
 	for {
-		var wstatus syscall.WaitStatus
-		if _, err := syscall.Wait4(-s.cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil); err != syscall.EINTR && err != nil {
-			if syscall.ECHILD == err {
-				return nil
-			}
-			return err
+		// __WALL (since Linux 2.4)
+		// Wait for all children, regardless of type ("clone" or "non-clone").
+		pid1, err = syscall.Wait4(-s.cmd.Process.Pid, &status, syscall.WALL, &rusage)
+		// https://linux.die.net/man/2/waitid
+		// wait(): on success, returns the process ID of the terminated child; on error, -1 is returned.
+		if pid1 == -1 {
+			break
 		}
+		logger.Noticef("Waited for child %d of PGID %d", pid1, s.cmd.Process.Pid)
 	}
+	// (for wait()) The calling process does not have any unwaited-for children.
+	if err == syscall.ECHILD {
+		return nil
+	}
+	return err
 }
 
 // backoffTimeElapsed is called when the current backoff's timer has elapsed,
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -23,13 +23,6 @@ import (
 	"github.com/canonical/pebble/internal/strutil/shlex"
 )
 
-func init() {
-	// Reap sub-children via setting the child subreaper
-	// that way we can reap "grandchildren" from our main process,
-	// effectively we can wait for the children of the terminating process
-	unix.Prctl(unix.PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0)
-}
-
 // TaskServiceRequest extracts the *ServiceRequest that was associated
 // with the provided task when it was created, reflecting details of
 // the operation requested.
@@ -363,9 +356,9 @@ func (s *serviceData) startInternal() error {
 	done := make(chan struct{})
 	go func() {
 		waitErr := s.cmd.Wait()
-		childErr := s.waitProcessGroup()
-		if childErr != nil {
-			logger.Noticef("Cannot wait for children processes after main process exit: %v", childErr)
+		err := waitChildren(s.cmd.Process.Pid)
+		if err != nil {
+			logger.Noticef("Cannot wait for children processes after main process exit: %v", err)
 		}
 		close(done)
 		err = s.exited(waitErr)
@@ -612,29 +605,20 @@ func (s *serviceData) stop() error {
 	return nil
 }
 
-func (s *serviceData) waitProcessGroup() error {
-	var (
-		status syscall.WaitStatus
-		rusage syscall.Rusage
-		pid1   int
-		err    error
-	)
+func waitChildren(pid int) error {
 	for {
-		// __WALL (since Linux 2.4)
 		// Wait for all children, regardless of type ("clone" or "non-clone").
-		pid1, err = syscall.Wait4(-s.cmd.Process.Pid, &status, syscall.WALL, &rusage)
-		// https://linux.die.net/man/2/waitid
+		wpid, err := syscall.Wait4(-pid, nil, syscall.WALL, nil)
 		// wait(): on success, returns the process ID of the terminated child; on error, -1 is returned.
-		if pid1 == -1 {
-			break
+		if wpid == -1 {
+			// (for wait()) The calling process does not have any unwaited-for children.
+			if err == syscall.ECHILD {
+				return nil
+			}
+			return err
 		}
-		logger.Noticef("Waited for child %d of PGID %d", pid1, s.cmd.Process.Pid)
+		logger.Debugf("Waited for child %d of PGID %d", wpid, pid)
 	}
-	// (for wait()) The calling process does not have any unwaited-for children.
-	if err == syscall.ECHILD {
-		return nil
-	}
-	return err
 }
 
 // backoffTimeElapsed is called when the current backoff's timer has elapsed,
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -356,8 +356,13 @@ func (s *serviceData) startInternal() error {
 	done := make(chan struct{})
 	go func() {
 		waitErr := s.cmd.Wait()
+		err := s.waitProcessGroup()
+		if err != nil {
+			logger.Noticef("Cannot wait for process group: %v", err)
+		}
+		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
 		close(done)
-		err := s.exited(waitErr)
+		err = s.exited(waitErr)
 		if err != nil {
 			logger.Noticef("Cannot transition state after service exit: %v", err)
 		}
@@ -588,13 +593,6 @@ func (s *serviceData) stop() error {
 			logger.Noticef("Cannot send SIGTERM to process: %v", err)
 		}
 		s.transition(stateTerminating)
-		for {
-			var wstatus syscall.WaitStatus
-			if _, err := syscall.Wait4(-s.cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil); err != syscall.EINTR && err != nil {
-				break
-			}
-		}
-		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
 
 	case stateBackoff:
 		logger.Noticef("Service %q stopped while waiting for backoff", s.config.Name)
@@ -605,6 +603,18 @@ func (s *serviceData) stop() error {
 		return fmt.Errorf("cannot stop service while %s", s.state)
 	}
 	return nil
+}
+
+func (s *serviceData) waitProcessGroup() error {
+	for {
+		var wstatus syscall.WaitStatus
+		if _, err := syscall.Wait4(-s.cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil); err != syscall.EINTR && err != nil {
+			if syscall.ECHILD == err {
+				return nil
+			}
+			return err
+		}
+	}
 }
 
 // backoffTimeElapsed is called when the current backoff's timer has elapsed,
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -588,6 +588,12 @@ func (s *serviceData) stop() error {
 			logger.Noticef("Cannot send SIGTERM to process: %v", err)
 		}
 		s.transition(stateTerminating)
+		for {
+			var wstatus syscall.WaitStatus
+			if _, err := syscall.Wait4(-s.cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil); err != syscall.EINTR && err != nil {
+				break
+			}
+		}
 		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
 
 	case stateBackoff:
--- a/internal/overlord/servstate/manager.go
+++ b/internal/overlord/servstate/manager.go
@@ -13,6 +13,7 @@ import (
 	"github.com/canonical/pebble/internal/overlord/state"
 	"github.com/canonical/pebble/internal/plan"
 	"github.com/canonical/pebble/internal/servicelog"
+	"golang.org/x/sys/unix"
 )
 
 type ServiceManager struct {
@@ -48,6 +49,11 @@ func (e *LabelExists) Error() string {
 }
 
 func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string, serviceOutput io.Writer, restarter Restarter) (*ServiceManager, error) {
+	// Set Pebble as a "child subreaper" so it becomes the parent of orphaned children
+	// rather than PID 1, allowing us to wait for processes that a service starts (see
+	// https://unix.stackexchange.com/a/250156/73491).
+	err := unix.Prctl(unix.PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0)
+
 	manager := &ServiceManager{
 		state:         s,
 		runner:        runner,
@@ -61,7 +67,7 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string, serv
 	runner.AddHandler("start", manager.doStart, nil)
 	runner.AddHandler("stop", manager.doStop, nil)
 
-	return manager, nil
+	return manager, err
 }
 
 func (m *ServiceManager) reloadPlan() error {
--- a/internal/overlord/servstate/manager_test.go
+++ b/internal/overlord/servstate/manager_test.go
@@ -926,7 +926,7 @@ func (s *S) TestStopWaitChildren(c *C) {
 services:
     test2:
         override: merge
-        command: bash -c "trap \"sleep 0.15; echo Hello Child\" SIGTERM; echo Test Child; sleep 0.1"
+        command: bash -c 'bash -c "{ trap \"echo child 1\" SIGTERM; sleep 0.4; } & { trap \"echo child 2\" SIGTERM; sleep 0.3; } & sleep 0.2"; sleep 0.1'
 `)
 	err := s.manager.AppendLayer(layer)
 	c.Assert(err, IsNil)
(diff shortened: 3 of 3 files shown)


benhoyt:
Hi @merkata Thanks for your submission. This definitely needs to be fixed - however, the way the code is structured right now isn't going to work, as each state machine event function, such as `stop()`, acquires the `servicesLock`. So these "event" functions have to return immediately. What we'd need to do is call `Wait4` in a background goroutine and have it call an event function when it finishes waiting.

We also need to consider what the behavior should be when the program exits for other reasons - should it wait for the children in those cases too?

Without giving it too much thought, maybe we could just add the wait-children logic after the existing `cmd.Wait()` call [here](https://github.com/canonical/pebble/blob/3a48156fcbf025462e4d063a9939307528017d7f/internal/overlord/servstate/handlers.go#L358), and then it'd just work? Like so:

```go
go func() {
	waitErr := s.cmd.Wait()
	err := waitProcessGroup(s.cmd.Process.Pid) // factor out your for loop into function
	if err != nil {
		logger.Noticef("Cannot wait for process group: %v", err)
	}
	close(done)
	err := s.exited(waitErr)
	if err != nil {
		logger.Noticef("Cannot transition state after service exit: %v", err)
	}
}()
```

Again, don't take that as gospel -- but that's a thought.

merkata:
Hi @benhoyt , thanks a lot for the feedback! I might have misunderstood something, so I have to ask :)

When you say it isn't going to work, do you mean in general or the code should be modified/refactored? I'm asking because the stop() function now respects children taking the time they need to finish up before getting killed by the time.AfterFunc, but this is with the cost of holding the lock longer. If we need to return immediately, then perhaps the time.AfterFunc does not belong there, as it imposes a hard limit on something taking longer than 5 sec in that function (although it should return immediately and release the lock..)?

With that new knowledge, I made a change to address this behaviour:

* every finish of a process, however it exits (runs to completion, gets SIGTERMd) is handled inside the go routine you mentioned - the os.Exec does its' Wait(), we wait for all children to exit and kick-off a time.AfterFunc in case these don't succeed.

* the stop() is now slimmer, it just sends a signal, sets a state and releases the lock, meaning it is not a responsibility to trigger a safety function like time.AfterFunc, just async try to terminate and if that doesn't succeed, the routine will handle the ungraceful termination from there.

I'm not sure about the use case with a program exiting and the program (or we) wouldn't want to wait for the children, an example would be really helpful. The current Wait4 implementation should ignore children in a hung state (so they should be killed by the time.AfterFunc) and retries the SIGTERM in case of an EINTR error which stops a blocking (system or library function) call in a child. Leaving children unattended is bad.

Thanks again!

benhoyt:
> When you say it isn't going to work, do you mean in general or the code should be modified/refactored?

Ah yes, that's exactly what I mean -- I don't mean in general, just that we needed to refactor this somewhat. I think what you have now is a lot closer -- thanks! However, after digging into this further today, I have a couple of questions.

1) I was thinking that "process group" was the same as "child processes", but it turns out they're two different things. When one process starts a child process, by default that new process has a new process group (but it's a child of the parent process). So we should actually be waiting on child processes, not the process group, right? Currently the error returned by Wait4 in the tests is ECHILD, "no child processes". But surely it should be waiting for at least one child process.

2) Because the Wait4 is not waiting on anything, I'm not sure the test is actually testing what we want. And that's why you had to move the line `time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })` -- so that the terminate time would elapse and Pebble would send SIGKILL. But I don't think that's what you want. I think we want to move the `time.AfterFunc` back to where it was, and increase the kill wait time if necessary.

So we need a test service that will create a child process (or two) and actually wait on that child process. I'm still not sure the right mechanism to wait on the child pids. Maybe reading `/proc/<pid>/task/<pid>/children` to get the child process IDs and then `wait` on each of them? Hmm, there's got to be a better way to do that.

Sorry for the back and forth on this, but I think now's the time to get this right.

merkata:
@benhoyt  thanks a lot for the feedback! I thought this would be easier, but it might not be the case :) Still a great discussion is forming and I really appreciate your thoughts and time. I think we will get this right in the end.
Yesterday I played around with the code and my understanding of how the solution should look right, I did manage to make things worse, but I think I made some good observations along the way.
My understanding for the child processes that are spawned by a process is that they can be (even by default) part of a newly formed process group by that process, so when you run for instance `bash -c "sleep 5"` you'd get PID 12345 and PGID 12345 for bash and PID 12367 and PGID 12345 for the spawned `sleep` child. Running it manually confirms it:

```shell
ubuntu@go-dev:~/pebble_configs$ ps efw -opgid,pid,command
   PGID     PID COMMAND
  24139   24139 -bash HOME=/home/ubuntu LANG=C.UTF-8 LESSCLOSE=/usr/bin/lesspipe %s %s LESSOPEN=| /usr/bin/lesspipe %s LOGNAME=ubunt
  31261   31261  \_ ps efw -opgid,pid,command SHELL=/bin/bash TMUX=/tmp//tmux-1000/default,2224,0 PEBBLE=/home/ubuntu/pebble_configs
   2225    2225 -bash HOME=/home/ubuntu LANG=C.UTF-8 LESSCLOSE=/usr/bin/lesspipe %s %s LESSOPEN=| /usr/bin/lesspipe %s LOGNAME=ubunt
  31236   31236  \_ pebble run SHELL=/bin/bash TMUX=/tmp//tmux-1000/default,2224,0 PEBBLE=/home/ubuntu/pebble_configs PWD=/home/ubun
  31255   31255      \_ bash -c trap "sleep 6; echo Hello from Child" SIGTERM; while true; do sleep 50; done SHELL=/bin/bash TMUX=/t
  31255   31256          \_ sleep 50 SHELL=/bin/bash TMUX=/tmp//tmux-1000/default,2224,0 PEBBLE=/home/ubuntu/pebble_configs PWD=/hom
   1751    1751 -bash USER=ubuntu LOGNAME=ubuntu HOME=/home/ubuntu PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
   2222    2222  \_ tmux SHELL=/bin/bash PEBBLE=/home/ubuntu/pebble_configs PWD=/home/ubuntu/dev/pebble LOGNAME=ubuntu XDG_SESSION_T
  30728   30728 bash -c trap "sleep 6; echo Hello from Child" SIGTERM; while true; do sleep 5; done SHELL=/bin/bash TMUX=/tmp//tmux-
  30728   31259  \_ sleep 5 SHELL=/bin/bash TMUX=/tmp//tmux-1000/default,2224,0 PEBBLE=/home/ubuntu/pebble_configs PWD=/home/ubuntu/
```

This is using the following config:

```yaml
ubuntu@go-dev:~/pebble_configs$ cat basic_pebble.yaml
summary: Simple layer

description: |
    A better description for a simple layer.

services:
    srv1:
        override: replace
        command: bash -c "trap \"sleep 6; echo Hello from Child\" SIGTERM; while true; do sleep 50; done"
        startup: enabled
    srv2:
        override: replace
        command: bash -c "trap \"cat /dev/zero; echo Done\" SIGTERM; while true; do sleep 50; done"
        startup: enabled
```

I've added srv2 under the assumption that srv1 is done and ready (it is not), as I wanted to see what happens with a "hanging" process that blocks for something (input that will never arrive in this case), so that it is ensured that it gets killed after waiting, but also that the wait does not hang itself.

I started looking into how and when does waiting occur in the go routine that is started in startInternal, specifically where should we wait for the children, it seems the cmd.Wait waits on the exit of the process and captures the exit status, so I thought we should wait subsequently on the children, though that did not work out.

I think your suggestion with having a test service that creates one two child processes is the right approach and maybe trapping SIGTERM (in the above config) and using scripts is not the right thing.

I'd like to dig a little bit more, but just to be sure, the goal of stop is to have a Lock on all services (via the servicemanager struct) and fire a stop for one service as fast as possible and let the outcome be dealt with externally (so in the go routine that has cmd.Wait or somewhere else) and release the Lock as fast as possible to prevent other services from waiting on acquiring the Lock for another operation, correct? So the behaviour we're after is - a service is one process that can spawn multiple children and when we stop the service, we should wait on the process and its children to exit (or be killed after a timeout has passed), right?

Again, thanks for all the great feedback!

benhoyt:
> So the behaviour we're after is - a service is one process that can spawn multiple children and when we stop the service, we should wait on the process and its children to exit (or be killed after a timeout has passed), right?

Yes, that's my basic understanding of the structure we want.

However, I've done some further digging, and I was mistaken earlier -- I think we *do* want to wait for the process group. In my tests I wasn't using `Setpgid`, but Pebble is starting service processes with `s.cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}`. For the following config which demonstrates sub and sub-sub-processes:

```
services:
    test:
        override: replace
        command: bash -c 'bash -c "sleep 301; sleep 1"; sleep 3'
        startup: enabled

    test2:
        override: replace
        command: bash -c 'sleep 401; sleep 2'
        startup: enabled
```

We get the following:

```
$ ps jx
   PPID     PID    PGID     SID TTY        TPGID STAT   UID   TIME COMMAND
...
 978154  986173  986173  978154 pts/4     986173 Sl+   1000   0:00 ./pebble run
 986173  986183  986183  978154 pts/4     986173 S     1000   0:00 bash -c bash -c "sleep 301; sleep 1"; sleep 3
 986183  986184  986183  978154 pts/4     986173 S     1000   0:00 bash -c sleep 301; sleep 1
 986184  986185  986183  978154 pts/4     986173 S     1000   0:00 sleep 301
 986173  986186  986186  978154 pts/4     986173 S     1000   0:00 bash -c sleep 401; sleep 2
 986186  986187  986186  978154 pts/4     986173 S     1000   0:00 sleep 401
 941754  986226  986226  941754 pts/2     986226 R+    1000   0:00 ps jx
```

So there's a new PID for every line (that makes sense), and a new PGID (process group ID) for every service (986183 for test and 986186 for test2), which is what we want for process groups to work. And the PID of the main process is the PGID of all he sub-processes. So sending SIGTERM/SIGKILL to the process group as we have now is what we want. Additionally, waiting for the process group to finish as you've done *is* I think what we want after all.

But back to code structure, yes, in the goroutine that does cmd.Wait(), we should add `waitProcessGroup(pid)` which would also wait for all the processes in that service's process group. So I think your existing code is close. So my point 1 in the [comment above](https://github.com/canonical/pebble/pull/99#issuecomment-1016063596) is incorrect, but I think point 2 is still valid.

merkata:
I did some further digging and I think the latest commit addresses the issue. I think using `bash -c` for the tests might lead to confusion, as bash itself waits for spawned commands:

```bash
ubuntu@go-dev:~/pebble_configs$ strace -f -e trace=waitid,waitpid,wait4 bash -c 'bash -c "sleep 10"; sleep 5'
wait4(-1, strace: Process 101215 attached
 <unfinished ...>
[pid 101215] +++ exited with 0 +++
<... wait4 resumed>[{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 101215
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=101215, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
wait4(-1, 0x7fff1f3ef3d0, WNOHANG, NULL) = -1 ECHILD (No child processes)
+++ exited with 0 +++
```

So sending a SIGTERM to the PGID where we start with a `bash -c` process, waiting on that process alone should be sufficient, as it waits for his children.

So I wanted to have processes that do not wait for their children necessarily and ended up with two approaches:

```go
func main() {
        var nAmount = flag.Int("n", 3, "sleep n amount of time in main")
        child1 := exec.Command("sleep", "10")
        child1.SysProcAttr = &syscall.SysProcAttr{Foreground: false}
        child1.Start()
        child2 := exec.Command("sleep", "20")
        child2.SysProcAttr = &syscall.SysProcAttr{Foreground: false}
        child2.Start()
        time.Sleep(time.Duration(*nAmount) * time.Second)
        fmt.Println("main done")
        os.Exit(1)
}
```

and 

```yaml
    test3:
        override: replace
        command: bash -c 'bash -c "sleep 40 & sleep 30 & sleep 20"; sleep 10'
        startup: enabled
```

These tests start children outside of job control (as background processes) within the same PGID but outside of "built-in" waiting in the case of bash. So for test3 you'd expect bash to wait on *sleep 20* and *sleep 10*, but the two child processes in the background (*sleep 40* and *sleep 30*) are not waited for by bash and we need to wait for them, and this is what happens in a test run:

```text
2022-01-23T22:23:03.117Z [pebble] Waited for child 99643 of PGID 99638
2022-01-23T22:23:03.117Z [pebble] Waited for child 99644 of PGID 99638
2022-01-23T22:23:03.118Z [pebble] Service "srv3" stopped
```

I also added the Prctl syscall to not let children escape to PID 1, with setting pebble as the sub-child reaper, so even when a child "escapes", it gets to PPID of pebble (run) and the PGID remains the one initially set when starting the process.

The current implementation thus assumes that there are no guarantees on the service that it is written in a way that children are waited on and cannot become zombies, so pebble assumes the role of a reaper (syscall.Prctl) and also waits explicitly on any children that are still waitable in the PGID of the process after it has been properly waited on (s.cmd.Wait()).

benhoyt:
Thanks. Looking into this now. Do you need to update the test? The new `TestStopWaitChildren` test currently fails.

internal/overlord/servstate/handlers.go
```diff
@@ -356,8 +363,12 @@ func (s *serviceData) startInternal() error {
 	done := make(chan struct{})
 	go func() {
 		waitErr := s.cmd.Wait()
+		childErr := s.waitProcessGroup()
+		if childErr != nil {
+			logger.Noticef("Cannot wait for children processes after main process exit: %v", childErr)
+		}
 		close(done)
-		err := s.exited(waitErr)
+		err = s.exited(waitErr)
```
benhoyt, line 371: This should be `:=` here to avoid setting the outer `err` (though if we change `childErr :=` above to just `err := s.waitProcessGroup()` then we won't need it.

internal/overlord/servstate/handlers.go
```diff
@@ -23,6 +23,13 @@ import (
 	"github.com/canonical/pebble/internal/strutil/shlex"
 )
 
+func init() {
+	// Reap sub-children via setting the child subreaper
+	// that way we can reap "grandchildren" from our main process,
+	// effectively we can wait for the children of the terminating process
+	unix.Prctl(unix.PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0)
+}
+
```
benhoyt, line 32: I think setting ourselves as a subreaper is the right thing for Pebble -- good find.

However, let's not do it in a package-level `init()`. That means importing the package has significant side effects. I think we should just do this in `servstate.NewManager` -- that way we can check and return an error too.

internal/overlord/servstate/handlers.go
```diff
@@ -356,8 +363,12 @@ func (s *serviceData) startInternal() error {
 	done := make(chan struct{})
 	go func() {
 		waitErr := s.cmd.Wait()
+		childErr := s.waitProcessGroup()
```
benhoyt, line 366: Nit: I think this should just be called `err`, as in `err := s.waitProcessGroup()`. (The `waitErr` is a bit of an anomaly as it's not tested right away but passed elsewhere.)

internal/overlord/servstate/handlers.go
benhoyt: I think we can also clarify the comment a bit. How about something like this?

```
// Set Pebble as a "child subreaper" so it becomes the parent of orphaned children
// rather than PID 1, allowing us to wait for processes that a service starts (see
// https://unix.stackexchange.com/a/250156/73491).
```

internal/overlord/servstate/handlers.go
```diff
@@ -601,6 +612,31 @@ func (s *serviceData) stop() error {
 	return nil
 }
 
+func (s *serviceData) waitProcessGroup() error {
```
benhoyt, line 615: Two small things:

1) I'd be inclined to call this `waitChildren()` now as it's waiting on all children, not just the process group -- let's make the name reflect what is does clearly
2) Let's make it a standalone function `func waitChildren(pid int) error { ... }`, as it only depends on `s.cmd.Process.Pid` from `s`, so it's cleaner to pass just the dependencies we use.

internal/overlord/servstate/handlers.go
```diff
@@ -601,6 +612,31 @@ func (s *serviceData) stop() error {
...
+	var (
+		status syscall.WaitStatus
+		rusage syscall.Rusage
+		pid1   int
+		err    error
+	)
+	for {
+		// __WALL (since Linux 2.4)
+		// Wait for all children, regardless of type ("clone" or "non-clone").
+		pid1, err = syscall.Wait4(-s.cmd.Process.Pid, &status, syscall.WALL, &rusage)
```
benhoyt, line 625: A couple of things:

* The docs say the wait functions check if `status` and `rusage` are not NULL, so we can just pass these in as `nil`, seeing we're not using the results.
* Also, let's rename `pid1` to `wpid` (what the `Wait4` docs call it).
* It's a bit simpler to use `wpid, err := syscall.Wait4(...)` and check err inside the loop to avoid having to define the vars at the top.
* Do we need to handle `EINTR` still?
* I think these scattered comments could be tidied up a bit. The `__WALL` comment and the die.net link probably aren't needed, for example.

internal/overlord/servstate/handlers.go
```diff
@@ -601,6 +612,31 @@ func (s *serviceData) stop() error {
...
+		// https://linux.die.net/man/2/waitid
+		// wait(): on success, returns the process ID of the terminated child; on error, -1 is returned.
+		if pid1 == -1 {
+			break
+		}
+		logger.Noticef("Waited for child %d of PGID %d", pid1, s.cmd.Process.Pid)
```
benhoyt, line 631: Should this be `Debugf`? We probably don't want to know about it in the normal case.

benhoyt (changes requested):
I've added a bunch of comments. Mainly style things, but I want to make sure we get the tests and logic right. Currently the tests fail intermittently, so something's not quite right. For example, `TestStopStartUpdatesConfig` and `TestStartStopServicesIdempotency` fail every few runs for me.

internal/overlord/servstate/handlers.go
merkata: This was a typo actually from trying out things, but I agree with having `childErr` be just `err`, so I changed that.

internal/overlord/servstate/handlers.go
merkata: Makes sense, I saw some `init()` funcs in the code and wasn't sure where this piece belongs best, agree on not having side effects and thanks for the pointer to `NewManager`, I've added it there now.
Comment is great and I just set it as is.

internal/overlord/servstate/handlers.go
merkata: Agree on both points, good remarks! Adjusted.

internal/overlord/servstate/handlers.go
merkata: Agreed on the points and adjusted the code, please have a look. For EINTR, it is about pebble receiving a signal that is not blocked and I don't think we need it in this case. From the documentation:
`If a blocked call to one of the following interfaces is interrupted by a signal handler, then the call will be automatically restarted after the signal handler returns if the SA_RESTART flag was used; otherwise the call will fail with the error EINTR.`

internal/overlord/servstate/handlers.go
merkata: Yes, changed to Debugf (this is what I was using it for in the first place) :)

benhoyt:
Thanks. Code looks good! However, the new test `TestStopWaitChildren` still doesn't pass. See: https://github.com/canonical/pebble/runs/4961072662?check_suite_focus=true

benhoyt:
You can run all the tests locally with `go test ./...`

internal/overlord/servstate/handlers.go
merkata: Agree, adjusted.

merkata:
Hi @benhoyt  sorry, I meant to leave a note about the test, I've noticed it failing and wanted to think about how the test should look like in order to verify the expected behaviour. The test now implements the logic of the manual testing I did (background child processes that are not waited for by the main "service" process) and the test also passes.

The test command:
```bash
bash -c 'bash -c "{ trap \"echo child 1\" SIGTERM; sleep 0.4; } & { trap \"echo child 2\" SIGTERM; sleep 0.3; } & sleep 0.2"; sleep 0.1'
```

Test behaviour:
```regex
2.* \[test2\] Terminated\n2.* \[test2\] child.*\n2.* \[test2\] Terminated\n2.* \[test2\] child.*\n
```

Actual result:
```text
2022-01-27T21:35:21.571Z [test4] Terminated
2022-01-27T21:35:21.572Z [test4] child 1
2022-01-27T21:35:21.572Z [test4] Terminated
2022-01-27T21:35:21.572Z [test4] child 2
```

Note that signals can be delivered in any order, so we're interested in having two children waited for (the rest is taken care of by `bash` as expected).

These resources were super helpful in wrapping my head around how to express what I want in bash next to the official (tldp.org) documentation:
https://mywiki.wooledge.org/ProcessManagement
https://mywiki.wooledge.org/SignalTrap

merkata:
I didn't implement fully what I wrote:

> Note that signals can be delivered in any order, so we're interested in having two children waited for (the rest is taken care of by bash as expected).

Now I've adjusted the test so that any order the signals are received (the expected ones), the test should pass.

benhoyt:
@merkata Can you please give me permission to push to your repo/branch so I can resolve conflicts against master on this PR? Thanks.

benhoyt:
Unfortunately I don't think the `TestStopWaitChildren` is doing what we want, as it passes just fine with the new code (Prctl and waitChildren) commented out. Looking into this further...

benhoyt:
Hi @merkata: I've dug into this some more and reproduced the issue locally with a little "zombie creator" program. I think what we want to do here is actually reap zombies globally (like other process managers such as tini and supervisord do), in order to fix the broader issues at https://github.com/canonical/pebble/issues/6.

So I'm going to close this PR and continue working on the larger fix on a different branch. However, I really appreciate the time you spent on this -- it really helped us focus on the issue, and the `PR_SET_CHILD_SUBREAPER` thing is spot on. I'll be working on this over the next couple of days -- please follow further work at https://github.com/canonical/pebble/issues/6. Thanks again.

hpidcock:
Would also like to repeat what @benhoyt said, @merkata thank-you for the PR, this prompted some good discussions internally and put us on the right path. If you have any future issues with Pebble we would be glad to help out either by opening an issue/PR here or you can reach us at https://chat.charmhub.io/charmhub/channels/pebble

merkata:
@benhoyt  @hpidcock  thanks, I really enjoyed the discussion and I'm glad the PR was useful in some way :)

PR #105, opened 2022-02-17
Add a comment to serviceState about the state diagram

benhoyt:
Was included as a drive-by update in https://github.com/canonical/pebble/pull/101, but making it a separate PR to avoid noise in that one.


benhoyt:
Trivial comment fix, merging directly.

PR #108, opened 2022-03-08
Fix panic when replan is called and there are no services to stop/start

### LINKED ISSUE #94: Replan crashes with panic when there are no active services
To reproduce, start `pebble run` with this config:

```
services:
    test:
        override: replace
        command: sleep 300
```

And then execute `pebble replan`. The server will panic. Here is the stack trace:

```
2021/12/13 12:26:59 http: panic serving pid=55794;uid=1000;socket=/home/ben/pebble/.pebble.socket;: runtime error: index out of range [0] with length 0
goroutine 50 [running]:
net/http.(*conn).serve.func1()
	/home/ben/sdk/go1.17/src/net/http/server.go:1801 +0xb9
panic({0x7ac960, 0xc000288030})
	/home/ben/sdk/go1.17/src/runtime/panic.go:1047 +0x266
github.com/canonical/pebble/internal/daemon.v1PostServices(0xa9c840, 0xc0002aa200, 0xc000282090)
	/home/ben/w/pebble/internal/daemon/api_services.go:186 +0x1072
github.com/canonical/pebble/internal/daemon.(*Command).ServeHTTP(0xa9c840, {0x856eb0, 0xc0002900c0}, 0xc0002aa200)
	/home/ben/w/pebble/internal/daemon/daemon.go:251 +0x27a
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0000b8000, {0x856eb0, 0xc0002900c0}, 0xc0002aa000)
	/home/ben/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210 +0x1cf
github.com/canonical/pebble/internal/daemon.logit.func1({0x8570c0, 0xc0002b4000}, 0xc0002aa000)
	/home/ben/w/pebble/internal/daemon/daemon.go:322 +0xdd
net/http.HandlerFunc.ServeHTTP(0x0, {0x8570c0, 0xc0002b4000}, 0xc000290078)
	/home/ben/sdk/go1.17/src/net/http/server.go:2046 +0x2f
net/http.serverHandler.ServeHTTP({0x8562e8}, {0x8570c0, 0xc0002b4000}, 0xc0002aa000)
	/home/ben/sdk/go1.17/src/net/http/server.go:2878 +0x43b
net/http.(*conn).serve(0xc00028c0a0, {0x8593a0, 0xc0002140c0})
	/home/ben/sdk/go1.17/src/net/http/server.go:1929 +0xb08
created by net/http.(*Server).Serve
	/home/ben/sdk/go1.17/src/net/http/server.go:3033 +0x4e8
```

benhoyt:
When a replan was executed and there were no services to stop/start
(for example, when there are no startup: enabled services), the
POST /v1/services endpoint would panic due to the unhandled case.
Handle this case, which is a change with no tasks -- it needs to be a
real change because the endpoint returns a change ID (the client is
expecting a change ID or an error). It shouldn't be an error because
the replan is actually successful, there's just nothing to do.

Also add a test for this case.

This also removes newChange, which was only used in one place, and
less clear than just including it inline.

Fixes #94

### DIFF
--- a/internal/daemon/api_services.go
+++ b/internal/daemon/api_services.go
@@ -180,12 +180,26 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 	// Use the original requested service name for the summary, not the
 	// resolved one. But do use the resolved set for the count.
 	var summary string
-	if len(services) == 1 {
+	switch len(services) {
+	case 0:
+		// Can happen with a replan that has no services to stop/start. A
+		// change with no tasks needs to be marked Done manually (normally a
+		// change is marked Done when its last task is finished).
+		summary = fmt.Sprintf("%s - no services", strings.Title(payload.Action))
+		change := st.NewChange(payload.Action, summary)
+		change.SetStatus(state.DoneStatus)
+		return AsyncResponse(nil, change.ID())
+	case 1:
 		summary = fmt.Sprintf("%s service %q", strings.Title(payload.Action), payload.Services[0])
-	} else {
+	default:
 		summary = fmt.Sprintf("%s service %q and %d more", strings.Title(payload.Action), payload.Services[0], len(services)-1)
 	}
-	change := newChange(st, payload.Action, summary, []*state.TaskSet{taskSet}, payload.Services)
+
+	change := st.NewChange(payload.Action, summary)
+	change.AddAll(taskSet)
+	if len(payload.Services) > 0 {
+		change.Set("service-names", payload.Services)
+	}
 
 	stateEnsureBefore(st, 0)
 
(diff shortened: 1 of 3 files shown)


internal/daemon/api_services.go
```diff
@@ -180,12 +180,26 @@ func v1PostServices(c *Command, r *http.Request, _ *userState) Response {
 	// Use the original requested service name for the summary, not the
 	// resolved one. But do use the resolved set for the count.
 	var summary string
-	if len(services) == 1 {
+	switch len(services) {
+	case 0:
```
hpidcock, line 184: Should we not instead check if the task set is empty? Has the same outcome, but we shouldn't rely on the service names used for summary generation as someone might add an operation that can affect more than services.

hpidcock (approved):
LGTM except one issue

internal/daemon/api_services.go
benhoyt: Fair - done

PR #109, opened 2022-03-10
Reduce server shutdown timeout from 25s to 1s

### LINKED ISSUE #55: `pebble logs -f` blocks shutdown of pebble
SIGINT/SIGTERM to pebble daemon while `pebble logs -f` is active does not behave as expected.

Pebble daemon waits for the `/v1/logs?follow=true` request to be cancelled.

benhoyt:
This is primarily so that when you're debugging locally and press
Ctrl-C to terminate the Pebble server, an open "pebble logs -f" stream
doesn't block shutdown for 25 seconds. There's no need for this timeout
to be so long, so reduce it.

Note that this is only the time it takes to shut down the API server
(perhaps it should be called something more specific like
`serverShutdownTimeout`).

After the shutdown timeout is elapsed, the server terminates the
connection and the client gets an "EOF" error.

Fixes #55

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -539,7 +539,7 @@ var (
 	rebootMaxTentatives    = 3
 )
 
-var shutdownTimeout = 25 * time.Second
+var shutdownTimeout = 3 * time.Second
 
 // Stop shuts down the Daemon.
 func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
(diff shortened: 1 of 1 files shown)


rwcarlsen (approved):
You can trust my review - I'm an expert...

hpidcock (approved):
Could be one second if this is just the http server.

benhoyt:
Merging this trivial change after @hpidcock's review.

Issue #56, opened 2021-07-12
`pebble logs -f` does not see services started after API call.


hpidcock:
If `pebble logs -f` happens before `pebble start <svc>`, logs from `<svc>` are omitted.

benhoyt:
@hpidcock I realize it's a bit of a cheat, but see https://github.com/canonical/pebble/pull/110 -- this just documents the current behaviour. It would be non-trivial to wire this up, and it doesn't seem like it'd be that important for debugging. Happy to leave this issue open if you think it's important, though.

benhoyt:
Per discussion, I've merged #110 to make the help match the current behaviour, but we'll keep this issue open.

benhoyt:
This would still be a useful feature, however, it's not something anyone else has asked for. Going to close for now, can always revisit.

PR #110, opened 2022-03-10
Document that "pebble logs -f" only follows currently-running services

benhoyt:
It would be quite a significant refactoring to have "pebble logs -f"
subscribe to new services being started and wire that through, so for
now at least, just document the current behaviour. It seems unlikely to
affect people in practice, as the service(s) you're debugging will
likely be running.

See https://github.com/canonical/pebble/issues/56


hpidcock (approved):
Obviously doesn't fix the problem but alerts users to the edge case.

PR #112, opened 2022-03-14
New SocketNotFoundError and use of %w for errors

tlm:
- This PR introduces a new error type SocketNotFoundError to the client
  so the dialer can report SocketNotFound errors. Needed by Juju to
  properly distinguish between errors form Pebble. The reason for adding
  it to Pebble and not Juju is it's not an easy situation to write tests
  for in Juju and by having it natively as part of the Pebble library
  allows for better error injection into the Juju tests.

- Have also swapped the usage of Errorf in the client over to wrapping
  errors with %w instead of the mix of %s & %v currently used.


client/client.go
```diff
@@ -32,6 +33,29 @@ import (
 	"github.com/canonical/pebble/internal/wsutil"
 )
 
+// SocketNotFoundError is the error type returned when the Pebble client fails
+// to find a unix socket at the specified path.
```
benhoyt, line 37: Nit: unix -> Unix

client/client.go
```diff
@@ -32,6 +33,29 @@ import (
...
+type SocketNotFoundError struct {
+	// Error wrapped error.
```
benhoyt, line 39: This doesn't read very well. Should it say `// Err is the wrapped error`?

client/client.go
```diff
@@ -32,6 +33,29 @@ import (
...
+	Err error
+
+	// SocketPath of the non existent socket.
+	SocketPath string
```
benhoyt, line 43: Naming: the type already includes the word "socket", so this could be called just `Path`.

client/client.go
```diff
@@ -32,6 +33,29 @@ import (
 	"github.com/canonical/pebble/internal/wsutil"
 )
 
+// SocketNotFoundError is the error type returned when the Pebble client fails
+// to find a unix socket at the specified path.
+type SocketNotFoundError struct {
+	// Error wrapped error.
+	Err error
+
+	// SocketPath of the non existent socket.
```
benhoyt, line 42: Nit: should this say `// SocketPath is the path of the non-existent socket.` (added "is the path" and hyphen)

client/client.go
```diff
@@ -32,6 +33,29 @@ import (
...
+	SocketPath string
+}
+
+// Error implements errors error interface.
```
benhoyt, line 46: Nit: "implements *the* error interface"?

client/client.go
```diff
@@ -32,6 +33,29 @@ import (
...
+func (s SocketNotFoundError) Error() string {
+	if s.SocketPath == "" && s.Err != nil {
```
benhoyt, line 48: Seems like this is dead code, as `s.Err` will always be non-nil and `s.SocketPath` will always be set.

client/client.go
```diff
@@ -32,6 +33,29 @@ import (
...
+		return s.Err.Error()
+	}
+	return fmt.Sprintf("socket %q not found", s.SocketPath)
+}
+
+// Unwrap implements errors Unwrap interface.
```
benhoyt, line 54: Nit: maybe `// Unwrap implements the errors.Unwrap interface.`?

client/client.go
```diff
@@ -48,6 +72,15 @@ func decodeWithNumber(r io.Reader, value interface{}) error {
...
 func unixDialer(socketPath string) func(string, string) (net.Conn, error) {
 	return func(_, _ string) (net.Conn, error) {
+		if _, err := os.Stat(socketPath); os.IsNotExist(err) {
+			return nil, &SocketNotFoundError{
+				Err:        err,
+				SocketPath: socketPath,
+			}
+		} else if err != nil {
+			return nil, fmt.Errorf("checking if pebble socket path %q exists: %w", socketPath, err)
+		}
```
benhoyt, line 82: IMO this would be slightly clearer flattened as:

```
_, err := os.Stat(socketPath)
if errors.Is(err, fs.ErrNotExist) {
	return nil, &SocketNotFoundError{Err: err, SocketPath: socketPath}
}
if err != nil {
	return ...
}
```

Note also the use of `errors.Is`, which is what the Go docs now recommend for new code, instead of `os.IsNotExist`.

client/client.go
```diff
@@ -48,6 +72,15 @@ func decodeWithNumber(r io.Reader, value interface{}) error {
 
 func unixDialer(socketPath string) func(string, string) (net.Conn, error) {
 	return func(_, _ string) (net.Conn, error) {
+		if _, err := os.Stat(socketPath); os.IsNotExist(err) {
+			return nil, &SocketNotFoundError{
+				Err:        err,
+				SocketPath: socketPath,
+			}
+		} else if err != nil {
+			return nil, fmt.Errorf("checking if pebble socket path %q exists: %w", socketPath, err)
```
benhoyt, line 81: This error message was a bit confusing to me at first. It should also basically never happen. What about dropping the word "pebble" and tightening it to just `fmt.Errorf("cannot stat %q: %w", socketPath, err)`?

client/client.go
```diff
@@ -180,6 +213,11 @@ func (e ConnectionError) Error() string {
 	return fmt.Sprintf("cannot communicate with server: %v", e.error)
 }
 
+// Unwrap implements errors Unwrap interface.
```
benhoyt, line 216: Same nit: `// Unwrap implements the errors.Unwrap interface.`

client/client.go
```diff
@@ -281,7 +319,7 @@ func decodeInto(reader io.Reader, v interface{}) error {
 		if err1 != nil {
 			buf = []byte(fmt.Sprintf("error reading buffered response body: %s", err1))
 		}
-		return fmt.Errorf("cannot decode %q: %s", buf, err)
+		return fmt.Errorf("cannot decode %q: %w", buf, err)
```
benhoyt, line 322: Good fixes, thanks.

client/client_test.go
```diff
@@ -311,3 +311,13 @@ func (cs *clientSuite) TestDebugGet(c *C) {
 	c.Check(cs.reqs[0].URL.Path, Equals, "/v1/debug")
 	c.Check(cs.reqs[0].URL.Query(), DeepEquals, url.Values{"action": []string{"do-something"}, "foo": []string{"bar"}})
 }
+
+func (cs *clientSuite) TestNonExistentSocketErrors(c *C) {
+	cli, err := client.New(&client.Config{Socket: "/tmp/not-the-droids-you-are-looking-for-bm90IHRoZSBkcm9pZHMgeW91IGFyZSBsb29raW5nIGZvcgo"})
```
benhoyt, line 316: Love the "not the droids" message. :-) Is the bm90I... stuff needed though?

client/client_test.go
```diff
@@ -311,3 +311,13 @@ func (cs *clientSuite) TestDebugGet(c *C) {
...
+	c.Check(err, IsNil)
+
+	_, err = cli.SysInfo()
+	c.Check(err, NotNil)
+	var notFoundErr *client.SocketNotFoundError
+	c.Check(errors.As(err, &notFoundErr), Equals, true)
```
benhoyt, line 322: Would be good to also test that the `notFoundErr.SocketPath` field is what we expect, and the `notFoundErr.Err` field is not nil.

benhoyt (approved):
A bunch of nit comments, but looks good to me, thanks!

client/client_test.go
tlm: Probably not I was just being overly verbose.

client/client.go
tlm: Agreed

client/client.go
tlm: I was just trying to add the cases if something ever uses this error outside of our current use case. It was more just cover the case for when and if this happens.

benhoyt: Fine with me. It's certainly not going to hurt. +1

benhoyt (approved):

niemeyer (approved):
LGTM, thanks!

PR #113, opened 2022-03-17
automatic timestamp trimming for service logs

### LINKED ISSUE #91: Strip timestamp from service logs to avoid double-timestamp issue
Currently when a service outputs timestamps itself we get double timestamps in the log, because Pebble also adds a timestamp-and-service-name prefix to every log line. We want to trim/strip the service's timestamp to avoid the double-up.

Some notes per discussion with Harry and Gustavo:

* Add a new "time-trim" field per service: defaults to "auto", can also be "disabled" (or a format string -- maybe later)
* How to do time.ParsePrefix prefix parsing? Go's time.Parse doesn't support parsing a timestamp as a prefix
   - See also https://pypi.org/project/python-dateutil/ and https://github.com/araddon/dateparse
* Auto-detect might detect format once at startup, or re-detect every so often
* Should we used the parsed time in place of the Pebble timestamp, or just drop it?
* Feedback from Thomas: fair bit of overhead, do performance testing (logs/s)

rwcarlsen:
This adds support for a user-specified regex for matching against service log line prefixes (e.g. redundant time stamps) that are then discarded before pebble attaches its own timestamp/prefix.  This can be specified in a pebble service section like this:
    
```
services:
    foo-serv:
        override: replace
        comand: foo
        time-trim: 1/2/2006
```
    
Before we would get:
    
```
3456-01-02T00:00:01.000Z07:00 [foo-serv] 1/2/3456 [foo]  foo the bar
3456-01-02T00:00:02.000Z07:00 [foo-serv] 1/2/3456 [foo]  bar the baz
...
```
    
and with the time-trim expression, we now get:
    
```
3456-01-02T00:00:01.000Z07:00 [foo-serv] foo the bar
3456-01-02T00:00:02.000Z07:00 [foo-serv] bar the baz
...
```

The current TrimWriter implementation buffers entire lines before trimming the prefix and forwarding the writes.  I initially started writing an implementation that started forwarding writes as soon as an entire timestamp/prefix match was found, but it didn't seem like a good tradeoff in code complexity vs potential mem/cpu gains. I don't expect most log lines will be particularly long (let me know if I'm wrong here).
    
Fixes #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,12 +246,13 @@ services:
         # half a minute ("30s").
         backoff-limit: <duration>
 
-        # (Optional) Regular expression for trimming prefixes from log
-        # messages.  This is used to discard matching prefixes from the
-        # service's log output before pebble attaches its own
-        # timestamp/prefix.  This can be useful for e.g. avoiding redundant
-        # timestamping.
-        log-trim: <regular-expression>
+        # (Optional) Time layout expression for trimming timestamps from log
+        # messages.  It must be specified using the reference-time syntax
+        # of the Go time package. This is used to discard redundant timestamp
+        # prefixes from the service's log output before pebble attaches its own
+        # timestamp/prefix. If omitted, Pebble will attempt to auto-detect and
+        # trim timestamps from the service's logs.
+        time-trim: <timestamp-layout>
 
 
 # (Optional) A list of health checks managed by this configuration layer.
@@ -387,7 +388,7 @@ Here are some of the things coming soon:
   - [x] Consider showing unified log as output of `pebble run` (use `-v`)
   - [x] Automatically restart services that fail
   - [x] Support for custom health checks (HTTP, TCP, command)
-  - [x] Support trimming prefixes (e.g. redundant timestamps) from logs
+  - [x] Automatically remove (double) timestamps from logs
   - [ ] Improve signal handling, e.g., sending SIGHUP to a service
   - [ ] Terminate all services before exiting run command
   - [ ] More tests for existing CLI commands
--- a/README.md
+++ b/README.md
@@ -246,6 +246,13 @@ services:
         # half a minute ("30s").
         backoff-limit: <duration>
 
+        # (Optional) Regular expression for trimming prefixes from log
+        # messages.  This is used to discard matching prefixes from the
+        # service's log output before pebble attaches its own
+        # timestamp/prefix.  This can be useful for e.g. avoiding redundant
+        # timestamping.
+        log-trim: <regular-expression>
+
 
 # (Optional) A list of health checks managed by this configuration layer.
 checks:
@@ -380,7 +387,7 @@ Here are some of the things coming soon:
   - [x] Consider showing unified log as output of `pebble run` (use `-v`)
   - [x] Automatically restart services that fail
   - [x] Support for custom health checks (HTTP, TCP, command)
-  - [ ] Automatically remove (double) timestamps from logs
+  - [x] Support trimming prefixes (e.g. redundant timestamps) from logs
   - [ ] Improve signal handling, e.g., sending SIGHUP to a service
   - [ ] Terminate all services before exiting run command
   - [ ] More tests for existing CLI commands
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"regexp"
 	"syscall"
 	"time"
 
@@ -313,7 +314,7 @@ func logError(err error) {
 }
 
 // startInternal is an internal helper used to actually start (or restart) the
-// command. It assumes the caller has ensures the service is in a valid state,
+// command. It assumes the caller has ensured the service is in a valid state,
 // and it sets s.cmd and other relevant fields.
 func (s *serviceData) startInternal() error {
 	args, err := shlex.Split(s.config.Command)
@@ -351,7 +352,13 @@ func (s *serviceData) startInternal() error {
 		outputIterator = s.logs.HeadIterator(0)
 	}
 	serviceName := s.config.Name
-	logWriter := servicelog.NewFormatWriter(s.logs, serviceName)
+	var logWriter io.Writer = servicelog.NewFormatWriter(s.logs, serviceName)
+	flush := func() error { return nil }
+	if s.config.LogTrim != "" {
+		tw := servicelog.NewTrimWriter(logWriter, regexp.MustCompile("^("+s.config.LogTrim+")"))
+		logWriter = tw
+		flush = tw.Flush
+	}
 	s.cmd.Stdout = logWriter
 	s.cmd.Stderr = logWriter
 
@@ -378,8 +385,12 @@ func (s *serviceData) startInternal() error {
 		} else {
 			logger.Debugf("Service %q exited with code %d.", serviceName, exitCode)
 		}
+		err := flush()
+		if err != nil {
+			logger.Noticef("Cannot flush service logs: %v", err)
+		}
 		close(done)
-		err := s.exited(exitCode)
+		err = s.exited(exitCode)
 		if err != nil {
 			logger.Noticef("Cannot transition state after service exit: %v", err)
 		}
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"regexp"
 	"syscall"
 	"time"
 
@@ -353,12 +352,9 @@ func (s *serviceData) startInternal() error {
 	}
 	serviceName := s.config.Name
 	var logWriter io.Writer = servicelog.NewFormatWriter(s.logs, serviceName)
-	flush := func() error { return nil }
-	if s.config.LogTrim != "" {
-		tw := servicelog.NewTrimWriter(logWriter, regexp.MustCompile("^("+s.config.LogTrim+")"))
-		logWriter = tw
-		flush = tw.Flush
-	}
+	trimmer := servicelog.NewTrimWriter(logWriter, s.config.TimeTrim)
+	logWriter = trimmer
+
 	s.cmd.Stdout = logWriter
 	s.cmd.Stderr = logWriter
 
@@ -385,7 +381,7 @@ func (s *serviceData) startInternal() error {
 		} else {
 			logger.Debugf("Service %q exited with code %d.", serviceName, exitCode)
 		}
-		err := flush()
+		err := trimmer.Flush()
 		if err != nil {
 			logger.Noticef("Cannot flush service logs: %v", err)
 		}
--- a/internal/overlord/servstate/manager_test.go
+++ b/internal/overlord/servstate/manager_test.go
@@ -100,6 +100,11 @@ services:
     test2:
         override: replace
         command: /bin/sh -c "echo test2 | tee -a %s; sleep 300"
+
+    trim-test:
+        override: replace
+        command: /bin/sh -c "echo prefix-to-trim  trim-test prefix-to-trim | tee -a %s; sleep 300"
+        log-trim: prefix-to-trim *
 `
 
 var planLayer2 = `
@@ -143,7 +148,7 @@ func (s *S) SetUpTest(c *C) {
 	os.Mkdir(filepath.Join(s.dir, "layers"), 0755)
 
 	s.log = filepath.Join(s.dir, "log.txt")
-	data := fmt.Sprintf(planLayer1, s.log, s.log)
+	data := fmt.Sprintf(planLayer1, s.log, s.log, s.log)
 	err := ioutil.WriteFile(filepath.Join(s.dir, "layers", "001-base.yaml"), []byte(data), 0644)
 	c.Assert(err, IsNil)
 	err = ioutil.WriteFile(filepath.Join(s.dir, "layers", "002-two.yaml"), []byte(planLayer2), 0644)
@@ -245,15 +250,15 @@ func (s *S) stopServices(c *C, services []string, nEnsure int) *state.Change {
 }
 
 func (s *S) startTestServices(c *C) {
-	chg := s.startServices(c, []string{"test1", "test2"}, 2)
+	chg := s.startServices(c, []string{"test1", "test2", "trim-test"}, 3)
 	s.st.Lock()
 	c.Check(chg.Status(), Equals, state.DoneStatus, Commentf("Error: %v", chg.Err()))
 	s.st.Unlock()
 
-	s.assertLog(c, ".*test1\n.*test2\n")
+	s.assertLog(c, ".*test1\n.*test2\n.*trim.*\n")
 
 	cmds := s.manager.RunningCmds()
-	c.Check(cmds, HasLen, 2)
+	c.Check(cmds, HasLen, 3)
 }
 
 func (s *S) TestStartStopServices(c *C) {
@@ -287,12 +292,12 @@ func (s *S) TestStartStopServicesIdempotency(c *C) {
 
 func (s *S) stopTestServices(c *C) {
 	cmds := s.manager.RunningCmds()
-	c.Check(cmds, HasLen, 2)
+	c.Check(cmds, HasLen, 3)
 
-	chg := s.stopServices(c, []string{"test1", "test2"}, 2)
+	chg := s.stopServices(c, []string{"test1", "test2", "trim-test"}, 3)
 
 	// Ensure processes are gone indeed.
-	c.Assert(cmds, HasLen, 2)
+	c.Assert(cmds, HasLen, 3)
 	for name, cmd := range cmds {
 		err := cmd.Process.Signal(syscall.Signal(0))
 		if err == nil {
@@ -311,7 +316,7 @@ func (s *S) stopTestServicesAlreadyDead(c *C) {
 	cmds := s.manager.RunningCmds()
 	c.Check(cmds, HasLen, 0)
 
-	chg := s.stopServices(c, []string{"test1", "test2"}, 2)
+	chg := s.stopServices(c, []string{"test1", "test2", "trim-test"}, 3)
 
 	c.Assert(cmds, HasLen, 0)
 
@@ -395,6 +400,15 @@ services:
 	c.Check(config.Summary, Equals, "A summary!")
 }
 
+func (s *S) TestServiceLogTrim(c *C) {
+	outputs := map[string]string{"trim-test": `2.* \[trim-test\] trim-test prefix-to-trim\n`}
+	s.testServiceLogs(c, outputs)
+
+	// Run test again, but ensure the logs from the previous run are still in the ring buffer.
+	outputs["trim-test"] += outputs["trim-test"]
+	s.testServiceLogs(c, outputs)
+}
+
 func (s *S) TestServiceLogs(c *C) {
 	outputs := map[string]string{
 		"test1": `2.* \[test1\] test1\n`,
@@ -415,9 +429,14 @@ func (s *S) testServiceLogs(c *C, outputs map[string]string) {
 		return
 	}
 
-	iterators, err := s.manager.ServiceLogs([]string{"test1", "test2"}, -1)
+	services := []string{}
+	for service := range outputs {
+		services = append(services, service)
+	}
+
+	iterators, err := s.manager.ServiceLogs(services, -1)
 	c.Assert(err, IsNil)
-	c.Assert(iterators, HasLen, 2)
+	c.Assert(iterators, HasLen, len(services))
 
 	for serviceName, it := range iterators {
 		buf := &bytes.Buffer{}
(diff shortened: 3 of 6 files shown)


internal/servicelog/formatter.go
```diff
@@ -52,6 +54,57 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
...
+		end := bytes.IndexRune(w.buf[pos:], '\n')
+		if end == -1 {
+			return written, nil // wait for rest of line
+		}
+		end += pos
+		loc := w.re.FindIndex(w.buf[:end])
+		start := 0
+		if loc != nil && loc[0] == 0 {
+			start = loc[1]
+		}
```
benhoyt, line 91: We only want it to match on the left, correct? Can we use a `^` regex anchor, to make it more efficient (so the regex match isn't doing a search, only to test for `loc[0] == 0` after)? We could take the user's `log-trim` value and make the actual regex `"^(" + logTrim + ")"` or similar?

internal/servicelog/formatter_test.go
```diff
@@ -46,6 +46,39 @@ func (s *formatterSuite) TestFormat(c *C) {
...
+		}
+
+		c.Assert(b.String(), Equals, fmt.Sprintf(`
+hello my name is joe
+and I work in a button factory
+this log entry is very old
+and dates in the middle 1/1/0033 are kept
+`[1:]))
+	}
+}
```
benhoyt, line 80: Are we getting full coverage with this one test? Even if so, seems like we should probably add a few more edge case tests: one with no writes, one with blank lines, one where the last line doesn't have a `\n` in it.

benhoyt (changes requested):
Good change overall! I do quite like this approach. Specific feedback in comments.

internal/overlord/servstate/handlers.go
```diff
@@ -349,7 +350,8 @@ func (s *serviceData) startInternal() error {
 		// started (previous logs have already been copied).
 		outputIterator = s.logs.HeadIterator(0)
 	}
-	logWriter := servicelog.NewFormatWriter(s.logs, s.config.Name)
+	formatWriter := servicelog.NewFormatWriter(s.logs, s.config.Name)
+	logWriter := servicelog.NewTrimWriter(formatWriter, regexp.MustCompile("^"+s.config.LogTrim))
```
benhoyt, line 354: I think we need parentheses around this ... consider if the `log-trim` regex is `foo|bar`, we'd end up with `^foo|bar`, which [matches](https://go.dev/play/p/pQwTQIWa3a1) "foo" at the start of the line and "bar" anywhere in the line, because `^` is higher precedence than `|`. So we should do `"^("+s.config.LogTrim+")"`, or `"^(?:"+s.config.LogTrim+")"` if we need the non-capturing version (I don't think you're using submatches here, so probably the first version is fine).

Also, why are we doing this unconditionally now? I quite liked the fact that previously you only got the performance penalty if `log-trim` was set.

internal/overlord/servstate/handlers.go
```diff
@@ -369,8 +371,12 @@ func (s *serviceData) startInternal() error {
 	done := make(chan struct{})
 	go func() {
 		waitErr := s.cmd.Wait()
+		err := logWriter.Flush()
+		if err != nil {
+			logger.Noticef("Failed to flush service logs: %v", err)
```
benhoyt, line 376: Nit: we've settled on "Cannot X" for log and error messages, so `"Cannot flush srevice logs: %v"`.

internal/servicelog/formatter.go
```diff
@@ -52,7 +54,83 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
...
+
+func (w *TrimWriter) Flush() error {
+	w.mut.Lock()
+	defer w.mut.Unlock()
+
+	if len(w.buf) == 0 {
+		return nil
+	}
+
+	loc := w.re.FindIndex(w.buf)
```
benhoyt, line 80: It seems like this is the same trimming code as below -- can we combine into an unexported helper method?

internal/overlord/servstate/handlers.go
```diff
@@ -350,6 +350,12 @@ func (s *serviceData) startInternal() error {
 		outputIterator = s.logs.HeadIterator(0)
 	}
 	logWriter := servicelog.NewFormatWriter(s.logs, s.config.Name)
+	if s.config.LogTrim != "" {
+		logWriter, err = servicelog.NewTrimWriter(logWriter, s.config.LogTrim)
+		if err != nil {
+			fmt.Errorf("invalid trim expression: %w", err)
```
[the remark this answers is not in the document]
rwcarlsen: fixed.

internal/servicelog/formatter.go
```diff
@@ -52,6 +54,57 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
...
 }
 
+type trimWriter struct {
+	dest     io.Writer
+	re       *regexp.Regexp
+	buf      []byte
+	postTrim bool
+}
+
+func NewTrimWriter(dest io.Writer, regex string) (io.Writer, error) {
```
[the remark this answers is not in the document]
rwcarlsen: Done (pass in Regexp, kill named returns, added Flush method, return concrete type).

internal/servicelog/formatter.go
```diff
@@ -52,6 +54,57 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
...
+}
+
+func (w *trimWriter) Write(p []byte) (nn int, ee error) {
+	// Buffered content has already been searched for newlines - track this.
+	pos := len(w.buf)
+
+	w.buf = append(w.buf, p...)
+	written := 0
+	for {
+		end := bytes.IndexRune(w.buf[pos:], '\n')
```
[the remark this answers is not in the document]
rwcarlsen: Done.

internal/servicelog/formatter.go
```diff
@@ -52,6 +54,57 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
...
+	if err != nil {
+		return nil, err
+	}
+	return &trimWriter{
+		dest: dest,
+		re:   re,
+	}, nil
+}
+
+func (w *trimWriter) Write(p []byte) (nn int, ee error) {
```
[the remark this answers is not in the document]
rwcarlsen: Done.

internal/servicelog/formatter.go
rwcarlsen: For now, I changed it to prepend `^` to the user-supplied regex.

internal/servicelog/formatter_test.go
rwcarlsen: Should be pretty decent coverage, but you're right - I'll work on some more tests.

internal/overlord/servstate/handlers.go
rwcarlsen: That occurred to me.  2 thoughts:

* If we add default regexes, it seems like it will be rare for users to explicitly blank it for performance reasons.  So in practice the performance win would almost never be realized.  Although it would at least make the workaround possible...
* To avoid the writer wrap, we either need to add a Flush method to the format writer (easy) or a type assertion where we call Flush.

Seems like maybe just adding a Flush method to the regular formatter would be the cleanest

internal/servicelog/formatter.go
rwcarlsen: Good catch - refactored.

internal/overlord/servstate/handlers.go
rwcarlsen: Done.

internal/overlord/servstate/handlers.go
benhoyt: I don't like adding a `Flush` method to the formatter if it's not needed. I think a type assertion where you call `Flush` is fine, like `if f, ok := writer.(interface{ Flush() }); ok { f.Flush() }`. Another approach I used recently was to use a func variable initialized to a no-op function, something like this:

```
flush := func() {}
if logTrim {
    logWriter := NewLogWriter(writer)
    writer = logWriter
    flush = func() { logWriter.Flush() }
}
...
flush()
```

rwcarlsen: I like the inline func var - done.

benhoyt (approved):
Looking good, thanks.

rwcarlsen:
Don't merge yet - I still need to make some better tests!

internal/plan/plan.go
```diff
@@ -65,6 +65,7 @@ type Service struct {
 	Startup     ServiceStartup `yaml:"startup,omitempty"`
 	Override    Override       `yaml:"override,omitempty"`
 	Command     string         `yaml:"command,omitempty"`
+	LogTrim     string         `yaml:"log-trim,omitempty"`
```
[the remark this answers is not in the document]
rwcarlsen: What time formats should be considered for a default regex?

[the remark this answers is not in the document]
benhoyt: Yeah, I don't know - that would require a bit of research. And on second thoughts, I wonder if the sane default is *not* to do that. It could get messy quite quickly. Maybe we just include a couple of example regexes in the README/docs?

It's also a performance hit, as we've discussed, though it would be nice to know how much of one. Would you be able to make a small Go benchmark of logging via just the `FormatWriter` vs logging via both `FormatWriter` and `TrimWriter` and see what the performance of each is like?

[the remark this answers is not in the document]
rwcarlsen: Yeah - I think defaults will either quickly get out of hand, or will not match very often despite always causing the performance hit.  Probably better to have some suggestions in docs.

internal/overlord/servstate/manager_test.go
```diff
@@ -79,7 +79,8 @@ services:
 
     test2:
         override: replace
-        command: /bin/sh -c "echo test2 | tee -a %s; sleep 300"
+        command: /bin/sh -c "echo prefix-to-trim   test2 | tee -a %s; sleep 300"
+        log-trim: prefix.*trim +
```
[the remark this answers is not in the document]
rwcarlsen: I pushed up changes to try to separate this out a bit.  However, the way the ServiceManager tests are set up is not very conducive to adding new cases cleanly.  A lot of the setup/helper functions for all the tests are hard-coded to specific details in the plan/layers.

internal/servicelog/formatter_test.go
rwcarlsen: I rounded out these tests a bit and checked coverage - we're basically hitting everything now.

internal/overlord/servstate/manager_test.go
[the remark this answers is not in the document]
benhoyt: Yeah, I agree with this! They should be much more standalone rather than depending on that one config at the top.

rwcarlsen:
Okay I've hacked together a rough first draft of what timestamp trimming.  I haven't yet compiled a list of good default timestamp formats to look for for detection.  There are a couple of open questions I have:

* How sophisticated do we want timestamp format auto-detection to be w.r.t. retries on failed attempts, skipping the first n log lines maybe, revisiting the autodetect periodically to double check nothing is broken/changed, etc.  The implementation I have now just tries to autodetect every line and gives up after N failures.
* What do we want to do if we have a layout (either auto detected or user-provided), but then later run into a line that doesn't have a correct time prefix?  I implemented just silently writing/logging the full line if the time doesn't parse correctly.
* Do we want to explicitly try to parse the time prefix for every line or just assume the field splitting is accurate and not worry about accidentally stripping non-timestamp content? I have the the latter implemented currently.

internal/plan/plan.go
```diff
@@ -475,9 +475,9 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 				Message: fmt.Sprintf(`plan must define "command" for service %q`, name),
 			}
 		}
-		if _, err := regexp.Compile(service.LogTrim); err != nil {
+		if _, err := regexp.Compile(service.TimeTrim); err != nil {
```
benhoyt, line 478: This is a time format string now right, not a regex?

internal/servicelog/formatter.go
```diff
@@ -151,20 +158,70 @@ func (w *TrimWriter) Write(p []byte) (int, error) {
...
 	}
 	return nil
 }
+
+func timeLayoutPrefix(layout string, line []byte) string {
+	if layout == "" {
+		return ""
+	}
+
+	nfields := len(strings.Split(layout, " "))
```
benhoyt, line 201: For efficiency, this could be done only when the layout changes.

internal/servicelog/formatter.go
```diff
@@ -151,20 +158,70 @@ func (w *TrimWriter) Write(p []byte) (int, error) {
...
+	sep := []byte{' '}
+	fields := bytes.SplitN(line, sep, nfields)
```
benhoyt, line 203: Is this enough, or do we actually have to use `bytes.Fields` (or a similar, custom `FieldsN` type of function) so that runs of whitespace get coalesced into one? For example, times in `time.UnixDate` format may have two spaces before the day number, like `Mon Jan  2 15:04:05 MST 2006`. I'm guessing this code would fail for that?

internal/servicelog/formatter.go
```diff
@@ -151,20 +158,70 @@ func (w *TrimWriter) Write(p []byte) (int, error) {
...
+		// last layout char must be a space otherwise we can't detect the
+		// prefix using whitespace split/fields approach. This causes the last
+		// split field to be an empty string in the layout which we don't want
+		// to map to keeping any log line content - so we exclude
+		// this by reducing the number of fields we keep by one.
+		for _, field := range fields[:nfields-1] {
+			size += len(field) + len(sep)
+		}
+	}
+	return string(line[:size])
```
benhoyt, line 215: Not sure how much difference it takes compared to the actual time parsing, but ideally this would be done with just slicing the original byte slice to avoid to-from string allocations.

internal/servicelog/formatter.go
```diff
@@ -151,20 +158,70 @@ func (w *TrimWriter) Write(p []byte) (int, error) {
 	}
 }
 
-func (w *TrimWriter) write(p []byte) error {
+func (w *TrimWriter) write(line []byte) error {
+	const maxDetectFailures = 10
+	if w.layout == "" && w.detectFailures < maxDetectFailures {
```
benhoyt, line 163: Nice and simple, I like it!

benhoyt:
Nice! Regarding your questions:

1) I think this simple approach is probably enough. The implementation is very simple, and it's also easy to understand. Both good features.
2) I think that's reasonable -- if later a line doesn't parse correctly, just silently write the full line.
3) I think we want to explicitly parse the time prefix for every line, so that we're honoring the time-trim layout. Seems too loosey-goosey otherwise.

internal/servicelog/formatter.go
rwcarlsen: Yeah - I can just cache the number of fields.

internal/servicelog/formatter.go
rwcarlsen: I thought about this a bit, but doesn't the UnixDate format still only have one space?  The `_` in the layout string just indicates that a zero prefix to the digit is omitted.  I was under the impression that it wasn't possible to have multi-space runs that aren't explicitly present in the layout string.  I did it this way because it's much simpler to determine the length of the prefix - using `Fields` makes a bit more onerous because it may gobble up runs of whitespace.

internal/servicelog/formatter.go
rwcarlsen: This is done.

internal/servicelog/formatter.go
benhoyt: No, I think the `_` indicates an additional space if needed. See example here: https://go.dev/play/p/DIF6Deq8C8X and from the `time` package docs:

> Within the format string, the underscores in "_2" and "__2" represent spaces that may be replaced by digits if the following number has multiple digits, for compatibility with fixed-width Unix time formats. A leading zero represents a zero-padded value.
>
> The formats __2 and 002 are space-padded and zero-padded three-character day of year; there is no unpadded day of year format.

rwcarlsen:
Unfortunately, the conditional space padding is turning out to be a real bugger here.  Because of it, there is no fixed number of space-separated fields we can count on for determining the time prefix :-(  For example the layout `2006/1/_2` covers dates such as `2006/3/ 4` (2 fields) and `2112/11/14` (1 field, no spaces).  Additionally, that same example has variable width dates.  To make matters even worse, `_` in the layout only represents leading spaces *sometimes*.  If `_` prefixes the month, for example, it is interpreted as a literal underscore rather than representing leading space for fixed widths; we can't just treat underscores as spaces when determining field counts from a layout without first understanding the semantics of each underscore in the layout.  If we resolved all this for a particular layout, then we'd still have to keep a set of field-count/prefix-length permutations that we checked for on every line.

rwcarlsen:
Another consideration for auto-detection: If the first lines of logs from a service have `2022/12/31`, we might assume that the format is `2006/01/02`, but it might actually be `2006/1/2`, or it might really be `2006/1/_2`.  Lots of ambiguity.  We might need to add in capability to re-try the auto-detection if we run into failed timestamp parsing (maybe only re-autodetect if we get consecutive timestamp parse failures).

rwcarlsen:
I've come up with an approach.  I'm not sure I'd call it elegant, but I think it may be able to work well enough.  For dates like 1/_2/2006 that could result in either "12/31/2112" or "1/ 1/2113" the number of whitespace separated fields is not fixed (1 vs.  2).  My latest mods handle that by replacing underscores in the layout with spaces for determining the maximum number of space-separated fields in a layout.  It then parses incoming log timestamps, potentially with extra content for e.g. cases where we collected 2 fields instead of the necessary 1 as explained above, while checking for and allowing "extra text" errors.  The "extra text" errors are used to determine the actual timestamp prefix length and the trimmed logging then proceeds as normal.

Because of these and other ambiguities, it is possible that we may fail to parse timestamps due to autodetecting the incorrect layout.  This is now handled by counting how many parse failures occur post-auto-detection.  When this count hits a certain threshold, the current auto-detected layout is discarded and we revert back to layout detection mode.

benhoyt:
It seems like an acceptable approach to me. And it seems like it's very unlikely to happen in practice. Will putting the `_` formats at the bottom of the list of formats to try help reduce the number of times this happens?

rwcarlsen:
Yes.  As long as we can make approx determinations about which timestamp formats are more likely, we can certainly put those in the list above ambiguous overlapping formats.

internal/plan/plan.go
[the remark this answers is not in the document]
rwcarlsen: Switching back to timestamp parsing - marking this as resolved.

internal/servicelog/formatter.go
rwcarlsen: This has been addressed with the latest changes.

README.md
```diff
@@ -246,6 +246,13 @@ services:
         # half a minute ("30s").
         backoff-limit: <duration>
 
+        # (Optional) Regular expression for trimming prefixes from log
+        # messages.  This is used to discard matching prefixes from the
+        # service's log output before pebble attaches its own
+        # timestamp/prefix.  This can be useful for e.g. avoiding redundant
+        # timestamping.
+        log-trim: <regular-expression>
```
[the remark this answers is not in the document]
rwcarlsen: done

internal/overlord/servstate/manager_test.go
```diff
@@ -393,9 +407,14 @@ func (s *S) testServiceLogs(c *C, outputs map[string]string) {
 		return
 	}
 
-	iterators, err := s.manager.ServiceLogs([]string{"test1", "test2"}, -1)
+	services := []string{}
+	for service := range outputs {
+		services = append(services, service)
+	}
```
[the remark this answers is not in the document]
rwcarlsen: Fancy

internal/servicelog/formatter.go
```diff
@@ -95,3 +97,150 @@ func (f *formatter) Write(p []byte) (nn int, ee error) {
...
+
+// TrimWriter removes a timestamp prefix from each line written (separated by '\n')
+// and forwards the trimmed writes to a destination writer.  Internal
+// buffering does occur so it is important to call Flush after all writes
+// are completed.  If no timestamp layout is explicitly specified,
+// auto-detection using a list of pre-determined formats is attempted.
+type TrimWriter struct {
+	dest           io.Writer
+	postTrim       bool
+	mut            sync.Mutex
```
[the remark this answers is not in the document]
rwcarlsen: done

rwcarlsen:
Here are samples of the non-pebble component of logs from a few observability and data platform teams' charms:

```
# loki-k8s-operator
level=info ts=2022-05-10T18:37:54.594835469Z caller=table_manager.go:169 msg="uploading tables"

# prometheus-k8s-operator
ts=2022-05-10T18:42:52.152Z caller=head.go:604 level=info component=tsdb msg="WAL segment loaded" segment=0 maxSegment=0

# grafana-k8s-operator
t=2022-05-10T18:47:15+0000 lvl=info msg="Executing migration" logger=migrator id="create index IDX_temp_user_status - v1-7"

# postgresql-k8s-operator
2022-05-10 22:07:00.917 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432

# mysql-k8s-operator
2022-05-10T22:09:30.701161Z 1 [System] [MY-013577] [InnoDB] InnoDB initialization has ended.
```

Notably some of the logs don't exactly have the timestamp as a prefix :-/

And an unrelated thought - perhaps we should include a plan config field/way to disable the auto-timestamp trimming...

benhoyt:
Closing this for the reasons given in https://github.com/canonical/pebble/issues/91#issuecomment-1198871930

PR #114, opened 2022-03-18
Stop all services when Ctrl+C or SIGTERM/SIGQUIT received

benhoyt:
This PR makes Pebble try to gracefully stop all services when SIGINT (Ctrl+C) or SIGTERM/SIGQUIT is received. This goes via the regular "stop" task process, which sends SIGTERM to each process, then waits up to the killWait time (5s) before sending SIGKILL.

The "stop" tasks are added to a new "stop" change. I considered whether a change is appropriate here (given the recent discussion about events), and I do think it is. For one, this is a user-initiated action (Ctrl-C pressed / signal sent), and besides, the tasks are ordinary "stop" tasks to stop the services involved, so they need to be part of an associated change to kick them off. Later when we add events, we can additionally fire a shutdown event of some kind.

Output of `pebble changes` and `pebble tasks` showing the record of the "stop" change and tasks (this is after stopping and restart the Pebble daemon, hence the "autostart" change):

```
$ pebble changes
1    Done    today at 17:01 NZST  today at 17:01 NZST  Autostart service "sleepy" and 1 more
2    Done    today at 17:01 NZST  today at 17:01 NZST  Stop all running services
3    Done    today at 17:01 NZST  today at 17:01 NZST  Autostart service "sleepy" and 1 more

$ pebble tasks 2
Done    today at 17:01 NZST  today at 17:01 NZST  Stop service "sleepy"
Done    today at 17:01 NZST  today at 17:01 NZST  Stop service "test"
```

See related issue: https://github.com/canonical/pebble/issues/6.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -551,6 +551,16 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 		return fmt.Errorf("internal error: no Overlord")
 	}
 
+	// Shut down the service manager, stopping all services. Must do this
+	// before overlord.Stop, as it creates a change and waits for the change,
+	// and overlord.Stop calls StateEngine.Stop, which locks, so Ensure would
+	// result in a deadlock.
+	err := d.overlord.ServiceManager().Shutdown()
+	if err != nil {
+		// This isn't fatal for exiting the daemon, so log and continue.
+		logger.Noticef("Cannot shut down service manager: %v", err)
+	}
+
 	d.tomb.Kill(nil)
 
 	d.mu.Lock()
@@ -601,7 +611,7 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 	}
 	d.overlord.Stop()
 
-	err := d.tomb.Wait()
+	err = d.tomb.Wait()
 	if err != nil {
 		// do not stop the shutdown even if the tomb errors
 		// because we already scheduled a slow shutdown and
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -41,6 +41,7 @@ import (
 	"github.com/canonical/pebble/internal/overlord"
 	"github.com/canonical/pebble/internal/overlord/checkstate"
 	"github.com/canonical/pebble/internal/overlord/restart"
+	"github.com/canonical/pebble/internal/overlord/servstate"
 	"github.com/canonical/pebble/internal/overlord/standby"
 	"github.com/canonical/pebble/internal/overlord/state"
 	"github.com/canonical/pebble/internal/systemd"
@@ -551,14 +552,13 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 		return fmt.Errorf("internal error: no Overlord")
 	}
 
-	// Shut down the service manager, stopping all services. Must do this
-	// before overlord.Stop, as it creates a change and waits for the change,
-	// and overlord.Stop calls StateEngine.Stop, which locks, so Ensure would
-	// result in a deadlock.
-	err := d.overlord.ServiceManager().Shutdown()
+	// Stop all running services. Must do this before overlord.Stop, as it
+	// creates a change and waits for the change, and overlord.Stop calls
+	// StateEngine.Stop, which locks, so Ensure would result in a deadlock.
+	err := d.stopRunningServices()
 	if err != nil {
 		// This isn't fatal for exiting the daemon, so log and continue.
-		logger.Noticef("Cannot shut down service manager: %v", err)
+		logger.Noticef("Cannot stop running services: %v", err)
 	}
 
 	d.tomb.Kill(nil)
@@ -632,6 +632,41 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 		return ErrRestartSocket
 	}
 
+	return nil
+}
+
+// This is a little more than servstate.killWait (but that isn't and shouldn't
+// be exported, so just duplicate it here).
+const stopRunningTimeout = 5*time.Second + 100*time.Millisecond
+
+// stopRunningServices stops all running services, waiting for a short time
+// for them all to stop.
+func (d *Daemon) stopRunningServices() error {
+	taskSet, err := servstate.StopRunning(d.state, d.overlord.ServiceManager())
+	if err != nil {
+		return err
+	}
+	if taskSet == nil {
+		logger.Debugf("No services to stop.")
+		return nil
+	}
+
+	// One change to stop them all.
+	logger.Noticef("Stopping all running services.")
+	st := d.state
+	st.Lock()
+	chg := st.NewChange("stop", "Stop all running services")
+	chg.AddAll(taskSet)
+	st.EnsureBefore(0) // start operation right away
+	st.Unlock()
+
+	// Wait for a limited amount of time for them to stop.
+	select {
+	case <-chg.Ready():
+		logger.Debugf("All services stopped.")
+	case <-time.After(stopRunningTimeout):
+		return errors.New("timeout stopping running services")
+	}
 	return nil
 }
 
--- a/internal/overlord/servstate/manager.go
+++ b/internal/overlord/servstate/manager.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/canonical/pebble/internal/logger"
 	"github.com/canonical/pebble/internal/overlord/restart"
 	"github.com/canonical/pebble/internal/overlord/state"
 	"github.com/canonical/pebble/internal/plan"
@@ -327,7 +328,7 @@ func (m *ServiceManager) StartOrder(services []string) ([]string, error) {
 }
 
 // StopOrder returns the provided services, together with any dependants,
-// in the proper order for starting them all up.
+// in the proper order for stopping them all.
 func (m *ServiceManager) StopOrder(services []string) ([]string, error) {
 	releasePlan, err := m.acquirePlan()
 	if err != nil {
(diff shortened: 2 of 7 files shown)


internal/overlord/servstate/manager.go
```diff
@@ -462,3 +463,60 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+	logger.Noticef("Stopping all running services.")
+	st := m.state
+	st.Lock()
+	taskSet, err := Stop(st, services)
+	if err != nil {
+		st.Unlock()
+		return err
+	}
+	chg := st.NewChange("shutdown", "Shut down service manager")
+	chg.AddAll(taskSet)
```
hpidcock, line 491: This looks good. Not sure on the change name/description, but the rest looks decent.

benhoyt: If you have better suggestions, let me know. This seemed direct. I also used the active voice "Shut down ..." rather than "Shutting down ..." in the summary because that's what existing Pebble and snapd changes do ("Start ...", "Download ...", and so on).

internal/overlord/servstate/manager_test.go
```diff
@@ -70,7 +70,7 @@ var planLayer1 = `
 services:
     test1:
         override: replace
-        command: /bin/sh -c "echo test1 | tee -a %s; sleep 300"
+        command: /bin/sh -c "echo test1 | tee -a %s; sleep 10"
```
benhoyt, line 73: Drive-by fix to reduce the unnecessarily sleep time in these tests. None of them are alive for more than a few hundred milliseconds, so 10s is ample.

rwcarlsen (commented):
Looks pretty good as far as my inexperienced pebble eyes can see.

internal/overlord/servstate/manager.go
```diff
@@ -477,3 +477,74 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+	st.EnsureBefore(0)
+	st.Unlock()
+
+	// Wait for a limited amount of time for them to stop.
+	select {
+	case <-chg.Ready():
+		logger.Debugf("All services stopped.")
+	case <-time.After(shutdownTimeout):
+		return fmt.Errorf("timed out after %v", shutdownTimeout)
+	}
```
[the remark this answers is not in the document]
benhoyt: Thanks. Yeah, the convention is the caller would add that context, so for example:

```go
err := Shutdown()
if err != nil {
    return fmt.Errorf("cannot shut down service manager: %w", err)
}
```

In actual fact for this case we log the shutdown error and add context in the log mesage. Here's the actual calling code:

```go
	err := d.overlord.ServiceManager().Shutdown()
	if err != nil {
		// This isn't fatal for exiting the daemon, so log and continue.
		logger.Noticef("Cannot shut down service manager: %v", err)
	}
```

So it would print something like:

```
2022-04-28T22:27:32.261Z [pebble] Cannot shut down service manager: timed out after 5.1s
```

internal/overlord/servstate/manager.go
```diff
@@ -477,3 +477,74 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+	services, err := m.servicesToStop()
+	if err != nil {
+		return err
+	}
+	if len(services) == 0 {
+		logger.Debugf("No services to stop.")
+		return nil
+	}
+
+	// One change to stop them all.
```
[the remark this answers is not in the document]
benhoyt: Movie?! _The Lord of the Rings_ is a _book_! :-)

internal/overlord/servstate/manager.go
```diff
@@ -477,3 +477,74 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+	defer releasePlan()
+
+	// Get all service names in plan.
+	var services []string
+	for name := range m.plan.Services {
+		services = append(services, name)
+	}
+
+	// Order according to dependency order.
+	stop, err := m.plan.StopOrder(services)
```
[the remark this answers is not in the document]
benhoyt: In this case `m.plan.Services` is a `map[string]*Service` and we need a slice of the service names / keys. Looking forward to https://pkg.go.dev/golang.org/x/exp/maps#Keys (maybe?) ending up in the stdlib.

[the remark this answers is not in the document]
rwcarlsen: :man_facepalming:

internal/overlord/servstate/manager.go
```diff
@@ -477,3 +477,74 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+	// One change to stop them all.
+	logger.Noticef("Stopping all running services.")
+	st := m.state
+	st.Lock()
+	taskSet, err := Stop(st, services)
+	if err != nil {
+		st.Unlock()
+		return err
+	}
+	chg := st.NewChange("shutdown", "Shut down service manager")
```
niemeyer, line 503: The service manager is an implementation detail which is supposed to be invisible to the user. To the user, what's going on here is that all services are going down, so maybe "Stop all services"?

internal/overlord/servstate/manager.go
niemeyer: Yeah, it's always on the imperative.

internal/overlord/servstate/manager.go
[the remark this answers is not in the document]
niemeyer: There's still a point about this being a complex operation, though, and the call site has no clue of what actually timed out. So maybe:

`"timeout waiting for change to complete"`

... would give some helpful hint.

internal/overlord/servstate/manager.go
```diff
@@ -477,3 +477,74 @@ func (m *ServiceManager) CheckFailed(name string) {
 		}
 	}
 }
+
+var shutdownTimeout = killWait + 100*time.Millisecond
+
+// Shutdown shuts down the service manager, stopping all running services.
+func (m *ServiceManager) Shutdown() error {
```
niemeyer, line 484: There seems to be some encapsulation puncturing here which look like a quick shortcut, and could perhaps be avoided. 

We generally have these concepts well organized:

- Manager: manages the underlying business logic of whatever is being taken care of
- Requests: create the task sets to perform specific roles 
- Handlers: take care of task execution itself, triggered by the state machinery

Meanwhile, the daemon, or anywhere else outside, creates the task sets, and injects into changes as desired for its purposes. Multiple task sets can be mixed and matched as appropriate for the task at hand, and the overlord machinery just makes sure it takes place. This pattern should be visible across the overlord and the entire API.

So what we have here is a bit unusual: we have a method in the manager doing the entire dance by itself, even blocking the entire overlord while its tasks execute. That doesn't feel great. Can we avoid it by doing it in a more traditional organization?

niemeyer (changes requested):
General idea looks sane, thanks!  Only point of concern is the atypical organization.

internal/daemon/daemon.go
```diff
@@ -551,6 +551,16 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 		return fmt.Errorf("internal error: no Overlord")
 	}
 
+	// Shut down the service manager, stopping all services. Must do this
```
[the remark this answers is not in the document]
benhoyt: I'll review to ensure I'm consistent in the rest of the PR, but note that they're two different things: "shut down" is the verb (or verb phrase) form and "shutdown" is the noun or adjective form.

For example, you'd say "Shut down the generator" (verb) but "Shutdown is complete" (noun) and "Shutdown sequence initiated" (adjective). Similar with "log in" (verb) and "login" (noun/adjective). See https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/s/shut-down-shutdown and https://grammarist.com/spelling/log-in-login/.

internal/overlord/servstate/manager.go
benhoyt: I'll take a look at reorganizing this -- thanks for the feedback.

internal/daemon/daemon.go
```diff
@@ -551,6 +551,16 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
...
+	// before overlord.Stop, as it creates a change and waits for the change,
+	// and overlord.Stop calls StateEngine.Stop, which locks, so Ensure would
+	// result in a deadlock.
+	err := d.overlord.ServiceManager().Shutdown()
```
benhoyt, line 558: To be consistent with the above note about "shut down" versus "shutdown", I guess this should be called `ShutDown`, because function names are usually verbs. I'll see where the refactoring ends up and rename if necessary.

internal/overlord/servstate/manager.go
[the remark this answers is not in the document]
benhoyt: Changed (in the refactoring) to "timeout stopping running services".

internal/overlord/servstate/manager.go
benhoyt: Changed to "Stop all running services".

Also changed the change kind to just "stop" -- it's part of the daemon shutdown sequence, but with the refactoring this change is really just stopping services like an ordinary "stop" change.

internal/overlord/servstate/manager.go
benhoyt: I've refactored this so the `servstate.StopRunning` function (in request.go) is the thing that creates the task set, and the caller (in the daemon code) calls that and adds it to a task as part of the shutdown sequence. Yeah, it's more consistent this way, and the diff got slightly shorter (mainly due to slightly shorter tests).

internal/daemon/daemon.go
benhoyt: Yeah, this was no longer relevant with the refactoring. It's `servstate.StopRunning` now.

niemeyer (approved):

Issue #104, opened 2022-02-11
Add a way to introspect failing/restarting services


benhoyt:
Per discussion at https://github.com/canonical/pebble/pull/86#discussion_r795964837, it wasn't obvious when `ServiceInfo.Restarts` was reset to 0 (if at all), and whether this was the right thing to expose. Knowing whether Pebble has restarted 1000 times vs 1001 isn't particularly useful, for example.

On the other hand, when debugging a f(l)ailing service, we want the user to be able to see how often it's restarting. We kind of want a differential, like "restarts per minute" or "restarts in last 10 minutes". But that might be just as hard to explain, so maybe the absolute number is simplest and the user can see how fast it's going up over by querying over time, or by knowing when Pebble was started.

I lean towards keeping this field -- it seems quite useful to me -- and having it never reset (until Pebble restarts). It would also be exposed in `pebble services`.

benhoyt:
Possible options:

1) Add `Restarts` and `RestartsSpan`, giving the number of restarts in the last N minutes (e.g., 6 restarts in the last 1 minute). Implementation: a sliding window of restart times, keeping only those in the last N minutes.
2) Add a `LastStart` timestamp or `Uptime` field. Trivial implementation: just store `time.Now()` whenever a service is started.
3) Add a Pebble "change" object whenever we restart the service (viewable with `pebble changes`). Restarting a service is a fairly invasive operation, so perhaps this isn't too heavy. Implementation: a new `backoff-restart` change type with a single `backoff-restart` task. In `servstate/handlers.go`, `backoffTimeElapsed` would create the `backoff-restart` change and immediately start the task, instead of calling `startInternal` directly.

benhoyt:
Notes from meeting just now:

* Have a change that's "recovering this service" when it enters the backoff loop, then one task per restart? 
* But the last task would stay open (in Doing state), meaning "I'm waiting and monitoring the service". Until the backoff is reset after backoff-limit time and you go Done.
* Alternatively: a single task with a number of restarts and a log? Maybe try them both and see what looks/works better.
* `StartTime` on the services API would be useful too.

benhoyt:
Per comment on https://github.com/canonical/pebble/pull/117#issuecomment-1128364829, we're planning to do this with events (a new concept) instead, and have started spec [JU048](https://docs.google.com/document/d/16PJ85fefalQd7JbWSxkRWn0Ye-Hs8S1yE99eW7pk8fA/edit#) to design that.

[...]

PR #115, opened 2022-04-28
Add "current-since" field to services API and CLI

benhoyt:
This PR adds a `current-since` field to the API, meaning the time at which the "current" field last changed, for example started (current changed from `inactive` to `active`), and so on. It's related to https://github.com/canonical/pebble/issues/104 (though in a previous iteration it was "start time", not "curren since").

Here's how it looks like the CLI:

```
$ pebble services
Service  Startup  Current  Since
test2    enabled  active   today at 17:08 NZST

$ pebble services --abs-time
Service  Startup  Current  Since
test2    enabled  active   2022-04-28T17:08:45+12:00
```

This PR also removes the "restarts" field which isn't being used and was internal-only. Per #104, we're going to use changes/tasks or events instead.

### DIFF
--- a/internal/daemon/api_services.go
+++ b/internal/daemon/api_services.go
@@ -28,10 +28,10 @@ import (
 )
 
 type serviceInfo struct {
-	Name      string     `json:"name"`
-	Startup   string     `json:"startup"`
-	Current   string     `json:"current"`
-	StartTime *time.Time `json:"start-time,omitempty"` // pointer as omitempty doesn't work with time.Time directly
+	Name         string     `json:"name"`
+	Startup      string     `json:"startup"`
+	Current      string     `json:"current"`
+	CurrentSince *time.Time `json:"current-since,omitempty"` // pointer as omitempty doesn't work with time.Time directly
 }
 
 func v1GetServices(c *Command, r *http.Request, _ *userState) Response {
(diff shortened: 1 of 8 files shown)


hpidcock (approved):

benhoyt:
Per discussion at London mini-sprint, we want to make the API field `current-since` and display it in the CLI as "Since" (a `-` if Since is not set). The `current-since` value would be updated whenever `current` changed. CLI output example:

```
$ pebble services
Service  Startup  Current  Since
test2    enabled  active   today at 17:08 NZST
```

internal/daemon/api_services.go
```diff
@@ -20,16 +20,18 @@ import (
...
 )
 
 type serviceInfo struct {
-	Name    string `json:"name"`
-	Startup string `json:"startup"`
-	Current string `json:"current"`
+	Name         string     `json:"name"`
+	Startup      string     `json:"startup"`
+	Current      string     `json:"current"`
+	CurrentSince *time.Time `json:"current-since,omitempty"` // pointer as omitempty doesn't work with time.Time directly
```
niemeyer, line 34: That's awkward, considering times have IsZero which fits properly with the notion of omitempty.

This is an internal detail, so I'm not blocking on it, but it's curious that the standard json package is misbehaving there.

internal/overlord/servstate/handlers.go
```diff
@@ -282,6 +282,13 @@ func (s *serviceData) transition(state serviceState) {
 
 // transitionRestarting changes the service's state and also sets the restarting flag.
 func (s *serviceData) transitionRestarting(state serviceState, restarting bool) {
+	// Update current-since time if derived status is changing.
+	oldStatus := stateToStatus(s.state)
+	newStatus := stateToStatus(state)
+	if oldStatus != newStatus {
+		s.currentSince = time.Now()
```
niemeyer, line 289: Nice, very clean logic.

niemeyer (approved):
Very clean PR, thanks Ben.

PR #116, opened 2022-04-29
Add missing state transition to state diagram

benhoyt:
Somehow I'd missed this when I added the state diagram originally.


benhoyt:
This is just a fix to the internal developer documentation, so I'm going to merge it directly.

PR #117, opened 2022-05-02
Record "recover-service" change when recovering/restarting a service

### LINKED ISSUE #104: Add a way to introspect failing/restarting services

benhoyt:
Record a "recover" change when recovering a service, and record one task per restart within that.

This will enable debugging/introspection on whether a service is failing and being auto-restarted. You can introspect it with the changes API or the "pebble changes" and "pebble tasks" CLI commands.

The kind (type) of the changes is "recover" and the kind of the tasks is "restart".

Example output while I service is begin recovered (change 3 is not yet ready):

```
$ pebble changes
1    Done    today at 16:08 NZST  today at 16:08 NZST  Autostart service "test2"
2    Done    today at 16:08 NZST  today at 16:09 NZST  Recover service "test2"
3    Done    today at 16:09 NZST  -                    Recover service "test2"
```

After it's recovered the change is marked done/Ready:

```
$ pebble changes
1    Done    today at 16:08 NZST  today at 16:08 NZST  Autostart service "test2"
2    Done    today at 16:08 NZST  today at 16:09 NZST  Recover service "test2"
3    Done    today at 16:09 NZST  today at 16:10 NZST  Recover service "test2"

$ pebble tasks 2
Done    today at 16:08 NZST  today at 16:08 NZST  Restart service "test2"
Done    today at 16:08 NZST  today at 16:08 NZST  Restart service "test2"
Done    today at 16:09 NZST  today at 16:09 NZST  Restart service "test2"

$ pebble tasks 3
Done    today at 16:09 NZST  -                    Restart service "test2"
```

Fixes #104


benhoyt:
Per discussion with @hpidcock (who had discussed with @niemeyer at the product sprint), we're leaning towards reserving changes and tasks only for user-initiated actions, and we're likely to add an "events" system and use events for this instead. Moving back to draft.

benhoyt:
Per above comment, we're likely to use events for this instead. Closing for now (can always reopen later if necessary).

PR #119, opened 2022-05-09
Record "recover-check" change and task logs when health check is failing

### LINKED ISSUE #103: Add a way to detect and introspect check failures
Per thread at https://github.com/canonical/pebble/pull/86#discussion_r795942968, I previously had `LastError` and `ErrorDetails` fields on the `CheckInfo` API struct. However, this was a bit ad-hoc and didn't fit our existing constructs like task logs. I've pulled those out of that PR for now, but we should design a better approach.

It would seem quite heavy to just use Pebble "changes" and "tasks" for checks. If we did that, I think each check would have to be a change (with one task?) and that would flood the changes list. Besides, these aren't really state changes, they're checks. ("Change represents a tracked modification to the system state.")

So instead I suggest (idea from @niemeyer) using the same "log" concept that a task has, which is a `Log []string` field, and adding that to the `CheckInfo` object. We could keep the last 10 like a `Task` does. Each log string would be a timestamp-formatted string with a prefix like `ERROR` and a message, like so:

```
2022-02-22T11:40:53+13:00 ERROR timed out after 100ms: context deadline exceeded
2022-02-22T11:50:53+13:00 ERROR timed out after 100ms: context deadline exceeded
```

This field would be in the `CheckInfo` struct for a particular named check, so no further context is needed. The log would contain check failure errors and the details that was previously in `ErrorDetails` (truncate HTTP body for `http` checks, last few line of command output for `exec` checks).

The `/v1/checks` API would return this information as a list of strings. The CLI `pebble checks` could return the (possibly truncated) first log line by default, and the full log if you specify `pebble checks --verbose`. Something like this:

```
$ pebble checks
Check  Level  Status  Failures  Log
chk1   -      up      0/3
chk2   alive  down    2/2       2022-02-22T11:40:53+13:00 ERROR timed out after 100ms...

$ pebble checks --verbose
Check  Level  Status  Failures
chk1   -      up      0/3
chk2   alive  down    2/2
    2022-02-22T11:40:53+13:00 ERROR timed out after 100ms: context deadline exceeded 
    2022-02-22T11:50:53+13:00 ERROR timed out after 100ms: context deadline exceeded 
```

It's an open question whether the log would be cleared/reset when the check became healthy again. Probably not, as it's a log/history?

benhoyt:
Also record a task for every failure until the threshold, and close
the change (set it to Done) when the health check goes successful
again. The error message and any details is included in the task log.

This is similar to the "recover-service" change in https://github.com/canonical/pebble/pull/117,
but for health checks.

Fixes #103.

### DIFF
--- a/internal/overlord/checkstate/manager.go
+++ b/internal/overlord/checkstate/manager.go
@@ -16,11 +16,13 @@ package checkstate
 
 import (
 	"context"
+	"fmt"
 	"sort"
 	"sync"
 	"time"
 
 	"github.com/canonical/pebble/internal/logger"
+	"github.com/canonical/pebble/internal/overlord/state"
 	"github.com/canonical/pebble/internal/plan"
 )
 
@@ -29,14 +31,15 @@ type CheckManager struct {
 	mutex           sync.Mutex
 	checks          map[string]*checkData
 	failureHandlers []FailureFunc
+	state           *state.State
 }
 
 // FailureFunc is the type of function called when a failure action is triggered.
 type FailureFunc func(name string)
 
 // NewManager creates a new check manager.
-func NewManager() *CheckManager {
-	return &CheckManager{}
+func NewManager(st *state.State) *CheckManager {
+	return &CheckManager{state: st}
 }
 
 // NotifyCheckFailed adds f to the list of functions that are called whenever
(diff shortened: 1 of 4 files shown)


benhoyt:
Per discussion with @hpidcock (who had discussed with @niemeyer at the product sprint), we're leaning towards reserving changes and tasks only for user-initiated actions, and we're likely to add an "events" system and use events for this instead. Moving back to draft.

benhoyt:
Per above comment, we're likely to use events for this instead. Closing for now (can always reopen later if necessary).

PR #120, opened 2022-06-17
Add configuration file validation

flotter:
Kerncraft (and other craft tools) needs to validate supplied or
generated Pebble configuration snippets. Instead of duplicating
schema validation in the craft tool, the safer and more
maintainable approach will be to ask the selected Pebble
version to validate the given set of configuration layers.

Add a configuration validation option to Pebble.

On failure Pebble will output existing configuration schema error
messages and return a non-zero exit code.

For example:

```
0 $ go build ./cmd/pebble
0 $ PEBBLE=~/pebble ./pebble validate
error: cannot parse layer "layer": yaml: unmarshal errors:
  line 1: field servicez not found in type plan.Layer
1 $ PEBBLE=~/pebble ./pebble validate
0 $ 
```

### DIFF
--- a/cmd/pebble/cmd_validate.go
+++ b/cmd/pebble/cmd_validate.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2014-2020 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/canonical/pebble/internal/plan"
+	"github.com/jessevdk/go-flags"
+)
+
+var shortValidateHelp = "Validate daemon configration"
+var longValidateHelp = `
+Perform validation of daemon configuration files and exit.
+`
+
+type cmdValidate struct {
+	clientMixin
+}
+
+func init() {
+	addCommand("validate", shortValidateHelp, longValidateHelp, func() flags.Commander { return &cmdValidate{} }, nil, nil)
+}
+
+func (cmd cmdValidate) Execute(args []string) error {
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	pebbleDir, _ := getEnvPaths()
+
+	_, err := plan.ReadDir(pebbleDir)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "configuration validation failed: %v\n", err)
+		panic(&exitStatus{1})
+	}
+
+	return nil
+}
--- a/cmd/pebble/cmd_validate.go
+++ b/cmd/pebble/cmd_validate.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2020 Canonical Ltd
+// Copyright (c) 2022 Canonical Ltd
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License version 3 as
@@ -15,16 +15,15 @@
 package main
 
 import (
-	"fmt"
-	"os"
+	"github.com/jessevdk/go-flags"
 
 	"github.com/canonical/pebble/internal/plan"
-	"github.com/jessevdk/go-flags"
 )
 
-var shortValidateHelp = "Validate daemon configration"
+var shortValidateHelp = "Validate daemon configuration"
 var longValidateHelp = `
-Perform validation of daemon configuration files and exit.
+Validate the Pebble configuration layers in the $PEBBLE/layers directory,
+exiting with an error message and non-zero exit code on failure.
 `
 
 type cmdValidate struct {
@@ -43,10 +42,5 @@ func (cmd cmdValidate) Execute(args []string) error {
 	pebbleDir, _ := getEnvPaths()
 
 	_, err := plan.ReadDir(pebbleDir)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "configuration validation failed: %v\n", err)
-		panic(&exitStatus{1})
-	}
-
-	return nil
+	return err
 }
--- a/cmd/pebble/cmd_validate_test.go
+++ b/cmd/pebble/cmd_validate_test.go
@@ -1,0 +1,26 @@
+// Copyright (c) 2014-2020 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main_test
+
+import (
+	. "gopkg.in/check.v1"
+
+	pebble "github.com/canonical/pebble/cmd/pebble"
+)
+
+func (s *PebbleSuite) TestValidateCommand(c *C) {
+	_, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"validate"})
+	c.Assert(err, IsNil)
+}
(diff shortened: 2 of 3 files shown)


flotter:
The actual choice of option name, i.e. 'validate' is not yet discussed by Gustavo. We spoke about dry-run, but analysing the code it felt like a cleaner approach not to start the Daemon at all, and only validate the set of configuration files.

cmd/pebble/cmd_validate.go
```diff
@@ -0,0 +1,52 @@
...
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	pebbleDir, _ := getEnvPaths()
+
+	_, err := plan.ReadDir(pebbleDir)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "configuration validation failed: %v\n", err)
+		panic(&exitStatus{1})
```
benhoyt, line 48: `main()` already prints the error to stderr and exits with exit code 1 if this returns an error, so there's no need to panic or use `exitStatus` here. You could just `return fmt.Errorf("configuration validation failed: %w", err)`.

Actually, I'd argue that we don't even need to wrap / add a message to the error here, as the point of this command is validating, so I don't think we don't need to add that. Could just `return err`.

cmd/pebble/cmd_validate.go
```diff
@@ -0,0 +1,52 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/canonical/pebble/internal/plan"
+	"github.com/jessevdk/go-flags"
```
benhoyt, line 22: Nit: we usually try to order imports in three groups, separated by a blank line: stdlib, 3rd party, then canonical/pebble.

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	. "gopkg.in/check.v1"
+
+	pebble "github.com/canonical/pebble/cmd/pebble"
+)
+
+func (s *PebbleSuite) TestValidateCommand(c *C) {
+	_, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"validate"})
+	c.Assert(err, IsNil)
```
benhoyt, line 25: It looks like we're only testing the happy path here. Can we add a test of validation failure?

benhoyt (commented):
Nice, I like this idea. Added a few small comments.

cmd/pebble/cmd_validate.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2014-2020 Canonical Ltd
```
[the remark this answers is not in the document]
flotter: Fixed.

cmd/pebble/cmd_validate.go
flotter: Makes sense. Fixed.

cmd/pebble/cmd_validate.go
flotter: Fixed.

cmd/pebble/cmd_validate_test.go
flotter: Done.

cmd/pebble/cmd_validate.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/canonical/pebble/internal/plan"
+	"github.com/jessevdk/go-flags"
+)
+
+var shortValidateHelp = "Validate daemon configration"
```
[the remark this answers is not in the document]
flotter: Fixed.

cmd/pebble/cmd_validate.go
```diff
@@ -0,0 +1,52 @@
...
+var longValidateHelp = `
+Perform validation of daemon configuration files and exit.
```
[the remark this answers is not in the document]
flotter: Fixed type. Ben I think it would be very helpful for me if we could have a quick brainstorm on your second proposal. I think it makes a lot of sense, so much so that I wonder if we should not add the option "globally" so the $PEBBLE directory can be specified from command-line. But there seems to be a bit of a startup-race between needing the socket path and resolving the command-line options.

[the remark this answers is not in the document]
flotter: Agreed we would split this out and see if an option such as "--pebble-dir" could be added globally to also allow the daemon to get a custom pebble dir from commandline.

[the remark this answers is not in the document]
benhoyt: @flotter and I discussed this on a call, and agreed that we should leave this PR as is (using the env var) but consider adding a `--pebble-dir` (and possibly `--pebble-socket`) command line argument globally, but do that in a separate PR.

benhoyt (approved):
Looks good, thanks for the updates. Let's get @niemeyer's review too.

atesburak (approved):
LGTM

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	. "gopkg.in/check.v1"
+
+	pebble "github.com/canonical/pebble/cmd/pebble"
+)
+
+func (s *PebbleSuite) TestValidateCommandYamlOK(c *C) {
```
benhoyt, line 27: I think name is overly verbose: it doesn't need "command" in it because we're in the `cmd/pebble` directory so all these tests are about commands. Also `Yaml` is unnecessary (and would be capitalized as `YAML` in Go style). So I'd go with `TestValidateOK`.

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,59 @@
...
+	var err error
```
benhoyt, line 28: No need for `var`; this should a short variable declaration `err := ` below. Same issue in the other test.

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	layersDir := filepath.Join(s.pebbleDir, "layers")
+	err = os.Mkdir(layersDir, 0755)
+	c.Assert(err, IsNil)
+
+	fpath := filepath.Join(layersDir, "001-kerncraft.yaml")
```
benhoyt, line 34: Naming nit: IMO `path` is enough, or `layerPath` (the `f` prefix doesn't really say much). Same in the other test.

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,59 @@
...
+	c.Assert(err, IsNil)
+
+	_, err = pebble.Parser(pebble.Client()).ParseArgs([]string{"validate"})
+	c.Assert(err, IsNil)
+}
+
+// TestValidateCommandYamlInvalidSchema does not attempt to test the plan validation
+// logic. The purpose of this test is to ensure the plumbing is in place to report
+// a YAML schema violation using the 'validate' command.
+func (s *PebbleSuite) TestValidateCommandYamlInvalidSchema(c *C) {
```
benhoyt, line 45: Same comment about long name. Could be just `TestValidateError`.

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,59 @@
...
+	err = ioutil.WriteFile(fpath, []byte("summary: Kerncraft Layer"), 0644)
+	c.Assert(err, IsNil)
+
+	_, err = pebble.Parser(pebble.Client()).ParseArgs([]string{"validate"})
+	c.Assert(err, IsNil)
+}
+
+// TestValidateCommandYamlInvalidSchema does not attempt to test the plan validation
+// logic. The purpose of this test is to ensure the plumbing is in place to report
+// a YAML schema violation using the 'validate' command.
```
benhoyt, line 44: I don't think this comment is necessary. Most of the command tests are like that (including the OK case above).

cmd/pebble/cmd_validate_test.go
```diff
@@ -0,0 +1,59 @@
...
+	err = os.Mkdir(layersDir, 0755)
+	c.Assert(err, IsNil)
+
+	fpath := filepath.Join(layersDir, "001-kerncraft.yaml")
+	err = ioutil.WriteFile(fpath, []byte("randomkey: Kerncraft Layer"), 0644)
+	c.Assert(err, IsNil)
+
+	_, err = pebble.Parser(pebble.Client()).ParseArgs([]string{"validate"})
+	c.Assert(err, ErrorMatches, "cannot parse layer \"kerncraft\": yaml: unmarshal errors:\n"+
+		"  line 1: field randomkey not found in type plan.Layer")
```
benhoyt, line 58: This would be slightly neater (and not require escaped quotes) with backtick raw strings.

benhoyt (commented):
Added a few more nit comments on the tests.

benhoyt:
@flotter Are you still hoping to have this merged? If so I can bring it up in our next Juju/Pebble spec meeting.

Issue #126, opened 2022-07-21
When starting a service with non-root user Pebble environment variables used own to root


gcalvinos:
We are creating a Pebble charm that creates a service with user and group: "appuser". That user exists in the image.

```
layer_config = {
            "summary": "lcm layer",
            "description": "pebble config layer for nbi",
            "services": {
                self.service_name: {
                    "override": "replace",
                    "summary": "lcm service",
                    "command": "python3 -m osm_lcm.lcm",
                    "startup": "enabled",
                    "user": "appuser",
                    "group": "appuser",
                    "environment": environments,
                }
            },
        }
```
The process in the container runs as "appuser", so that's fine:
```
root@lcm-0:/app/osm_lcm# ps -fe
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 14:15 ?        00:00:01 /charm/bin/pebble run --create-dirs --hold --http :38813 --verbose
appuser       16       1  1 14:15 ?        00:00:21 python3 -m osm_lcm.lcm
```
But in the code we execute we use the `$HOME` variable, and instead of getting the home from "appuser" we get the home from "root", and we get a "Permission error" for that reason:

`2022-07-21T14:17:25.355Z [lcm] PermissionError: [Errno 13] Permission denied: '/root/.ssh'`

Is that the expected behaviour?

benhoyt:
Thanks for the report!

You're right, we should be setting `USER` and `HOME` when user/group is set in the service config. Incidentally, we do this for one-shot commands but not for services. I can repro this locally fairly easily, and I have some temporary code that fixes it (I'll add tests and then push up a proper PR):

```diff
diff --git a/internal/overlord/servstate/handlers.go b/internal/overlord/servstate/handlers.go
index bbf8170..a28480b 100644
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -5,6 +5,8 @@ import (
        "io"
        "os"
        "os/exec"
+       "os/user"
+       "strconv"
        "syscall"
        "time"
 
@@ -325,6 +327,12 @@ func (s *serviceData) startInternal() error {
        s.cmd = exec.Command(args[0], args[1:]...)
        s.cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 
+       // Copy environment to avoid update original.
+       environment := make(map[string]string)
+       for k, v := range s.config.Environment {
+               environment[k] = v
+       }
+
[...]

PR #128, opened 2022-07-24
Actually run the user/group tests (as root) so we're testing that code

benhoyt:
These tests were somewhat tricky to run locally, and we forget, so let's at least run them automatically in GitHub actions. I'll be using this same GitHub Actions check to test https://github.com/canonical/pebble/issues/126 as well.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,10 +45,12 @@ jobs:
     - name: Test
       run: |
         echo user=$USER group=$GROUP
+        cat /etc/passwd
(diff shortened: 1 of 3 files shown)


hpidcock (approved):
Seems an ok way to run the tests for now, but if there is more of these tests, might need a better way to run them.

benhoyt:
Yeah, definitely not an elegant way of running these tests. Not sure of a hugely better way off the top of my head. Seeing this is just test-running script, we can always iterate later. Merging for now.

PR #130, opened 2022-07-28
Make write/push and make-dir chown intermediate dirs when setting user

### LINKED ISSUE #124: Container.push() won't use passed user:group to create directories.
When calling `container.push(make_dirs=True, user=user, group=group)`, the directories are not created using the `user:group`. This can lead to the file being unreachable when Pebble is running a workload with specific `user:group`.

I created a custom path for a file on `/mydir`, pushing a file with:
```
container.push(path, rendered, permissions=0o660, make_dirs=True, user="redis", group="redis")
```
After container is running, I can inspect the permissions of the directory:
```
root@redis-k8s-0:/mydir# ls -la                         
total 12                                                
drwxr-xr-x 2 root  root  4096 Jul 15 06:56 .            
drwxr-xr-x 1 root  root  4096 Jul 15 06:56 ..           
-rw-r----- 1 redis redis  539 Jul 15 06:56 sentinel.conf
``` 
Only the file has the correct `user:group`

benhoyt:
Per user report, write/push and make-dir are not setting the user/group
on intermediate directories when the user and group is set. This can
lead to the file being unreachable when Pebble is running a workload
with a specific user:group.

Fix these API calls to chown the intermediate directories as needed.
Use the existing osutil.MkdirAllChown where appropriate.

Add tests for the above, including integration tests that require root.

Fixes #124

### DIFF
--- a/internal/daemon/api_files.go
+++ b/internal/daemon/api_files.go
@@ -470,9 +470,14 @@ func writeFile(item writeFilesItem, source io.Reader) error {
 		return nonAbsolutePathError(item.Path)
 	}
 
-	// Create parent directory if needed
+	uid, gid, err := normalizeUidGid(item.UserID, item.GroupID, item.User, item.Group)
+	if err != nil {
+		return fmt.Errorf("cannot look up user and group: %w", err)
+	}
+
+	// Create parent directory if needed.
 	if item.MakeDirs {
-		err := os.MkdirAll(pathpkg.Dir(item.Path), 0o755)
+		err := mkdirAllUserGroup(pathpkg.Dir(item.Path), 0o755, uid, gid)
 		if err != nil {
 			return fmt.Errorf("cannot create directory: %w", err)
 		}
@@ -483,16 +488,19 @@ func writeFile(item writeFilesItem, source io.Reader) error {
 	if err != nil {
 		return err
 	}
-	uid, gid, err := normalizeUidGid(item.UserID, item.GroupID, item.User, item.Group)
-	if err != nil {
-		return fmt.Errorf("cannot look up user and group: %w", err)
-	}
 	sysUid, sysGid := sys.UserID(osutil.NoChown), sys.GroupID(osutil.NoChown)
 	if uid != nil && gid != nil {
 		sysUid, sysGid = sys.UserID(*uid), sys.GroupID(*gid)
 	}
-
 	return atomicWriteChown(item.Path, source, perm, osutil.AtomicWriteChmod, sysUid, sysGid)
+}
+
+func mkdirAllUserGroup(path string, perm os.FileMode, uid, gid *int) error {
+	if uid != nil && gid != nil {
+		return mkdirAllChown(path, perm, sys.UserID(*uid), sys.GroupID(*gid))
+	} else {
+		return os.MkdirAll(path, perm)
+	}
 }
 
 func parsePermissions(permissions string, defaultMode os.FileMode) (os.FileMode, error) {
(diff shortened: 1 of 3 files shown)


internal/daemon/api_files.go
```diff
@@ -538,22 +546,25 @@ func makeDir(dir makeDirsItem) error {
...
 	}
-	if uid != nil && gid != nil {
-		err = chown(dir.Path, *uid, *gid)
+	if dir.MakeParents {
+		err := mkdirAllUserGroup(dir.Path, perm, uid, gid)
+		if err != nil {
+			return err
+		}
+	} else {
+		err := os.Mkdir(dir.Path, perm)
```
hpidcock, line 559: This should be implemented in the osutil package so that it can use the same mutex as the MkdirAllUserGroup and also employ the same Mkdir with temp name, Chown and Rename to final name logic.

hpidcock (approved):

internal/daemon/api_files.go
benhoyt: Done, thanks

internal/daemon/api_files.go
```diff
@@ -483,18 +488,29 @@ func writeFile(item writeFilesItem, source io.Reader) error {
...
 	}
-
 	return atomicWriteChown(item.Path, source, perm, osutil.AtomicWriteChmod, sysUid, sysGid)
 }
 
+func mkdirAllUserGroup(path string, perm os.FileMode, uid, gid *int) error {
+	if uid != nil && gid != nil {
+		return mkdirAllChown(path, perm, sys.UserID(*uid), sys.GroupID(*gid))
+	} else {
+		return os.MkdirAll(path, perm)
```
hpidcock, line 502: just call mkdirAllChown with osutil.NoChown that way the directory create is always within the mutex

internal/daemon/api_files.go
```diff
@@ -483,18 +488,29 @@ func writeFile(item writeFilesItem, source io.Reader) error {
...
+	}
+}
+
+func mkdirUserGroup(path string, perm os.FileMode, uid, gid *int) error {
+	if uid != nil && gid != nil {
+		return mkdirChown(path, perm, sys.UserID(*uid), sys.GroupID(*gid))
+	} else {
+		return os.Mkdir(path, perm)
```
hpidcock, line 510: just call mkdirChown with osutil.NoChown that way the directory create is always within the mutex

internal/osutil/mkdirallchown.go
```diff
@@ -56,25 +59,39 @@ func mkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupI
...
+func mkdirChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
 	cand := path + ".mkdir-new"
 
 	if err := os.Mkdir(cand, perm); err != nil && !os.IsExist(err) {
 		return err
 	}
 
-	if err := sys.ChownPath(cand, uid, gid); err != nil {
-		return err
+	if uid != NoChown || gid != NoChown {
```
hpidcock, line 72: we don't need this if-statement now

internal/osutil/mkdirallchown.go
```diff
@@ -56,25 +59,39 @@ func mkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupI
...
 	defer fd.Close()
 
 	return fd.Sync()
 }
+
+// MkdirChown is like os.Mkdir but it also calls os.Chown on the directory it
+// creates.
+func MkdirChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
+	mu.Lock()
+	defer mu.Unlock()
```
hpidcock, line 95:
```
if uid == NoChown && gid == NoChown {
   return os.Mkdir(path, perm)
}
```

hpidcock (approved):

internal/osutil/mkdirallchown.go
```diff
@@ -24,6 +24,9 @@ import (
 )
 
 // XXX: we need to come back and fix this; this is a hack to unblock us.
+// Have a lock so that if one goroutine tries to mkdirallchown /foo/bar, and
+// another tries to mkdirallchown /foo/baz, they can't both decide they need
+// to make /foo and then have one fail.
 var mu sync.Mutex
 
 // MkdirAllChown is like os.MkdirAll but it calls os.Chown on any
```
hpidcock, line 32:
```
func MkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
	mu.Lock()
	defer mu.Unlock()
	if uid == NoChown && gid == NoChown {
		return os.Mkdir(path, perm)
	}
	return mkdirAllChown(filepath.Clean(path), perm, uid, gid)
}
```

internal/daemon/api_files.go
benhoyt: Done

internal/daemon/api_files.go
benhoyt: Done

internal/osutil/mkdirallchown.go
benhoyt: Done

internal/osutil/mkdirallchown.go
benhoyt: Done

PR #134, opened 2022-08-11
Add last failure (error and details) to checks API and CLI command

benhoyt:
This is a similar but slightly cleaned up version of what we discussed [here](https://github.com/canonical/pebble/pull/86#discussion_r795942968). We decided against it at the time, but now after having considered using changes/tasks and then [events](https://docs.google.com/document/d/16PJ85fefalQd7JbWSxkRWn0Ye-Hs8S1yE99eW7pk8fA/edit#) for check diagnostics, we've decided that it's probably better to separate out the events/hooks use case from the debugging use case, and we're back to having a simple way to view the check errors.

The response from the API now includes a `last-failure` key as follows (it's only present if `failures` is greater than 0):

```
{
  "name": "chk1",
  "status": "down",
  "failures": 1,
  "threshold": 1,
  "last-failure": {
    "error": "an error message",
    "details": "extended details ..."  // this may or may not be present, depending on the failure
  }
}
```

And the `pebble checks` CLI command is extended with a `Last Failure` field and a `--verbose` option as follows:

```
# In non-verbose mode, the "Last Failure" field is truncated to 60 chars max (with "..." suffix if truncated)
$ pebble checks
Check  Level  Status  Failures  Last Failure
chk1   -      down    1/1       an error message: extended details ...

# In --verbose mode, the first line is the error message and subsequent lines show the details, if any
$ pebble checks --verbose
Check  Level  Status  Failures  Last Failure
chk1   -      down    1/1       an error message:
                                extended details ...
                                may be multiple lines ...
```

### DIFF
--- a/internal/overlord/checkstate/checkers.go
+++ b/internal/overlord/checkstate/checkers.go
@@ -75,7 +75,7 @@ func (c *httpChecker) check(ctx context.Context) error {
 			details = strings.Join(lines, "\n")
 		}
 		return &detailsError{
-			error:   fmt.Errorf("received non-20x status code %d", response.StatusCode),
+			error:   fmt.Errorf("non-20x status code %d", response.StatusCode),
 			details: details,
 		}
 	}
--- a/client/checks.go
+++ b/client/checks.go
@@ -65,6 +65,16 @@ type CheckInfo struct {
 	// Threshold is this check's failure threshold, from the layer
 	// configuration.
 	Threshold int `json:"threshold"`
+
+	// LastFailure describes the last failure that occurred for this check
+	// (only present if the number of failures is nonzero).
+	LastFailure *CheckFailure `json:"last-failure"`
+}
+
+// CheckFailure holds information about a single check failure.
+type CheckFailure struct {
+	Error   string `json:"error"`
+	Details string `json:"details"`
 }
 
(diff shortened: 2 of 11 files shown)


internal/overlord/checkstate/checkers.go
```diff
@@ -75,7 +75,7 @@ func (c *httpChecker) check(ctx context.Context) error {
 			details = strings.Join(lines, "\n")
 		}
 		return &detailsError{
-			error:   fmt.Errorf("received non-20x status code %d", response.StatusCode),
+			error:   fmt.Errorf("non-20x status code %d", response.StatusCode),
```
benhoyt, line 78: I shortened this check error message slightly as "received" seemed redundant.

hpidcock: usually we call them 2xx 3xx 4xx etc when referring to http status codes

benhoyt: Good point. Fixed.

benhoyt:
Per discussion at London mini-sprint, with the new approach of doing checks via changes and tasks properly, we likely don't need this. Though we could show error information pulled from the change/task in `pebble checks`.

benhoyt:
Per above, we're planning to do this a different way via changes+tasks.

PR #135, opened 2022-08-11
Bump test coverage for some commands

anpep:
This is a relatively small patch that improves upon the existing test suites for well-tested Pebble commands to achieve near-complete code coverage. I'm working on adding tests for the rest of commands but as these will be slightly larger patches I will do separate PRs for each of them.

### DIFF
--- a/cmd/pebble/cmd_plan_test.go
+++ b/cmd/pebble/cmd_plan_test.go
@@ -25,16 +25,27 @@ import (
 )
 
 func (s *PebbleSuite) TestGetPlan(c *check.C) {
+	fail := false
+
 	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
 		c.Check(r.Method, check.Equals, "GET")
 		c.Check(r.URL.Path, check.Equals, "/v1/plan")
 		c.Check(r.URL.Query(), check.DeepEquals, url.Values{"format": []string{"yaml"}})
-		fmt.Fprint(w, `{
+
+		if fail {
+			fmt.Fprint(w, `{
+    "type": "error",
+    "result": {"message": "could not do the thing"}
+}`)
+		} else {
+			fmt.Fprint(w, `{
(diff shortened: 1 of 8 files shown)


cmd/pebble/cmd_plan_test.go
```diff
@@ -25,16 +25,27 @@ import (
...
 func (s *PebbleSuite) TestGetPlan(c *check.C) {
+	fail := false
+
 	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
 		c.Check(r.Method, check.Equals, "GET")
 		c.Check(r.URL.Path, check.Equals, "/v1/plan")
 		c.Check(r.URL.Query(), check.DeepEquals, url.Values{"format": []string{"yaml"}})
-		fmt.Fprint(w, `{
+
+		if fail {
```
benhoyt, line 35: This is one of those that I'd prefer to avoid the flag and just add another test (even if it's a few more lines of code). It's simpler to reason about, and avoids the need for `s.ResetStdStreams()`.

cmd/pebble/cmd_version_test.go
```diff
@@ -35,4 +35,8 @@ func (s *PebbleSuite) TestVersionCommand(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(s.Stdout(), Equals, "client  4.56\nserver  7.89\n")
 	c.Assert(s.Stderr(), Equals, "")
+
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"version", "extra", "args"})
```
benhoyt, line 39: Once again, let's make this a separate test, `TestVersionExtraArgs` or something.

benhoyt (commented):
Looks good, thanks for the additional test coverage. Just a couple of minor comments.

benhoyt (approved):
LGTM, thanks!

PR #136, opened 2022-08-11
Add tests for autostart command

anpep:
This patch adds coverage for the `pebble autostart` command.


benhoyt (commented):
Thanks again. Exact same review comments here as per https://github.com/canonical/pebble/pull/137#pullrequestreview-1072137669, so I won't repeat them here.

benhoyt (approved):
LGTM now, thanks!

PR #137, opened 2022-08-12
Add tests for the replan command

anpep:
This patch adds coverage for the `pebble replan` command.

### DIFF
--- a/cmd/pebble/cmd_replan_test.go
+++ b/cmd/pebble/cmd_replan_test.go
@@ -1,0 +1,108 @@
+// Copyright (c) 2022 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main_test
+
+import (
+	"fmt"
+	pebble "github.com/canonical/pebble/cmd/pebble"
+	"gopkg.in/check.v1"
+	"net/http"
+)
+
+func (s *PebbleSuite) TestReplan(c *check.C) {
+	failGetChange := false
+	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
+		fakeChange := `{
+  "id":   "43",
+  "kind": "replan",
+  "summary": "...",
+  "status": "Done",
+  "ready": true,
+  "spawn-time": "2016-04-21T01:02:03Z",
+  "ready-time": "2016-04-21T01:02:04Z",
+  "tasks": []
+}`
+		if r.URL.Path == "/v1/changes" {
+			c.Check(r.Method, check.Equals, "GET")
+			fmt.Fprintf(w, `{"type": "sync", "result": [%s]}`, fakeChange)
+			return
+		} else if r.URL.Path == "/v1/changes/43" {
+			c.Check(r.Method, check.Equals, "GET")
+			if failGetChange {
+				fmt.Fprintf(w, `{"type": "error", "result": {"message": "could not foo"}}`)
+			} else {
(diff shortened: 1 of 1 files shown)


cmd/pebble/cmd_replan_test.go
```diff
@@ -0,0 +1,108 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main_test
+
+import (
+	"fmt"
+	pebble "github.com/canonical/pebble/cmd/pebble"
+	"gopkg.in/check.v1"
+	"net/http"
```
benhoyt, line 21: Nit: we like to keep imports in three separate groups, separated by a blank line: first standard library imports, next third party packages, then local Pebble imports. [Example.](https://github.com/canonical/pebble/blob/46b3a83075bf7329fae3b732b322a56b01a328c7/cmd/pebble/cmd_plan_test.go#L17-L25)

cmd/pebble/cmd_replan_test.go
```diff
@@ -0,0 +1,108 @@
...
+  "id":   "43",
+  "kind": "replan",
+  "summary": "...",
+  "status": "Done",
+  "ready": true,
+  "spawn-time": "2016-04-21T01:02:03Z",
+  "ready-time": "2016-04-21T01:02:04Z",
+  "tasks": []
+}`
+		if r.URL.Path == "/v1/changes" {
```
benhoyt, line 37: It's unclear to me why `/v1/changes` is needed (the test passes fine without it, so I don't think it is).

cmd/pebble/cmd_replan_test.go
```diff
@@ -0,0 +1,108 @@
...
+}`)
+	})
+
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"replan"})
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
+
+	rest, err = pebble.Parser(pebble.Client()).ParseArgs([]string{"replan", "--no-wait"})
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
```
benhoyt, line 72: I think if we're going to check `--no-wait` we should ensure that it *doesn't* send `/v1/changes/43` in that case, and that it prints out the change ID on stdout, as in `c.Check(s.Stdout(), check.Equals, "43\n")`. I realize it's a bit of redundancy, but it'd probably be simpler to do that in a separate test function to avoid having flags like "did we call this URL".

cmd/pebble/cmd_replan_test.go
```diff
@@ -0,0 +1,108 @@
...
+		fmt.Fprintf(w, `{
+    "type": "async",
+    "status-code": 202,
+    "change": "43"
+}`)
+	})
+
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"replan"})
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
```
benhoyt, line 68: Should we check that stdout/stderr are empty, like we do in a bunch of other tests, for example: `c.Check(s.Stderr(), Equals, "")`.

cmd/pebble/cmd_replan_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	rest, err = pebble.Parser(pebble.Client()).ParseArgs([]string{"replan", "--no-wait"})
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
+
+	failGetChange = true
+	rest, err = pebble.Parser(pebble.Client()).ParseArgs([]string{"replan"})
+	c.Assert(err, check.ErrorMatches, "could not foo")
+	c.Assert(rest, check.HasLen, 1)
+	failGetChange = false
```
benhoyt, line 78: Again here, I think it'd be simpler if this was a separate test, avoiding the need for the `failGetChange` flag. A bit more redundancy and lines of code, but much simpler to reason about individual tests.

benhoyt (commented):
Thanks for this! Just some minor comments.

cmd/pebble/cmd_replan_test.go
anpep: Now that you speak of it, I don't remember what was my reasoning behind that. You are absolutely correct, thanks for your feedback!

benhoyt (approved):
LGTM now, thanks!

PR #140, opened 2022-08-16
Add tests for stop command

anpep:


benhoyt (approved):
These are boring, I know -- but it's good to have coverage of these commands, thanks.

PR #141, opened 2022-08-16
Add tests for help command

anpep:


cmd/pebble/cmd_help_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	pebble "github.com/canonical/pebble/cmd/pebble"
+)
+
+func (s *PebbleSuite) TestHelpCommandWorks(c *C) {
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"help", "add"})
+
+	flagsErr, ok := err.(*flags.Error)
+	c.Assert(ok, Equals, true)
+	c.Assert(flagsErr.Type, Equals, flags.ErrHelp)
```
benhoyt, line 29: Re `c.Assert` vs `c.Check`, I know it's a bit confusing, but I generally try to use `Assert` only when the test can't continue, so in this case we would use `Assert` for the `ok == true` check (because the following tests would fail/panic if you didn't), but then `c.Check` for the rest, because then the test failure will show all the check failures at once.

cmd/pebble/cmd_help_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	c.Assert(rest, HasLen, 0)
+	c.Assert(s.Stdout(), Equals, "")
```
benhoyt, line 32: It was surprising to me that stdout was empty here, because it should be printing a help message. Then I looked at the code and realized the `help foo` case (and some other help cases) are handled in `cmd/pebble/main.go:run`.

So I think for the help tests we should change the tests to call run(). You can do this by calling `RunMain` which is defined to be `run` in `export_test.go` (see `TestErrorResult` for an example). This also uses `fakeArgs()` -- which IMO should be done as an argument to `run()` like `run(args []string)` explicitly, so feel free to change that and get rid of `fakeArgs` if you want.

Then we can test that `s.Stdout()` actually has something in it. I think we shouldn't test for the entire help text -- that's annoying to maintain -- but that a couple of small snippets are present, for example `Usage:` and `pebble add` and `The add command` or something.

cmd/pebble/cmd_help_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (s *PebbleSuite) TestHelpManWorks(c *C) {
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"help", "--man"})
+	c.Assert(err, Equals, nil)
+	c.Assert(rest, HasLen, 0)
+	c.Assert(s.Stdout(), Not(Equals), "")
+	c.Assert(s.Stderr(), Equals, "")
+}
+
+func (s *PebbleSuite) TestHelpCommandFailsCommandRequired(c *C) {
```
benhoyt, line 76: These `CommandRequired` tests will go away (or turn into something else) once we change the tests to go via `run()`, because `run` handles that case.

cmd/pebble/cmd_help_test.go
```diff
@@ -0,0 +1,114 @@
...
+func (s *PebbleSuite) TestHelpAllWorks(c *C) {
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"help", "--all"})
+	c.Assert(err, Equals, nil)
+	c.Assert(rest, HasLen, 0)
+	c.Assert(s.Stdout(), Not(Equals), "")
+	c.Assert(s.Stderr(), Equals, "")
+}
+
+func (s *PebbleSuite) TestHelpManWorks(c *C) {
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"help", "--man"})
```
benhoyt, line 69: Huh, I was not aware this existed. Interesting!

cmd/pebble/cmd_help_test.go
```diff
@@ -0,0 +1,114 @@
...
+	c.Assert(rest, HasLen, 0)
+	c.Assert(s.Stdout(), Equals, "")
+	c.Assert(s.Stderr(), Equals, "")
+}
+
+func (s *PebbleSuite) TestHelpAllWorks(c *C) {
+	rest, err := pebble.Parser(pebble.Client()).ParseArgs([]string{"help", "--all"})
+	c.Assert(err, Equals, nil)
+	c.Assert(rest, HasLen, 0)
+	c.Assert(s.Stdout(), Not(Equals), "")
```
benhoyt, line 64: Probably we want to test for a small snippet, something like `c.Check(s.Stdout(), Matches, "(?s)Pebble lets you control services.*Commands can be classified as follows.*")`.

benhoyt (changes requested):

benhoyt (approved):

benhoyt:
Thanks again!

PR #142, opened 2022-08-17
Rework tests for the changes command

anpep:
This small patch revamps the tests for the actual `pebble changes` command (and not only the `pebble tasks` command, which is mostly targeted by the previous test suite).


benhoyt (approved):
LGTM. Thanks again!

Issue #107, opened 2022-02-22
Add Go client and CLI commands for files API


tunahanertekin:
Progress on this:

* [x] https://github.com/canonical/pebble/pull/143
* [x] https://github.com/canonical/pebble/pull/144
* [x] https://github.com/canonical/pebble/pull/146
* [x] https://github.com/canonical/pebble/pull/147
* [x] https://github.com/canonical/pebble/pull/148

For completeness, we need to Go back and implement the Go client and CLI commands for the files API. Our previously discussed approach to this is shown in the [spec](https://docs.google.com/document/d/1NPPZZB_qCwtTru-YsKkRSAb6mXJzBiE0fG3GTZnx4fM/edit#heading=h.1tsi183ijqx). See there for details, but in summary:

```
# Go client. For now, we're not proposing to support multi-file push/pull (same as Python client).
func (*Client) Pull(opts *PullOptions) (PushResult, error)
func (*Client) Push(opts *PushOptions) (PushResult, error)
func (*Client) ListFiles(opts *ListFilesOptions) (ListFilesResult, error)
func (*Client) MakeDir(opts *MakeDirOptions) (MakeDirResult, error)
func (*Client) RemovePath(opts *RemovePathOptions) (RemovePathResult, error)

# Pebble CLI.
pebble push file.txt /path [-m|--mode=644 --user=USER|uid --group=GROUP|gid]
pebble pull /path file.txt
pebble cat /path # same as pull, but write to stdout
pebble mkdir /path [-p|--parents -m|--mode=755 --user=USER|uid --group=GROUP|gid]
pebble rm /path [-r|--recursive]
pebble ls '/etc/*.conf'  # path can be dir, file, or glob
```

@tunahanertekin's original message:

> Hi there, Pebble's File API usage is explained [here](https://juju.is/docs/sdk/pebble#heading--the-files-api) when using it on Juju operator framework. Is there a way to use Pebble's File API on Golang? I've tried to find it in [Pebble's Go client](https://pkg.go.dev/github.com/canonical/pebble) but I couldn't.

benhoyt:
Thanks, @tunahanertekin -- yeah, at this point we don't have a Go client or CLI commands for the files API (there's only the Python one as part of the Python Operator Framework). I'm going to update this issue to request adding that. We should have it for completeness, though we haven't needed it to date for Juju.

Out of interest, what's your use case, and what APIs specifically would be most important? Are you trying to write a Juju operator in Go?

[...]

PR #143, opened 2022-08-18
Implement file listing functionality

anpep:
I went ahead and started implementing #107, starting with the `pebble ls` command for listing files.

I'm a bit uncomfortable about parts of my solution and would really appreciate feedback about those (in particular about the `fileInfo` struct and that struct implementing `fs.FileInfo`)

### DIFF
--- a/client/files.go
+++ b/client/files.go
@@ -15,18 +15,15 @@
 package client
 
 import (
-	"io/fs"
 	"net/url"
 	"os"
 	"strconv"
 	"time"
 )
 
-// ListFilesOptions holds the options for a call to ListFiles
+// ListFilesOptions holds the options for a call to ListFiles.
 type ListFilesOptions struct {
-	// Path is the absolute path of the fiole system entry to list. Note that
-	// when Pattern is specified, it should specify the absolute path of the
-	// parent directory (required)
+	// Path is the absolute path of the file system entry to be listed.
 	Path string
 
 	// Pattern is the glob-like pattern string to filter results by. When
@@ -42,50 +39,69 @@ type ListFilesOptions struct {
 type FileInfo struct {
 	name    string
 	size    int64
-	mode    fs.FileMode
+	mode    os.FileMode
 	modTime time.Time
-
-	// Path is the full absolute path of the file
-	Path string
-	// UserID is the ID of the owner user
-	UserID int
-	// GroupID is the ID of the owner group
-	GroupID int
-	// User is the string representing the owner user name
-	User string
-	// Group is the string representing the owner user group
-	Group string
+	path    string
+	userID  int
+	groupID int
+	user    string
+	group   string
 }
 
-// Name returns the base name of the file
+// Name returns the base name of the file.
 func (fi FileInfo) Name() string {
 	return fi.name
 }
 
 // Size returns the length in bytes for regular files. For others, its
-// behavior is system-dependent
+// behavior is system-dependent.
 func (fi FileInfo) Size() int64 {
 	return fi.size
 }
 
-// Mode returns the file mode and permission bits
-func (fi FileInfo) Mode() fs.FileMode {
+// Mode returns the file mode and permission bits.
+func (fi FileInfo) Mode() os.FileMode {
 	return fi.mode
 }
 
-// ModTime returns the file modification time
+// ModTime returns the file modification time.
 func (fi FileInfo) ModTime() time.Time {
 	return fi.modTime
 }
 
-// IsDir is an abbreviation for Mode().IsDir()
+// IsDir is an abbreviation for Mode().IsDir().
 func (fi FileInfo) IsDir() bool {
 	return fi.mode.IsDir()
 }
 
-// Sys returns the underlying data source (can return nil)
+// Sys returns the underlying data source (can return nil).
 func (fi FileInfo) Sys() interface{} {
 	return nil
+}
+
+// Path is the full absolute path of the file.
+func (fi FileInfo) Path() string {
+	return fi.path
+}
+
+// UserID is the ID of the owner user.
+func (fi FileInfo) UserID() int {
+	return fi.userID
+}
+
+// GroupID is the ID of the owner group.
+func (fi FileInfo) GroupID() int {
+	return fi.groupID
+}
+
+// User is the string representing the owner user name.
+func (fi FileInfo) User() string {
+	return fi.user
+}
+
+// Group is the string representing the owner user group.
+func (fi FileInfo) Group() string {
+	return fi.group
 }
 
 // ListFiles obtains the contents of a directory or glob, or information about a file.
@@ -136,8 +152,9 @@ func calculateFileMode(fileType string, permissions string) (mode os.FileMode, e
 		return 0, err
 	}
 
-	mode = os.FileMode(p)
+	mode = os.FileMode(p) & os.ModePerm
 	switch fileType {
+	case "file":
 	case "directory":
 		mode |= os.ModeDir
 	case "symlink":
@@ -172,17 +189,17 @@ func resultToFileInfo(result fileInfoResult) (*FileInfo, error) {
 		fi.size = *result.Size
 	}
 	if result.UserID != nil {
-		fi.UserID = *result.UserID
+		fi.userID = *result.UserID
 	}
 	if result.GroupID != nil {
-		fi.GroupID = *result.GroupID
+		fi.groupID = *result.GroupID
 	}
 
-	fi.Path = result.Path
+	fi.path = result.Path
 	fi.name = result.Name
 	fi.mode = mode
-	fi.User = result.User
-	fi.Group = result.Group
+	fi.user = result.User
+	fi.group = result.Group
 
 	return fi, nil
 }
--- a/client/files.go
+++ b/client/files.go
@@ -44,8 +44,8 @@ type FileInfo struct {
 	mode    os.FileMode
 	modTime time.Time
 	path    string
-	userID  int
-	groupID int
+	userID  *int
+	groupID *int
 	user    string
 	group   string
 }
@@ -86,13 +86,13 @@ func (fi *FileInfo) Path() string {
 	return fi.path
 }
 
-// UserID is the ID of the owner user.
-func (fi *FileInfo) UserID() int {
+// UserID is the ID of the owner user (can be nil).
+func (fi *FileInfo) UserID() *int {
 	return fi.userID
 }
 
-// GroupID is the ID of the owner group.
-func (fi *FileInfo) GroupID() int {
+// GroupID is the ID of the owner group (can be nil).
+func (fi *FileInfo) GroupID() *int {
 	return fi.groupID
 }
 
@@ -190,13 +190,9 @@ func resultToFileInfo(result fileInfoResult) (*FileInfo, error) {
 	if result.Size != nil {
 		fi.size = *result.Size
 	}
-	if result.UserID != nil {
-		fi.userID = *result.UserID
-	}
-	if result.GroupID != nil {
-		fi.groupID = *result.GroupID
-	}
 
+	fi.userID = result.UserID
+	fi.groupID = result.GroupID
 	fi.path = result.Path
 	fi.name = result.Name
 	fi.mode = mode
--- a/client/files.go
+++ b/client/files.go
@@ -15,7 +15,6 @@
 package client
 
 import (
-	"errors"
 	"io/fs"
 	"net/url"
 	"os"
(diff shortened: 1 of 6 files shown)


benhoyt:
Thanks! Using `fs.FileInfo` is an interesting idea. I'm actually off work today but I'll take a look at this on my Monday.

benhoyt (changes requested):

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,66 @@
...
+`
+
+func (cmd *cmdLs) Execute(args []string) error {
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	opts := client.ListFilesOptions{
+		Path:    cmd.Positional.Path,
+		Pattern: "",
```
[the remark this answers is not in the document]
anpep: Yes, I just didn't implement it yet. I was thinking on parsing the argument passed as "path" and detect some of the common characters for globs (*, ?, etc.). The pattern would be the glob without the directory path, and the path would be the actual directory. I think this is the easiest way to implement it, too, having seen the source code for the files API

client/files.go
```diff
@@ -0,0 +1,195 @@
...
+func (fi fileInfo) User() string {
+	return fi.user
+}
+
+func (fi fileInfo) Group() string {
+	return fi.group
+}
+
+type FileInfo interface {
+	os.FileInfo
```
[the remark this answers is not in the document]
anpep: About option 1, I think it would be a bit counter-intuitive to have it return a `fs.FileInfo` which can be type asserted to `client.FileInfo`. How does the user of our API know this information? Also, since the `client.FileInfo` struct would implement `fs.FileInfo`, it would be usable with functions where `fs.FileInfo` is accepted anyway.

I like the option 2 better which, as you say, is more consistent with the rest of the API. Regarding the struct implementing `fs.FileInfo`, would it look like the following?

```go
type FileInfo struct {
	fs.FileInfo

	Path            string
	UserID, GroupID int
	User, Group     string
}
```

I'm asking because the contrast between having `FileInfo.Name()`, `FileInfo.Mode()`, et al. but then `FileInfo.UserID` and not `FileInfo.UserID()` looks a bit inconsistent to me—but if it's considered more idiomatic I'd go with that (my Go is still a bit bad, also 😅).

Thank you very much for your feedback!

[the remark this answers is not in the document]
benhoyt: Yeah, I'd probably make them all methods, even `Path()`, `User()`, and so on that aren't part of the `fs.FileInfo` interface, just for consistency. It does seem weird to have some of them methods and some fields.

cmd/pebble/cmd_ls.go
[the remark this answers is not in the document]
benhoyt: That's reasonable and works for me. We can at least start with that approach and see what Gustavo and others think.

client/files.go
```diff
@@ -15,82 +15,106 @@
 package client
 
 import (
-	"errors"
 	"io/fs"
```
benhoyt, line 18: Oh, this is interesting -- now that I think of it, `io/fs` was added in Go 1.16, and we currently support Go versions down to 1.14. So we should probably make this `os.FileInfo` after all.

client/files.go
```diff
@@ -15,82 +15,106 @@
...
-	size            int64
-	mode            fs.FileMode
-	modTime         time.Time
-	path            string
-	userID, groupID int
-	user, group     string
-}
-
-func (fi fileInfo) Name() string {
+	// Path is the absolute path of the fiole system entry to list. Note that
```
benhoyt, line 27: Typo: fiole

client/files.go
```diff
@@ -15,82 +15,106 @@
...
+	// when Pattern is specified, it should specify the absolute path of the
+	// parent directory (required)
```
benhoyt, line 29: I'm not sure what this second sentence about Pattern means.

cmd/pebble/cmd_ls.go
```diff
@@ -16,24 +16,32 @@ package main
...
 	"github.com/canonical/pebble/client"
 )
 
 type cmdLs struct {
 	clientMixin
-	Itself     bool `long:"itself"`
+	timeMixin
+
+	Directory  bool `short:"d" long:"directory"`
+	LongFormat bool `short:"l" long:"long"`
```
benhoyt, line 32: Note that `ls` doesn't have the long version of this argument, so probably best to leave `--long` off.

cmd/pebble/cmd_ls.go
```diff
@@ -47,20 +55,26 @@ func (cmd *cmdLs) Execute(args []string) error {
...
-		Itself:  cmd.Itself,
+		Itself:  cmd.Directory,
 	}
 	files, err := cmd.client.ListFiles(&opts)
 	if err != nil {
 		return err
 	}
 
-	fmt.Println("Name\tUser\tGroup\t")
+	w := tabwriter.NewWriter(os.Stdout, 1, 1, 1, ' ', 0)
```
benhoyt, line 65: Use the package-level `Stdout` here for testability (see example in `cmd/pebble/format.go`).

benhoyt (approved):
Still a few small things, but updating my overall to approve.

client/files.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"io/fs"
+	"net/url"
+	"os"
+	"strconv"
+	"time"
+)
+
+// ListFilesOptions holds the options for a call to ListFiles
```
benhoyt, line 25: Nit: sentence should end with a full stop (period).

client/files.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+type FileInfo struct {
+	name    string
+	size    int64
+	mode    fs.FileMode
+	modTime time.Time
+
+	// Path is the full absolute path of the file
+	Path string
```
benhoyt, line 49: Per other discussion, I think these should be methods too, for consistency (so not half of them are methods and half fields).

client/files.go
```diff
@@ -0,0 +1,188 @@
...
+	UserID int
+	// GroupID is the ID of the owner group
+	GroupID int
+	// User is the string representing the owner user name
+	User string
+	// Group is the string representing the owner user group
+	Group string
+}
+
+// Name returns the base name of the file
```
benhoyt, line 60: Nit: sentences should end with a `.` (and below).

client/files.go
```diff
@@ -0,0 +1,188 @@
...
+	case "symlink":
+		mode |= os.ModeSymlink
+	case "socket":
+		mode |= os.ModeSocket
+	case "named-pipe":
+		mode |= os.ModeNamedPipe
+	case "device":
+		mode |= os.ModeDevice
+	default:
+		mode |= os.ModeIrregular
```
benhoyt, line 152: This may have partly been me being unclear, but I think we still need a `case "file"` that sets no bits, otherwise `"file"` will become `os.ModeIrregular`. I'd do like so:

```go
switch fileType {
case "file":
    // no mode bits set for regular file
case "directory":
    mode |= os.ModeDir
...
default:
    mode |= os.ModeIrregular
}
```

anpep: Further testing revealed that issue rather quickly, thank you!

cmd/pebble/cmd_ls.go
anpep: That explains why my `s.Stdout()` assertions didn't work, thank you!

client/files_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/canonical/pebble/client"
+	. "gopkg.in/check.v1"
+)
+
+func (cs *clientSuite) TestListFiles(c *C) {
+	_, err := cs.cli.ListFiles(&client.ListFilesOptions{
+		Path: "/",
+	})
+	c.Assert(err, IsNil)
```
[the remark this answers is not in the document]
anpep: Isn't this already tested on the daemon? https://github.com/canonical/pebble/blob/a755cf20f4baba6f81717d2c82e28518fa9e7926/internal/daemon/api_files_test.go#L100-L118

Also, I haven't found a way to test the client like this (setting up a Pebble API and spin up the client, instead of mocking the API responses)

client/files.go
```diff
@@ -0,0 +1,205 @@
...
+	name    string
+	size    int64
+	mode    os.FileMode
+	modTime time.Time
+	path    string
+	userID  int
+	groupID int
+	user    string
+	group   string
+}
```
benhoyt, line 49: It would probably be good to add a compile-time check that this implements `os.FileInfo`. In Go a common idiom to do that is adding the line (at the package level):

```go
var _ os.FileInfo = (*FileInfo)(nil)
```

Oh wait, that makes me realize: I think the receiver for all of these should be a pointer receiver, to avoid copying this relatively large struct (and to match the `[]*FileInfo` return value of `ListFiles`).

client/files.go
```diff
@@ -0,0 +1,205 @@
...
+func (fi FileInfo) ModTime() time.Time {
+	return fi.modTime
+}
+
+// IsDir is an abbreviation for Mode().IsDir().
+func (fi FileInfo) IsDir() bool {
+	return fi.mode.IsDir()
+}
+
+// Sys returns the underlying data source (can return nil).
```
benhoyt, line 77: Should probably be more specific, like "(always nil for client.FileInfo)".

client/files_test.go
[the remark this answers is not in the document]
benhoyt: Fair enough. However, I think we want a bit more coverage of the conversion routines. For example, I think we should check at least some of the results here, and check that the handling of the pointers / optional fields like UserID/GroupID is working correctly.

client/files_test.go
```diff
@@ -0,0 +1,280 @@
...
+  "status-code": 200,
+  "status": "OK",
+  "result": [
+    {
+      "path": "/bin",
+      "name": "bin",
+      "type": "symlink",
+      "permissions": "777",
+      "last-modified": "2022-04-21T03:02:51Z",
+      "user-id": 0,
```
benhoyt, line 164: We should probably test a value other than 0 for user-id/group-id, because 0 is the zero value so could happen "by accident". (Or we can test it above in `TestListFiles`.)

client/files_test.go
```diff
@@ -0,0 +1,280 @@
...
+      "user": "root",
+      "group-id": 0,
+      "group": "root"
+	}
+  ]
+}`
+	_, err := cs.cli.ListFiles(&client.ListFilesOptions{
+		Path: "/irreg",
+	})
+	c.Assert(err, ErrorMatches, "*invalid syntax*")
```
benhoyt, line 227: I'm not sure I understand how this matches, as `*` isn't a valid regular expression, and `ErrorMatches` takes a regex, right? Wouldn't it be `.*`?

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,106 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	w := tabWriter()
+	defer w.Flush()
+	for _, fi := range files {
+		if cmd.LongFormat {
+			fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\n", fi.Mode().String(), fi.User(), fi.Group(), cmd.fmtTime(fi.ModTime()), fi.Name())
```
benhoyt, line 77: I think we should add the size before the date column for type=file (probably make it "-" for other types).

benhoyt (approved):
Nice! Very close now, just a few minor comments. Also, Gustavo still wants to review larger/API changes, so we'll need his eyes after that.

client/files_test.go
anpep: Actually, I was thinking that the API can return a `nil` uid/gid, but `client.FileInfo` actually uses `int` and not `*int` for those. Should we keep using `*int` to allow for `nil` values on those fields?

https://github.com/canonical/pebble/blob/4855fa64b0a3e89fca33958306c20dbe19ab39f1/client/files.go#L191-L196

I don't think `userID` and `groupID` should be 0 (a perfectly valid uid/gid) if the API returns `nil` for those. For the `size` attribute things are a bit different because we need to implement `os.FileInfo.Size() int64` and I don't know if using a sentinel value like -1 works in this context.

What do you think?

client/files_test.go
anpep: Mere luck ;)

Captura de pantalla 2022-08-24 a las 12 19 15

client/files_test.go
benhoyt: Good point. We should probably do `*int` for UserID and GroupID to match the API and semantics.

I think size is fine as is (no need to return -1) because it's explicitly marked as implementation-defined if type!=file.

client/files.go
```diff
@@ -0,0 +1,203 @@
...
+	Group        string `json:"group"`
+}
+
+func calculateFileMode(fileType string, permissions string) (mode os.FileMode, err error) {
+	p, err := strconv.ParseUint(permissions, 8, 32)
+	if err != nil {
+		return 0, err
+	}
+
+	mode = os.FileMode(p) & os.ModePerm
```
flotter, line 157: We are masking out all bits here except for 0777. I just want to make sure its intentional that we do not want to show setuid, setgid or sticky.

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,113 @@
...
+	defer w.Flush()
+	for _, fi := range files {
+		if cmd.LongFormat {
+			var size string
+			if fi.Mode().IsRegular() {
+				size = quantity.FormatAmount(uint64(fi.Size()), -1) + "B"
+			} else {
+				size = "-"
+			}
+			fmt.Fprintf(w, "%s\t%s\t%s\t%6s\t%s\t%s\n", fi.Mode().String(), fi.User(), fi.Group(), size, cmd.fmtTime(fi.ModTime()), fi.Name())
```
flotter, line 84: I think mode here will print out as decimal values, not octal. Perhaps this should even look like what the "ls" command would print for various mode options instead?

anpep: I wish it was better documented, but [`FileMode.String()`](https://pkg.go.dev/io/fs#FileMode.String) returns something like `ls` would (e.g. `drwxrw-r--`)

client/files.go
anpep: I did not take that into account! We should fix this along [the 4-digit octal issue you mentioned in the other PR](https://github.com/canonical/pebble/pull/144#discussion_r959414764).

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// ErrPatternOutsideFileName is returned if a pattern is found anywhere but the file name of a path
+var ErrPatternOutsideFileName = errors.New("patterns can only be applied to file names, not directories")
+
+type cmdLs struct {
+	clientMixin
+	timeMixin
+
+	Directory  bool `short:"d" long:"directory"`
```
benhoyt, line 36: Per discussion with Gustavo: let's just use the shorthand notation `-d`, not support `--directory`.

Rationale: over time we've learned that it doesn't help to offer ambiguity on such options. If it is something that's unclear or we only use it rarely, we use the long form so people understand what it is. If you use it all the time then a short notation is good, and people only ever use the short `-d` form. It's quite rare to have an option that's both unclear and used all the time.

benhoyt (commented):

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+var lsDescs = map[string]string{
+	"d": `Display information about the file system entry, instead of listing directory contents.`,
+	"l": `Display file system entries in a list format.`,
+}
+
+var shortLsHelp = "List path contents"
+var longLsHelp = `
+The ls command takes a path (a file, directory or glob) and obtains its
```
niemeyer, line 51:
```
The ls command lists entries in the filesystem at the specified path. A glob pattern
may be specified for the last path element.
```

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return nil
+}
+
+func init() {
+	addCommand("ls", shortLsHelp, longLsHelp, func() flags.Commander { return &cmdLs{} }, merge(lsDescs, timeDescs), nil)
+}
+
+func parseGlob(path string) (parsedPath string, parsedPattern string, err error) {
```
niemeyer, line 97: The results may be shortened to `(parsedPath, parsedPattern string, err error)`

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,113 @@
...
+	dir, file := pathpkg.Split(strings.TrimRight(path, "/"))
+
+	const patternCharacters = "*?["
+	if strings.ContainsAny(dir, patternCharacters) {
+		// Patterns can not be applied recursively, only on file names
+		return "", "", ErrPatternOutsideFileName
```
niemeyer, line 103: The comment above can go once the error message is inline, as it's self-documenting.

cmd/pebble/cmd_ls.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if strings.ContainsAny(file, patternCharacters) {
+		// File name contains a pattern
```
niemeyer, line 107: This comment here is unnecessary, as it's spelling out the code right above it. (literally: _"if contains file pattern"_).

As mentioned, this is clearly nitpicking, but since these are the first few PRs, I'm taking the chance to be a bit more detailed to help us establishing a pattern for future work.

cmd/pebble/cmd_ls_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (s *PebbleSuite) TestLsLongFormat(c *C) {
+	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
+		c.Assert(r.Method, Equals, "GET")
+		c.Assert(r.URL.Path, Equals, "/v1/files")
+		c.Assert(r.URL.Query(), DeepEquals, url.Values{"action": {"list"}, "path": {"/"}})
+		fmt.Fprintln(w, `{
+	"type": "sync",
+	"result": [
+		{
```
niemeyer, line 72: Same as earlier about the style.

niemeyer (changes requested):
Thanks for the PR, Angel, and for the review Ben. That series of PRs are some of the cleanest work I've seen lately.

I'm sending below a number of comments, and some of them can certainly be considered over-zealous, but since that's the first series of PRs there are some details that would be nice to share so that future work we do together can be all on the same page. Many of these are simple patterns, that once we agree on will get out of the way for good.

PR #144, opened 2022-08-25
Implement mkdir functionality

anpep:
This is a WIP PR with the aim of implementing the `pebble mkdir` command and the associated client-side functionality. Feedback is greatly appreciated! (specially around error handling logic)


client/files.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if len(result) > 0 && result[0].Error != nil {
+		return errors.New(result[0].Error.Message)
+	}
+
+	return nil
+}
+
+func convertOptionsToPayload(opts *MakeDirOptions) (*makeDirPayload, error) {
```
benhoyt, line 107: This is in the `client` package` so needs a more unique name. I'd probably just include this logic in the `MakeDir` function, or call it `buildMakeDirPayload` or something.

client/files.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	var result []fileResult
+	_, err = client.doSync("POST", "/v1/files", nil, nil, &body, &result)
+	if err != nil {
+		return err
+	}
+
+	if len(result) > 1 {
+		panic(fmt.Sprintf("expected at most 1 result from API, got %d", len(result)))
```
benhoyt, line 97: If this was a programming error, I'd agree with a panic for this. However, it's a server error so we shouldn't panic but return an ordinary error.

client/files.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func convertOptionsToPayload(opts *MakeDirOptions) (*makeDirPayload, error) {
+	user := opts.User
+	group := opts.Group
+
+	// UserID/GroupID take precedence over User/Group
+	// If we don't do this, the call to MakeDir can fail if the user/group
+	// name is not consistent with the ID, which can be counter-intuitive
+	if opts.User != "" && opts.UserID != nil {
```
benhoyt, line 114: I think we should pass these directly through, as the API does that by design. It's to ensure that if you pass through both ID and name, it tells you if they don't match -- so I think we want to forward that error through to the user.

client/files.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Permissions specifies the 3-digit octal UNIX permissions the created
+	// directories will have.
+	Permissions string
+
+	// UserID indicates the ID of the owner user for the created directories.
+	UserID *int
+
+	// User indicates the name of the owner user for the created directories.
+	// If used together with UserID, UserID takes precedence.
```
benhoyt, line 42: Per comment below, I think we should make this match the API, and behave like "If used together with UserID, the value must match UserID's username."

client/files.go
```diff
@@ -0,0 +1,135 @@
...
+	Error *errorResult `json:"error,omitempty"`
+}
+
+type errorResult struct {
+	Message string `json:"message"`
+}
+
+// MakeDir creates a directory or directory tree.
+func (client *Client) MakeDir(opts *MakeDirOptions) error {
+	payload, err := convertOptionsToPayload(opts)
```
benhoyt, line 79: For a local helper function that can't fail, I don't see any reason to return an error. We can easily add an error if it's needed.

client/files.go
benhoyt: It also seems to me the test should be `len(result) != 1` here, as there's always a result (with a `Path` field) even on success.

client/files_test.go
```diff
@@ -0,0 +1,121 @@
...
+	MakeParents bool   `json:"make-parents"`
+	Permissions string `json:"permissions"`
+	UserID      *int   `json:"user-id"`
+	User        string `json:"user"`
+	GroupID     *int   `json:"group-id"`
+	Group       string `json:"group"`
+}
+
+func (cs *clientSuite) TestMakeDir(c *C) {
+	cs.rsp = `{"type":"sync","result":[]}`
```
benhoyt, line 41: It seems to me there should be one result with a `"path"` field here.

client/files_test.go
```diff
@@ -0,0 +1,121 @@
...
+	cs.rsp = `
+{
+	"type": "sync",
+	"result": [
+		{
+			"path": "/foobar",
+			"error": {
+				"message": "could not bar",
+				"kind": "permission-denied",
+				"value": 42
```
benhoyt, line 97: Hmm, this makes me think we should probably have a more specific error type, like `FileError` with the kind broken out as an enum. Or we could use the existing `client.Error` type (and add to the `ErrorKind...` values).

benhoyt (commented):
Overall looking fine! Left some comments re error handling and a few other things.

client/files_test.go
anpep: We actually do! I think it's `errorKindGenericFileError`

client/files.go
```diff
@@ -0,0 +1,127 @@
...
+	// be created or not. If one of the parent directories of the specified
+	// path does not exist and MakeParents is false, the call will fail.
+
+	// MakeParents, if true, specifies that any non-existent parent directories
+	// should be created. If false (the default), the call will fail if the
+	// directory to be created has at least one parent directory that does not
+	// exist.
+	MakeParents bool
+
+	// Permissions specifies the permission bits of the directories to be created.
```
benhoyt, line 40: We should probably document that the directory is created with permission 0o755 if this is nil.

Alternatively, it might simplify the API to make this a non-pointer, with 0 meaning "use default". It seems to me it's not useful to use 0 as a permission value.

client/files.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	var body bytes.Buffer
+	err := json.NewEncoder(&body).Encode(&payload)
+	if err != nil {
+		return err
+	}
+
+	var result []fileResult
+	_, err = client.doSync("POST", "/v1/files", nil, nil, &body, &result)
```
benhoyt, line 113: This currently fails against a real Pebble with the error `invalid Content-Type ""`. We need to add a `Content-Type` header:

```diff
$ git diff
diff --git a/client/files.go b/client/files.go
index 1404dce..45bf8f9 100644
--- a/client/files.go
+++ b/client/files.go
@@ -110,7 +110,10 @@ func (client *Client) MakeDir(opts *MakeDirOptions) error {
        }
 
        var result []fileResult
-       _, err = client.doSync("POST", "/v1/files", nil, nil, &body, &result)
+       headers := map[string]string{
+               "Content-Type": "application/json",
+       }
+       _, err = client.doSync("POST", "/v1/files", nil, headers, &body, &result)
        if err != nil {
                return err
        }
```

To catch issues like this, it would also be good if we had at least one client-to-real-server test to test this (same with the other file APIs).

I tested locally by running `pebble run` in one terminal window, then `pebble mkdir` in another.

client/files_test.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+		}
+	]
+}`
+
+	err := cs.cli.MakeDir(&client.MakeDirOptions{
+		Path: "/foobar",
+	})
+	c.Assert(err, ErrorMatches, "could not bar")
```
benhoyt, line 135: Here's where we would test that `err` type-asserts to `*client.Error` with `Kind == "permission-denied"`.

cmd/pebble/cmd_mkdir.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/canonical/pebble/client"
+)
+
+type cmdMkdir struct {
+	clientMixin
+
+	MakeParents bool   `short:"p" long:"parents"`
+	Permissions string `short:"m" long:"mode"`
+	User        string `short:"u" long:"user"`
+	Group       string `short:"g" long:"group"`
```
benhoyt, line 33: With the `exec` command we discussed this, and ended up going for two params, `--uid` and `--user` (and `--gid` and `--group`). Let's do the same here for consistency, and drop the short `u` and `g` variants.

client/files.go
```diff
@@ -78,14 +78,16 @@ type fileResult struct {
...
+	Value   interface{} `json:"value,omitempty"`
 }
 
 // MakeDir creates a directory or directory tree.
 func (client *Client) MakeDir(opts *MakeDirOptions) error {
 	var permissions string
-	if opts.Permissions != nil {
-		permissions = fmt.Sprintf("%o", *opts.Permissions)
+	if opts.Permissions != 0 {
+		permissions = fmt.Sprintf("%3o", opts.Permissions)
```
benhoyt, line 90: Per my original comment, this needs to be `%03o`, so it's zero-padded rather than space-padded. Otherwise you get:

```
$ pebble mkdir /home/ben/w/pebble/foo -m 25
error: permissions must be a 3-digit octal string, got " 25"
exit status 1
```

We should probably add a test for a 1 or 2 digit mode to catch this.

client/files.go
```diff
@@ -78,14 +78,16 @@ type fileResult struct {
 }
 
 type errorResult struct {
-	Message string `json:"message"`
+	Message string      `json:"message"`
+	Kind    string      `json:"kind,omitempty"`
+	Value   interface{} `json:"value,omitempty"`
 }
 
 // MakeDir creates a directory or directory tree.
```
benhoyt, line 86: Can we add something a paragraph like:

```
// The error returned is a *Error if the request went through successfully
// but there was an OS-level error creating the directory, with the Kind
// field set to the specific error kind, for example "permission-denied".
```

benhoyt (approved):

client/files.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// MakeDir creates a directory or directory tree.
+// The error returned is a *Error if the request went through successfully
+// but there was an OS-level error creating the directory, with the Kind
+// field set to the specific error kind, for example "permission-denied".
+func (client *Client) MakeDir(opts *MakeDirOptions) error {
+	var permissions string
+	if opts.Permissions != 0 {
+		permissions = fmt.Sprintf("%03o", opts.Permissions)
```
flotter, line 89: I was not sure why we are printing this in at least a 3-rune form. It looks like the daemon tests for either zero or 3 rune length string when validating permissions.

Note that treating the directory permissions as only 3 octal values may not be correct. At least chisel recently added support for the sticky bit if I read this correctly. 

https://github.com/canonical/chisel/commit/c06b76261f1399256655dddc99196c28cb592291

This means a change in the daemon handler.

Also see: https://github.com/golang/go/issues/25539

client/files.go
```diff
@@ -0,0 +1,134 @@
...
+	// should be created. If false (the default), the call will fail if the
+	// directory to be created has at least one parent directory that does not
+	// exist.
+	MakeParents bool
+
+	// Permissions specifies the permission bits of the directories to be created.
+	// If 0 or unset, defaults to 755.
+	Permissions os.FileMode
+
+	// UserID indicates the ID of the owner user for the created directories.
```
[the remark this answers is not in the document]
anpep: I completely agree. Thank you!

client/files.go
anpep: This is the common logic I've found across pebble, but I was secretly thinking on unifying that behavior once all the file-related PRs are merged, perhaps by adding this utility function to x-go:

https://github.com/canonical/pebble/blob/b13464b97acd35b140b1ad2758ce745f0735daf8/internal/daemon/api_files.go#L514-L523

And, of course, treat the permissions properly there. So, I'm not sure if I should directly fix it here, implement a second `parsePermissions()` function on files.go with the correct behavior, or leave it like this and fix it once these PRs are merged.

What do you think?

flotter: I guess breaking it up makes sense.

client/files.go
```diff
@@ -86,7 +86,7 @@ type errorResult struct {
 func (client *Client) MakeDir(opts *MakeDirOptions) error {
 	var permissions string
 	if opts.Permissions != 0 {
-		permissions = fmt.Sprintf("%03o", opts.Permissions)
+		permissions = fmt.Sprintf("%04o", opts.Permissions)
```
benhoyt, line 89: We should still make this `%03o` so the client works with older Pebble servers (which checks for exactly 3). That will make it 3 digits, unless you pass in a higher number like 0o1755, in which case it'll be 4 digits (which won't work on older clients anyway).

internal/daemon/api_files.go
```diff
@@ -250,7 +250,7 @@ func fileInfoToResult(fullPath string, info os.FileInfo, userCache, groupCache m
 		Name:         pathpkg.Base(fullPath),
 		Type:         fileModeToType(mode),
 		Size:         psize,
-		Permissions:  fmt.Sprintf("%03o", mode.Perm()),
+		Permissions:  fmt.Sprintf("%04o", mode.Perm()),
```
benhoyt, line 253: Similar to above, I'd prefer this to be `%03o` so it's 3 digit by default and older, tightly-validating clients don't reject it. I realize this probably won't happen in reality, because the main client (Python Operator Framework) looks like it handles any number of digits, but better to be safe and backwards-compatible. Also, go's `mode.Perm()` will never return more than 3 digits now anyway, because `ModePerm FileMode = 0777`.

internal/osutil/permissions.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2014-2020 Canonical Ltd
```
benhoyt, line 1: We've been putting just "(c) 2022" in the header for new files.

internal/osutil/permissions.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	// Allow 777, 0777 or 1777, but never 77 or 1
+	if len(permissions) != 3 && len(permissions) != 4 {
+		return 0, fmt.Errorf("file permissions must be a 3- or 4-digit octal string, got %q", permissions)
+	}
+
+	v, err := strconv.ParseUint(permissions, 8, 32)
+	if err != nil {
+		return 0, fmt.Errorf("file permissions must be a valid %d-digit octal string, got %q", len(permissions), permissions)
```
benhoyt, line 38: I think it's a bit confusing to have this return a specific length. Let's make this error string the same as above.

internal/osutil/permissions.go
```diff
@@ -0,0 +1,42 @@
...
+
+package osutil
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+)
+
+// ParsePermissions parses an octal string representing UNIX file permissions
```
benhoyt, line 23: Nit: usually spelled "Unix" these days.

internal/osutil/permissions.go
```diff
@@ -0,0 +1,42 @@
...
+// into a os.FileMode value. If permissions is an empty string, defaultMode will
+// be used instead.
+func ParsePermissions(permissions string, defaultMode os.FileMode) (os.FileMode, error) {
```
benhoyt, line 26: We also need to add a `permissions_test.go` file testing these various cases.

benhoyt:
I also noticed the tests are failing not due to intermittent failures but due to the `%04o` change. Once we revert that as per my comments above that should resolve itself. You can also run all Go tests locally with `go test ./...`

benhoyt (approved):

benhoyt:

internal/osutil/permissions.go
```diff
@@ -0,0 +1,42 @@
...
+// ParsePermissions parses an octal string representing Unix file permissions
+// into a os.FileMode value. If permissions is an empty string, defaultMode will
+// be used instead.
+func ParsePermissions(permissions string, defaultMode os.FileMode) (os.FileMode, error) {
+	if permissions == "" {
+		return defaultMode, nil
+	}
+
+	// Allow 777, 0777 or 1777, but never 77 or 1
+	if len(permissions) != 3 && len(permissions) != 4 {
```
niemeyer, line 32: Major red-flag here: this PR is about adding mkdir functionality on the client, yet it silently moved this function while changing its behavior, and this is not even documented in the PR description.  So this is silently modifying the API of Pebble in a way that is very easy to miss. The change in behavior needs to be reverted from this PR, and moved into its own PR with an explanation about why it's required and potential consequences for Pebble usage.

To be extra clear, moving the function to osutil is fine, and I'm not saying the change in behavior is wrong. I haven't even stopped to think about it. I'm just highlighting that we need to be careful with such silent external API changes as part of unrelated PRs. Once an API changes, typically there's no going back.

In addition, this particular change also seems to disagree with what's documented elsewhere in the PR.

niemeyer (changes requested):
Thanks, this looks very nice in general. One important detail, and many trivial ones. Please have a look at #143 first as this refers to some earlier comments there.

niemeyer (approved):
Thanks for the tweaks. For future changes, please avoid force-pushing once the PR has open reviews, as that makes it harder for the reviewers to incrementally review (the "View changes since last review" link is gone).

This LGTM.. Just needs de-conflicting and a since we're there, another trivial:

benhoyt:
Excellent, thanks for your work on this! Merging.

PR #146, opened 2022-08-30
Implement RemovePath and pebble rm command

anpep:


client/files.go
```diff
@@ -0,0 +1,98 @@
...
+	"encoding/json"
+	"fmt"
+)
+
+// RemovePathOptions holds the options for a call to RemovePath.
+type RemovePathOptions struct {
+	// Path is the absolute path to be deleted (required).
+	Path string
+
+	// MakeParents, if true, specifies that any non-existent parent directories
```
benhoyt, line 28: Comment needs updating.

client/files.go
```diff
@@ -0,0 +1,98 @@
...
+type fileResult struct {
+	Path  string       `json:"path"`
+	Error *errorResult `json:"error,omitempty"`
+}
+
+type errorResult struct {
+	Message string      `json:"message"`
+	Kind    string      `json:"kind,omitempty"`
+	Value   interface{} `json:"value,omitempty"`
+}
```
benhoyt, line 54: There's going to be quite a lot of merge conflicts when we merge all of these. :-) But that's fine -- I think `fileResult` and `errorResult` should be general across all the file APIs.

client/files.go
```diff
@@ -0,0 +1,98 @@
...
+
+// RemovePath deletes files and directories.
```
benhoyt, line 56: What about "deletes a file or directory", as it only deletes one (possibly recursively).

client/files.go
```diff
@@ -0,0 +1,98 @@
...
+// The error returned is a *Error if the request went through successfully
+// but there was an OS-level error creating the directory, with the Kind
```
benhoyt, line 58: This comment needs updating.

cmd/pebble/cmd_rm.go
```diff
@@ -0,0 +1,57 @@
...
+
+	Recursive bool `short:"r" long:"recursive"`
+
+	Positional struct {
+		Path string `positional-arg-name:"<path>"`
+	} `positional-args:"yes" required:"yes"`
+}
+
+var rmDescs = map[string]string{
+	"recursive": "Create parent directories as needed.",
```
benhoyt, line 34: Text needs updating.

cmd/pebble/cmd_rm.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+var shortRmHelp = "Remove files and directories"
+var longRmHelp = `
+The rm command removes a file or directory.
+If --recursive is specified, all files and directories contained
+within the specified path will also be removed.
```
benhoyt, line 41: IMO the second sentence can be removed, as it should be documented under the `--recursive` argument's help.

benhoyt (commented):
Wasn't sure whether this was ready for review yet, but I did a quick pass. I'll review more thoroughly later -- just let me know.

client/files.go
anpep: You're completely right (: I didn't find any alternative to this while working on multiple related functionalities at once, and didn't want any PR to be dependent on each other.

client/files.go
anpep: That's slightly better, thank you!

cmd/pebble/cmd_rm.go
anpep: I agree with you. Here I mimicked the help text of other commands such as `add` which sometimes include a sentence about an option that's relevant to the command.

https://github.com/canonical/pebble/blob/b13464b97acd35b140b1ad2758ce745f0735daf8/cmd/pebble/cmd_add.go#L40-L45

client/files.go
benhoyt: Yeah, I think that's the right choice for review -- thanks!

client/files.go
```diff
@@ -0,0 +1,96 @@
...
+	Path      string `json:"path"`
+	Recursive bool   `json:"recursive"`
+}
+
+type fileResult struct {
+	Path  string       `json:"path"`
+	Error *errorResult `json:"error,omitempty"`
+}
+
+type errorResult struct {
```
benhoyt, line 48: Thinking about this now, we should be able to JSON-decode this directly to `*client.Error`, which already has these fields and the JSON tags. Similar for the other PRs.

cmd/pebble/cmd_rm.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/jessevdk/go-flags"
+
+	"github.com/canonical/pebble/client"
+)
+
+type cmdRm struct {
+	clientMixin
+
+	Recursive bool `short:"r" long:"recursive"`
```
benhoyt, line 26: Similar here as other PRs: let's remove `--recursive` and just have `-r`.

benhoyt (commented):

niemeyer (approved):
This looks good, thanks Angel.  Looks ready for merging, just pending conflict resolution with the prior list files PR.

PR #147, opened 2022-09-01
feat: push client functionality and CLI command

anpep:
This PR previously contained the implementations of both the push and pull commands and client functionalities. This PR has been refactored to now include only the push relevant code, with the pull functionality present in #148. This allows for the individual PRs to be kept small. Thanks for your feedback!


client/files.go
```diff
@@ -0,0 +1,204 @@
...
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		if err := u.uploadFiles(); err != nil {
+			panic(err)
```
benhoyt, line 189: Per above, we probably won't need this goroutine, but normally for error handling you'd use a channel to send the err value back to the caller.

client/files_test.go
```diff
@@ -0,0 +1,60 @@
...
+	c.Assert(err, IsNil)
+
+	err = cs.cli.PushFile(&client.PushFileOptions{
+		LocalPath:  filePath,
+		RemotePath: "/file.dat",
+	})
+	c.Assert(err, IsNil)
+
+	c.Assert(cs.req.URL.Path, Equals, "/v1/files")
+	c.Assert(cs.req.Method, Equals, "POST")
```
benhoyt, line 59: In terms of testing, I think you'll want to parse `cs.req.Body` with a `multipart.Reader` and use `cs.req.Header` to get the boundary. See https://pkg.go.dev/mime/multipart#example-NewReader for an example of how to parse the boundary from the `Content-Type` header. You can then check that there are two parts (the JSON metadata and the file content) and that each part is what we expect.

benhoyt (changes requested):
I think this is a good start, thanks! I've suggested a significant simplification to the implementation to avoid the goroutine -- let me know what you think.

For `Pull`, we *could* us the signature from the [spec](https://docs.google.com/document/d/1NPPZZB_qCwtTru-YsKkRSAb6mXJzBiE0fG3GTZnx4fM/edit#heading=h.1tsi183ijqx), like so:

```go
func (*Client) Pull(opts *PullOptions) (*PushResult, error)

type PullOptions struct {
    Path string
}

type PullResult struct {
    Reader io.ReadCloser
}
```

This is similar to the Python client's [`pull`](https://ops.readthedocs.io/en/latest/#ops.pebble.Client.pull) method, which takes a path and returns a reader.

However, I'm thinking now that for the Go client it would be simpler to implement and to use in the normal case (writing the pulled content to a file) if instead of returning a `ReadCloser`, it takes a `Dest io.Writer` in the `PullOptions` struct. To write to a file, you just pass in the file open for writing as `Dest`. So it's easy to use, plus the implementation can again be simple, without goroutines.

So I guess I'm advocating for:

```go
func (*Client) Pull(opts *PullOptions) error

type PullOptions struct {
    Path string
    Dest io.Writer
}
```

client/files.go
```diff
@@ -0,0 +1,204 @@
...
+	Kind    string      `json:"kind,omitempty"`
+	Value   interface{} `json:"value,omitempty"`
+}
+
+type fileUpload struct {
+	localPath string
+	info      writeFilesItem
+}
+
+type fileUploader struct {
```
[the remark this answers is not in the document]
anpep: That's really elegant, thank you for your feedback!

anpep:
I've implemented `client.Pull()` and the `pebble pull` command, but there are a lot of things that can be ironed out, in particular:
- Due to a dependency on `http.Response.Body` for parsing the multipart body of the response, I was unable to use any of the `client.raw()`/`client.do()`/`client.doSync()` methods and [basically had to copy over the logic from `client.raw()`](https://github.com/canonical/pebble/pull/147/files#diff-c6dfd31f1fc23f5fa255eafba0fec5faa9f4605faf0bab68cca47f40e48d34f6R196-R215). Due to the lack of a better mechanism (please correct me if I'm wrong), I propose the `Body` member from the `http.Response` object on `client.raw()` to be exposed in the currently-empty `client.ResultInfo` struct. Maybe there's a better way to achieve this, feedback welcome!
- The `client.Pull()` method returns an struct with an `io.ReadCloser` associated to the multipart file. However, calling `.Close()` on this struct will not close the `http.Response.Body` that is left open. Maybe we can wrap this `io.ReadCloser` in a struct like this?
  ```go
  type PullReadCloser struct {
      reader io.ReadCloser
      body io.Closer
  }
  func (p PullReadCloser) Read(b []byte) (n int, err error) {
      return p.reader.Read(b)
  }
  func (p PullReadCloser) Close() error {
      if err := p.reader.Close(); err != nil {
          return err
      }
      if err := p.body.Close(); err != nil {
          return err
      }
  }
  ``` 
- I still need to write some tests, but I wanted to get this commit a bit early so I could gather feedback before I go on.

cmd/pebble/cmd_push.go
```diff
@@ -0,0 +1,137 @@
...
+		return ErrExtraArgs
+	}
+
+	f, err := os.Open(cmd.Positional.LocalPath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	p, err := newPushProgress(f, cmd.Positional.RemotePath)
```
benhoyt, line 118: To keep this PR smaller, it might be good to leave this out for now, and add in a follow-up PR. Personally I'm not sure it's worth it. On all the files I've tried it's been so fast I just see the progress bar flash (or not see it at all) and then it says, "45MB transferred in 150ms" or whatever. Though of course that's on a local network.

cmd/pebble/cmd_pull.go
```diff
@@ -0,0 +1,125 @@
...
+
+		size := quantity.FormatAmount(uint64(p.size), 0)
+		duration := quantity.FormatDuration(time.Since(p.started).Seconds())
+		p.pb.Notify(fmt.Sprintf("Transferred %sB in %s", size, duration))
+	}
+
+	return
+}
+
+func newPullProgress(f *os.File, remotePath string) (*pullProgress, error) {
```
benhoyt, line 74: Similar comment here as below -- maybe we leave the progress bar to a separate PR to keep this one smaller (and maybe not do it at all). It also requires the new `Content-Length` header you've added, which won't be available on older servers (see related comments).

benhoyt (commented):
Thanks, good work! Add a number of comments. The main thing we need to lock down is the signature of `Pull` and `PullOptions`.

client/files.go
```diff
@@ -0,0 +1,275 @@
...
+	// that might be created will also be owned by this group.
+	Group string
+}
+
+// PullOptions contains the options for a call to Pull.
+type PullOptions struct {
+	// Path indicates the absolute path of the file in the remote system
+	// (required).
+	Path string
+}
```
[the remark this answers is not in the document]
anpep: Absolutely, I went straight to the spec Google Doc and forgot to review past comments on this PR, so I missed your feedback on that, which solves a bunch of things. Thanks!

client/files.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if mediaType != "multipart/form-data" {
+		// Probably JSON-encoded error response
+		var res response
+		var fr []fileResult
+
+		if err := decodeInto(rsp.Body, &res); err != nil {
```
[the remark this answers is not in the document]
anpep: I think you are right, I just copied the implementation from `client.go` but didn't take this into account. And, yes, I think the problem goes away with the `Dest io.Writer`

benhoyt (approved):
Changes look good, thanks for that!

cmd/pebble/cmd_push.go
```diff
@@ -0,0 +1,98 @@
...
+		UserID:      cmd.UserID,
+		User:        cmd.User,
+		GroupID:     cmd.GroupID,
+		Group:       cmd.Group,
+	})
+	if err != nil {
+		return err
+	}
+
+	size := quantity.FormatAmount(uint64(st.Size()), -1)
```
benhoyt, line 89: Per discussion with Gustavo: let's just drop this output for now (and hence drop the Stat above too). We can revisit later, looking at how snapd does progress bar output for reference.

cmd/pebble/cmd_push.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return err
+	}
+
+	t := time.Now()
+	err = cmd.client.Push(&client.PushOptions{
+		Source:      f,
+		Path:        cmd.Positional.RemotePath,
+		MakeDirs:    cmd.MakeDirs,
+		Permissions: cmd.Permissions,
```
benhoyt, line 79: Per discussion with Gustavo, the default for `-p` / permissions should be the permission bits of the local file (only the 777 part / low 9 bits).

cmd/pebble/cmd_push.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/canonical/pebble/client"
+	"github.com/canonical/pebble/internal/strutil/quantity"
+)
+
+type cmdPush struct {
+	clientMixin
+
+	MakeDirs    bool   `short:"p" long:"parents"`
+	Permissions string `short:"m" long:"mode"`
```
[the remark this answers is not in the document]
benhoyt: Per discussion with Gustavo at mini-sprint, we think the default permissions (mode bits) for `pebble push` should be based on the local file mode, as that's the most likely desired outcome ("once an executable, always an executable").

However, if `-m <nnn>` is specified, obviously that would override the local file mode.

benhoyt (commented):

client/files.go
```diff
@@ -0,0 +1,162 @@
...
+		"Content-Type":        {"application/json"},
+		"Content-Disposition": {`form-data; name="request"`},
+	})
+	if err != nil {
+		return fmt.Errorf("cannot encode metadata in request payload: %w", err)
+	}
+
+	payload := writeFilesPayload{
+		Action: "write",
+		Files: []writeFilesItem{{
```
flotter, line 110: I see the daemon (and the slices [] of writeFileItem confirms) can handle a body with multiple file parts. Any idea how that would or could map to the push command we expose? Perhaps in the future we could support transfer multiple files by specifying a directory, or recursively write files from a base directory?

I guess this is not important, just noticed the difference between the daemon and client.

flotter (approved):

client/files.go
anpep: Interesting feedback. I limited myself to implement the API that was already on the spec, but I assume we could take a `[]*PushOptions` slice instead of just `*PushOptions` in order to support this without much more complexity.

benhoyt: Yeah, we spent time designed the underlying API to be multi-file friendly, but we decided to only implement single-file Go functions in the Client (Go client, as well as the Python client in `ops`) for simplicity for now. We can add them separately later if need be.

benhoyt:
@anpep Care to bring this branch up to date, and we'll see if we can get it over the goal line before too long?

benhoyt:
FWIW, this has been requested a few times, by @sed-i on the Observability team most recently.

client/files.go
flotter: Sounds like this can be a future discussion.

client/files.go
```diff
@@ -364,3 +368,135 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
 
 	return nil
 }
+
+// PushOptions contains the options for a call to Push.
```
benhoyt, line 372: Nit: we've moved towards not adding doc comments for "obvious" types/functions, like the comment for `PushOptions` basically saying "options for Push". So let's remove this.

client/files.go
```diff
@@ -364,3 +368,135 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	// Source is the source of data to write (required).
+	Source io.Reader
+
+	// Path indicates the absolute path of the file in the destination
+	// machine (required).
+	Path string
+
+	// MakeDirs, if true, will create any non-existing directories in the path
+	// to the remote file. If false, the default, the call to Push will
+	// fail if any non-existing directory is found on the remote path.
```
benhoyt, line 383:
```suggestion
	// MakeDirs, if true, will create any non-existing directories in the path
	// to the remote file. If false (the default) the call to Push will
	// fail if any of the parent directories of path do not exist.
```

client/files.go
```diff
@@ -364,3 +368,135 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	MakeDirs bool
+
+	// Permissions indicates the mode of the file in the destination machine.
+	// Defaults to 0644. Note that, when used together with MakeDirs, the
+	// directories that might be created will not use this mode, but 0755.
+	Permissions string
```
benhoyt, line 389: It seems to me this should be of type `os.FileMode`, like the similar `Permissions` field in `MakeDirOptions`. Can you remember why we made it a string here?

```suggestion
	// Permissions indicates the mode of the file on the destination machine.
	// If 0 or unset, defaults to 0644. Note that, when used together with MakeDirs, the
	// directories that might be created will not use this mode, but 0755.
	Permissions os.FileMode
```

client/files.go
```diff
@@ -364,3 +368,135 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+
+	// UserID indicates the user ID of the owner for the file in the destination
+	// machine. When used together with MakeDirs, the directories that might be
+	// created will also be owned by this user.
```
benhoyt, line 393: Suggestion (and similar below):

```suggestion
	// UserID indicates the user ID of the owner for the file on the destination
	// machine. When used together with MakeDirs, the directories that are
	// created will also be owned by this user.
```

client/files.go
```diff
@@ -364,3 +368,135 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	if err != nil {
+		return fmt.Errorf("cannot encode metadata in request payload: %w", err)
+	}
+
+	payload := writeFilesPayload{
+		Action: "write",
+		Files: []writeFilesItem{{
+			Path:        opts.Path,
+			MakeDirs:    opts.MakeDirs,
+			Permissions: opts.Permissions,
```
benhoyt, line 453: This will become:
```suggestion
			Permissions: fmt.Sprintf("%03o", opts.Permissions),
```

client/files.go
```diff
@@ -364,3 +368,135 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	b.Reset()
+	mw.Close()
+	footer := b.String()
+
+	var result []fileResult
+	body := io.MultiReader(strings.NewReader(header), opts.Source, strings.NewReader(footer))
+	headers := map[string]string{
+		"Content-Type": mw.FormDataContentType(),
+	}
+	if _, err := client.doSync("POST", "/v1/files", nil, headers, body, &result); err != nil {
```
benhoyt, line 486: Let's use the new `client.Requester().Do()` interface here.

internals/cli/cmd_push.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func init() {
+	AddCommand(&CmdInfo{
+		Name:        "push",
+		Summary:     cmdPushSummary,
+		Description: cmdPushDescription,
+		ArgsHelp: map[string]string{
+			"-p":      "Create parent directories for the file",
+			"-m":      "Set permissions for the file (e.g. 0644)",
```
benhoyt, line 54: What about "override" (because the default is to copy the permission bits):

```suggestion
			"-m":      "Override permissions for the file (3-digit octal)",
```

internals/cli/cmd_push.go
```diff
@@ -0,0 +1,96 @@
...
+	defer f.Close()
+
+	st, err := f.Stat()
+	if err != nil {
+		return err
+	}
+
+	var permissions string
+	if cmd.Permissions != "" {
+		permissions = cmd.Permissions
```
benhoyt, line 81: With the change to `os.FileMode` this will need to use `strconv.ParseUint(cmd.Permissions, 8, 32)` like the `mkdir` command does.

benhoyt (changes requested):
Hi @anpep, I'd love to bring this up to date and merged soon, as the CLI `push` and `pull` commands have been requested by more than one person. Let's merge in the latest master and update to using the new `client.Requester().Do()` method.

I've also left a few minor comments -- some tweaks, and one real change: to make `Permissions` an integer `os.FileMode` rather than a string.

client/files.go
anpep: I wrote this long ago, so I don't really recall why I did this in the first place (: 
But that's a good catch! Thank you

internals/cli/cmd_push.go
anpep: I agree with this, but the other file-related commands use exactly the same help text. Should we update them all to match or keep the current one?

anpep: As per our agreement today, let's use your suggestion and make a separate PR to propagate the changes to the other commands!

client/files.go
```diff
@@ -364,3 +369,147 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	// Encode multipart footer
+	body.Reset()
+	mw.Close()
+	footer := body.String()
+
+	resp, err := client.Requester().Do(context.Background(), &RequestOptions{
+		Type:    SyncRequest,
+		Method:  "POST",
+		Path:    "/v1/files",
+		Query:   nil,
```
benhoyt, line 491: No need to specify `Query`, as it's optional (the zero value is nil).

internals/cli/cmd_push.go
```diff
@@ -0,0 +1,102 @@
...
+		Path:        cmd.Positional.RemotePath,
+		MakeDirs:    cmd.MakeDirs,
+		Permissions: st.Mode().Perm(),
+		UserID:      cmd.UserID,
+		User:        cmd.User,
+		GroupID:     cmd.GroupID,
+		Group:       cmd.Group,
+	}
+
+	if cmd.Permissions != "" {
```
benhoyt, line 93: It fills a bit odd to always call stat and use `st.Mode().Perm()`, only to override it here. Let's structure this a bit differently, have a variable `var permissions os.FileMode` and only call stat in the else case where `cmd.Permissions == ""`:

```go
var permissions os.FileMode
if cmd.Permissions != "" {
    // ... parse it ...
    permissions = os.FileMode(p)
} else {
    // ... call stat ...
    permissions = st.Mode().Perm()
}

opts := client.PushOptions{
    // ...
}
```

benhoyt (changes requested):
Thanks for the updates. Just a couple of minor tweaks requested.

internals/cli/cmd_push.go
```diff
@@ -74,29 +74,27 @@ func (cmd *cmdPush) Execute(args []string) error {
...
-	st, err := f.Stat()
-	if err != nil {
+	var permissions os.FileMode
+	if cmd.Permissions != "" {
+		p, err := strconv.ParseUint(cmd.Permissions, 8, 32)
+		if err != nil {
+			return fmt.Errorf("invalid mode for directory: %q", cmd.Permissions)
+		}
+		permissions = os.FileMode(p)
+	} else if st, err := f.Stat(); err == nil {
```
benhoyt, line 84: Sorry to be nit-picky, but I'd prefer we write this as follows to make the if/else variants balanced and each handling one case, rather than the stat error handling put into a 3rd clause, and to use the usual `!= nil` expression rather than `== nil`.

```go
} else {
    st, err := f.Stat()
    if err != nil {
        return err
    }
    permissions = st.Mode().Perm()
}
```

anpep: Thanks! Will take into account from now on (:

benhoyt (approved):
Looks good, thanks for the updates. Approving -- I'll aim to merge this in the next couple of days.

internals/cli/cmd_push.go
```diff
@@ -0,0 +1,102 @@
...
+const cmdPushSummary = "Transfer a file to the remote system"
+const cmdPushDescription = `
+The push command transfers a file to the remote system.
+`
+
+type cmdPush struct {
+	client *client.Client
+
+	Parents bool   `short:"p"`
+	Mode    string `short:"m"`
```
benhoyt, line 36: @anpep FYI, I changed these fields names to `Parents` and `Mode` so they matched the 'p' and 'm' (before it was confusing because **M**akeDirs went with 'p' and **P**ermissions went with 'm'). I've done the similar change in `cmd_mkdir.go`.

I realize this makes the names different from the client field names, but that's okay.

PR #148, opened 2022-09-08
feat: pull client functionality and CLI command

anpep:
This PR isolates the pull functionality that was removed from the previous PR (#147) and introduces the following changes:

 * Use a destination `io.Writer` instead of a local file path or `io.Reader`.
 * Includes close to perfect coverage of the `pebble pull` command and client functionality.

Feedback appreciated!


client/files.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// PullOptions contains the options for a call to Pull.
+type PullOptions struct {
+	// Path indicates the absolute path of the file in the remote system
+	// (required).
+	Path string
+
+	// Dest is the destination io.Writer that will receive the data (required).
+	Dest io.Writer
```
benhoyt, line 35: Per discussion with Gustavo at mini-sprint: we bike-shedding on this for about 10 minutes :-) and ended up deciding on `Target` rather than `Dest` or "destination". Rationale is that "target" is a complete word and short (in fact, same number of letters as "source" to boot).

client/files.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type fileResult struct {
+	Path  string       `json:"path"`
+	Error *errorResult `json:"error,omitempty"`
+}
+
+type errorResult struct {
```
benhoyt, line 43: Same as other PR: let's use `*client.Error` (there shouldn't be any need for an `errorResult` separately).

benhoyt (commented):

client/files.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid Content-Type: %w", err)
+	}
+	if mediaType != "multipart/form-data" {
+		// Not an error response after all
+		return fmt.Errorf("expected a multipart response but didn't get one")
+	}
+
+	mr := multipart.NewReader(rsp.Body, params["boundary"])
+	for {
```
benhoyt, line 72: We're expecting one and only one file here (two multipart parts, `files` and `response`), so it seems to me we shouldn't use a `for` loop, but instead just call `NextPart()` twice in a row, ensuring we get `FormName() == "files"` for the first one, and `FormName() == "response"` for the second.

With the current approach, it will succeed silently and fall to the bottom if there are 0 or 1 parts (but there should be exactly 2).

So something like this:

```go
mr := multipart.NewReader(rsp.Body, params["boundary"])

filesPart, err := mr.NextPart()
if err != nil {
    return fmt.Errorf("cannot decode multipart payload: %w", err)
}
defer filesPart.Close()
if filesPart.FormName() != "files" {
    return fmt.Errorf(`expected first field name to be "files", got %q`, filesPart.FormName())
}
if _, err = io.Copy(opts.Target, part); err != nil {
    return fmt.Errorf("cannot write: %w", err)
}

responsePart, err := mr.NextPart()
if err != nil {
    return fmt.Errorf("cannot decode multipart payload: %w", err)
}
defer responsePart.Close()
if responsePart.FormName() != "response" {
    return fmt.Errorf(`expected second field name to be "response", got %q`, responsePart.FormName())
}
// process response metadata
```

client/files.go
```diff
@@ -0,0 +1,119 @@
...
+	"net/url"
+)
+
+// PullOptions contains the options for a call to Pull.
+type PullOptions struct {
+	// Path indicates the absolute path of the file in the remote system
+	// (required).
+	Path string
+
+	// Target is the destination io.Writer that will receive the data (required).
```
benhoyt, line 34: We might want to add something like: "During a call to Pull, Target may be written to even if an error is returned."

client/files.go
```diff
@@ -0,0 +1,119 @@
...
+		defer part.Close()
+
+		if part.FormName() == "files" {
+			if _, err = io.Copy(opts.Target, part); err != nil {
+				return fmt.Errorf("cannot write: %w", err)
+			}
+		} else if part.FormName() == "response" {
+			// Process response metadata
+			var res response
+			var fr []fileResult
```
benhoyt, line 89: Better to move `var fr []fileResult` down below to just above where it's first used (just above the `decodeWithNumber` line).

client/files.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+			if err := res.err(client); err != nil {
+				return err
+			}
+			if res.Type != "sync" {
+				return fmt.Errorf("expected sync response, got %q", res.Type)
+			}
+			if err := decodeWithNumber(bytes.NewReader(res.Result), &fr); err != nil {
+				return fmt.Errorf("cannot unmarshal result: %w", err)
+			}
```
benhoyt, line 103: Coming back to this it was unclear to me why we need to use `decodeWithNumber`, and then I realized these lines are copied from `client.do`. Probably good to group those lines into a block (just add a blank line above `if err := res.err(client)` with a comment at the top of the block saying "This checking/decoding logic is copied from client.do" or similar).

client/files.go
```diff
@@ -0,0 +1,119 @@
...
+	// Obtain Content-Type to check for a multipart payload and parse its value
+	// in order to obtain the multipart boundary
+	contentType := rsp.Header.Get("Content-Type")
+	mediaType, params, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return fmt.Errorf("invalid Content-Type: %w", err)
+	}
+	if mediaType != "multipart/form-data" {
+		// Not an error response after all
+		return fmt.Errorf("expected a multipart response but didn't get one")
```
benhoyt, line 68: How about: `fmt.Errorf("expected a multipart response, got %q", mediaType)`?

client/files_test.go
```diff
@@ -0,0 +1,327 @@
...
+	cs.rsp = srcBuf.String()
+
+	// Check response
+	err = cs.cli.Pull(&client.PullOptions{
+		Path: "/foo/bar.dat",
+	})
+	c.Assert(err, ErrorMatches, "expected exactly one result from API, got 2")
+}
+
+func (cs *clientSuite) TestPullFailsWithClientError(c *C) {
```
benhoyt, line 288: Thorough suite of tests -- thanks!

client/files.go
anpep: You are completely right about it failing silently, that must be fixed. However, my rationale for using a loop here is that I didn't want to make any assumption on the ordering of the multipart data. One could place the `response` part before `files` and it would be as valid as doing it the other way around. Your approach is more clear and surely works on Pebble, but should we impose this "hidden" restriction on future Pebble client implementations?

client/files_test.go
anpep: (^:

client/files.go
benhoyt: Interesting rationale. However, the order is significant: the `files` part (can be multiple in the API, but in this case always 1) is guaranteed to come before the `response` part. This is a design decision, because the files have to be opened and read and written to the HTTP connection, which can error in various ways. So the errors are only known afterward trying to write the file, hence including the `response` with the metadata+errors at the bottom.

When pushing (writing), it's the opposite. The client must include the `request` metadata part first, then the `files` parts.

anpep: Thanks for your feedback!

flotter (approved):
I do not yet have the Go experience give any authoritative approvals. However, I deeply studied both the pull and push code, and compared the expectations of the daemon side, and it looks good for me.

The only observation, which I highlighted on the push PR, is the daemons ability to handle multiple files in a request, and the client side only ever requests a single file, which seems like a sensible starting point.

anpep:
> I do not yet have the Go experience give any authoritative approvals. However, I deeply studied both the pull and push code, and compared the expectations of the daemon side, and it looks good for me.
> 
> The only observation, which I highlighted on the push PR, is the daemons ability to handle multiple files in a request, and the client side only ever requests a single file, which seems like a sensible starting point.

Thanks for your feedback on both PRs!

benhoyt (approved):
@anpep Looking good -- no further comments. Could you also bring this one up to date, and maybe we can get it merged before too long as well? (Though I realize it'll conflict again with #147 if we merge that first.)

benhoyt:
FWIW, this has been requested a few times, by @sed-i on the Observability team most recently.

anpep:
I've brought this PR up to date. I'd like to receive feedback on this one, especially on the part where I handle response headers. I decided to write a simple `headerToMap()` function which converts a `http.Header` to a `map[string]string`. Note that the underlying type behind `http.Header` is `map[string][]string` (in order to handle multiple headers with the same name), but I wanted to retain the same type we use for `RequestOptions`:

```go
type RequestOptions struct {
    // ...
    Headers map[string]string
}

type RequestResponse struct {
    // ...
    Headers map[string]string
}
```

client/client.go
```diff
@@ -375,11 +380,20 @@ func (rq *defaultRequester) Do(ctx context.Context, opts *RequestOptions) (*Requ
 	// Common response
 	return &RequestResponse{
 		StatusCode: serverResp.StatusCode,
+		Headers:    headerToMap(httpResp.Header),
 		ChangeID:   serverResp.Change,
 		Result:     serverResp.Result,
 	}, nil
 }
 
+func headerToMap(h http.Header) map[string]string {
```
benhoyt, line 389: This should go away once we switch to using an `http.Header`.

client/files.go
```diff
@@ -364,3 +368,96 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+		Method: "GET",
+		Path:   "/v1/files",
+		Query: map[string][]string{
+			"action": {"read"},
+			"path":   {opts.Path},
+		},
+		Headers: map[string]string{
+			"Accept": "multipart/form-data",
+		},
+		Body: nil,
```
benhoyt, line 396: No need for this line -- it's implied.

client/files.go
```diff
@@ -364,3 +368,96 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	})
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	// Obtain Content-Type to check for a multipart payload and parse its value
+	// in order to obtain the multipart boundary.
+	mediaType, params, err := mime.ParseMediaType(resp.Headers["Content-Type"])
```
benhoyt, line 405:
```suggestion
	mediaType, params, err := mime.ParseMediaType(resp.Headers.Get"Content-Type"))
```

client/files.go
```diff
@@ -364,3 +368,96 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	if err != nil {
+		return fmt.Errorf("invalid Content-Type: %w", err)
```
benhoyt, line 407:
```suggestion
		return fmt.Errorf("cannot parse Content-Type: %w", err)
```

client/files.go
```diff
@@ -364,3 +368,96 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	if err != nil {
+		return fmt.Errorf("cannot decode multipart payload: %w", err)
+	}
+	defer filesPart.Close()
+
+	if filesPart.FormName() != "files" {
+		return fmt.Errorf(`expected first field name to be "files", got %q`, filesPart.FormName())
+	}
+	if _, err := io.Copy(opts.Target, filesPart); err != nil {
+		return fmt.Errorf("cannot write: %w", err)
```
benhoyt, line 425:
```suggestion
		return fmt.Errorf("cannot write to target: %w", err)
```

client/files.go
```diff
@@ -364,3 +368,96 @@ func (client *Client) RemovePath(opts *RemovePathOptions) error {
...
+	responsePart, err := mr.NextPart()
+	if err != nil {
+		return fmt.Errorf("cannot decode multipart payload: %w", err)
+	}
+	defer responsePart.Close()
+	if responsePart.FormName() != "response" {
+		return fmt.Errorf(`expected second field name to be "response", got %q`, responsePart.FormName())
+	}
+
+	// Process response metadata (see defaultRequester.Do() in client package)
```
benhoyt, line 437:
```suggestion
	// Process response metadata (see defaultRequester.Do)
```

client/files_test.go
```diff
@@ -574,3 +580,356 @@ func (cs *clientSuite) TestRemovePathFailsWithMultipleAPIResults(c *C) {
...
+	// Check response
+	var targetBuf bytes.Buffer
+	err := cs.cli.Pull(&client.PullOptions{
+		Path:   "/foo/bar.dat",
+		Target: &targetBuf,
+	})
+	c.Assert(err, ErrorMatches, "cannot decode multipart payload: .*")
+}
+
+type defectiveWriter struct{}
```
benhoyt, line 661:
```suggestion
type errWriter struct{}
```

internals/cli/cmd_help.go
```diff
@@ -192,7 +192,7 @@ var HelpCategories = []HelpCategory{{
 }, {
 	Label:       "Files",
 	Description: "work with files and execute commands",
-	Commands:    []string{"ls", "mkdir", "rm", "exec"},
+	Commands:    []string{"pull", "ls", "mkdir", "rm", "exec"},
```
benhoyt, line 195: Note: let's put `push` first in this list once we've merged that from master. (I think that'll be the more commonly used one.)

internals/cli/cmd_pull_test.go
```diff
@@ -0,0 +1,206 @@
...
+	args := []string{"pull", "/foo/bar.dat", filePath}
+	rest, err := cli.Parser(cli.Client()).ParseArgs(args)
+	c.Assert(err, IsNil)
+	c.Assert(rest, HasLen, 0)
+	c.Check(s.Stdout(), Equals, "")
+	c.Check(s.Stderr(), Equals, "")
+
+	f, err := os.Open(filePath)
+	c.Assert(err, IsNil)
+	b, err := ioutil.ReadAll(f)
```
benhoyt, line 82: `Open` ... `ReadAll` can be replaced by `os.ReadFile`.

benhoyt (changes requested):
Just the `Headers` field type change (as discussed) and some very nitty things to change.

client/client.go
```diff
@@ -61,6 +61,7 @@ type RequestOptions struct {
 
 type RequestResponse struct {
 	StatusCode int
+	Header     http.Header
```
benhoyt, line 64: Sorry to nitpick, but I think I'd like this called `Headers` plural to match the field name in `RequestOptions` (I've always found it weird that the `net/http` field name is singular anyway).

internals/cli/cmd_help.go
benhoyt: The merge from master hasn't been done yet.

benhoyt (changes requested):
One more tweak `Header` -> `Headers`, and also we need to merge from master and fix conflicts. Thanks!

anpep:
Sorry about the incomplete changes @benhoyt! I absolutely forgot we already merged `push` into master (:

benhoyt (approved):
Looking good!

PR #151, opened 2022-09-28
Implement boot-firmware command

anpep:
The `boot-firmware` command, when run as init process for Termus, should perform all the tasks needed to leave the system in a working state.

This is a draft PR for implementing this functionality. It currently lacks tests but those can be added later if you feel good about the overall quality of this patch (:

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -55,7 +55,7 @@ snappass  disabled  inactive
 
 ## Using Curl to hit the API
 
-For debugging, you can also use [curl](https://curl.se/) in Unix socket mode to hit the Pebble API:
+For debugging, you can also use [curl](https://curl.se/) in unix socket mode to hit the Pebble API:
 
 ```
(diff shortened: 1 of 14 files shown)


cmd/pebble/cmd_boot.go
```diff
@@ -0,0 +1,61 @@
...
+var shortBootHelp = `Bootstrap a system with Pebble running as PID 1`
+
+var longBootHelp = `
+The boot command checks for PID and working directory restrictions
+for Pebble and prepares required mountpoints before running the
+configured Pebble environment.
+`
+
+func (cmd *cmdBoot) Execute(args []string) error {
+	if len(args) > 1 {
```
atesburak, line 43: The command itself is a way to consume boot but we will have other means of consuming the feature. I think it would help avoid redundant code if we move all the checks to the internal pebble daemon.

anpep: Oops! I think I made a typo here, and the expression would be `len(args) > 0`. I include this check as it's a common pattern I've seen in all other pebble commands.

PR #152, opened 2022-09-28
[JUJU-1885] Moves Pebble plan types and helpers to public.

tlm:
This change moves around Pebble types and utils associated with plans into public packages (not internal). This change was done so that other software configuring Pebble plans can have access to the concrete types to build up configuration, for example Juju.

The Juju team would like this so we aren't redefining types that run the risk of becoming out of sync with Pebble.


internal/tarjan/tarjan.go
```diff
@@ -24,13 +24,13 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-package plan
+package tarjan
```
benhoyt, line 27: I suggest we call this package `sortutil` in case we add more sorting utilities here (and it's similar naming to the existing `strutil` and the like).

internal/tarjan/tarjan_test.go
```diff
@@ -24,7 +24,7 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-package plan
+package tarjan
```
benhoyt, line 27: Looks like you need to update the function name in the tests.

benhoyt (commented):
Rename suggestion, and tests need fixing, but otherwise I agree this is a reasonable thing to do.

benhoyt (approved):
LGTM

hpidcock (approved):

PR #155, opened 2022-10-06
[JUJU-1889] Log forwarding

barrettj12:
Continues work done in #123. I will make changes as per review comments on that PR, as well as the [spec](https://docs.google.com/document/d/12tJFhIbVIlOcblG6l9_u-K5bCib78lojRJ-hdGBxSsw/edit?usp=sharing).


internal/overlord/servstate/handlers.go
```diff
@@ -373,10 +374,14 @@ func (s *serviceData) startInternal() error {
 		// started (previous logs have already been copied).
 		outputIterator = s.logs.HeadIterator(0)
 	}
+
+	// set up output logging for the command
 	serviceName := s.config.Name
 	logWriter := servicelog.NewFormatWriter(s.logs, serviceName)
-	s.cmd.Stdout = logWriter
-	s.cmd.Stderr = logWriter
+	forwarder := logstate.NewLogForwarder(s.manager.logMgr.GetTargets, serviceName)
```
hpidcock, line 381: so we need to remove this log forwarder and the io.MultiWriter here, we should be using s.logs

niemeyer:
We have a new PR being pushed with "take two", so I'm closing this so we can clean up the queue a bit.

Please note I have not gone through any of these PRs. This is indeed just some janitoring.

PR #156, opened 2022-10-10
Refactor plan package

barrettj12:
#152 exported the `plan` package so that external Go projects could use the types defined there to easy write Pebble layers. However, this package also dragged in a bunch of extra utilities, causing compilation problems on non-Unix. This PR does the following:

- Create a new package `internal/planutil`
- Move the following methods (and associated constants / helper methods) from `plan` into `internal/planutil`:
  - CombineLayers
  - StartOrder
  - StopOrder
  - ParseLayer
  - ReadLayersDir
  - ReadDir
- Move the `plan` tests into `internal/planutil`
- Update imports for things which moved from `plan` to `internal/planutil`.

Now, importing the `github.com/canonical/pebble/plan` package will give just the struct definitions, which should fix the compilation issues.


tlm (commented):
lgtm @barrettj12

layer/layer.go
```diff
@@ -0,0 +1,355 @@
+package layer
```
hpidcock, line 1: Missing license.

barrettj12: Good catch, a few other files were missing licenses too so I added these.

internal/overlord/servstate/manager.go
```diff
@@ -118,7 +133,7 @@ func (m *ServiceManager) Plan() (*plan.Plan, error) {
 // AppendLayer appends the given layer to the plan's layers and updates the
 // layer.Order field to the new order. If a layer with layer.Label already
 // exists, return an error of type *LabelExists.
-func (m *ServiceManager) AppendLayer(layer *plan.Layer) error {
+func (m *ServiceManager) AppendLayer(layer *layer.Layer) error {
```
benhoyt, line 136: Hmm, yeah, `layer` is such a good (and common) variable name. This shadowing is allowed in Go, but it's a bit odd and confusing: "in this piece of code, is `layer` the package or a variable?"

Perhaps would should call the package `layers` instead? Or even `types`? (`types.Layer` reads pretty well.)

internal/overlord/servstate/manager.go
```diff
@@ -177,36 +192,36 @@ func findLayer(layers []*plan.Layer, label string) (int, *plan.Layer) {
 // CombineLayer combines the given layer with an existing layer that has the
 // same label. If no existing layer has the label, append a new one. In either
 // case, update the layer.Order field to the new order.
-func (m *ServiceManager) CombineLayer(layer *plan.Layer) error {
+func (m *ServiceManager) CombineLayer(l *layer.Layer) error {
```
benhoyt, line 195: Sort of proves the point above about the package name.

tlm (approved):

internal/planutil/plan.go
```diff
@@ -0,0 +1,506 @@
+// Copyright (c) 2021 Canonical Ltd
```
benhoyt, line 1: This file should probably be called `planutil.go` now.

internal/planutil/plan_test.go
```diff
@@ -13,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package plan_test
+package planutil_test
```
benhoyt, line 15: And this file `planutil_test.go`.

benhoyt (approved):
Looks good to me! Let's just change the `planutil` filenames to `planutil*.go`.

niemeyer:
Actually, given go.mod supposedly this should not be a problem since no code is supposed to break upfront without an update, so I going to revert it so that we can stack the upcoming PRs without this change.

PR #160, opened 2022-10-25
Flesh out client package docs; small README improvements

benhoyt:
This add a brief package doc comment to the client package, as well as short doc comment to (I think!) every type and method in the client package. This is part of a larger effort to document Pebble more thoroughly -- this PR focusses on the Go reference docs for the `client` package.

Also make a few updates to the README. Removed two bullet points from "Some details worth highlighting", as those are discussed in detail below. Link to API docs instead of providing an inline example. Update "TODO" section items.

Also add an example for the client package docs. It's not really run (in the "too hard basket" for now; I've tested it locally though), but at least it's type checked.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -35,7 +35,7 @@ import (
 )
 
 // SocketNotFoundError is the error type returned when the client fails
-// to find a Unix socket at the specified path.
+// to find a unix socket at the specified path.
 type SocketNotFoundError struct {
 	// Err is the wrapped error.
 	Err error
@@ -44,7 +44,6 @@ type SocketNotFoundError struct {
 	Path string
 }
 
-// Error implements the error interface.
 func (s SocketNotFoundError) Error() string {
 	if s.Path == "" && s.Err != nil {
 		return s.Err.Error()
@@ -52,7 +51,6 @@ func (s SocketNotFoundError) Error() string {
 	return fmt.Sprintf("socket %q not found", s.Path)
 }
 
-// Unwrap implements errors.Unwrap interface.
 func (s SocketNotFoundError) Unwrap() error {
 	return s.Err
 }
@@ -92,10 +90,10 @@ type doer interface {
 // Config allows the user to customize client behavior.
 type Config struct {
 	// BaseURL contains the base URL where the Pebble daemon is expected to be.
-	// It can be empty for a default behavior of talking over a Unix socket.
+	// It can be empty for a default behavior of talking over a unix socket.
 	BaseURL string
 
-	// Socket is the path to the Unix socket to use.
+	// Socket is the path to the unix socket to use.
 	Socket string
 
 	// DisableKeepAlive indicates that the connections should not be kept
@@ -133,7 +131,6 @@ type jsonWriter interface {
 	WriteJSON(v interface{}) error
 }
 
-// New returns a new Client with the given configuration.
 func New(config *Config) (*Client, error) {
 	if config == nil {
 		config = &Config{}
@@ -143,7 +140,7 @@ func New(config *Config) (*Client, error) {
 	var transport *http.Transport
 
 	if config.BaseURL == "" {
-		// By default talk over a UNIX socket.
+		// By default talk over a unix socket.
 		transport = &http.Transport{Dial: unixDialer(config.Socket), DisableKeepAlives: config.DisableKeepAlive}
 		baseURL := url.URL{Scheme: "http", Host: "localhost"}
 		client = &Client{baseURL: baseURL}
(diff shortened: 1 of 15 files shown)


client/files.go
```diff
@@ -39,6 +39,7 @@ type ListFilesOptions struct {
 	Itself bool
 }
 
+// FileInfo holds information about a single file.
```
niemeyer, line 42: Same.

client/client.go
```diff
@@ -201,14 +202,20 @@ func (client *Client) WarningsSummary() (count int, timestamp time.Time) {
 	return client.warningCount, client.warningTimestamp
 }
 
+// RequestError is returned when there's an error creating an http.Request.
```
niemeyer, line 205: This sounds too specific. We can return this for anything related to the request processing before the actual request is made. We don't need to mention the underlying implementation of performing the request here.

client/client.go
```diff
@@ -133,7 +133,7 @@ type jsonWriter interface {
 	WriteJSON(v interface{}) error
 }
 
-// New returns a new instance of Client
+// New returns a new Client with the given configuration.
```
niemeyer, line 136: This was already there, but also a good example of unnecessary commenting.

I'll stop mentioning such cases now as the idea should be clear, but there are other instances.

client/client.go
```diff
@@ -91,22 +91,22 @@ type doer interface {
...
 // Config allows the user to customize client behavior.
 type Config struct {
-	// BaseURL contains the base URL where the pebble daemon is expected to be.
-	// It can be empty for a default behavior of talking over a unix socket.
+	// BaseURL contains the base URL where the Pebble daemon is expected to be.
+	// It can be empty for a default behavior of talking over a Unix socket.
 	BaseURL string
 
-	// Socket is the path to the unix socket to use
+	// Socket is the path to the Unix socket to use.
```
niemeyer, line 98: This should probably be kept lowercased, as we are not talking about the Actual Unix, but rather a unix-typed socket, so more of an adjective than the product name. People often call it *nix, but that seems over the top. Lowercasing should be enough.

niemeyer (changes requested):
That's super nice, thanks Ben!

Just a few minors:

client/client.go
```diff
@@ -201,14 +202,20 @@ func (client *Client) WarningsSummary() (count int, timestamp time.Time) {
...
 	return fmt.Sprintf("cannot build request: %v", e.error)
 }
 
-type ConnectionError struct{ error }
+// ConnectionError represents a connection or communication error.
+type ConnectionError struct {
+	error
+}
 
+// Error implements the error interface.
```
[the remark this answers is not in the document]
benhoyt: Thanks for the review! I know what you're getting at, and in regular code I can't stand `i++ // increment i` type of comments.

However, these feed directly into the online pkg.go.dev docs, and I think it's important to be complete there, even if sometimes redundant and obvious. One person's obvious will be another person's "why isn't this documented? does this do what I think it does?".

Another thing my suggested approach has going for it: no more debating which methods should be commented and which not. The only exceptions I think we should make are for `String()` and `Error()`, as they're builtin/stdlib interfaces common to all Go projects.

Thoughts?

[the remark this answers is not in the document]
niemeyer: It sounds like we are in complete agreement, so it's not clear what the push-back or the "However" here is.  The rule I follow and encourage is simple: don't write comments which are obvious. This is important because not only it prevents us writing stuff that we and others will need to read without adding anything to their understanding, but it actually encourages writing good comments: if the comment for a function or type is obvious, that's often a hint that we're not doing a good job explaining what's not obvious. If there's really nothing to explain, then the comment is irrelevant.

These points apply equally to both the code and the doc site.

client/example_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"log"
+
+	"github.com/canonical/pebble/client"
+)
+
+var runExample = false
+
+func Example() {
+	if !runExample {
+		return // don't run this example under "go test"
```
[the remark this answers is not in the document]
benhoyt: Done, though I updated the comment and moved it to beside the `runExample`, to say why we're including it here if it's not run (because it will be type-checked at least).

client/client.go
benhoyt: Fair -- updated similar instances to match.

benhoyt:
@niemeyer Thanks. I've removed obvious comments now, such as comments on `String()` and `Error()` and comments like "WaitChangeOptions holds the options for the WaitChange call." Ready for re-review.

niemeyer (approved):
Thank you. This all looks good now. :+1:

We just need the conflicts resolved so this can go in.

PR #161, opened 2022-10-26
Remove test APIS (FakeDoRetry, DebugGet, DebugPost) from client package

benhoyt:
Move them to export_test.go, which makes them only available for tests, which is what they're intended and used for. Strictly speaking this is a backwards-incompatible change, as these were exported, but it seems extremely unlikely anyone is using these test utilities.

In our README we say: "We try to never change the underlying API itself in a backwards-incompatible way, however, we may sometimes change the Go client in backwards-incompatible ways." I think this is one of those "sometimes".

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -251,18 +251,6 @@ var (
 	doTimeout = 5 * time.Second
 )
 
-// FakeDoRetry fakes the delays used by the do retry loop.
-func FakeDoRetry(retry, timeout time.Duration) (restore func()) {
-	oldRetry := doRetry
-	oldTimeout := doTimeout
-	doRetry = retry
-	doTimeout = timeout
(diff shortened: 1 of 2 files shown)


niemeyer (changes requested):
These functions were intended to be public so they can help packages that are using the client to test and debug, rather than being artifacts for the local client testing itself alone.

That said, the name or FakeDoRetry is awkward and might be worth changing. That sounds more like SetRetryDelay or similar.

benhoyt:
Fair. If these are (also) intended for use in tests of packages that use the client, it makes sense to leave them. And in that case, I don't think it's worth changing the name. I may tweak the doc comments slightly in https://github.com/canonical/pebble/pull/160. Closing.

PR #162, opened 2022-10-26
Add "Using Pebble" documentation to README

benhoyt:
This PR adds a large section of documentation to the README about how to use Pebble, and how some of its major features work. This is intended as a mini user manual; several people have asked about how these things work, and I've previously had to say "sorry, it's documented in the code". :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ designed with unique features that help with more specific use cases.
 
   - [General model](#general-model)
   - [Layer configuration examples](#layer-configuration-examples)
-  - [Running pebble](#running-pebble)
+  - [Using pebble](#using-pebble)
   - [Layer specification](#layer-specification)
   - [API and clients](#api-and-clients)
   - [Roadmap/TODO](#roadmap--todo)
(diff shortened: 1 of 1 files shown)


barrettj12 (commented):
This is a really nice addition. What do you think about putting it in a separate file though? This README is already getting quite bloated - I'd even support moving some of the stuff currently in the README to separate files.

README.md
```diff
@@ -118,26 +118,248 @@ services:
...
+
+Separate from the service manager, Pebble implements custom "health checks" that can be configured to restart services when they fail.
+
+Each check can be one of three types. The types and their success criteria are:
+
+* `http`: an HTTP `GET` request to the URL specified must return an HTTP 2xx status code
+* `tcp`: opening the given TCP port must be successful
+* `exec`: executing the specified command must yield a zero exit code
+
+Checks are configured in the layer configuration using the top-level field `checks`. Below is an example layer showing the three different types of checks:
```
tmihoc, line 238: This belongs in the layer configuration spec section. You can reference it from here. Or maybe have the generic spec there and a link to that + a concrete example here.

README.md
```diff
@@ -118,26 +118,248 @@ services:
         command: cmd
 ```
```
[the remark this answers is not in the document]
tmihoc: I see you actually do give the full spec starting with line 359. It should be higher up and the examples should be nested under it.

README.md
```diff
@@ -9,7 +9,7 @@ designed with unique features that help with more specific use cases.
 
   - [General model](#general-model)
   - [Layer configuration examples](#layer-configuration-examples)
-  - [Running pebble](#running-pebble)
+  - [Using pebble](#using-pebble)
   - [Layer specification](#layer-specification)
```
tmihoc, line 13: In the Table of Contents / the organization of the sections, please consider using Diataxis. That is, you should the 5 main sections below, with your existing sections nested within along the following lines (at least, at first glance):

- Home
- Tutorial
- How-to guides
    - Install (i.e., How to install Pebble)
    - Set up (i.e., How to set up the Pebble directory)
    - Use (i.e., How to use Pebble)
    - Run the daemon (i.e., How to run the Pebble daemon)
    - Start and stop services (i.e., How to start and stop Pebble services)
- Reference
    - General model
    - Layer configuration spec + examples
    - `pebble` CLI
        - List of `pebble` CLI commands
    - Service dependencies
    - Health checks
    - Changes and tasks
    - ...
- Explanation

At first glance, all the materials you have are either How-to guides or Reference, so you can leave out the Tutorial and Explanation for now. As for how to organize How-to guides vs. Reference: I'd put any material that defines things into Reference and any material that gives instructions into How-to guides. I'd also try to organize the Reference docs around primitive concepts and bits and then the How-to guides also around that where, for a product that has a CLI, the how-to guides are usually about manipulating a primitive entity in various ways using various CLI commands. See, for example, https://juju.is/docs/olm/the-juju-client and https://juju.is/docs/olm/manage-juju, or https://juju.is/docs/olm/network-spaces and https://juju.is/docs/olm/manage-spaces , or https://juju.is/docs/olm/action and https://juju.is/docs/olm/manage-actions .

tmihoc (commented):
I've made a bunch of comments, mainly regarding the structure.

README.md
```diff
@@ -9,7 +9,7 @@ designed with unique features that help with more specific use cases.
```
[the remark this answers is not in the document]
benhoyt: Thanks for looking at the doc as a whole, and I agree this could use improvement. However, the scope of this PR is just adding the "Using Pebble" documentation, so I'm going to leave this for another PR.

README.md
benhoyt: This is a good discussion, however, it's a big structural change to the rest of the document -- I'm going to leave this for future work.

README.md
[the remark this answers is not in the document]
benhoyt: Yeah, that's right, the YAML spec is under "Layer specification" in the format of https://juju.is/docs/sdk/metadata-yaml -- I put it below "Using Pebble" as I felt reference material should be after the general "using Pebble" section.

README.md
benhoyt: I realize there's a bit of duplication here, however, I think it's important to know the check types up-front. But I've added a phrase and link to the full spec below.

benhoyt:
> What do you think about putting it in a separate file though? This README is already getting quite bloated - I'd even support moving some of the stuff currently in the README to separate files.

@barrettj12 I thought about this, but on balance I think I prefer it in a single file. If anything I think we could split out "Using Pebble" into `manual.md` or something, and "Layer specification" into `layer-spec.md`. Also interested to hear @niemeyer's opinion here.

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+
+This will start the pebble daemon itself, as well as starting all the services that
+are marked as `startup: enabled` (if you don't want that, use `--hold`). Then
+other pebble commands may be used to interact with the running daemon, for example,
+in another terminal window.
+
+To override the default configuration directory, set the `PEBBLE` environment variable when running:
+
+```
+$ PEBBLE=~/pebble pebble run
```
niemeyer, line 151: I suggest breaking this down into two separate lines, first exporting the variable, and then calling the command. This way follow up commands would be able to work correctly as well, and it also avoids the cacophony of pebble pebble pebble.

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+2022-10-26T01:18:26.904Z [pebble] Started daemon.
+...
+```
+
+The daemon's service manager stores the most recent stdout and stderr from each service (using a 100KB ring buffer per service). These are accessible via the logs API or using `pebble logs`. If you want to also write service logs to Pebble's own stdout, run the daemon with `--verbose`:
```
niemeyer, line 156: This seems to be going too fast into debugging. Instead, we should be talking about the model: statuses, plan, changes, etc. Otherwise this smells like a boring process runner, instead of a nicely structured service support system.

To improve this, I suggest cutting off right before the run command output, continuing through the rest of the experience, and only later mentioning the ability to see logs.

Also, before we start/stop and change the state, let's talk about reading the current state that took place right after running.

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+
+To start or stop specific services, type `pebble start` or `pebble stop` followed by one or more service names, for example:
+
+```
+$ pebble start srv1 srv2  # start two services (and any dependencies)
+
+$ pebble stop srv1        # stop one service
+```
+
+When starting a service, Pebble records a "Start" [change](#changes-and-tasks) for the operation, executes the service's `command`, and waits 1 second to ensure the command doesn't exit too quickly. Assuming the command doesn't exit within 1 second, the start is considered successful, otherwise `pebble start` will exit with an error.
```
niemeyer, line 180: It seems premature to dive so deeply into the details of process management here. I'd reserve such details for an advanced section, and even then probably avoid promising things like exact timings as that can easily change.

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+The "Startup" column shows whether this service should be automatically started when Pebble starts ("enabled" means auto-start, "disabled" means don't auto-start).
+
+The "Current" column shows the current status of the service, and can be one of the following:
+
+* `active`: starting or running
+* `inactive`: not yet started, being stopped, or stopped
+* `backoff`: in a [backoff-restart loop](#service-auto-restart)
+* `error`: in an error state (currently: exited and exit action is "ignore")
+
+Implementation note: Pebble's service manager uses a [state machine](https://github.com/canonical/pebble/blob/a1f52239079e58cb6e3b6c7472ef5438b91eefab/internal/overlord/servstate/handlers.go#L67-L82) to manage the life cycle of a service. You can see these internal states and their transitions in the [state diagram](https://raw.githubusercontent.com/canonical/pebble/master/internal/overlord/servstate/state-diagram.svg).
```
niemeyer, line 206: Similar to the above point, we don't want to be documenting internal details of the implementation and pointing people to the code in a section that is intended for a new user that has never seen Pebble.

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+
+For example, if service `nginx` requires `logger`, `pebble start nginx` will start `logger` and then start `nginx`. Running `pebble stop logger` will stop `nginx` and then `logger`; however, running `pebble stop nginx` will only stop `nginx` (`nginx` depends on `logger`, not the other way around).
+
+If multiple dependencies need to be started at once, they're started in order according to the `before` and `after` configuration: `before` is a list of services that must be started before this one (but it doesn't `require` them). Or if it's easier to specify the other way around, `after` is a list of services that must be started after this one.
+
+If the configuration of `requires`, `before`, and `after` for a service results in a cycle or "loop", an error will be returned when attempting to start or stop the service.
+
+### Service auto-restart
+
+Pebble's service manager automatically restarts services that exit unexpectedly. By default, this is done whether the exit code is zero or non-zero, but you can change this using the `on-success` and `on-failure` fields in a configuration layer. The possible values for these fields are:
```
niemeyer, line 220: Do we explain configuration layers before talking about them?

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+
+```
+$ pebble exec --timeout 1s -- sleep 3
+error: cannot perform the following tasks:
+- exec command "sleep" (timed out after 1s: context deadline exceeded)
+```
+
+### File management
+
+Pebble provides various API calls and commands to manage files and directories on the server. The simplest way to use these is with the commands below, several of which should be familiar:
```
niemeyer, line 349: Some of these should be under a different header, such as "Container usage" with a nice intro, as they won't make much sense for someone using Pebble locally.

README.md
```diff
@@ -118,26 +118,250 @@ services:
...
+
+# TODO -- the following commands are coming soon
+$ pebble rm <path>              # remove a file or directory (like "rm")
+$ pebble push <local> <remote>  # copy file to server (like "cp")
+$ pebble pull <remote> <local>  # copy file from server (like "cp")
+```
 
 ## Layer specification
 
+Below is the full specification for a Pebble configuration layer. Layers are added statically using a file in `$PEBBLE/layers`, or dynamically via the layers API or `pebble add`.
```
niemeyer, line 363: Seems a bit weird to cover so many aspects of Pebble without even mentioning how to set it up at all. It's impossible to use Pebble before reading the section below.

We might have a trivial config close to the top, to enable experimentation with the remaining commands. Bonus points if the flow actually works for real (different headers build on what came before, and commands actually work).

niemeyer (changes requested):
It's a good idea to improve our docs, but the proposal feels a bit too dense and too shallow at the same time. Some points related to this:

README.md
benhoyt: We do: configuration layers are discussed in the second main section of the doc (that I presume you wrote originally), "Layer configuration examples".

README.md
benhoyt: Good point. I've added a new "Container usage" section with exec and file management as sub-headings of that.

README.md
benhoyt: See above -- we do lead with the "Layer configuration examples" section near the top of the doc. It gives a basic example and describes the replace vs merge override feature.

Re bonus points: I agree it'd be better if this was more real-world and flow-on. However, given that section is already in place, I'm going to leave that for a future PR. I think it's fine to make incremental progress here.

README.md
benhoyt: Good point - done.

README.md
benhoyt: I agree the logging stuff was going too deep (and tangential) here -- fixed. I've added a separate sub-section "Logs" at the bottom of the "Using Pebble" section. Regarding "statuses, plan, changes, etc": We talk about the plan/config in the first section as mentioned, and changes and tasks in their own section below.

README.md
benhoyt: I'm refactored this section a little, but I disagree that this is too much detail here. This README is what we have for a user manual right now, so I think it should explain it, not try to stuff important details in an "advanced" section.

I at first removed the exact timings, but the current timings are part of said user manual, and as such should be documented. I think it's important, otherwise if we say something like "waits a short time" users will be thinking, "Okay, but how long is short?"

README.md
benhoyt: Fair -- I've removed this paragraph.

README.md
niemeyer: Got it, thanks.

README.md
niemeyer: Thank you.

README.md
niemeyer: Sounds good, thank you.

niemeyer (approved):
Thanks for the changes. There's more work to do in the README, but I also wonder how much we should improve in until it becomes worthy of its own site.

PR #173, opened 2022-12-13
CI improvements

barrettj12:
- Disable fail-fast for unit tests, so that failing on one version of Go doesn't cancel the tests for all other versions.
- Upgrade actions/setup-go to v3 - as a result, we can now get the Go version from `go.mod`.
- Run unit tests on Go 1.18 and 1.19 as well.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,8 +11,9 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Nice -- works for me!

benhoyt:
Hmm, looks like the cache is failing for some reason. I haven't used that before. Is it a configuration issue, or would it be safer to disable it?

PR #175, opened 2023-01-04
Add --dry option to run command

anpep:
This option is useful for validating the Pebble layers without having to actually start the Pebble daemon.
Note that, in any case, pebble will exit. If the plan was valid, pebble will exit with an error code of 0; otherwise, with a non-zero exit code.

### DIFF
--- a/cmd/pebble/cmd_run.go
+++ b/cmd/pebble/cmd_run.go
@@ -138,11 +138,15 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 	if err != nil {
 		return err
 	}
-	if err := d.Init(); err != nil {
+	if err := d.Validate(); err != nil {
 		return err
 	}
 	if rcmd.Dry {
 		return nil
+	}
+
+	if err := d.Init(); err != nil {
+		return err
 	}
 
 	// Run sanity check now, if anything goes wrong with the
--- a/cmd/pebble/cmd_run.go
+++ b/cmd/pebble/cmd_run.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -51,7 +50,7 @@ func init() {
 	addCommand("run", shortRunHelp, longRunHelp, func() flags.Commander { return &cmdRun{} },
 		map[string]string{
 			"create-dirs": "Create pebble directory on startup if it doesn't exist",
-			"dry":         "Don't start the pebble daemon. This option is incompatible with --create-dirs",
+			"dry":         "Validate plan without starting the pebble daemon.",
 			"hold":        "Do not start default services automatically",
 			"http":        `Start HTTP API listening on this address (e.g., ":4000")`,
 			"verbose":     "Log all output from services to stdout",
@@ -61,10 +60,6 @@ func init() {
 func (rcmd *cmdRun) Execute(args []string) error {
 	if len(args) > 0 {
 		return ErrExtraArgs
-	}
-
-	if rcmd.CreateDirs && rcmd.Dry {
-		return errors.New("cannot run pebble: --create-dirs and --dry are mutually exclusive")
 	}
 
 	sigs := make(chan os.Signal, 2)
@@ -133,7 +128,6 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 	dopts := daemon.Options{
 		Dir:        pebbleDir,
 		SocketPath: socketPath,
-		Dry:        rcmd.Dry,
 	}
 	if rcmd.Verbose {
 		dopts.ServiceOutput = os.Stdout
@@ -146,6 +140,9 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 	}
 	if err := d.Init(); err != nil {
 		return err
+	}
+	if rcmd.Dry {
+		return nil
 	}
 
 	// Run sanity check now, if anything goes wrong with the
@@ -162,10 +159,6 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 	}
 	d.Version = cmd.Version
 	d.Start()
-
-	if rcmd.Dry {
-		return nil
-	}
 
 	watchdog, err := runWatchdog(d)
 	if err != nil {
--- a/cmd/pebble/cmd_run.go
+++ b/cmd/pebble/cmd_run.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


cmd/pebble/cmd_run.go
```diff
@@ -49,6 +51,7 @@ func init() {
 	addCommand("run", shortRunHelp, longRunHelp, func() flags.Commander { return &cmdRun{} },
 		map[string]string{
 			"create-dirs": "Create pebble directory on startup if it doesn't exist",
+			"dry":         "Don't start the pebble daemon. This option is incompatible with --create-dirs",
```
benhoyt, line 54: I'd prefer a help message that was a bit more direct ("Does X" rather than "Don't ..."). Perhaps "Dry run: validate plan without starting daemon"? Or would the argument better be called `--validate` or `--validate-only`, as it seems that's the intent, not really a dry *run*?

If it is "dry run", why is it incompatible (only) with `--create-dirs`? It seems like it's also incompatible with `--hold` and `--http` and `--verbose`, but we don't check for those. IMO we should just drop this check.

cmd/pebble/cmd_run.go
```diff
@@ -152,10 +160,13 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 		tic = time.NewTicker(checkRunningConditionsRetryDelay)
 		checkTicker = tic.C
 	}
-
 	d.Version = cmd.Version
 	d.Start()
 
+	if rcmd.Dry {
+		return nil
+	}
```
benhoyt, line 168: This seems a bit of an odd way to wire it up: by passing the flag through to the `Daemon`, then calling `Daemon.Start()`, but having it do nothing if dry is set. Why not just check `rcmd.Dry` after calling `Init`?

In fact, why not avoid passing the dry flag down at all? We could add a new `Daemon.Validate()` function which just checks `d.overlord.ServiceManager().Plan()`, and then call that directly from `runDaemon` if `rcmd.Dry` is set (and then exit). Otherwise call `Init()` (which could be updated to call `Validate`).

Note that this PR also has the side effect of making `pebble run` exit immediately if the existing plan is invalid. I think that's a good thing, but it is a behavioural change, so should probably be noted explicitly in the PR description.

internal/daemon/daemon.go
```diff
@@ -354,6 +357,13 @@ func logit(handler http.Handler) http.Handler {
 // Init sets up the Daemon's internal workings.
 // Don't call more than once.
 func (d *Daemon) Init() error {
+	if _, err := d.overlord.ServiceManager().Plan(); err != nil {
+		return err
+	}
```
benhoyt, line 362: Per above, these three lines could be moved to a `Validate` method, and the `dry` field removed.

benhoyt (changes requested):

internal/daemon/daemon.go
```diff
@@ -351,9 +353,21 @@ func logit(handler http.Handler) http.Handler {
 	})
 }
 
+// Validate performs the checks needed to ensure that the Pebble daemon can be started.
+func (d *Daemon) Validate() error {
+	if _, err := d.overlord.ServiceManager().Plan(); err != nil {
+		return err
+	}
+	return nil
```
benhoyt, line 361: Nit, but could just do:

```go
_, err := d.overlord.ServiceManager().Plan()
return err
```

internal/daemon/daemon.go
```diff
@@ -72,6 +72,8 @@ type Options struct {
 	// ServiceOuput is an optional io.Writer for the service log output, if set, all services
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
+
+	Dry bool
```
benhoyt, line 76: We don't need this anymore, right?

benhoyt (approved):
Couple of minor comments, but approved assuming those changes are mode.

cmd/pebble/cmd_run.go
```diff
@@ -139,6 +141,9 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 	if err := d.Init(); err != nil {
```
flotter, line 141: It feels a little bit weird to call Init() here, which creates all the socket listeners, only to return and exit knowing those will never be used.

Would it not be cleaner to call d.Validate() in runDaemon() before we call d.Init(), and only proceed to Init() if not --dry?

benhoyt: We've removed the `Dry` flag from the daemon, but I agree with your concern here. What about we explicitly call `d.Validate()` in `cmd_run.go` above the `d.Init()` call, then check `rcmd.Dry` after that, then call `d.Init()` (the plan is cached, so that's fine to do in terms of efficiency).

internal/daemon/daemon.go
```diff
@@ -351,9 +351,19 @@ func logit(handler http.Handler) http.Handler {
...
+// Validate performs the checks needed to ensure that the Pebble daemon can be started.
+func (d *Daemon) Validate() error {
+	_, err := d.overlord.ServiceManager().Plan()
+	return err
+}
+
 // Init sets up the Daemon's internal workings.
 // Don't call more than once.
 func (d *Daemon) Init() error {
+	if err := d.Validate(); err != nil {
```
flotter, line 363: If I read the code correctly, we will always perform all the validation steps twice for a normal run? It does not feel clean to me if I understand it correctly.

niemeyer: @anpep!?

anpep: That's correct. My opinion on that, as we discussed previously, is that a more elegant implementation would add too much complexity to the existing code, so I thought this is a good compromise. Will stay tuned to Gustavo's feedback once he reviews this PR. Thanks!

cmd/pebble/cmd_run.go
```diff
@@ -49,6 +50,7 @@ func init() {
 	addCommand("run", shortRunHelp, longRunHelp, func() flags.Commander { return &cmdRun{} },
 		map[string]string{
 			"create-dirs": "Create pebble directory on startup if it doesn't exist",
+			"dry":         "Validate plan without starting the pebble daemon.",
```
niemeyer, line 53: "Attempt to run without actually running"

This keeps it open ended, as we'll want to do more than just validate the plan there.

As a convention trivial, please note the lack of full stop on these sentences.

internal/daemon/daemon.go
```diff
@@ -351,9 +351,19 @@ func logit(handler http.Handler) http.Handler {
 	})
 }
 
+// Validate performs the checks needed to ensure that the Pebble daemon can be started.
+func (d *Daemon) Validate() error {
+	_, err := d.overlord.ServiceManager().Plan()
```
niemeyer, line 356: I agree with Fred's earlier comment, in that this is doing logic twice, but also this is a bit limited in what it can do. So in a way, it does too much, and too little at the same time.

What might be nice is to have an actual drymode that starts the overlord and all the managers, so all the little areas have a chance of attempting to run, and then shut it all down. We'll of course need to have a way to inform the managers that they're running dry, so they don't actually start work that could change the system.

Happy to discuss ideas here or elsewhere before you spent too much time on an uncertain avenue.

niemeyer (changes requested):

PR #176, opened 2023-01-09
Add boot-firmware command

anpep:
This PR adds support for the `boot-firmware` command. Which does some initialization prior to starting the Pebble daemon on Termus.

### DIFF
--- a/cmd/pebble/cmd_boot_firmware.go
+++ b/cmd/pebble/cmd_boot_firmware.go
@@ -1,0 +1,128 @@
+// Copyright (c) 2022 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+	"time"
+
+	"github.com/jessevdk/go-flags"
+
+	"github.com/canonical/pebble/client"
+	. "github.com/canonical/pebble/cmd"
+	"github.com/canonical/pebble/internal/boot"
+	"github.com/canonical/pebble/internal/daemon"
+	"github.com/canonical/pebble/internal/logger"
+	"github.com/canonical/pebble/internal/plan"
+)
+
+type cmdBootFirmware struct {
+	clientMixin
+
+	Force   bool `long:"force"`
+	Verbose bool `short:"v" long:"verbose"`
+}
+
+var bootFirmwareDescs = map[string]string{
+	"force":   `Skip all checks`,
+	"verbose": `Log all output from services to stdout`,
+}
+
+var shortBootFirmwareHelp = `Bootstrap a system with Pebble running as PID 1`
+
+var longBootFirmwareHelp = `
+The boot-firmware command performs checks on the running system, prepares the
+environment to get a working system and starts the Pebble daemon.
+`
+
+func (cmd *cmdBootFirmware) Execute(args []string) error {
+	if len(args) > 1 {
+		return ErrExtraArgs
+	}
+
+	if !cmd.Force {
+		if err := boot.CheckBootstrap(); err != nil {
+			return err
+		}
+	}
+
+	if err := boot.Bootstrap(); err != nil {
+		return err
+	}
+
+	t0 := time.Now().Truncate(time.Millisecond)
+
+	pebbleDir, socketPath := getEnvPaths()
+	if err := os.MkdirAll(pebbleDir, 0755); err != nil {
+		return err
+	}
+	if _, err := plan.ReadDir(pebbleDir); err != nil {
+		return err
+	}
+
+	dopts := daemon.Options{
+		Dir:        pebbleDir,
+		SocketPath: socketPath,
+	}
+	if cmd.Verbose {
+		dopts.ServiceOutput = os.Stdout
+	}
+
+	d, err := daemon.New(&dopts)
+	if err != nil {
+		return err
+	}
+	if err := d.Init(); err != nil {
+		return err
+	}
+
+	d.Version = Version
+	d.Start()
+
+	logger.Debugf("activation done in %v", time.Now().Truncate(time.Millisecond).Sub(t0))
+
+	servopts := client.ServiceOptions{}
+	changeID, err := cmd.client.AutoStart(&servopts)
+	if err != nil {
+		logger.Noticef("Cannot start default services: %v", err)
+	} else {
+		logger.Noticef("Started default services with change %s.", changeID)
+	}
+
+out:
+	for {
+		select {
+		case <-d.Dying():
+			logger.Noticef("Server exiting!")
+			break out
+		}
+	}
+
+	cmd.client.CloseIdleConnections()
+	if err := d.Stop(nil); err != nil {
+		return err
+	}
+
+	if err := syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART); err != nil {
+		return fmt.Errorf("cannot reboot: %w", err)
+	}
+	return nil
+}
+
+func init() {
+	addCommand("boot-firmware", shortBootFirmwareHelp, longBootFirmwareHelp, func() flags.Commander { return &cmdBootFirmware{} }, bootFirmwareDescs, nil)
+}
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -30,6 +30,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/canonical/pebble/client"
+	"github.com/canonical/pebble/internal/boot"
 	"github.com/canonical/pebble/internal/logger"
 )
 
@@ -43,11 +44,6 @@ var (
 	// set to logger.Panicf in testing
 	noticef = logger.Noticef
 )
-
-// defaultPebbleDir is the Pebble directory used if $PEBBLE is not set. It is
-// created by the daemon ("pebble run") if it doesn't exist, and also used by
-// the pebble client.
-const defaultPebbleDir = "/var/lib/pebble/default"
 
 type options struct {
 	Version func() `long:"version"`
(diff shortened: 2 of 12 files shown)


cmd/pebble/main.go
```diff
@@ -348,8 +344,15 @@ func run() error {
 		return fmt.Errorf("cannot create client: %v", err)
 	}
 
+	var xtra []string
 	parser := Parser(cli)
-	xtra, err := parser.Parse()
+	if err2 := boot.CheckBootstrap(); err2 == nil {
```
benhoyt, line 349: I didn't understand this at first, then I saw in the spec we want `boot-firmware` to be the default if no command is specified.

To help clarify, perhaps we can make this something other than `CheckBootstrap`. Perhaps `boot.IsTermus() bool` that just checks for the environment variable (or just check for the environment variable directly in main)?

internal/boot/bootstrap_termus.go
```diff
@@ -0,0 +1,59 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package boot
+
+import (
+	"errors"
+	"os"
+)
+
+var Getpid = os.Getpid
```
benhoyt, line 25: Nit: this shouldn't be capitalized/exported. Perhaps call it `getpid` or `osGetpid`. (it's an internal package, so it doesn't matter much, but we might as well be consistent).

internal/boot/bootstrap_termus_test.go
```diff
@@ -0,0 +1,95 @@
...
+// Hook up check.v1 into the "go test" runner
+func Test(t *testing.T) { TestingT(t) }
+
+func (s *bootstrapSuite) TestCheckBootstrap(c *C) {
+	err := CheckBootstrap()
+	c.Assert(err, ErrorMatches, "must run as PID 1. Use --force to suppress this check")
+}
+
+func (s *bootstrapSuite) TestCheckBootstrapPID1(c *C) {
+	Getpid = func() int { return 1 }
```
benhoyt, line 41: To avoid surprises, this should be returned to its old value at the end of the test with `defer func() { Getpid = os.Getpid }()` or probably better, by fetching the old value and setting it to that value in the defer, like you've done with the environment variable.

internal/boot/bootstrap_termus_test.go
```diff
@@ -0,0 +1,95 @@
...
+	err := CheckBootstrap()
+	c.Assert(err, ErrorMatches, "TERMUS environment variable must be set to 1. Use --force to suppress this check")
+}
+
+func (s *bootstrapSuite) TestCheckBootstrapPID1AndEnv(c *C) {
+	Getpid = func() int { return 1 }
```
benhoyt, line 47: Same as above, this should be reset at the end of the test.

internal/boot/bootstrap_termus_test.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (s *bootstrapSuite) TestBootstrapFails(c *C) {
+	var attemptedMounts []mount
+
+	oldMountImpl := MountImpl
+	defer func() { MountImpl = oldMountImpl }()
+	MountImpl = func(source string, target string, fstype string, flags uintptr, data string) error {
+		attemptedMounts = append(attemptedMounts, mount{source, target, fstype, flags, data})
+		if len(attemptedMounts) == 2 {
```
benhoyt, line 83: For simplicity in this failure test, I'd probably just avoid `attemptedMounts` and return an error on the first call.

internal/boot/util_termus.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+var (
+	MountImpl   = syscall.Mount
+	UnmountImpl = syscall.Unmount
```
benhoyt, line 28: These should not be exported. Perhaps name them `syscallMount` and so on?

internal/boot/util_termus_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+var _ = Suite(&utilSuite{})
+
+func buildMountImpl(err error) func(string, string, string, uintptr, string) error {
+	return func(string, string, string, uintptr, string) error {
+		return err
+	}
+}
+
+func buildUnmountImpl(err error) func(string, int) error {
```
benhoyt, line 36: IMO these functions don't add much value, and even obscure the code (they're really "buildMountErrorImpl"). Seems clearer just to inline the func literals in the test.

internal/boot/util_termus_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (s *utilSuite) TestMount(c *C) {
+	oldMountImpl := MountImpl
+	defer func() { MountImpl = oldMountImpl }()
+	MountImpl = buildMountImpl(nil)
+
+	m := &mount{"/dev/nvme0n1p3", "/boot", "ext4", 0, ""}
+	err := m.mount()
+	c.Assert(err, IsNil)
```
benhoyt, line 49: We're not really testing much here. Shouldn't we have an `attemptedMounts` slice we check like in the tests above? Do we even need these tests given that we're testing this in `Bootstrap`?

benhoyt (commented):
Thanks for this! I've left a bunch of comments, some minor, a few more major. My main question is whether we need the build tag at all (it seems to me removing it would simplify quite a bit).

cmd/pebble/cmd_boot_firmware_common_test.go
```diff
@@ -0,0 +1,48 @@
+//go:build !termus
+// +build !termus
```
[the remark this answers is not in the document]
benhoyt: Okay, I caught up with Gustavo about this, and he said he'd recommended a build tag a while ago, because of all the things that may be added to `boot-firmware` in future (potentially drivers and so on). That's fair enough.

cmd/pebble/cmd_boot_firmware.go
```diff
@@ -0,0 +1,128 @@
...
+
+type cmdBootFirmware struct {
+	clientMixin
+
+	Force   bool `long:"force"`
+	Verbose bool `short:"v" long:"verbose"`
+}
+
+var bootFirmwareDescs = map[string]string{
+	"force":   `Skip all checks`,
```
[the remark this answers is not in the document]
anpep: I think it can be useful for debugging purposes. Sometimes I find myself booting a firmware to a shell and run pebble interactively from there after inspecting the environment variables or some aspects from the live system.

cmd/pebble/cmd_boot_firmware.go
```diff
@@ -0,0 +1,128 @@
...
+	changeID, err := cmd.client.AutoStart(&servopts)
+	if err != nil {
+		logger.Noticef("Cannot start default services: %v", err)
+	} else {
+		logger.Noticef("Started default services with change %s.", changeID)
+	}
+
+out:
+	for {
+		select {
```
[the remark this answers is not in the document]
anpep: That would be a nice improvement, thanks!

cmd/pebble/cmd_boot_firmware_common_test.go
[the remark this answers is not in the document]
anpep: I say let's adopt this early and guide the architecture towards build tags, since other teams might also require it soon enough. I'm also not a fan of a pebble binary having logic that's essentially dead code for many targets, especially considering that Termus logic is particularly specific and irrelevant for the common Pebble use case.

cmd/pebble/main.go
anpep: I don't see the need to duplicate the bootstrap checks here: these checks are not expensive in terms of CPU time and also we should not do partial checks (i.e. checking *only* the environment variable) because we could be in a scenario in which a user invokes Pebble from a shell (and the `TERMUS` env var will be `1` since it's inherited down in the process tree), thus triggering the unwanted behavior of Pebble running the `boot-firmware` command by default.

I do agree that we should name it something more appropriate. Thanks for your feedback!

internal/boot/bootstrap_termus.go
```diff
@@ -0,0 +1,59 @@
+//go:build termus
```
flotter, line 1: How about just renaming this package bootstrap?

Then we could expose:

bootstrap.Validate()
bootstrap.Run()

cmd/pebble/cmd_boot_firmware.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) 2022 Canonical Ltd
```
flotter, line 1: If you have two variants of cmd_boot_firmware*, then the bootstrap package does not have to be termus specific, which feels nicer, as its very generic PID1 behaviour?

cmd/pebble/main.go
anpep: On second thought, after replacing all build tag boilerplate from `internal/boot` (now `internal/bootstrap`), it has become impossible to continue with this approach (because these symbols simply do not exist on the non-`termus` variant of Pebble). I've implemented your solution and would love to hear your opinion (:

internal/boot/bootstrap_termus.go
anpep: I suggest `bootstrap.Validate()` and `bootstrap.Do()` (since you don't "run the bootstrap", but you "do" the bootstrapping or "perform" the bootstrapping).

internal/boot/util_termus.go
```diff
@@ -0,0 +1,54 @@
+//go:build termus
```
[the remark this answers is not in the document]
anpep: The mountinfo/mount structs refer to different things even in the kernel API, so I think mixing those up would add up to some confusion at some point.

cmd/pebble/cmd_boot_firmware_common.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+var bootFirmwareDescs = map[string]string{
+	"force":   `Skip all checks`,
+	"verbose": `Log all output from services to stdout`,
+}
+
+var shortBootFirmwareHelp = `Bootstrap a system with Pebble running as PID 1`
+
+var longBootFirmwareHelp = `
```
flotter, line 40: There is a block of common code duplicated between *_common and *_termus. Would it not be cleaner and leaner to have a common file with only the help command stuff, and then two variants with different Executes ?

flotter: It also feels a bit strange that the common case describes a scenario in the help it does not support, in the same file.

anpep: At first I went with a similar approach but ended up with a lot of boilerplate and extra files with no real content. As far as I understand, it is a convention in Golang to prefer a bit of redundancy instead of more boilerplate.

anpep: Also, I don't think it is a good idea to have help messages change conditionally. The idea is not to treat the commands as different, but to treat the unsupported scenarios as run-time errors during the execution of the command that's "unique" to the user (even if it's two commands technically)

PR #179, opened 2023-01-12
allow Pebble to be packaged as a snap

cjdcordeiro:
## Motivation

Pebble has become a core component of multiple *craft tools (eg. [Rockcraft](https://github.com/canonical/rockcraft)).

In order to facilitate the installation of Pebble into such tools, we'd rather have a mechanism that allows for a clean pinpointing of the Pebble release/risk that is being installed and shipped with the tool

## Problem statement

While installing it from the source is a viable option (and what is [being done at the moment](https://github.com/canonical/rockcraft/blob/main/rockcraft/project.py#L491)), such an approach makes upgrade paths impractical and creates more friction when it comes to tracking security (it basically makes it harder for users to trust it).

## Proposal

While Pebble could be packaged as a DEB (or similar open-source artefact), a Snap seems to be the easiest, fastest, most automated and most familiar way to achieve it.

**In this PR:**
 - a `snapcraft.yaml` recipe is being added, to generate a classic Pebble snap
    - the classic confinement for Pebble has already been [accepted](https://forum.snapcraft.io/t/classic-confinement-for-pebble/33059)
    - the snap is currently owned by me (https://snapcraft.io/pebble). I'll make the request to move it to "Canonical" once this PR is merged
    - atm, the snap is graded as devel, and relying on `VERSION` and `mkversion.sh` to infer the Snap version
        - if we decide later on to promote the Snap to stable, we should talk about how to version it, and we can then change its `grade` to be dynamic in `snapcraft.yaml`
 - a new GitHub workflow for building the Pebble snap on every PR. This snap is only uploaded to the GH worfklow's execution. The automatic upload to the Snap store can be done once this PR is merged.

---

## Update

 - the snap's `grade` is also dynamic
 - the snap uses a Pebble wrapper script to handle the default `PEBBLE` directory when running `pebble`

### Next steps

Once this PR is merged:

 - I will connect the store builds with the canonical/pebble repository (atm, the [existing snap](https://snapcraft.io/pebble) is building automatically from my [fork](https://github.com/cjdcordeiro/pebble/))
 - I will [submit a request](https://forum.snapcraft.io/c/store-requests/19) for changing the ownership of this snap to "Canonical", keeping me, @niemeyer, @jnsgruk and @benhoyt (someone else?)
 - finally, I'll propose changes to the existing GH workflows to automate the promotion of this snap to `candidate` on every official release


cjdcordeiro:
It seems to me that some of the unit tests randomly fail from time to time, and I cannot re-trigger those jobs without re-committing, so I'll let the reviewers re-trigger the failed job manually.

benhoyt:
@cjdcordeiro Hmm, yes. I think we have a bit of work to do w.r.t. unit test reliability. I just re-ran the failed ones.

benhoyt:
@cjdcordeiro Just coming back to this. Assuming https://github.com/canonical/pebble/pull/204 gets merged, we can probably use GoReleaser to build the snap, as [GoReleaser can generate snaps](https://goreleaser.com/customization/snapcraft/).

cjdcordeiro:
@benhoyt thanks! Thank sounds like a perfect fit. I'll be watching that PR.

jnsgruk:
@benhoyt @cjdcordeiro indeed Goreleaser works quite well for snaps in cases like this one where the build is trivial - I've used it to quite good effect in the [parca](https://github.com/parca-dev/parca/blob/0a37cc966ca433914379a1a0cf0fd72b35357426/.goreleaser.yml#L36-L84) project for building their snap, including adding hooks and config etc. if it's a useful reference.

cjdcordeiro:
Yes I agree. And if not discussed yet, we should also discuss the lifecycle management of Pebble, i.e. how to version, when to release and where (eg. "on `tag` -> "promote" + "on push: main" -> "release to edge")

jnsgruk:
@cjdc this looks like it's good to me. Now we have the versioning set up, we can introduce the dynamic grade as you mention in the comments.

We should also consider a workflow that publishes the latest version to `edge` on merge to main,

cjdcordeiro:
@jnsgruk yep I'll need to revisit this. I think the goal would be to release the first semantic version once https://github.com/canonical/pebble/pull/191 is in. 

Based on https://github.com/canonical/pebble/pull/204, we need to:
 - decide if we should only release versioned snaps or, as you say, "publish the latest version to edge on merge to main",
 - change ownership of the Snap from me to `canonical`
 - add snap tests

I'll try to include this for the next Pulse, and in the meantime get some of these discussions ongoing.

.github/workflows/snap.yml
```diff
@@ -0,0 +1,23 @@
+name: Pebble snap
+
+on: [pull_request]
```
benhoyt, line 3: It seems like this should happen on push to master and on new release, rather than on pull request. Maybe this?

```
on:
  push:
    branches: [master]
  release:
    types: [published]
```

.github/workflows/snap.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      - name: Build Pebble Snap
+        id: build-pebble-snap
+        uses: snapcore/action-build@v1
+
+      - name: Attach pebble snap to GH workflow execution
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ steps.build-pebble-snap.outputs.snap }}
+          path: ${{ steps.build-pebble-snap.outputs.snap }}
```
benhoyt, line 23: Will we be adding `snapcore/action-publish` to publish it to Charmhub too?

snap/local/pebble-wrapper
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -eu
+
+# The default Pebble directory should sit within the $SNAP_USER_DATA because:
+# - access to the system's /var/lib/pebble/default isn't guaranteed, and
+# - the default $PEBBLE path should also be removed when removing the snap.
+DEFAULT_PEBBLE_DIRECTORY="${SNAP_USER_DATA}/var/lib/pebble/default"
```
benhoyt, line 6: Does it need to be under `/var/lib/pebble/default` *within* the Snap user data directory? What about just `DEFAULT_PEBBLE_DIRECTORY="${SNAP_USER_DATA}"`?

This would also avoid the need for the user to create the directory or use `--create-dirs`, because (I presume) Snap ensures it exists.

snap/local/pebble-wrapper
```diff
@@ -0,0 +1,11 @@
...
+# We don't create this directory because the user can execute Pebble
+# with --create-dirs or even choose a different directory.
+export PEBBLE="${PEBBLE:-$DEFAULT_PEBBLE_DIRECTORY}"
```
benhoyt, line 9: I don't know too much about snaps and wasn't part of the previous discussion, but presumably we have to do it with this wrapper script instead of using `environment` in the `snapcraft.yaml` so that we can set `PEBBLE` but only if it hasn't already been set by the user?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+parts:
+  pebble:
+    plugin: go
+    source: .
+    build-snaps:
+      - go
+    build-environment:
+      - CGO_ENABLED: 0
+      - GOFLAGS: -ldflags=-w -ldflags=-s
```
benhoyt, line 40: FWIW, we use `-trimpath -ldflags='-s -w'` in the GitHub Actions that builds the official binaries, so probably best to replicate that here.

I'm curious, how does this know to build the `./cmd/pebble` package?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  pebble-release-data:
+    plugin: nil
+    source: .
+    override-build: |
+      version="$(cat cmd/VERSION || ./cmd/mkversion.sh --output-only)"
```
benhoyt, line 46: I presume the built Pebble binary is not yet available? If so we could use `pebble version --client`.

snap/snapcraft.yaml
benhoyt: Alternatively I suppose we could download one of the official binaries, now that those are being built?

.github/workflows/snap.yml
cjdcordeiro: this specific action, atm, is purely a test for making sure the snapcraft.yaml is sane and that the snap can be successfully built. after all, we shouldn't merge changes if those are going to make the store builds fail.

the build on `push: master` will happen from the store, and on `release` will be implemented separately later on (see "Next Steps" above)

.github/workflows/snap.yml
cjdcordeiro: To Charmhub or the snap store? The latter will be done from the store itself, and promotions will be done here (coming in a future PR)

snap/local/pebble-wrapper
cjdcordeiro: Yes that's it. The problem with the `app.<name>.environment` is that (very much like a Pebble service) it will not be a default, but rather enforce an environment variable for the app. So you wouldn't be able to change it via the parent's environment.

snap/snapcraft.yaml
cjdcordeiro: Good catch. Thanks, I'll add `-trimpath` to the GOFLAGS (see https://github.com/canonical/pebble/pull/179/commits/c49bb04a426edf10009b5b62c4ea4f0521a4c177). The `go` plugin installs for `./...` ([ref](https://github.com/snapcore/snapcraft/blob/0594429b75bfc663a59f583e42639239c1568e0a/snapcraft_legacy/plugins/v2/go.py#L79)).

Wrt to taking the binaries instead of re-compiling...I'd need to think about it. I'd say it's not possible atm the moment, mainly for 2 reasons:
 1. the snap is built by the store, automatically. We can't customize (afaik) the triggering conditions for that store-side build, which means we can't guarantee the Pebble binary has already been built whenever the store build takes place
 2. the pebble binaries are only released to a predictable place `on: release`, otherwise they are only uploaded as a build artifact of the GH action, which means: 1) it would be cumbersome for snapcraft.yaml to infer where to get that binary from, and 2) those binaries will only be there for a limited amount of time (the default retention time is 90 days)

snap/local/pebble-wrapper
cjdcordeiro: This is possible and an easy change. The only thing I'm not too comfortable with is the user-facing information. I.e. `pebble help` will print

```
Set the PEBBLE environment variable to override the configuration directory 
(which defaults to /var/lib/pebble/default). Set PEBBLE_SOCKET to override 
the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).
```

So that's why keeping the `/var/lib/pebble/default` as a path suffix sounded more natural to me, despite it requiring the explicit creation of said path (which is actually what you'd expect if you were installing Pebble from source or via the existing release binaries).

So having said that, if you still think it would be better to just have `$SNAP_USER_DATA`, I'll change it.

snap/snapcraft.yaml
cjdcordeiro: Yes we can. That's better indeed. Thanks. Applied in https://github.com/canonical/pebble/pull/179/commits/c21e9c5cbad88164f677f9923190f2e7efe1c573

cjdcordeiro:
UPDATE: I've added one more job to the snap tests, to make sure the Pebble snap can be installed and to add a black box test.

.github/workflows/snap.yml
benhoyt: Got it. Yeah, sorry, I meant the Snap store. I live in the charm world. :-)

snap/local/pebble-wrapper
benhoyt: I do think it'd be better, but I see your concern. How about we tweak the help message while we're at it. Something like this (with better wrapping):

```
Set the PEBBLE environment variable to override the configuration directory (which
defaults to /var/lib/pebble/default, or to SNAP_USER_DATA if being used via the Snap).
Set PEBBLE_SOCKET to override the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).
```

Or perhaps say `or to $HOME/snap/pebble/<rev> if being used via the Snap` -- I'm not sure which is better from a Snap user point of view.

snap/snapcraft.yaml
benhoyt: Yep, that's fair -- thanks

.github/workflows/snap.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Install the Pebble snap
+        run: |
+          # Install the Pebble snap from the artifact built in the previous job
+          sudo snap install --dangerous --classic ${{ needs.build.outputs.pebble-snap }}
+
+          # Make sure Pebble is installed
+          pebble version
+
+      - name: Run sanity test
```
jnsgruk, line 44:
```suggestion
      - name: Run smoke test
```

Per inclusive naming guidelines

jnsgruk (approved):

.github/workflows/snap.yml
cjdcordeiro: thanks ;)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,61 @@
...
+      craftctl default
+
+  pebble-release-data:
+    plugin: nil
+    source: .
+    override-build: |
+      version="$(${CRAFT_STAGE}/bin/pebble version --client)"
+
+      # Set snap version
+      snapcraftctl set-version "$version"
```
flotter, line 51: Is there a reason you prefer using snapcraftctl here (and further down)?

```craftctl set version="$version"```

flotter (approved):

snap/snapcraft.yaml
cjdcordeiro: thanks for noticing @flotter . `craftctl` is indeed preferred. I've changed it.

cjdcordeiro:
fyi @jnsgruk to transfer the ownership of this snap to Canonical, at least 2 requirements must be met:
 1. multiple contributors, since Canonical-owned snaps must have a strong support commitment. This seems to be the case as the idea is to have @niemeyer @jnsgruk @benhoyt and @cjdcordeiro as maintainers
 2. have a stable release. This is not yet the case. So we can either wait for this PR to be merged and then we promote the snap, or I can promote it now, but it is still based on my fork

jnsgruk:
> fyi @jnsgruk to transfer the ownership of this snap to Canonical, at least 2 requirements must be met:
> 
>  1. multiple contributors, since Canonical-owned snaps must have a strong support commitment. This seems to be the case as the idea is to have @niemeyer @jnsgruk @benhoyt and @cjdcordeiro as maintainers
> 
>  2. have a stable release. This is not yet the case. So we can either wait for this PR to be merged and then we promote the snap, or I can promote it now, but it is still based on my fork

Ack. I think this is good to merge now, and we could push the 1.2.0 release into stable once it is.

@niemeyer thoughts?

jnsgruk (approved):

PR #184, opened 2023-01-25
[JUJU-1887] Introduces service grace periods for shutdown and pebble shut down signals.

tlm:
- This commit adds a new cmd arg to pebble run that allows the user to specify the behaviour to take place when Pebble receives sigterm. The two supported values are ignore and terminate.

  This change is needed for Kubernetes and Juju to address bug lp1951415. When Kubernetes decides to remove a pod from the cluster to reschedule else where it sends a sigterm to all containers in the pod at once. Having Pebble ignore this sigterm and giving the charm a chance to run it's tear down hooks to completion before the workload under control exits is critical for the day to day operations of the charm.

- The second change in this commit adds a new api endpoint to Pebble called shutdown. If Pebble is ignoring sigterm messages from controlling applications such as Kubernetes then we need a way to signal to Pebble to shutdown.

- Third change is the ability to define a grace period for services and how long they receive to handle SIGTERM and shut down before being killed by Pebble.

This work is being proposed as part of the overall fix in Juju for https://bugs.launchpad.net/juju/+bug/1951415

Testing:

The easiest way to perform manual testing against this change is to create a pebble service definition and run this with `pebble run --sigterm=ignore`. From there you should be able to send term to the process and watch the log output indicate that it is ignoring the signal. `kill -s TERM <pid>`

### DIFF
--- a/cmd/pebble/cmd_run.go
+++ b/cmd/pebble/cmd_run.go
@@ -19,10 +19,13 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
+	"github.com/canonical/x-go/strutil"
 	"github.com/jessevdk/go-flags"
+	"golang.org/x/sys/unix"
 
 	"github.com/canonical/pebble/client"
 	"github.com/canonical/pebble/cmd"
@@ -39,20 +42,39 @@ The run command starts pebble and runs the configured environment.
 type cmdRun struct {
 	clientMixin
 
-	CreateDirs bool   `long:"create-dirs"`
-	Hold       bool   `long:"hold"`
-	HTTP       string `long:"http"`
-	Verbose    bool   `short:"v" long:"verbose"`
+	CreateDirs      bool   `long:"create-dirs"`
+	Hold            bool   `long:"hold"`
+	HTTP            string `long:"http"`
+	ShutdownSignals string `long:"shutdown-signals" default:"INT,QUIT,TERM"`
+	Verbose         bool   `short:"v" long:"verbose"`
 }
 
 func init() {
 	addCommand("run", shortRunHelp, longRunHelp, func() flags.Commander { return &cmdRun{} },
 		map[string]string{
-			"create-dirs": "Create pebble directory on startup if it doesn't exist",
-			"hold":        "Do not start default services automatically",
-			"http":        `Start HTTP API listening on this address (e.g., ":4000")`,
-			"verbose":     "Log all output from services to stdout",
+			"create-dirs":      "Create pebble directory on startup if it doesn't exist",
+			"hold":             "Do not start default services automatically",
+			"http":             `Start HTTP API listening on this address (e.g., ":4000")`,
+			"shutdown-signals": "Signals that trigger shutdown",
+			"verbose":          "Log all output from services to stdout",
 		}, nil)
+}
+
+func processSignalNames(signalNames string) ([]os.Signal, error) {
+	sigs := strutil.CommaSeparatedList(signalNames)
+	signals := make([]os.Signal, 0, len(sigs))
+	for _, userSig := range sigs {
+		signal := strings.ToUpper(userSig)
+		if !strings.HasPrefix(signal, "SIG") {
+			signal = "SIG" + signal
+		}
+		sig := unix.SignalNum(signal)
+		if sig == 0 { /* error */
+			return signals, fmt.Errorf("unrecognised signal %q", userSig)
+		}
+		signals = append(signals, sig)
+	}
+	return signals, nil
 }
 
 func (rcmd *cmdRun) Execute(args []string) error {
@@ -60,8 +82,22 @@ func (rcmd *cmdRun) Execute(args []string) error {
 		return ErrExtraArgs
 	}
 
+	// We purposely ignore certain signals here. This is so the default go
+	// handlers don't process the signals. They may be added back in to our
+	// signal handler via the user provided ShutdownSignals.
+	signal.Ignore(syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM)
+	shutdownSignals, err := processSignalNames(rcmd.ShutdownSignals)
+	if err != nil {
+		return err
+	}
+
 	sigs := make(chan os.Signal, 2)
-	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	if len(shutdownSignals) != 0 {
+		// We don't want to call Notify with an empty slice of signals. Doing
+		// this will cause us to receive every signal when the user has asked
+		// us to ignore them all.
+		signal.Notify(sigs, shutdownSignals...)
+	}
 
 	if err := runDaemon(rcmd, sigs); err != nil {
 		if err == daemon.ErrRestartSocket {
--- a/internal/daemon/api_shutdown.go
+++ b/internal/daemon/api_shutdown.go
@@ -1,0 +1,27 @@
+// Copyright (c) 2023 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package daemon
+
+import (
+	"net/http"
+)
+
+// v1PostShutdown is a handler for initiating a Pebble shutdown via the daemon.
+func v1PostShutdown(c *Command, r *http.Request, _ *userState) Response {
+	// We call tomb.Kill as calling stop on the daemon will block and create a
+	// dead lock condition between this handler and stop.
+	c.d.tomb.Kill(nil)
+	return SyncResponse(nil)
+}
--- a/internal/overlord/servstate/export_test.go
+++ b/internal/overlord/servstate/export_test.go
@@ -72,11 +72,13 @@ func FakeOkayWait(wait time.Duration) (restore func()) {
 	}
 }
 
-func FakeKillWait(kill, fail time.Duration) (restore func()) {
-	old1, old2 := killWait, failWait
-	killWait, failWait = kill, fail
+// FakeGraceKillWait changes both the defaultGracePeriod and killWaitDuration
+// respectively for testing purposes.
+func FakeGraceKillWait(gracePeriod, killWait time.Duration) (restore func()) {
+	old1, old2 := defaultGracePeriod, killWaitDuration
+	defaultGracePeriod, killWaitDuration = gracePeriod, killWait
 	return func() {
-		killWait, failWait = old1, old2
+		defaultGracePeriod, killWaitDuration = old1, old2
 	}
 }
 
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -54,9 +54,15 @@ func TaskServiceRequest(task *state.Task) (*ServiceRequest, error) {
 }
 
 var (
-	okayWait = 1 * time.Second
-	killWait = 5 * time.Second
-	failWait = 10 * time.Second
+	// defaultGracePeriod is the duration afforded to services for processing
+	// SIGTERM signals and shutting down cleanly if the service hasn't specified
+	// their own duration.
+	defaultGracePeriod = 5 * time.Second
+	okayWait           = 1 * time.Second
+
+	// killWaitDuration is the duration given to services for shutting down when
+	// Pebble sends a SIGKILL signal.
+	killWaitDuration = 5 * time.Second
 )
 
 const (
@@ -592,6 +598,17 @@ func (s *serviceData) sendSignal(signal string) error {
 	return nil
 }
 
+// gracePeriod reports the duration that this service should be given when being
+// asked to shutdown gracefully before being force terminated. The value
+// returned will either be the services pre configured value or the default
+// grace period for pebble.
+func (s *serviceData) gracePeriod() time.Duration {
+	if s.config.GracePeriod.IsSet {
+		return s.config.GracePeriod.Value
+	}
+	return defaultGracePeriod
+}
+
 // stop is called to stop a running (or backing off) service.
 func (s *serviceData) stop() error {
 	s.manager.servicesLock.Lock()
@@ -606,7 +623,7 @@ func (s *serviceData) stop() error {
 			logger.Noticef("Cannot send SIGTERM to process: %v", err)
 		}
 		s.transition(stateTerminating)
-		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
+		time.AfterFunc(s.gracePeriod(), func() { logError(s.terminateTimeElapsed()) })
 
 	case stateBackoff:
 		logger.Noticef("Service %q stopped while waiting for backoff", s.config.Name)
@@ -654,8 +671,9 @@ func (s *serviceData) terminateTimeElapsed() error {
 		if err != nil {
 			logger.Noticef("Cannot send SIGKILL to process: %v", err)
 		}
+
 		s.transitionRestarting(stateKilling, s.restarting)
-		time.AfterFunc(failWait-killWait, func() { logError(s.killTimeElapsed()) })
+		time.AfterFunc(killWaitDuration, func() { logError(s.killTimeElapsed()) })
 
 	default:
 		// Ignore if timer elapsed in any other state.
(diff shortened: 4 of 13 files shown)


benhoyt:
@cjdcordeiro Thanks for the pointer to Docker's `--stop-signal` ([docs here](https://docs.docker.com/engine/reference/commandline/run/#-stop-container-with-signal---stop-signal)). Yeah, I think that's a nice design. Just like Docker, our default is SIGTERM. From their docs:

> The --stop-signal flag sets the system call signal that will be sent to the container to exit. This signal can be a signal name in the format SIG<NAME>, for instance SIGKILL, or an unsigned number that matches a position in the kernel’s syscall table, for instance 9.
>
> The default is SIGTERM if not specified.

Pebble calls this "shutdown" (for example, the "shutdown" action in the Pebble config), and the term "stop signal" in Pebble's case would more mean the signal Pebble sends to services to stop them, which is a different thing. So I suggest we call this `--shutdown-signal` and have it default to `SIGTERM` (allow names or numbers).

To disable (as Tom wants here), we could just set it to an arbitrary signal like `--shutdown-signal=SIGUSR1`, or perhaps better, allow `--shutdown-signal=none` as a special case.

client/pebble.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2023 Canonical Ltd
```
benhoyt, line 1: 2023, nice! :-)

IMO we should name this `shutdown.go`. We've tried to avoid using the name "Pebble" itself in contexts like this, as everything is Pebble here, after all. So IMO `shutdown.go`, but we could think of a broader name (or just call it shutdown.go now, and wait until there are multiple things that apply to think of a better name.

client/pebble.go
```diff
@@ -0,0 +1,27 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package client
+
+// ShutdownOptions specifies the options struct to use when command Pebble to
```
benhoyt, line 17: Grammar nit: "when command Pebble to shutdown" reads oddly. What about just "when calling Shutdown".

client/pebble.go
```diff
@@ -0,0 +1,27 @@
...
+// shutdown. Currently no options are available. This struct is left here for
+// future proofing.
+type ShutdownOptions struct {
+}
+
+// Shutdown issues a call to Pebble instructing Pebble to shutdown.
+func (client *Client) Shutdown(opts *ShutdownOptions) error {
+	_, err := client.doSync("POST", "/v1/shutdown", nil, nil, nil, nil)
+	return err
+}
```
benhoyt, line 27: I think we should have a `shutdown_test.go` too. It'll be ultra barebones, but just use the same style as the other tests and ensure the `/v1/shutdown` API call is sent.

cmd/pebble/cmd_run.go
```diff
@@ -31,6 +31,12 @@ import (
 	"github.com/canonical/pebble/internal/systemd"
 )
 
+const (
+	defaultSigTimeout      = 10 * time.Second
```
benhoyt, line 35: Looks like `defaultSigTimeout` is now unused?

cmd/pebble/cmd_run.go
```diff
@@ -39,10 +45,11 @@ The run command starts pebble and runs the configured environment.
...
 	clientMixin
 
-	CreateDirs bool   `long:"create-dirs"`
-	Hold       bool   `long:"hold"`
-	HTTP       string `long:"http"`
-	Verbose    bool   `short:"v" long:"verbose"`
+	CreateDirs    bool   `long:"create-dirs"`
+	Hold          bool   `long:"hold"`
+	HTTP          string `long:"http"`
+	SigtermAction string `long:"sigterm"`
```
benhoyt, line 51: Consider changing to `--shutdown-signal=SIGFOO` as per @cjdcordeiro's comment. Can discuss what you think of that at standup.

internal/daemon/api_pebble_test.go
```diff
@@ -0,0 +1,42 @@
...
+	c.Assert(err, IsNil)
+	rsp := v1PostShutdown(shutdownCmd, req, nil).(*resp)
+	rec := httptest.NewRecorder()
+	rsp.ServeHTTP(rec, req)
+	c.Assert(rec.Code, Equals, 200)
+	c.Assert(rsp.Status, Equals, 200)
+	c.Assert(rsp.Type, Equals, ResponseTypeSync)
+
+	_, open := <-daemon.tomb.Dying()
+	fmt.Println(open)
```
benhoyt, line 40: Leftover debugging?

cmd/pebble/cmd_run.go
hpidcock: I think I agree. But I am concerned `--shutdown-signal=SIGFOO` reads different, in the docker case, --stop-signal is the signal which docker sends to the container to stop it. In this case, pebble does not send SIGFOO when it shuts down but rather listens for SIGFOO to trigger the shutdown.

I am concerned also that the signals could then used for multiple actions. For example, in future we may add `--restart-signal=SIGFOO`, when both restart-signal and shutdown-signal use SIGFOO, what is the outcome of that?
`--sigterm=shutdown`, `--sigint=restart`, is less ambiguous/less prone to undefined behaviour.

So now after writing this, I think I've convinced myself that perhaps the way it is now is the better way.

cmd/pebble/cmd_run.go
```diff
@@ -60,6 +68,15 @@ func (rcmd *cmdRun) Execute(args []string) error {
 		return ErrExtraArgs
 	}
 
+	if rcmd.SigtermAction == "" {
+		rcmd.SigtermAction = sigtermActionTerminate
```
hpidcock, line 72: I think `--sigterm=shutdown` is better than `--sigterm=terminate`.

cmd/pebble/cmd_run.go
```diff
@@ -180,6 +197,10 @@ out:
 	for {
 		select {
 		case sig := <-ch:
+			if sig == syscall.SIGTERM && rcmd.SigtermAction == sigtermActionIgnore {
```
hpidcock, line 200: This will probably need more in the future (if we want to respond to other signals) but this is ok for now.

internal/daemon/api_pebble.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"net/http"
+)
+
+// v1PostShutdown is a handler for initiating a Pebble shutdown via the daemon.
+func v1PostShutdown(c *Command, r *http.Request, _ *userState) Response {
+	// We call tomb.Kill as calling stop on the daemon will block and create a
+	// dead lock condition between this handler and stop.
+	c.d.tomb.Kill(nil)
```
hpidcock, line 25: I think we need a log here with the shutdown reason?

hpidcock (commented):
Thanks for this.

cmd/pebble/cmd_run.go
tlm: I've already done the work to move it to the docker method. See updates. What is the decision @benhoyt @hpidcock ?

cmd/pebble/cmd_run.go
```diff
@@ -36,34 +37,92 @@ var longRunHelp = `
...
-			"verbose":     "Log all output from services to stdout",
+			"create-dirs":      "Create pebble directory on startup if it doesn't exist",
+			"hold":             "Do not start default services automatically",
+			"http":             `Start HTTP API listening on this address (e.g., ":4000")`,
+			"shutdown-signals": "Symbolic signal(s) pebble listens to for shutting down. Defaults to TERM,INT & QUIT ",
+			"verbose":          "Log all output from services to stdout",
 		}, nil)
 }
 
+func processShutdownSignals(rawUserSignals string) ([]os.Signal, error) {
```
benhoyt, line 77: Naming nit: would `signalNames` be clearer than "raw user" signals (a UeSR signal is something else)?

cmd/pebble/cmd_run.go
```diff
@@ -36,34 +37,92 @@ var longRunHelp = `
...
+}
+
 func (rcmd *cmdRun) Execute(args []string) error {
 	if len(args) > 0 {
 		return ErrExtraArgs
 	}
 
+	var err error
+	shutdownSignals := defaultShutdownSignals
+	if rcmd.ShutdownSignals != "" {
```
benhoyt, line 110: Once we're using `default:`, how about we remove this `if` so that empty string is supported and means "no shutdown signals"?

cmd/pebble/cmd_run.go
```diff
@@ -36,34 +37,92 @@ var longRunHelp = `
...
+	}
+
 	sigs := make(chan os.Signal, 2)
-	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	// We always want to watch for defaultShutdownSignals otherwise default go
+	// handlers will watch them for us. By appending defaultShutdownSignals we
+	// run the risk of doubling up the signals from the user and our default
+	// collection. This doesn't matter to notify and the list is small so we
+	// don't worry about it.
+	signal.Notify(sigs, append(shutdownSignals, defaultShutdownSignals...)...)
```
benhoyt, line 123: Instead of this, can we make use of `signal.Ignore`, either `signal.Ignore()` to ignore all, then call `signal.Notify(...)`, or just `signal.Ignore(SIGINT, SIGQUIT, SIGTERM)` and then Notify? I haven't tested this, but that seems cleaner than doubling up and having to test below in the handler.

cmd/pebble/cmd_run.go
```diff
@@ -180,8 +239,13 @@ out:
...
 		case sig := <-ch:
-			logger.Noticef("Exiting on %s signal.\n", sig)
-			break out
+			for _, shutdownSig := range shutdownSignals {
+				if shutdownSig.String() == sig.String() {
+					logger.Noticef("Exiting on %s signal.\n", sig)
+					break out
+				}
+			}
+			logger.Noticef("Signal %s is not a shut down signal. Ignoring.\n", sig)
```
benhoyt, line 248: I'm not sure this should be logged, as they didn't ask for it. Or maybe `logger.Debugf`.

cmd/pebble/cmd_run.go
```diff
@@ -36,34 +37,92 @@ var longRunHelp = `
...
+			"http":             `Start HTTP API listening on this address (e.g., ":4000")`,
+			"shutdown-signals": "Symbolic signal(s) pebble listens to for shutting down. Defaults to TERM,INT & QUIT ",
+			"verbose":          "Log all output from services to stdout",
 		}, nil)
 }
 
+func processShutdownSignals(rawUserSignals string) ([]os.Signal, error) {
+	sigs := strings.Split(rawUserSignals, ",")
+	signals := make([]os.Signal, len(sigs))
+	i := 0
```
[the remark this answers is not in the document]
tlm: ah good pickup. I was going for a different algorithm then change my mind.

cmd/pebble/cmd_run.go
tlm: Think we have this now.

internal/daemon/api_shutdown.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"net/http"
+
+	"github.com/canonical/pebble/internal/logger"
+)
+
+// v1PostShutdown is a handler for initiating a Pebble shutdown via the daemon.
+func v1PostShutdown(c *Command, r *http.Request, _ *userState) Response {
+	logger.Noticef("received shut down api request, proceeding to shut down daemon")
```
benhoyt, line 25: Nit: we've tried to be careful to keep log message formatting consistent, starting log messages with a capital letter (unlike errors), and using proper capitalization like "API".

However, I don't think we actually need this log, as all API requests are logged. So even without this we get something like this (the `POST /v1/shutdown` line is the one I'm referring to):

```
2023-02-06T23:03:57.176Z [pebble] POST /v1/shutdown 50.11µs 200
2023-02-06T23:03:57.176Z [pebble] Server exiting!
2023-02-06T23:03:57.201Z [pebble] Stopping all running services.
2023-02-06T23:03:57.231Z [pebble] Service "srv1" stopped
```

internal/overlord/servstate/handlers.go
```diff
@@ -606,7 +606,12 @@ func (s *serviceData) stop() error {
 			logger.Noticef("Cannot send SIGTERM to process: %v", err)
 		}
 		s.transition(stateTerminating)
-		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
+
+		gracePeriod := killWait
```
benhoyt, line 610: Assuming we're calling the plan field "grace-period", why don't we rename this `defaultGracePeriod` to match?

internal/overlord/servstate/handlers.go
```diff
@@ -654,6 +659,7 @@ func (s *serviceData) terminateTimeElapsed() error {
 		if err != nil {
 			logger.Noticef("Cannot send SIGKILL to process: %v", err)
 		}
+
 		s.transitionRestarting(stateKilling, s.restarting)
 		time.AfterFunc(failWait-killWait, func() { logError(s.killTimeElapsed()) })
```
benhoyt, line 664: We'll need to also update the other uses of `killWait` (there's one here and also in `checkFailed`). Perhaps we can have a `s.gracePeriod()` method that gets the plan grace-period or default if not set, to keep that logic in one place?

We also need to change how the `failWait` constant works. It should probably be an additional duration (5 seconds) after the grace period, rather than being the total time (you might have to update the tests). Maybe call it something meaningful for the new usage, like `timeBeforeKill`?

internal/plan/plan.go
```diff
@@ -59,12 +59,13 @@ type Layer struct {
...
-	Startup     ServiceStartup `yaml:"startup,omitempty"`
-	Override    Override       `yaml:"override,omitempty"`
-	Command     string         `yaml:"command,omitempty"`
+	Name        string           `yaml:"-"`
+	Summary     string           `yaml:"summary,omitempty"`
+	Description string           `yaml:"description,omitempty"`
+	Startup     ServiceStartup   `yaml:"startup,omitempty"`
+	Override    Override         `yaml:"override,omitempty"`
+	Command     string           `yaml:"command,omitempty"`
+	GracePeriod OptionalDuration `yaml:"grace-period,omitempty"`
```
benhoyt, line 68: `GracePeriod` should probably not be under the "Basic details" section. Move it to the "Auto-restart and backoff functionality" section?

internal/plan/plan.go
```diff
@@ -130,6 +131,9 @@ func (s *Service) Merge(other *Service) {
 	if other.Command != "" {
 		s.Command = other.Command
 	}
+	if other.GracePeriod.IsSet {
+		s.GracePeriod = other.GracePeriod
+	}
```
benhoyt, line 136: Thanks for catching this and adding it to `Merge`!

benhoyt (changes requested):
Thanks: re-reviewed and added a few more comments. I think there's a few things we need to fix yet, particularly the handling of `failWait` and other uses of `killWait`.

cmd/pebble/cmd_run.go
```diff
@@ -19,49 +19,113 @@ import (
...
 	clientMixin
 
-	CreateDirs bool   `long:"create-dirs"`
-	Hold       bool   `long:"hold"`
-	HTTP       string `long:"http"`
-	Verbose    bool   `short:"v" long:"verbose"`
+	CreateDirs      bool   `long:"create-dirs"`
+	Hold            bool   `long:"hold"`
+	HTTP            string `long:"http"`
+	ShutdownSignals string `long:"shutdown-signals" defaults:"INT,QUIT,TERM"`
```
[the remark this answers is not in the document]
tlm: I was thinking about it tests. I'll dive into it and see what I can come up with.

cmd/pebble/cmd_run.go
```diff
@@ -19,49 +19,113 @@ import (
...
 func (rcmd *cmdRun) Execute(args []string) error {
 	if len(args) > 0 {
 		return ErrExtraArgs
 	}
 
+	var err error
+	shutdownSignals := []os.Signal{
+		syscall.SIGINT,
+		syscall.SIGQUIT,
+		syscall.SIGTERM,
```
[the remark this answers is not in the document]
tlm: We are still going to need these 3 hard coded outside of the `default` specifier for the ignore call. I've reworked this a bit to hopefully read better.

internal/overlord/servstate/handlers.go
tlm: I have called it `killWaitDuration` as that is what it's doing. Changed to 5 seconds as that is the amount of time afforded previously and it's been working for Pebble thus far.

internal/overlord/servstate/export_test.go
```diff
@@ -72,11 +72,13 @@ func FakeOkayWait(wait time.Duration) (restore func()) {
...
-	old1, old2 := killWait, failWait
-	killWait, failWait = kill, fail
+// FakeGraceKillWait changes both the defaultGracePeriod and killWaitDuration
+// respectively for testing purposes.
+func FakeGraceKillWait(gracePeriod, killWait time.Duration) (restore func()) {
+	old1, old2 := defaultGracePeriod, killWaitDuration
+	defaultGracePeriod, killWaitDuration = gracePeriod, killWait
 	return func() {
-		killWait, failWait = old1, old2
+		defaultGracePeriod, killWaitDuration = old1, old2
```
benhoyt, line 81: Cool, thanks. One more thing. These (at the top of `servstate/manager_test.go`):

```go
	shortKillWait = 100 * time.Millisecond
	shortFailWait = 200 * time.Millisecond
```

Should probably be changed to `shortGracePeriod` and `shortKillWaitDuration`, and the latter could be changed to `100 * time.Millisecond` to match what it was before (now that it's additive).

benhoyt (approved):
This looks good to me now, thanks! Just one small comment re the `short*Wait` constants.

tlm:
Closing in favor of #190

PR #188, opened 2023-01-31
cmd: Add enter command

woky:
The enter command behaves just like run when run without positional
  arguments, with the exception that the --hold flag is not supported.
  That is, it runs as service manager. When run with positional arguments,
  the service manager is started in background without starting autostart
  services (i.e. like run with the --hold flag) and positional arguments
  are executed as pebble command. Only subset of pebble commands is
  supported, namely exec, help, ls, plan, services, start, stop and
  version. Autostart services can be started with the --run flag. start
  command must be run with the --run flag and ls, plan, services and stop
  commands cannot be run with the --run flag. Normally, the service
  manager is shut down when the command finishes, except for start and
  stop commands where the service manager continues running. The enter
  command is intended to be used as entrypoint in container images.


woky:
@niemeyer @benhoyt Note that I've included my https://github.com/woky/go-flags/ fork which supports setting `PassAfterNonOption` to subset of commands. I've a opened a PR against upstream: https://github.com/jessevdk/go-flags/pull/393 But given the activity of the maintainer on the GitHub, I suspect it won't get any attention anytime soon. There's also other PR to go-flags incoming from @rebornplusplus adding support to `find -exec`-like terminated option arguments. Should we consider forking go-flags?

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,164 @@
...
+		logger.SetLogger(logger.NullLogger)
+	}
+
+	runReadyCh := make(chan func(), 1)
+	runPanicCh := make(chan interface{})
+	var runStop func()
+	var runPanic interface{}
+
+	go func() {
+		defer func() { runPanicCh <- recover() }()
```
benhoyt, line 140: Why don't we just let a panic crash the process?

cmd/pebble/main.go
```diff
@@ -255,53 +256,18 @@ func Parser(cli *client.Client) *flags.Parser {
 	}
 	// Add all the sub-commands of the debug command
 	for _, c := range debugCommands {
-		obj := c.builder()
```
benhoyt, line 258: Why does all this code go away? Was it duplicated from above previously?

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,164 @@
...
+service manager. When run with positional arguments, the service manager is
+started in background without starting autostart services (i.e. like run with
+the --hold flag) and positional arguments are executed as pebble command. Only
+subset of pebble commands is supported, namely exec, help, ls, plan, services,
+start, stop and version. Autostart services can be started with the --run flag.
+stop command must be run with the --run flag and ls, plan and services commands
+cannot be run with the --run flag. Normally, the service manager is shut down
+when the command finishes. With start and stop commands the service manager
+continues running. The enter command is intended to be used as entrypoint in
+container images.
```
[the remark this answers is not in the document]
woky: I've split the description into multiple paragraphs. These peculiarities are part of a specification so the command has to behave this way. I haven't included the `ENTRYPOINT` example as it'd be just `ENTRYPOINT ["pebble", "enter"]`. The first paragraph covers that use-case briefly. I'd complete the description of the use-case once we have some use-cases.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,164 @@
...
+`
+
+type enterFlags int
+
+const (
+	enterFlagSilcenceLogging enterFlags = 1 << iota
```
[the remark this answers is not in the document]
woky: Fixed.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,164 @@
...
+	enterFlagNoServiceManager
+	enterFlagKeepServiceManager
+	enterFlagRequireServiceAutostart
+	enterFlagProhibitServiceAutostart
+)
+
+type haveEnterFlags interface {
+	enterFlags() enterFlags
```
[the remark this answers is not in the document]
woky: I've reverted the changes to individual comments and moved the logic into single type switch. Unfortunately, it's a type switch, and not a data structure, because of the way how go-flags `flags.Commander` works. This is an interface with `Execute(args []string) error` as the only method. By default, when the parser runs, this method is executed for the last command given. The only way to intercept this is to set `Parser.CommandHandler` to a function with the following signature:
```
func(command Commander args []string) error
````
IMHO, the signature should've been:
```
func(command Command, commander Commander args []string) error
```
This would give me a way to index some data structure via `command.Name`.

File: cmd/pebble/cmd_enter.go [hunk 6]
```diff
@@ -0,0 +1,164 @@
...
+	var (
+		subcCommander flags.Commander
+		subcExtraArgs []string
+		enterFlags    enterFlags
+	)
+
+	subcParser := createParser(cmd.client, enterCommands, enterDebugCommands)
+	subcParser.CommandHandler = func(commander flags.Commander, args []string) error {
+		if withFlags, ok := commander.(haveEnterFlags); ok {
+			enterFlags = withFlags.enterFlags()
```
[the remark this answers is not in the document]
woky: Removed, though the type switch is still used, as commented above.

cmd/pebble/cmd_enter.go
woky: This is because in `cmd/` we actually use `panic({exitStatus(1)})` as kind of _long jump_. When `exitStatus` error is recovered in `main`, the main just exits with the given status code. The run command uses it for restart functionality and also to bail when something goes wrong. If I let the goroutine panic, the `exitStatus` error wouldn't be recovered in main. And, since `exitStatus` has `Error` method, if run command wanted to exit with status 1, then the following would be printed
```
panic: internal error: exitStatus{1} being handled as normal error

goroutine 6 [running]:
main.(*cmdRun).run(0x0?, 0x0?)
	/home/woky/devel/pebble/cmd/pebble/cmd_run.go:82 +0xa9
main.(*cmdEnter).Execute.func2()
	/home/woky/devel/pebble/cmd/pebble/cmd_enter.go:140 +0x25
created by main.(*cmdEnter).Execute
	/home/woky/devel/pebble/cmd/pebble/cmd_enter.go:138 +0x46a
exit status 2
```

instead of just

```
exit status 1
```

cmd/pebble/main.go
woky: There were minimal differences (posted below) which IMHO weren't worth to have this code duplicated. But I've reverted this refactor.
```
--- cmd.normal	2023-02-06 00:51:00.803178868 +0100
+++ cmd.debug	2023-02-06 00:51:10.103288970 +0100
@@ -2,19 +2,11 @@
 		if x, ok := obj.(clientSetter); ok {
 			x.setClient(cli)
 		}
-		if x, ok := obj.(parserSetter); ok {
-			x.setParser(parser)
-		}
-
-		cmd, err := parser.AddCommand(c.name, c.shortHelp, strings.TrimSpace(c.longHelp), obj)
+		cmd, err := debugCommand.AddCommand(c.name, c.shortHelp, strings.TrimSpace(c.longHelp), obj)
 		if err != nil {
-			logger.Panicf("cannot add command %q: %v", c.name, err)
+			logger.Panicf("cannot add debug command %q: %v", c.name, err)
 		}
 		cmd.Hidden = c.hidden
-		if c.alias != "" {
-			cmd.Aliases = append(cmd.Aliases, c.alias)
-		}
-
 		opts := cmd.Options()
 		if c.optDescs != nil && len(opts) != len(c.optDescs) {
 			logger.Panicf("wrong number of option descriptions for %s: expected %d, got %d", c.name, len(opts), len(c.optDescs))
@@ -49,6 +41,3 @@
 			arg.Name = name
 			arg.Description = desc
 		}
-		if c.extra != nil {
-			c.extra(cmd)
-		}
```

woky:
@benhoyt I've addressed your comments. It won't be ready to merge unless either the PR to go-flags is merged or we decide to fork it (and how), but I think it's ready to review.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
...
+a single pebble command after starting the service manager. It's intended to be
+used as entrypoint in container images.
+
+When no subcommand is given, the behavior is identical to the run command.
+
+When a subcommand is given, the service manager is started in the background
+without starting "autostart" services (i.e. run --hold, unless --run flag is
+given), and the subcommand is run as a pebble client command that can
+communicate with the service manager instance in the background. The service
+manager is shut down after the commands finishes unless noted otherwise below.
```
niemeyer, line 22: This is doing a reasonable job of describing the behavior, but as Ben already spotted it is also very confusing because the actual behavior of any one command is hard to grasp because it's an ANDing of a number of different conditional paragraphs that may affect the given command or not.  The agreement we've reached at the end of our several meetings was supposed to be one where the behavior of enter would be the most natural option for the given command at hand, so if we did that justice, we shoud be able to have a more straightforward description here too.

With that said, let's leave that aside for now and focus on the implementation. Once this is ready to go in, we can do a pass together and find the best way to word this out.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
...
+	case *cmdHelp:
+		enterFlags = enterFlagNoServiceManager
+	case *cmdLs, *cmdPlan, *cmdServices, *cmdVersion:
+		enterFlags = enterFlagSilenceLogging | enterFlagProhibitServiceAutostart
+	case *cmdStart:
+		enterFlags = enterFlagKeepServiceManager
+	case *cmdStop:
+		enterFlags = enterFlagKeepServiceManager | enterFlagRequireServiceAutostart
+	default:
+		supported = false
```
niemeyer, line 96: I was thinking about allowing each of the commands to have their own flags defined locally, but the clarity of this function is so nice and readable. It indeed makes it feel like a better approach.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if _, err := parser.ParseArgs(cmd.Positional.Cmd); err != nil {
+		return err
+	}
+
+	enterFlags, supported := commandEnterFlags(commander)
+
+	if !supported {
+		return fmt.Errorf("enter: this command is not supported")
```
niemeyer, line 137: `enter: subcommand %q is not supported`

Otherwise when we get the error while running a script or something we don't know what is the actual issue.

It also seems better to call these as "subcommands" since "enter" is the command here.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if enterFlags&enterFlagRequireServiceAutostart != 0 && !cmd.Run {
+		return fmt.Errorf("enter: this command can only run with autostarted services (hint: add --run to enter)")
```
niemeyer, line 141: Perhaps more directly:

```
enter: must use --run before %q subcommand
```

With that said, the memory I have of our final agreement (which arguably took a bit of back and forth, so I may easily be missing something) was that we'd make subcommands behave in their most natural way, and when there was no natural way at all we'd prevent the subcommand from running.

In particular here, the fact we'd have _start_ and _stop_ disagreeing on the requirement of _--run_, and the fact we're forcing _--run_ to be provided to a subcommand that could not work without it, are both bad symptoms that this isn't quite what we need. Let's talk about it elsewhere to revisit this topic please.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if enterFlags&(enterFlagProhibitServiceAutostart|enterFlagNoServiceManager) != 0 && cmd.Run {
+		return fmt.Errorf("enter: this command cannot run with autostarted services (hint: remove --run from enter)")
```
niemeyer, line 145: This seems good, though. We could just make it more direct as well:
```
enter: cannot provide --run before %q subcommand
```

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
...
+		if runPanic == nil {
+			panic("internal error: pebble daemon stopped early")
+		}
+		panic(runPanic)
+	}
+
+	err := commander.Execute(extraArgs)
+
+	if err != nil || enterFlags&enterFlagKeepServiceManager == 0 {
+		runStop()
```
niemeyer, line 178: Haven't reached the end of the code yet, so maybe this will be answered later, but how are we keeping the run command going if we're returning from `enter` early here?  In other words, the fact we're not asking the run command to stop explicitly won't do much for us if this command is about to return and exit from Pebble itself.

niemeyer (changes requested):
That's surprisingly tidy, thanks Tomas.

Just a few minors and some high-level questions:

cmd/pebble/cmd_enter.go
woky: I agree it's not exactly straightforward read.

cmd/pebble/cmd_enter.go
woky: I wholeheartedly agree that the sub-command behavior as currently specified is confusing and hard to convey.

cmd/pebble/cmd_enter.go
woky: This is handled by `runResultCh` (previously called `runPanicCh`) that's written to in the goroutine closure above after the `run` command ends, and that's being read from right after this statement.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,188 @@
...
+  ls
+  plan
+  services
+  start
+  version
+
+Furthermore, some commands have restrictions and idiosyncrasies:
+
+  exec, ls, plan, services, version
+    Service manager logging is disabled unless -v/--verbose flag is provided.
```
benhoyt, line 37: The wording still needs a few style/grammar tweaks I think. But for now, just note that these indentations are removed when displaying the help (i.e., running `pebble enter -h`).

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,188 @@
...
+`
+
+type enterFlags int
+
+const (
+	enterFlagSilenceLogging enterFlags = 1 << iota
+	enterFlagNoServiceManager
+	enterFlagKeepServiceManager
+	enterFlagRequireServiceAutostart
+	enterFlagProhibitServiceAutostart
```
benhoyt, line 56: I don't think the `Flag` word here adds anything to these names, and removing it would shorten them up a bit. For example, `enterFlagNoServiceManager` to `enterNoServiceManager`.

benhoyt (approved):
This looks good to me. Can we add some simple tests for a few of the `pebble enter` use cases? Might be more trouble than it's worth, but worth some thought IMO.

cmd/pebble/cmd_enter.go
woky: Noted, thanks. This is done by `wrapText` function in go-flags. Let's keep it indented and fix go-flags later.

woky:
@niemeyer @benhoyt I've addressed most of your comments (and replied to the rest). Please take a look again.

Note the CI failures. This because of [data race detector](https://go.dev/doc/articles/race_detector) detecting race condition when setting logger with `logger.SetLogger()`: <details>
<summary>Data race detector output</summary>

```
==================
WARNING: DATA RACE
Write at 0x00000117add0 by goroutine 336:
  github.com/canonical/pebble/internal/logger.SetLogger()
      /home/woky/canonical/devel/pebble/internal/logger/logger.go:83 +0x15e
  github.com/canonical/pebble/cmd/pebble.run()
      /home/woky/canonical/devel/pebble/cmd/pebble/main.go:343 +0x36
  github.com/canonical/pebble/cmd/pebble.main()
      /home/woky/canonical/devel/pebble/cmd/pebble/main.go:324 +0x58
  github.com/canonical/pebble/cmd/pebble.RealMain()
      /home/woky/canonical/devel/pebble/cmd/pebble/export_test.go:77 +0xdc
  github.com/canonical/pebble/cmd/pebble_test.(*PebbleSuite).TestEnterHelpCommand()
      /home/woky/canonical/devel/pebble/cmd/pebble/cmd_enter_test.go:79 +0x22c
  runtime.call16()
      /usr/lib/go/src/runtime/asm_amd64.s:728 +0x48
  reflect.Value.Call()
      /usr/lib/go/src/reflect/value.go:370 +0xc7
  gopkg.in/check%2ev1.(*suiteRunner).forkTest.func1()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:775 +0xadd
  gopkg.in/check%2ev1.(*suiteRunner).forkCall.func1()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:669 +0xec

Previous read at 0x00000117add0 by goroutine 319:
  github.com/canonical/pebble/internal/logger.Debugf()
      /home/woky/canonical/devel/pebble/internal/logger/logger.go:66 +0x6e
  github.com/canonical/pebble/internal/overlord/cmdstate.(*execution).controlLoop()
      /home/woky/canonical/devel/pebble/internal/overlord/cmdstate/handlers.go:464 +0x10e4
  github.com/canonical/pebble/internal/overlord/cmdstate.(*execution).do.func9()
      /home/woky/canonical/devel/pebble/internal/overlord/cmdstate/handlers.go:272 +0x84

Goroutine 336 (running) created at:
  gopkg.in/check%2ev1.(*suiteRunner).forkCall()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:666 +0x5c4
  gopkg.in/check%2ev1.(*suiteRunner).forkTest()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:757 +0x16b
  gopkg.in/check%2ev1.(*suiteRunner).runTest()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:812 +0x439
  gopkg.in/check%2ev1.(*suiteRunner).run()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:618 +0x3e6
  gopkg.in/check%2ev1.Run()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/run.go:92 +0x49
  gopkg.in/check%2ev1.RunAll()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/run.go:84 +0x12b
  gopkg.in/check%2ev1.TestingT()
      /home/woky/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/run.go:72 +0x565
  github.com/canonical/pebble/cmd/pebble_test.Test()
      /home/woky/canonical/devel/pebble/cmd/pebble/main_test.go:24 +0x2e
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:1576 +0x216
  testing.(*T).Run.func1()
      /usr/lib/go/src/testing/testing.go:1629 +0x47

Goroutine 319 (finished) created at:
  github.com/canonical/pebble/internal/overlord/cmdstate.(*execution).do()
      /home/woky/canonical/devel/pebble/internal/overlord/cmdstate/handlers.go:272 +0xfc5
  github.com/canonical/pebble/internal/overlord/cmdstate.(*CommandManager).doExec()
      /home/woky/canonical/devel/pebble/internal/overlord/cmdstate/handlers.go:118 +0xb09
  github.com/canonical/pebble/internal/overlord/cmdstate.(*CommandManager).doExec-fm()
      <autogenerated>:1 +0x4d
  github.com/canonical/pebble/internal/overlord/state.(*TaskRunner).run.func1()
      /home/woky/canonical/devel/pebble/internal/overlord/state/taskrunner.go:195 +0xb3
  gopkg.in/tomb%2ev2.(*Tomb).run()
      /home/woky/go/pkg/mod/gopkg.in/tomb.v2@v2.0.0-20161208151619-d5d1b5820637/tomb.go:163 +0x51
  gopkg.in/tomb%2ev2.(*Tomb).Go.func2()
      /home/woky/go/pkg/mod/gopkg.in/tomb.v2@v2.0.0-20161208151619-d5d1b5820637/tomb.go:159 +0x47
==================
```
</details>

I've never debugged data races in Golang so I could use some help. However, my theory is that this is a non-issue and it only surfaced because this is the first test where pebble is started as service manager (overlord). The `logger.SetLogger()` call is made only in `main()` and only once. In actual `pebble enter` execution the `main()` is called exactly once (in the same process). The sequence of the steps in data race detector report (write after previous read) indicates that this is an issue with tests not being properly isolated between each other.

I could disable the the data race detector in the new test file with `// +build !race` but that doesn't seem right. Synchronizing access to the logger also doesn't seem right as that'd make logging expensive. What are your thoughts?

woky:
I've fixed the data race by synchronizing access to the global `logger` variable. After discussion with @niemeyer we concluded it's fine.

The currently failing test in `Go 1.15` is unrelated:
```
2023-03-02T15:20:33.4744199Z START: daemon_test.go:727: daemonSuite.TestRebootHelper
2023-03-02T15:20:33.4744628Z START: daemon_test.go:64: daemonSuite.SetUpTest
2023-03-02T15:20:33.4744979Z PASS: daemon_test.go:64: daemonSuite.SetUpTest	0.000s
2023-03-02T15:20:33.4745184Z 
2023-03-02T15:20:33.4745346Z using shellcheck: "/usr/bin/shellcheck"
2023-03-02T15:20:33.4745658Z daemon_test.go:744:
2023-03-02T15:20:33.4745942Z     c.Assert(err, check.IsNil)
2023-03-02T15:20:33.4746406Z ... value *os.SyscallError = &os.SyscallError{Syscall:"waitid", Err:0xa} ("waitid: no child processes")
2023-03-02T15:20:33.4746683Z 
2023-03-02T15:20:33.4746866Z START: daemon_test.go:73: daemonSuite.TearDownTest
2023-03-02T15:20:33.4747262Z PASS: daemon_test.go:73: daemonSuite.TearDownTest	0.000s
2023-03-02T15:20:33.4747479Z 
2023-03-02T15:20:33.4747635Z FAIL: daemon_test.go:727: daemonSuite.TestRebootHelper
```

benhoyt (approved):

benhoyt:
Looks good to me (apart from perhaps the help wording which it sounds like you and Gustavo will work on). And thanks for the logging fix!

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"fmt"
+
+	"github.com/canonical/pebble/internal/logger"
+	"github.com/jessevdk/go-flags"
+)
+
+const shortEnterHelp = "Run pebble command in new pebble environment"
+const longEnterHelp = `
```
niemeyer, line 11: Okay, here is the proposal
Short help: "Run subcommand under a container environment"

Long help:
```
The enter command facilitates the use of Pebble as an entrypoint for containers.
When used without a subcommand, it mimics the behavior of the run command
alone, while if used with a subcommand it runs that subcommand in the most
appropriate environment taking in account its purpose.

These subcommands are currently supported:

  help      (1)(2)
  version   (1)(2)
  plan      (1)(2)
  services  (1)(2)
  ls        (1)(2)
  exec      (2)
  start     (3)
  stop      (3)

(1) Services are not started.
```

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,188 @@
...
+    Service manager logging is disabled unless -v/--verbose flag is provided.
+
+  help, ls, plan, services, version
+    Service autostart (the --run flag) is prohibited.
+
+  start, stop
+    Service manager continues running after the command finishes.
+
+  stop
+    Service autostart (the --run flag) is required.
```
niemeyer, line 46: We forgot to address this wart in our call yesterday. We need to figure why start/stop were proposed to be different, as it's very awkward to require a flag at all times for stop and not for start.

niemeyer (changes requested):
Here is the proposed help text, and we forgot to talk about `stop` vs. `--run` yesterday.

Otherwise, this looks good to go.

cmd/pebble/cmd_enter.go
cjdcordeiro: looks good.

nitpick:

```
< taking in account 
> taking into account
```

cmd/pebble/cmd_enter.go
cjdcordeiro: Recalling the spec (RK007) discussions, the logic was that `pebble enter stop [svc ...]` would have an equivalent behaviour to `pebble run --hold; pebble stop [svc ...]`, which was deemed to be undesired and thus marked as an "error" combination. 

> Service autostart (the --run flag) is required.

Putting this in order words, `--run` is required as a consequence of `pebble enter` acting as `pebble run --hold`. Without it,  `enter stop` doesn't do anything cause there are no "started" services.

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,172 @@
...
+	sharedRunEnterOpts
+	Run        bool `long:"run"`
+	Positional struct {
+		Cmd []string `positional-arg-name:"<subcommand>"`
+	} `positional-args:"yes"`
+}
+
+func init() {
+	optsHelp := map[string]string{
+		"run": "Start \"autostart\" services before executing subcommand",
```
niemeyer, line 54: s/Start "autostart" services/Start default services/

This is the same terminology used in the autostart command itself.

niemeyer (approved):
Alright, thanks for all the reviews and help here.

This is good to go, pending only the go-flags support.

niemeyer:
Given the lack of response from upstream, let's temporarily fork go-flags at canonical/go-flags, and submit/review the required extensions there please.

woky:
> Given the lack of response from upstream, let's temporarily fork go-flags at https://github.com/canonical/go-flags, and submit/review the required extensions there please.

Thanks. Opened PR https://github.com/canonical/go-flags/pull/1

cjdcordeiro (approved):

woky:
@niemeyer @benhoyt There was a bug in `subcommand --help` and `help command` handling. I've fixed it in latest commit: https://github.com/canonical/pebble/pull/188/commits/d331c3c88c85d39c9fdab72805a347bf18e947cf

woky:
Updated go.{mod|sum} for canonical/go-flags fork. Tests pass! Should be good to go. :o)

cmd/pebble/cmd_enter.go
```diff
@@ -0,0 +1,186 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/canonical/pebble/internal/logger"
+	"github.com/jessevdk/go-flags"
```
niemeyer, line 7: This needs to be changed to Canonical's version of go-flags instead of just using go.mod. There's a long term commitment here as we have no indication of upstream's intention to address the changes required, and it shoud be clear to people reading these files that they are indeed looking at something else.

niemeyer (changes requested):
Thanks, Tomas.  One trivial only before we merge:

cmd/pebble/cmd_enter.go
niemeyer: Please note this is true of all the files under cmd/, not just this one.

woky: Mentioned this in more detail out of band. Module name of our fork has to be changed first: https://github.com/canonical/go-flags/pull/3#pullrequestreview-1368307459

I will update all the imports after the forked module name change.

niemeyer:
One more trivial, sorry about that: formatting checks are failing.

woky:
> One more trivial, sorry about that: formatting checks are failing.

Whoops. Fixed. (And squashed into first commit.)

PR #190, opened 2023-02-08
[JUJU-1887] Adds kill delay duration for service shutdown.

tlm:
- This change adds a new key to the Pebble service plan spec called kill-delay. This allows the author of a service definition to control the amount of time Pebble allows before sending SIGTERM and SIGKILL to a process.

- This change also increases the default duration afforded to pebble for kill-delay from 5s to 25s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -510,6 +510,10 @@ services:
         # half a minute ("30s").
         backoff-limit: <duration>
 
+        # (Optional) The amount of time afforded to this service to handle
+        # SIGTERM and exit gracefully before SIGKILL terminates it forcefully.
+        # Default is 5 seconds ("5s").
+        kill-delay: <duration>
 
 # (Optional) A list of health checks managed by this configuration layer.
 checks:
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -635,10 +635,6 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 	return nil
 }
 
-// This is a little more than servstate.killWait (but that isn't and shouldn't
-// be exported, so just duplicate it here).
-const stopRunningTimeout = 5*time.Second + 100*time.Millisecond
-
 // stopRunningServices stops all running services, waiting for a short time
 // for them all to stop.
 func (d *Daemon) stopRunningServices() error {
@@ -664,7 +660,7 @@ func (d *Daemon) stopRunningServices() error {
 	select {
 	case <-chg.Ready():
 		logger.Debugf("All services stopped.")
-	case <-time.After(stopRunningTimeout):
+	case <-time.After(d.overlord.ServiceManager().StopTimeout()):
 		return errors.New("timeout stopping running services")
 	}
 	return nil
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -54,9 +54,18 @@ func TaskServiceRequest(task *state.Task) (*ServiceRequest, error) {
 }
 
 var (
-	okayWait = 1 * time.Second
-	killWait = 5 * time.Second
-	failWait = 10 * time.Second
+	// okayDelay is the time to wait after starting a service before we conclude
+	// that it's running successfully.
+	okayDelay = 1 * time.Second
+
+	// killDelayDefault is the duration afforded to services for processing
+	// SIGTERM signals and shutting down cleanly if the service hasn't specified
+	// their own duration.
+	killDelayDefault = 5 * time.Second
+
+	// failDelay is the duration given to services for shutting down when Pebble
+	// sends a SIGKILL signal.
+	failDelay = 5 * time.Second
 )
 
 const (
@@ -307,7 +316,7 @@ func (s *serviceData) start() error {
 			return err
 		}
 		s.transition(stateStarting)
-		time.AfterFunc(okayWait, func() { logError(s.okayWaitElapsed()) })
+		time.AfterFunc(okayDelay, func() { logError(s.okayWaitElapsed()) })
 
 	default:
 		return fmt.Errorf("cannot start service while %s", s.state)
@@ -592,6 +601,17 @@ func (s *serviceData) sendSignal(signal string) error {
 	return nil
 }
 
+// killDelay reports the duration that this service should be given when being
+// asked to shutdown gracefully before being force terminated. The value
+// returned will either be the services pre configured value or the default
+// kill delay for pebble.
+func (s *serviceData) killDelay() time.Duration {
+	if s.config.KillDelay.IsSet {
+		return s.config.KillDelay.Value
+	}
+	return killDelayDefault
+}
+
 // stop is called to stop a running (or backing off) service.
 func (s *serviceData) stop() error {
 	s.manager.servicesLock.Lock()
@@ -606,7 +626,7 @@ func (s *serviceData) stop() error {
 			logger.Noticef("Cannot send SIGTERM to process: %v", err)
 		}
 		s.transition(stateTerminating)
-		time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
+		time.AfterFunc(s.killDelay(), func() { logError(s.terminateTimeElapsed()) })
 
 	case stateBackoff:
 		logger.Noticef("Service %q stopped while waiting for backoff", s.config.Name)
@@ -654,8 +674,9 @@ func (s *serviceData) terminateTimeElapsed() error {
 		if err != nil {
 			logger.Noticef("Cannot send SIGKILL to process: %v", err)
 		}
+
 		s.transitionRestarting(stateKilling, s.restarting)
-		time.AfterFunc(failWait-killWait, func() { logError(s.killTimeElapsed()) })
+		time.AfterFunc(failDelay, func() { logError(s.killTimeElapsed()) })
 
 	default:
 		// Ignore if timer elapsed in any other state.
@@ -732,7 +753,7 @@ func (s *serviceData) checkFailed(action plan.ServiceAction) {
 					logger.Noticef("Cannot send SIGTERM to process: %v", err)
 				}
 				s.transitionRestarting(stateTerminating, true)
-				time.AfterFunc(killWait, func() { logError(s.terminateTimeElapsed()) })
+				time.AfterFunc(s.killDelay(), func() { logError(s.terminateTimeElapsed()) })
 			case stateBackoff:
 				logger.Noticef("Service %q %s action is %q, waiting for current backoff",
 					s.config.Name, onType, action)
--- a/internal/overlord/servstate/manager.go
+++ b/internal/overlord/servstate/manager.go
@@ -300,6 +300,30 @@ func (m *ServiceManager) Services(names []string) ([]*ServiceInfo, error) {
 	return services, nil
 }
 
+// StopTimeout returns the worst case duration that will have to be waited for
+// to have all services in this manager stopped.
+func (m *ServiceManager) StopTimeout() time.Duration {
+	m.servicesLock.Lock()
+	defer m.servicesLock.Unlock()
+
+	maxDuration := killDelayDefault
+	for _, service := range m.services {
+		if service == nil {
+			continue
+		}
+		switch service.state {
+		case stateStarting, stateRunning, stateTerminating:
+			if service.killDelay() > maxDuration {
+				maxDuration = service.killDelay()
+			}
+		}
+	}
+
+	// We add a little extra time here to allow for signals to be sent and
+	// processed.
+	return maxDuration + failDelay + 100*time.Millisecond
+}
+
 func stateToStatus(state serviceState) ServiceStatus {
 	switch state {
 	case stateStarting, stateRunning:
--- a/internal/overlord/servstate/manager_test.go
+++ b/internal/overlord/servstate/manager_test.go
@@ -46,9 +46,9 @@ import (
 )
 
 const (
-	shortOkayWait = 50 * time.Millisecond
-	shortKillWait = 100 * time.Millisecond
-	shortFailWait = 200 * time.Millisecond
+	shortOkayDelay = 50 * time.Millisecond
+	shortKillDelay = 100 * time.Millisecond
+	shortFailDelay = 100 * time.Millisecond
 )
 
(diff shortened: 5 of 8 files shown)


README.md
```diff
@@ -445,6 +445,11 @@ services:
         after:
             - <other service name>
 
+        # (Optional) The amount of time afforded to this service to handle
+        # SIGTERM and exit gracefully before being force killed.
+        # Default is 25 seconds ("25s").
+        kill-delay: <grace-period>
+
```
benhoyt, line 452: It's weird to place this between "after" and "before". I would probably put it last, after "backoff-limit".

Also, can we clarify that this is connected to stop/restart? Perhaps:

```
        # (Optional) When a service is being stopped or restarted, this is the
        # amount of time allowed for the service to handle SIGTERM and exit
        # gracefully before being force-killed with SIGKILL. Default is 25
        # seconds ("25s").
```

internal/daemon/daemon.go
```diff
@@ -635,10 +635,6 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 	return nil
 }
 
-// This is a little more than servstate.killWait (but that isn't and shouldn't
-// be exported, so just duplicate it here).
-const stopRunningTimeout = 5*time.Second + 100*time.Millisecond
-
```
benhoyt, line 641: Do we still want to remove this? If so, why?

internal/overlord/servstate/handlers.go
```diff
@@ -54,9 +54,15 @@ func TaskServiceRequest(task *state.Task) (*ServiceRequest, error) {
...
 
 var (
-	okayWait = 1 * time.Second
-	killWait = 5 * time.Second
-	failWait = 10 * time.Second
+	// defaultKillDelay is the duration afforded to services for processing
+	// SIGTERM signals and shutting down cleanly if the service hasn't specified
+	// their own duration.
+	defaultKillDelay = 25 * time.Second
+	okayWait         = 1 * time.Second
```
benhoyt, line 61: Because `okayWait` is quite separate (and used when starting a service), can we move it to the top of this block? And perhaps add a brief comment while we're here:

```go
var (
    // okayWait is the time to wait after starting a service before we conclude
    // that it's running successfully.
    okayWait = 1 * time.Second
    ...
)
```

internal/overlord/servstate/manager_test.go
```diff
@@ -320,6 +328,32 @@ func (s *S) stopTestServicesAlreadyDead(c *C) {
 	s.st.Unlock()
 }
 
+func (s *S) TestGracePeriodIsUsed(c *C) {
```
benhoyt, line 331: This test takes 10 seconds to run, which more than doubles our test run time. :-) Can we shorten it to a couple hundred milliseconds at most? But we also need to ensure it's not flakey.

benhoyt (commented):
Looks good to me overall. Couple of nit comments and a couple of questions.

README.md
benhoyt: Oh, one more thing: Instead of saying `<grace-period>` as the value we should say `<duration>` (like other duration fields).

internal/daemon/daemon.go
tlm: Because it triggers well before the service is stopped and then forces the reaper to be stopped. Doesn't make a huge amount of sense in my opinion.

tlm: Try that out based on our discussions.

internal/overlord/servstate/manager.go
```diff
@@ -300,6 +300,24 @@ func (m *ServiceManager) Services(names []string) ([]*ServiceInfo, error) {
...
+	defer m.servicesLock.Unlock()
+
+	maxDuration := defaultKillDelay
+	for _, svc := range m.services {
+		if svc.killDelay() > maxDuration {
+			maxDuration = svc.killDelay()
+		}
+	}
+
+	// We add a millisecond on to the maximum here to allow for signals to be
```
benhoyt, line 316: Nit: it's 100 milliseconds, so comment doesn't match code. What about just commenting "We add a little..."

benhoyt (approved):
Looks good to me now, thank you!

paulomach:
Just passing by to say that's super critical for data platform deployments and we are looking forward to it. :grimacing:

delgod (approved):

internal/overlord/servstate/manager_test.go
```diff
@@ -46,9 +46,9 @@ import (
 )
 
 const (
-	shortOkayWait = 50 * time.Millisecond
-	shortKillWait = 100 * time.Millisecond
-	shortFailWait = 200 * time.Millisecond
+	shortOkayWait         = 50 * time.Millisecond
+	shortKillDelay        = 100 * time.Millisecond
+	shortKillWaitDuration = 100 * time.Millisecond
```
niemeyer, line 51: Needs updating to reflect previous points.

internal/overlord/servstate/manager_test.go
```diff
@@ -320,6 +328,51 @@ func (s *S) stopTestServicesAlreadyDead(c *C) {
 	s.st.Unlock()
 }
 
+func (s *S) TestMaximumStopRunningDuration(c *C) {
```
niemeyer, line 331: TestStopTimeout after previous points.

internal/overlord/servstate/manager_test.go
```diff
@@ -320,6 +328,51 @@ func (s *S) stopTestServicesAlreadyDead(c *C) {
...
+	_, _, err = s.manager.Replan()
+	c.Assert(err, IsNil)
+	s.startServices(c, []string{"test9"}, 1)
+	s.waitUntilService(c, "test9", func(svc *servstate.ServiceInfo) bool {
+		return svc.Current == servstate.StatusActive
+	})
+	c.Assert(s.manager.MaximumStopRunningDuration(), Equals, time.Minute*60+time.Millisecond*100)
+}
+
+func (s *S) TestKillWaitIsUsed(c *C) {
```
niemeyer, line 350: KillDelay

internal/overlord/servstate/manager.go
```diff
@@ -300,6 +300,24 @@ func (m *ServiceManager) Services(names []string) ([]*ServiceInfo, error) {
...
 
+// MaximumStopRunningDuration returns the worst case duration that will have to
+// be waited for to have all services in this manager stopped.
+func (m *ServiceManager) MaximumStopRunningDuration() time.Duration {
+	m.servicesLock.Lock()
+	defer m.servicesLock.Unlock()
+
+	maxDuration := defaultKillDelay
+	for _, svc := range m.services {
+		if svc.killDelay() > maxDuration {
```
niemeyer, line 311: A few comments here:

1. Shouldn't we be looking only at `s.state == stateRunning` here?
2. Other code that look into the map seem to consider the case where the map value is nil, which I assume means a quick shutdown. Do we have a panic waiting to happen here?
3. The code in this file (and in tests?) doesn't seem to use the `svc` naming. It's nice to keep these consistent when possible (or more consistent, if there are already multiple cases).

internal/overlord/servstate/manager_test.go
```diff
@@ -167,7 +175,7 @@ func (s *S) SetUpTest(c *C) {
 
 	restore := servstate.FakeOkayWait(shortOkayWait)
 	s.AddCleanup(restore)
-	restore = servstate.FakeKillWait(shortKillWait, shortFailWait)
+	restore = servstate.FakeKillDelayWait(shortKillDelay, shortKillWaitDuration)
```
niemeyer, line 178: That emphasizes one of the earlier points: the delta is "kill" vs. "fail", not "wait" vs. "delay". vs. "duration

niemeyer (changes requested):
Looks very nice, Thomas. Quite a few comments here, but almost all of them are cosmetic. Maybe one item worth pondering on.

internal/overlord/servstate/handlers.go
```diff
@@ -592,6 +601,17 @@ func (s *serviceData) sendSignal(signal string) error {
...
 
+// killDelay reports the duration that this service should be given when being
+// asked to shutdown gracefully before being force terminated. The value
+// returned will either be the services pre configured value or the default
+// grace period for pebble.
+func (s *serviceData) killDelay() time.Duration {
+	if s.config.KillDelay.IsSet {
+		return s.config.KillDelay.Value
+	}
+	return defaultKillDelay
```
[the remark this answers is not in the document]
tlm: Ta, I have removed the reference to grace period in the paragraph and replaced with kill delay to keep the language consistent.

internal/overlord/servstate/manager.go
tlm: Thanks, I have re-worked the code a little and updated tests.

I am considering starting, running and terminating.

Terminating might seem like an odd choice my justification is if you have service a that has a kill delay of 5 minutes and you tell it to stop then immediately stop Pebble the next wait time will be what ever the maximum is of the running services we still might fall outside of that 5 minute window.

Hope that makes sense.

beliaev-maksim:
hi @niemeyer @tlm !
looks like all the comments were addressed, are we good to merge ?

README.md
```diff
@@ -510,6 +510,10 @@ services:
         # half a minute ("30s").
         backoff-limit: <duration>
 
+        # (Optional) The amount of time afforded to this service to handle
+        # SIGTERM and exit gracefully before SIGKILL terminates it forcefully.
+        # Default is 25 seconds ("25s").
```
niemeyer, line 515: This feels like too much for a default. Imagine someone sitting on the terminal for 25 seconds while they run "pebble stop foo"... they will likely CTRL-C out before the job is done as it seems stuck anyway.

internal/daemon/daemon.go
```diff
@@ -635,10 +635,6 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 	return nil
 }
 
-// This is a little more than servstate.killWait (but that isn't and shouldn't
-// be exported, so just duplicate it here).
-const stopRunningTimeout = 5*time.Second + 100*time.Millisecond
```
niemeyer, line 640: That's where the 5 seconds came from, btw.  It's a reasonable time to be waiting for an action to get done before believing that it's actually dead.

internal/overlord/servstate/handlers.go
```diff
@@ -54,9 +54,18 @@ func TaskServiceRequest(task *state.Task) (*ServiceRequest, error) {
...
-	killWait = 5 * time.Second
-	failWait = 10 * time.Second
+	// okayDelay is the time to wait after starting a service before we conclude
+	// that it's running successfully.
+	okayDelay = 1 * time.Second
+
+	// killDelayDefault is the duration afforded to services for processing
+	// SIGTERM signals and shutting down cleanly if the service hasn't specified
+	// their own duration.
+	killDelayDefault = 25 * time.Second
```
niemeyer, line 64: Per earlier note, seems high for the default.

internal/overlord/servstate/manager.go
```diff
@@ -300,6 +300,30 @@ func (m *ServiceManager) Services(names []string) ([]*ServiceInfo, error) {
...
+		case stateStarting, stateRunning, stateTerminating:
+			if service.killDelay() > maxDuration {
+				maxDuration = service.killDelay()
+			}
+		}
+	}
+
+	// We add a little extra time here to allow for signals to be sent and
+	// processed.
+	return maxDuration + 100*time.Millisecond
```
niemeyer, line 324: We need to add failDelay here, as that's the amount of time we're willing to wait before the service is supposed to actually be terminated. Otherwise, 100 milliseconds is way too short and may generate bugs due to the process dying while it's still attempting to deliver the final KILL signals.

internal/overlord/servstate/manager_test.go
```diff
@@ -46,9 +46,9 @@ import (
 )
 
 const (
-	shortOkayWait = 50 * time.Millisecond
-	shortKillWait = 100 * time.Millisecond
-	shortFailWait = 200 * time.Millisecond
+	shortOkayDelay = 50 * time.Millisecond
+	shortKillDelay = 100 * time.Millisecond
+	shortFailDelay = 100 * time.Millisecond
```
niemeyer, line 51: Why was this changed?

niemeyer (changes requested):
All the logic looks fine, thank you.

Only points remaining are about the tuning of times.

internal/overlord/servstate/manager_test.go
benhoyt: It's because the original `shortFailWait` included both the time from first SIGTERM to SIGKILL and the second wait after that (100+100=200). Now the code is such that "fail delay" only includes that second part (which I think is a bit clearer in any case). So the total wait time here wasn't changed.

README.md
tlm: This was changed to 25s to be inline with Kubernetes default 30s. Allowing enough time by default for processes that require a little extra time. My understanding was this was an agreed upon item.

However happy to change. Should we take this back to 5s or another value? @benhoyt do you have any thoughts on what this value should be?

internal/overlord/servstate/manager.go
tlm: Thanks good pick up!

internal/daemon/daemon.go
tlm: Thanks have adjusted below.

internal/overlord/servstate/handlers.go
tlm: As per above.

README.md
jnsgruk: I think the point is that in cases where there is more direct user interaction, 25s feels too long, where 5s doesn't.

Presumably this can still be set to 25/30s when Juju writes the plan for the unit agent either way, and still solve our original issue?

benhoyt: Yeah, agreed re a user interacting with `pebble stop`. I reckon let's keep the default 5s and Juju can easily override.

tlm: Easy done.

internal/overlord/servstate/manager_test.go
niemeyer: Ack, thanks.

niemeyer (approved):
Thanks for the collaboration here, everyone.

PR #191, opened 2023-02-08
feat: Provide arguments to service via ``pebble run --args``

rebornplusplus:
The goal is to introduce the concept of optional and overrideable
arguments for a Pebble service, but we must ensure that the layering
properties of Pebble are preserved and respected. The proposed syntax
re-uses therefore the existing service “command” attribute:

	services:
	    myservice:
	        command: myservice --db=/var/db [ --quiet ]

The CLI is extended to support replacing the default command 
arguments defined in the plan, with the ones given at execution time.
Inspired by the ``-exec`` syntax of the ``find`` command-line utility,
we propose adding a new option to ``pebble run``, called ``--args``.

	Usage:
	    pebble run --args SERVICE [ARGS ...] [TERMINATOR]

	SERVICE:      required name of the Pebble service that will
	              receive the arguments;
	[ARGS ...]:   arguments to be passed to the Pebble service named
	              “SERVICE”;
	[TERMINATOR]: a “;” symbol (backslash-escaped if using the 
	              shell) to indicate that whatever follows is not 
	              an argument for “SERVICE”
	                  - this symbol is only required whenever other 
	                    Pebble-specific options and/or commands are 
	                    given after --args, otherwise can be omitted.

With this, we can provide additional arguments to multiple services 
defined in the plan, whether they hold default arguments or not.

Examples:

Start the Pebble server and only pass some arguments to “svc”:

	pebble run --args svc --verbose --foo "multi str arg"

Start the Pebble server with all services on hold, and give some 
  args to “svc”:

	pebble run --args svc --verbose --foo "multi str arg" \; --hold

Start the Pebble server and pass arguments to multiple services:

	pebble run --args svc1 --verbose --foo "multi str arg" \; \
               --args svc2 --somearg 1 --payload '{"foo": "bar"}'


rebornplusplus:
Like @woky mentioned in https://github.com/canonical/pebble/pull/188,
I have included https://github.com/rebornplusplus/go-flags
which supports terminated options like --args here. There is a PR
against upstream https://github.com/jessevdk/go-flags/pull/395. But
I am not sure when it will get any response from the maintainer. 
Should we consider forking ``go-flags`` and use that for now?

The conflicts in this PR come due to the usage of fork. Also, I have
a few scattered commits here. I intend to squash them before they
are okay to go in.

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse service %q command: %s", s.Name, err)
+	}
+	fargs := make([]string, 0)
+	for _, arg := range args {
+		if arg == "[" || arg == "]" {
+			if len(s.CmdArgs) > 0 {
+				break
+			} else {
+				continue
```
benhoyt, line 203: I'm pretty sure this isn't the logic we need: if there are `CmdArgs` we should insert them at this point but then keep appending the remaining args, right? For example:

```
command: echo foo [ bar ] baz
```

With `--args arg1 arg2 ;` should yield `echo foo arg1 arg2 baz`, right?

We should also handle various error cases like multiple `[ ... ]` groups, attempts to nest, etc (should probably be disallowed).

I had a quick play with it and came up with https://go.dev/play/p/CH25ToKC1q1 (but improvements/rewrite welcome).

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
...
+			}
+		}
+		fargs = append(fargs, arg)
+	}
+	for _, arg := range s.CmdArgs {
+		fargs = append(fargs, arg)
```
benhoyt, line 209: Note that you can do `fargs = append(fargs, s.CmdArgs...)` to append a slice all at once without the loop.

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
...
+			}
+		}
+		return str
+	}
+	args := make([]string, 0)
+	for _, arg := range cmd {
+		args = append(args, f(arg))
+	}
+	return strings.Join(args, " ")
+}
```
benhoyt, line 230: This seems like a halfway house that's neither unquoted nor properly-quoted. Seeing this is just used for logging I guess we could just log the slice with `%q` and be done with it.

It's a pity there's no `shlex.Join` function! Make it wouldn't be hard to make one based on https://github.com/python/cpython/blob/fd155b91392c5bd238781bb11e040cf46c94725c/Lib/shlex.py#L323C31-L334 -- should only be a few lines of code in Go too.

internal/plan/plan.go
benhoyt: Oh, I see you're explicitly only allowed `[ ... ]` at the end of the command. Sorry I'd missed that -- is that by design?

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
...
+
+func (s *Service) setArgs(args []string) error {
+	s.CmdArgs = append([]string(nil), args...)
+	return nil
+}
+
+func (s *Service) checkCommand() error {
```
benhoyt, line 237: Ah, I see you're checking some of the error cases here. However, it looks like this checks that the number of `[` and `]` is balanced (and equal to 1) but not that the `[` comes before the `]`. I think my code linked above handles that.

It is good to validate command here in the plan parsing code -- thanks. Perhaps we can share and reuse a function like `GetCommand()` here and in the service manager, and just pass `nil` to it for `cmdArgs` here when validating.

internal/plan/plan.go
```diff
@@ -622,6 +695,33 @@ func (p *Plan) StopOrder(names []string) ([]string, error) {
 	return order(p.Services, names, true)
 }
 
+// SetServiceArgs sets the service arguments provided by "pebble run --args"
+// to their respective services. Additionally, for a service the arguments
+// are also updated in the layer of the highest order that defines the service
+// so that the arguments to a service can persist if it is not affected
+// by a "pebble add" layer or similar.
+func (p *Plan) SetServiceArgs(serviceArgs map[string][]string) error {
```
benhoyt, line 703: Yeah, this is the logic (or similar logic) that I think should be separated from the plan and go in the service manager.

internal/plan/plan_test.go
```diff
@@ -1017,3 +1046,135 @@ func (s *S) TestMarshalLayer(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(string(out), Equals, string(layerBytes))
 }
+
+var svcArgsTests = []struct {
```
benhoyt, line 1050: I think these tests will become significantly simpler when moved out of plan. I think we can mostly test the `GetCommand()` (or whatever we call it) function in isolation as a pure function.

cmd/pebble/cmd_run.go
```diff
@@ -166,6 +168,20 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
 
 	logger.Debugf("activation done in %v", time.Now().Truncate(time.Millisecond).Sub(t0))
 
+	if rcmd.Args != nil {
+		mappedArgs, err := convertArgs(rcmd.Args)
+		if err != nil {
+			return err
+		}
+		servopts := client.ServiceOptions{Args: mappedArgs}
+		changeID, err := rcmd.client.PassServiceArgs(&servopts)
```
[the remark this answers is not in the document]
benhoyt: I realize that's how we do it for auto-starting services (below). That is a little different as we want that to be async. But still, I'm not quite sure why we're doing it there either. I've asked @niemeyer on Mattermost [here](https://chat.charmhub.io/charmhub/pl/ofsb9rpn938t8d5qnnr5zw11sa) ... but if we want to change that it can be a separate issue/PR.

internal/plan/plan.go
benhoyt: We could add this to our share x-go library?

benhoyt (changes requested):
This looks like a good first pass, thanks! I've left a bunch of comments/suggestions, with two main structural concerns:

1) I think we can simplify significantly just by avoiding the API/client stuff, and pass `--args` to the daemon (and from there, to the service manager) directly.
2) I think we should separate out the `CmdArgs` handling from the plan, as it's not really related to the plan, but a separate override of the plan. I think it'll clean it up and probably simplify to extract this.

internal/plan/plan.go
rebornplusplus: Thanks for the implementation. Yes, allowing `[ ... ]` at the very end was by design. 
I was initially considering support for `echo foo [ bar ] baz` as well, but had a quick chat within the team and `[ ... ]` at the end seemed to be the spec.
Let me know if you think `echo foo [ bar ] baz` would be a better idea and we can chat about it.
Cool implementation as well, thanks! I am adding some replies in the `checkCommand` comments.

internal/plan/plan.go
rebornplusplus: Yeah, something like `shlex.Join` would be nice indeed. I will try to write that up, and probably make a PR to x-go. Thanks.
I am ditching the `GetCommandStr` function, it was indeed a means for somewhat better logging.

internal/plan/plan.go
rebornplusplus: Yes, your code does indeed handle it. 
Mine does as well, but I guess, quite subtly.
```
	if leftCnt > 0 || rightCnt > 0 {
		if leftCnt != 1 || rightCnt != 1 || args[len(args)-1] != "]" {
			return fmt.Errorf("bad syntax regarding optional/overridable arguments")
		}
	}
```
Here, if `]` somehow comes before `[` and the count of them is still 1, the `args[len(args)-1] != "]"` check will fail. It's not quite transparent nor intuitive, sorry about that. Tried to be a bit cheeky.
I did plan on writing something like your proposed function, but couldn't resist to be a bit cheeky. :stuck_out_tongue: 
I will write it up nicely, using references from your code, thanks!

internal/plan/plan.go
rebornplusplus: Thanks!

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
 	return reflect.DeepEqual(s, other)
 }
 
+// GetCommand overrides the default arguments (inside [ ]) if
+// CmdArgs is present and returns the command as a stream of strings.
+func (s *Service) GetCommand() ([]string, error) {
+	args, err := shlex.Split(s.Command)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse service %q command: %s", s.Name, err)
```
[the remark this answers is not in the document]
rebornplusplus: Oh, thank you.

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
...
+	}
+	fargs := make([]string, 0)
```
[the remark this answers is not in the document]
rebornplusplus: Thanks!

internal/overlord/servstate/handlers.go
```diff
@@ -386,7 +383,7 @@ func (s *serviceData) startInternal() error {
 	s.cmd.Stderr = logWriter
 
 	// Start the process!
-	logger.Noticef("Service %q starting: %s", serviceName, s.config.Command)
+	logger.Noticef("Service %q starting: %v", serviceName, s.config.GetCommandStr(args))
```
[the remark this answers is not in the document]
rebornplusplus: So I was printing the slice before with `%v` and forgot to change that back. Thanks for picking this up!

cmd/pebble/cmd_run.go
```diff
@@ -199,3 +215,26 @@ out:
...
+
+// convert args from [][]string type to map[string][]string
+// and check for empty or duplicated --args usage
+func convertArgs(args [][]string) (map[string][]string, error) {
+	mappedArgs := make(map[string][]string)
+
+	for _, arg := range args {
+		switch len(arg) {
+		case 0:
+			return nil, fmt.Errorf("--args cannot have empty arguments")
```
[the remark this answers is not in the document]
rebornplusplus: Right, that's more tidy. Thanks.

cmd/pebble/cmd_run.go
```diff
@@ -166,6 +168,20 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal) error {
...
+		mappedArgs, err := convertArgs(rcmd.Args)
+		if err != nil {
+			return err
+		}
+		servopts := client.ServiceOptions{Args: mappedArgs}
+		changeID, err := rcmd.client.PassServiceArgs(&servopts)
+		if err != nil {
+			return fmt.Errorf("cannot pass arguments to services: %v", err)
+		} else {
+			logger.Noticef("Passed arguments to services with change %s.", changeID)
```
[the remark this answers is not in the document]
rebornplusplus: Right, thank you. Will leave the logging out for this.

cmd/pebble/cmd_run.go
```diff
@@ -52,6 +53,7 @@ func init() {
 			"hold":        "Do not start default services automatically",
 			"http":        `Start HTTP API listening on this address (e.g., ":4000")`,
 			"verbose":     "Log all output from services to stdout",
+			"args":        `Provide arguments to service (e.g., "SERVICE ARGS.. ; ")`,
```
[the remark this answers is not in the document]
rebornplusplus: That is a better plan, I agree. 

I am just wondering if we can provide arguments to **startup-enabled** services only. :thinking:

internal/plan/plan.go
benhoyt: I do think the `[ ... ]` syntax makes me think it'd work anywhere in the command, because there's a start and an end character. If we have an arbitrary restriction ("it only works at the end") then we have document that as a restriction. Unless there's a good reason to restrict it to only the end, I don't think we should restrict it (I thought it was a bug).

cjdcordeiro: Indeed, optional args were considered only at the end, as a sort of analogy to ENTRYPOINT + CMD.

I agree that it could work anywhere, and while this initial design choice falls on the conservative side, I do not see any major problems about allowing it to be anywhere in the `command`. I'm a bit sceptical about allowing `[...]` to be a prefix though, but apart from that, y, I guess it could be anywhere in the code

benhoyt: Okay, that's fine -- if we're going that route let's just make it clear in the help. We can always make it less conservative later if a need arises (but harder to go the other way!).

Oh, that reminds me, we should also update the `command` field in README.md under [Layer specification](https://github.com/canonical/pebble/#layer-specification) to mention the new `[ ... args ... ]` feature.

client/services.go
```diff
@@ -25,50 +25,64 @@ import (
...
+	services    []string
+	serviceArgs map[string][]string
+}
+
 type multiActionData struct {
-	Action   string   `json:"action"`
-	Services []string `json:"services"`
+	Action      string              `json:"action"`
+	Services    []string            `json:"services"`
+	ServiceArgs map[string][]string `json:"service-args"`
```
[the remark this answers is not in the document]
rebornplusplus: Agreed. But this will be gone, since we are avoiding api calls altogether.

client/services.go
```diff
@@ -25,50 +25,64 @@ import (
 
 type ServiceOptions struct {
 	Names []string
+	Args  map[string][]string
```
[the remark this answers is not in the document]
rebornplusplus: Thanks. But this will not be needed, since I am trying to pass the service arguments from the daemon itself.

client/services.go
```diff
@@ -25,50 +25,64 @@ import (
...
 // since they were started. opts.Names must be empty for this call.
 func (client *Client) Replan(opts *ServiceOptions) (changeID string, err error) {
-	_, changeID, err = client.doMultiServiceAction("replan", opts.Names)
+	_, changeID, err = client.doMultiServiceAction("replan", &multiActionOptions{services: opts.Names})
 	return changeID, err
 }
 
+// PassServiceArgs sets the service arguments provided by "pebble run --args"
+func (client *Client) PassServiceArgs(opts *ServiceOptions) (changeID string, err error) {
+	_, changeID, err = client.doMultiServiceAction("pass-args", &multiActionOptions{serviceArgs: opts.Args})
```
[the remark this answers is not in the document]
rebornplusplus: Thank you! That would be better, yes.

cmd/pebble/cmd_run.go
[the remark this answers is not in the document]
rebornplusplus: Took your advice and passing via daemon. Thank you.
Also I don't seem to have the access to the charmhub.io mattermost server. :confused:

cmd/pebble/cmd_run.go
```diff
@@ -199,3 +215,26 @@ out:
...
+		switch len(arg) {
+		case 0:
+			return nil, fmt.Errorf("--args cannot have empty arguments")
+		case 1:
+			return nil, fmt.Errorf("cannot pass empty arguments to service %q", arg[0])
+		}
+
+		name := arg[0]
+		if _, ok := mappedArgs[name]; ok {
+			return nil, fmt.Errorf("cannot pass args twice to service %q", name)
```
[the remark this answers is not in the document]
rebornplusplus: It might be handy that way, yeah. Let's chat about this on mattermost along with a few other confusions.

internal/plan/plan.go
```diff
@@ -65,6 +65,7 @@ type Service struct {
 	Startup     ServiceStartup `yaml:"startup,omitempty"`
 	Override    Override       `yaml:"override,omitempty"`
 	Command     string         `yaml:"command,omitempty"`
+	CmdArgs     []string       `yaml:"-"`
```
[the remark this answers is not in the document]
rebornplusplus: Thanks! It is nicer to not have this in plan, I agree.

cmd/pebble/cmd_run.go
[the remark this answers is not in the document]
rebornplusplus: Hi Ben, I removed the `startup-enabled` part, because right now, arguments can be passed even if they are not startup-enabled. Let me know if you think it should only be passed for startup-enabled services.

cmd/pebble/cmd_run.go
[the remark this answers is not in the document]
rebornplusplus: I did allow this. Thanks.

internal/plan/plan.go
```diff
@@ -184,6 +187,76 @@ func (s *Service) Equal(other *Service) bool {
 	return reflect.DeepEqual(s, other)
 }
 
+// GetCommand overrides the default arguments (inside [ ]) if
+// CmdArgs is present and returns the command as a stream of strings.
+func (s *Service) GetCommand() ([]string, error) {
```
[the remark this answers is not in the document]
rebornplusplus: I didn't move this and kept it here, with an additional parameter however. Helps a tiny bit with plan validation.
Let me know if you feel I should move this to service manager anyway.

internal/plan/plan.go
rebornplusplus: This should be resolved now, but would be glad if you took another look at it. :D

internal/plan/plan.go
rebornplusplus: Added, thanks!

internal/plan/plan_test.go
rebornplusplus: I added the new tests, please take a look.

rebornplusplus:
Hi @benhoyt, I tried to address your comments (they were super helpful!) in https://github.com/canonical/pebble/pull/191/commits/de12f37e97a3dd74f56f83243fae9c54c34c9124. Please take a look and let me know what you think. Thank you!

cmd/pebble/cmd_run.go
```diff
@@ -34,15 +34,25 @@ import (
...
 The run command starts pebble and runs the configured environment.
+
+Additional arguments to a service command can be provided by the
+--args option in the format:
+    --args <service-name> <arguments..> ;
+The arguments will override the default arguments if specified in
+the plan within [ ... ] group. The semi-colon at the end acts as a
+terminator (end-marker) for the --args option, and needs to be
+backslash-escaped if used at the shell. If no other options follow
+after --args, the terminator ";" may be omitted.
```
benhoyt, line 45:
```suggestion
Additional arguments can be provided to a service command with the
--args option, in the following format:

    --args <service-name> <arguments...> ;

If the command field in the service's plan has a [ <default-arguments...> ]
list, the --args arguments will replace those default arguments. If not, the
--args arguments will be appended to the command.

The semicolon terminator at the end of the --args list is only required
if other "run" arguments follow. If used at the shell, the ';' must be
backslash-escaped.
```

cmd/pebble/cmd_run.go
```diff
@@ -52,6 +62,7 @@ func init() {
 			"hold":        "Do not start default services automatically",
 			"http":        `Start HTTP API listening on this address (e.g., ":4000")`,
 			"verbose":     "Log all output from services to stdout",
+			"args":        `Provide arguments to a service`,
```
benhoyt, line 65:
```suggestion
			"args":        `Provide additional arguments to a service`,
```

internal/daemon/daemon.go
```diff
@@ -800,6 +803,9 @@ func New(opts *Options) (*Daemon, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err = ovld.PassServiceArgs(opts.ServiceArgs); err != nil {
+		return nil, err
+	}
```
benhoyt, line 808: Instead of adding a `PassServiceArgs` method that's called after `overlord.New()`, why don't we just *pass the service args* directly to `overlord.New()` as a new parameter?

And the same for `ServiceManager.SetServiceArgs()` -- just pass them down via `servstate.NewManager()`.

internal/overlord/servstate/handlers.go
```diff
@@ -431,6 +428,11 @@ func (s *serviceData) startInternal() error {
 	return nil
 }
 
+// buildCmdArgs returns the service command as a stream of arguments
+func (s *serviceData) buildCmdArgs() ([]string, error) {
+	return s.config.GetCommand(s.manager.serviceArgs[s.config.Name])
```
benhoyt, line 433: Seems like we might as well just do this one-liner inline rather than have a method that just calls another method?

internal/overlord/servstate/manager.go
```diff
@@ -146,6 +148,12 @@ func (m *ServiceManager) appendLayer(layer *plan.Layer) error {
 		return err
 	}
 	layer.Order = newOrder
+
+	// Remove the existing arguments of this layer's services
+	for _, service := range layer.Services {
+		delete(m.serviceArgs, service.Name)
+	}
```
benhoyt, line 155: Yeah, this behaviour seems weird to me. I think we should just keep the service args when a layer is added. They're explicitly an override the user has requested, so seems weird to just drop them. Similar below.

internal/plan/plan.go
```diff
@@ -184,6 +184,61 @@ func (s *Service) Equal(other *Service) bool {
 	return reflect.DeepEqual(s, other)
 }
 
+// GetCommand returns the service command as a stream of strings.
+// It adds the arguments in cmdArgs to the command if no default
+// argument is specified in [ ... ] group in the plan.
+// If default arguments are specified in the plan, the default
+// arguments are added to the command if cmdArgs is empty.
+// If not empty, the default arguments are overrided by arguments in cmdArgs.
+func (s *Service) GetCommand(cmdArgs []string) ([]string, error) {
```
benhoyt, line 193: Nit: I'd probably prefer this as a stand-alone helper function (not a method), so it's more explicit about the dependencies, and so callers don't necessarily need a `Service` instance lying around. What about this:

```go
func CommandArgs(command string, additionalArgs []string) ([]string, error) { ... }
```

internal/plan/plan.go
```diff
@@ -184,6 +184,61 @@ func (s *Service) Equal(other *Service) bool {
...
+	args, err := shlex.Split(s.Command)
+	if err != nil {
+		return nil, err
+	}
+
+	var inBrackets, gotBrackets bool
+	var fargs []string
```
benhoyt, line 200: Nit: "fargs" doesn't mean much to me. How about `result`?

benhoyt (approved):
Looking much better, thanks! I'm approving but with a few minor suggestions for some text clarity and a simplification about how you're passing the args down to overlord and service manager.

internal/daemon/daemon.go
rebornplusplus: Changed it, thanks. Updated the tests as well.

internal/overlord/servstate/handlers.go
rebornplusplus: That works as well.

internal/overlord/servstate/manager.go
rebornplusplus: I am keeping this deletion, per our chat in Mattermost.

benhoyt (approved):
Excellent, thanks for that -- looks good to me now.

internal/plan/plan_test.go
```diff
@@ -1017,3 +1064,65 @@ func (s *S) TestMarshalLayer(c *C) {
...
+
+func (s *S) TestGetCommand(c *C) {
+	cmdTests := []struct {
+		summary         string
+		command         string
+		cmdArgs         []string
+		expectedCommand []string
+		error           string
+	}{
+		{
```
niemeyer, line 1076: Let's please join these two lines so we get `}{{`, and move the slice out of the function, so we gain two indentation level for everything below and keep the function small. See the example of planTests above for reference.

internal/plan/plan_test.go
```diff
@@ -1017,3 +1064,65 @@ func (s *S) TestMarshalLayer(c *C) {
...
+			cmdArgs:         []string{"-v", "--opt"},
+			expectedCommand: []string{"cmd", "--foo", "bar", "-v", "--opt"},
+		}, {
+			summary: "[ ... ] should be a suffix",
+			command: "cmd [ --foo ] --bar",
+			error:   `cannot have any args after \[ ... \] group`,
+		}, {
+			summary: "[ ... ] should not be prefix",
+			command: "[ cmd --foo ]",
+			error:   `cannot have \[ ... \] group as prefix`,
```
niemeyer, line 1110: Thanks for the comprehensive testing.

niemeyer (changes requested):
Thanks for this Rafid. It's looking very promising.

Here is a first pass:

internal/plan/plan.go
```diff
@@ -184,6 +184,61 @@ func (s *Service) Equal(other *Service) bool {
 	return reflect.DeepEqual(s, other)
 }
 
+// CommandArgs returns a service command as a stream of strings.
+// It adds the arguments in cmdArgs to the command if no default
+// argument is specified in [ ... ] group in the plan.
+// If default arguments are specified in the plan, the default
+// arguments are added to the command if cmdArgs is empty.
+// If not empty, the default arguments are overrided by arguments in cmdArgs.
+func CommandArgs(command string, cmdArgs []string) ([]string, error) {
```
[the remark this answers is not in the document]
niemeyer: The suggestion is as outlined above. If you have a different idea we can talk about it, of course, but intuitively that felt like a reasonable design.

rebornplusplus:
Hi @niemeyer, thanks for your insightful comments. I have added a commit addressing them. Will you please take another look?

Additionally, the new commit will require the PR https://github.com/canonical/x-go/pull/19 to go in. Will you please take a look there as well? That one is awaiting one more approval.

rebornplusplus:
Also, I have added a PR in https://github.com/canonical/go-flags/pull/2 which will be needed for this PR as well. It will be really helpful if you can check it as well.

internal/overlord/servstate/manager.go
```diff
@@ -481,6 +485,44 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+			if _, ok := layers[i].Services[serviceName]; ok {
+				layer = layers[i]
+				break
+			}
+		}
+		if layer == nil {
+			return fmt.Errorf("service %q not found in plan", serviceName)
+		}
+
+		service := layer.Services[serviceName]
```
woky, line 515: Nitpick: To avoid extra `layer.Services` lookup, could be:
```
		var service *plan.Service

		// search for the topmost layer which contains the service
		// assuming the layers are sorted by order ASC
		for i := len(layers) - 1; i >= 0; i-- {
			if service = layers[i].Services[serviceName]; service != nil {
				break
			}
		}
		if service == nil {
			return fmt.Errorf("service %q not found in plan", serviceName)
		}
```

internal/plan/plan.go
```diff
@@ -184,6 +184,59 @@ func (s *Service) Equal(other *Service) bool {
...
+			}
+			extra = append(extra, arg)
+			continue
+		}
+		if gotBrackets {
+			return nil, nil, fmt.Errorf("cannot have any args after [ ... ] group")
+		}
+		if arg == "[" {
+			if idx == 0 {
+				return nil, nil, fmt.Errorf("cannot have [ ... ] group as prefix")
```
woky, line 215: I would say just `"command cannot be ["` here.

However, note:
```
% whereis -b [ 
[: /usr/bin/[ /opt/toybox/bin/[
```
But I think we can just rule out support for `/usr/bin/[`. :-)

woky (approved):
Nice. LGTM.

internal/plan/plan.go
rebornplusplus: Technically, `/usr/bin/[` will still be supported if `[` by itself is not an argument. `[5 -gt 4]`, `/usr/bin/[ 5 -gt 4]`, `[[ 5 > 4]]` will all work. What won't work is `[ 5 -gt 4 ]`.

That being said, I think the much more frequent case would be people adding all their commands inside `[ ... ]` to make it all overridable rather than people using `[` command. So I think the "cannot have [ ... ] group as prefix" error would be more relevant? Let me know if you think otherwise and I will change so.

internal/overlord/servstate/manager.go
rebornplusplus: Updated, thanks!

internal/plan/plan.go
woky: So, in your examples, only `/usr/bin/[ 5 -gt 4]` would work. `[5 -gt 4]` and `[[ 5 > 4 ]]` would look for `[5` and `[[` executable in `$PATH` respectively, which most likely doesn't exist. However, we can just rule out `[` as a service. :-) (Or suggest `test` as alternative.)

I find the message a bit confusing but maybe it's just me, so it's up to you (and others).

cjdcordeiro: a less technical but more accurate message would be: `command cannot be started with default arguments`

rebornplusplus: Updated, thanks.

internal/daemon/daemon.go
```diff
@@ -780,6 +780,19 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
...
 
+// PassServiceArgs passes the provided service arguments to
+// the service manager responsible for services under daemon overlord.
+func (d *Daemon) PassServiceArgs(serviceArgs map[string][]string) error {
+	if d.overlord == nil {
+		return fmt.Errorf("internal error: no Overlord")
+	}
+	serviceMgr := d.overlord.ServiceManager()
+	if serviceMgr == nil {
+		return fmt.Errorf("internal error: no Service Manager")
```
niemeyer, line 791: Same here.  The ServiceManager is "contractually" defined to exist there. If it doesn't, everything is _really_ broken and the panic will help us more than the error message.

go.mod
```diff
@@ -9,8 +9,12 @@ require (
...
-	golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4
+	golang.org/x/sys v0.0.0-20210320140829-1e4c9ba3b0c4
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
 	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
 	gopkg.in/yaml.v3 v3.0.1
 )
+
+replace github.com/jessevdk/go-flags v1.4.0 => github.com/canonical/go-flags v0.0.0-20230327193835-01157a7cbb96
+
+replace github.com/canonical/x-go => github.com/rebornplusplus/canonical-x-go v0.0.0-20230309051359-1ec2b2ee698e
```
niemeyer, line 20: Changes to go.mod and go.sum need to be reverted before the PR is final.

internal/overlord/servstate/manager.go
```diff
@@ -115,9 +115,10 @@ func (m *ServiceManager) Plan() (*plan.Plan, error) {
 	return m.plan, nil
 }
 
-// AppendLayer appends the given layer to the plan's layers and updates the
-// layer.Order field to the new order. If a layer with layer.Label already
-// exists, return an error of type *LabelExists.
+// AppendLayer appends the given layer to the plan's layers, updates the
+// layer.Order field to the new order and removes the existing arguments
+// of this layer's services. If a layer with layer.Label already exists,
+// return an error of type *LabelExists.
```
niemeyer, line 121: This looks like a left over from prior changes, and needs to be reverted (including the spare line below).

internal/overlord/servstate/manager.go
```diff
@@ -176,7 +178,8 @@ func findLayer(layers []*plan.Layer, label string) (int, *plan.Layer) {
 
 // CombineLayer combines the given layer with an existing layer that has the
 // same label. If no existing layer has the label, append a new one. In either
-// case, update the layer.Order field to the new order.
+// case, update the layer.Order field to the new order and remove existing
+// arguments of this layer's services.
```
niemeyer, line 182: Same here, and below.

internal/overlord/servstate/manager.go
```diff
@@ -481,6 +485,42 @@ func (m *ServiceManager) CheckFailed(name string) {
 	}
 }
 
+// SetServiceArgs sets the service arguments provided by "pebble run --args"
+// to their respective services. It updates the service command in the top most
+// layer which contains the service in the plan.
```
niemeyer, line 490: This is a lot closer, but still disagrees with what was suggested in the previous review and during our conversationo. The problem here is that it is still corrupting the plan and modifying something completely unrelated in a way that will be misleading to those inspecting the system. It also means that if someone takes off their own layer later, they'll mistakingly remove the extra arguments unexpectedly.

That's the sort of reason why the suggestion was to _append_ an additional layer that contains the arguments changed, in a clear and clean way.

niemeyer (changes requested):
Thanks for the changes, Rafid, this is much closer.

Here is a second pass:

cmd/pebble/cmd_run.go
```diff
@@ -34,15 +34,30 @@ import (
...
+    --args <service-name> <arguments...> ;
+
+If the command field in the service's plan has a [ <default-arguments...> ]
+list, the --args arguments will replace those default arguments. If not, the
+--args arguments will be appended to the command.
+
+The optional --args flags is followed by a service name and its arguments, and must be
+terminated by ";" unless there are no further Pebble options. For example:
+
+    pebble run --args myservice --port=8080 \; --hold
```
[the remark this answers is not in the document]
rebornplusplus: Thanks, done.

internal/daemon/daemon.go
```diff
@@ -780,6 +780,19 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 	return errExpectedReboot
 }
 
+// PassServiceArgs passes the provided service arguments to
+// the service manager responsible for services under daemon overlord.
+func (d *Daemon) PassServiceArgs(serviceArgs map[string][]string) error {
+	if d.overlord == nil {
+		return fmt.Errorf("internal error: no Overlord")
```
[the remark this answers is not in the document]
rebornplusplus: Thanks for the insight. Removed this one and the next as well.

go.mod
rebornplusplus: For sure, I will do that once https://github.com/canonical/x-go/pull/19 goes in. I intend to squash the commits as well, before merging. Once the changes are approved, I will get going on these.

internal/overlord/servstate/manager.go
rebornplusplus: Right, missed it. Thanks for picking up on it!

internal/overlord/servstate/manager.go
rebornplusplus: Thank you!

internal/overlord/servstate/manager.go
rebornplusplus: Ah okay. Perhaps I misunderstood a bit. I have updated this in https://github.com/canonical/pebble/pull/191/commits/8e6ef62bef2c5afd5f3bce225a5d0995bc0e0d71. Added a new layer as you suggested.

rebornplusplus:
Note about the failing "Tests / Root Tests": It passes in my GitHub repo and local workstation. https://github.com/rebornplusplus/pebble/actions/runs/4675467143 (successful GH action for the same set of commits)

woky (approved):
@niemeyer @rebornplusplus @cjdcordeiro The test failure was not fault of this PR. Now all checks are passing.

LGTM!

internal/overlord/servstate/manager.go
```diff
@@ -513,6 +513,38 @@ func (m *ServiceManager) CheckFailed(name string) {
 	}
 }
 
+// SetServiceArgs sets the service arguments provided by "pebble run --args"
+// to their respective services. It adds a new layer in the plan, the layer
+// consisting of services with commands having their arguments changed.
+func (m *ServiceManager) SetServiceArgs(serviceArgs map[string][]string) error {
```
benhoyt, line 519: Let's try to add a test for this in `manager_test.go` to ensure we're exercising and testing this code.

benhoyt (changes requested):
Thanks! Two small requests: 1) label, and 2) test `SetServiceArgs`

go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
 	gopkg.in/yaml.v3 v3.0.1
 )
+
+replace github.com/canonical/x-go => github.com/rebornplusplus/canonical-x-go v0.0.0-20230309051359-1ec2b2ee698e
```
benhoyt, line 18: Note the use of Rafid's fork+branch here -- we'll need to merge the `shlex.Quote` and `shlex.Join` PR (https://github.com/canonical/x-go/pull/19) to avoid this.

internal/overlord/servstate/manager.go
```diff
@@ -513,6 +513,38 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+	releasePlan, err := m.acquirePlan()
+	if err != nil {
+		return err
+	}
+	defer releasePlan()
+
+	newLayer := &plan.Layer{
+		Services: make(map[string]*plan.Service),
```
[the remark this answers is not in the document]
rebornplusplus: I set the label to "pebble-service-args" as @niemeyer suggested during our chat.

internal/overlord/servstate/manager.go
rebornplusplus: Done, thanks!

internal/overlord/servstate/manager.go
```diff
@@ -524,6 +524,8 @@ func (m *ServiceManager) SetServiceArgs(serviceArgs map[string][]string) error {
 	defer releasePlan()
 
 	newLayer := &plan.Layer{
+		// TODO: Consider making "pebble-service-args" a reserved label.
```
benhoyt, line 527: Or more generally, consider making any labels starting with the `pebble-` prefix reserved.

benhoyt (approved):

woky (approved):

go.mod
benhoyt: @rebornplusplus Does this need to be updated to 61c9f268b5f0?

go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
 	gopkg.in/yaml.v3 v3.0.1
 )
+
+replace github.com/canonical/x-go => github.com/rebornplusplus/canonical-x-go v0.0.0-20230425082201-61c9f268b5f0
```
niemeyer, line 18: This needs to be dropped.

go.sum
```diff
@@ -13,6 +11,8 @@ github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/pkg/term v1.1.0 h1:xIAAdCMh3QIAy+5FrE8Ad8XoDhEU4ufwbaSozViP9kk=
 github.com/pkg/term v1.1.0/go.mod h1:E25nymQcrSllhX42Ok8MRm1+hyBdHY0dCeiKZ9jpNGw=
+github.com/rebornplusplus/canonical-x-go v0.0.0-20230425082201-61c9f268b5f0 h1:LuYZQAYCYeQSmws6pZGIfAbr8fg+zSATmmyLNrjPAas=
+github.com/rebornplusplus/canonical-x-go v0.0.0-20230425082201-61c9f268b5f0/go.mod h1:A0/Jvt7qKuCDss37TYRNXSscVyS+tLWM5kBYipQOpWQ=
```
niemeyer, line 15: Same. Needs reverting.

internal/overlord/servstate/manager.go
niemeyer: Indeed, the latter. Let's please add this in a separate PR, @rebornplusplus.

internal/overlord/servstate/manager.go
```diff
@@ -513,6 +513,42 @@ func (m *ServiceManager) CheckFailed(name string) {
...
+		if err != nil {
+			return err
+		}
+		newLayer.Services[name] = &plan.Service{
+			Override: plan.MergeOverride,
+			Command:  plan.CommandString(base, args),
+		}
+	}
+
+	return m.appendLayer(newLayer)
```
niemeyer, line 549: Logic above looks great now, thanks.

internal/plan/plan.go
```diff
@@ -211,6 +211,59 @@ func (s *Service) Equal(other *Service) bool {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var inBrackets, gotBrackets bool
+
+	for idx, arg := range args {
+		if inBrackets {
+			if arg == "[" {
+				return nil, nil, fmt.Errorf("cannot nest [ ... ] groups")
```
niemeyer, line 228: The errors in this function lack context. That is, they need to say "cannot parse service %q command: ..." as a prefix so that when the error pops up in the user interface it is understandable. Typically we do that with a defer at the top that works on the global err variable so that we don't need to repeat that every single time.

internal/plan/plan.go
```diff
@@ -211,6 +211,59 @@ func (s *Service) Equal(other *Service) bool {
...
+			}
+			if arg == "]" {
+				inBrackets = false
+				continue
+			}
+			extra = append(extra, arg)
+			continue
+		}
+		if gotBrackets {
+			return nil, nil, fmt.Errorf("cannot have any args after [ ... ] group")
```
niemeyer, line 238: s/args/arguments/, as below.

internal/plan/plan.go
```diff
@@ -211,6 +211,59 @@ func (s *Service) Equal(other *Service) bool {
...
+		}
+		if arg == "[" {
+			if idx == 0 {
+				return nil, nil, fmt.Errorf("command cannot be started with default arguments")
```
niemeyer, line 242: `cannot start command with [ ... ] group`, as above and below.

niemeyer (changes requested):
Looking good now, thank you.

Probably the last pass:

go.mod
rebornplusplus: Dropped it, thanks. Please note that the tests are failing since `shlex.Join()` is missing. I will update Line 7 (quoted below) as soon as https://github.com/canonical/x-go/pull/19 gets merged.
```go
	github.com/canonical/x-go v0.0.0-20230113154138-0ccdb0b57a43
```

internal/overlord/servstate/manager.go
rebornplusplus: Thanks for the help!

internal/plan/plan.go
rebornplusplus: Added in the new commit https://github.com/canonical/pebble/pull/191/commits/d686b6f3a76b21b1cde43b4565dc1fcb565aafcc.

internal/plan/plan.go
rebornplusplus: Done, thanks.

internal/plan/plan.go
rebornplusplus: Done.

internal/overlord/servstate/manager.go
rebornplusplus: Created https://github.com/canonical/pebble/issues/220 to track it and will add a PR following this. Thanks.

go.mod
rebornplusplus: Updated to the new version in https://github.com/canonical/pebble/pull/191/commits/43ef4ef1b60f77561f8da9b016fdf81f412f34b3.

go.mod
rebornplusplus: The `shlex.Quote` and `shlex.Join` PR (https://github.com/canonical/x-go/pull/19) PR has been merged recently.
I updated the module version in https://github.com/canonical/pebble/pull/191/commits/43ef4ef1b60f77561f8da9b016fdf81f412f34b3.

woky (approved):
LGTM

niemeyer (approved):
Thank you!  Great to see this going in.

PR #192, opened 2023-02-21
Add `partinfo` with support for vfat and ext4

anpep:
The `partinfo` package adds support for enumerating the partitions from multiple disks available through sysfs. It also features the ability to read the partition label from the filesystem superblock on vfat and ext4. See [this doc](https://docs.google.com/document/d/1S1AXOUBrsoLEIww39Oy0lBItbUlnB_3-d2SAbekUqTA/edit?usp=share_link) for more on why we need this.

We could have used some pre-existing solutions from the Go ecosystem (see [diskfs/go-diskfs](https://github.com/diskfs/go-diskfs)), however these lack support for filesystems such as Ext, and in general have too much logic that we don't really need (e.g. parsing MBR/GPT partition table--the kernel already does that for us, support for filesystems other than vfat/ext4 which are the only ones we need currently or support for anything that's not reading the partition label).

In the end, this PR is intended to have a very low footprint, and existing solutions don't meet that requirement.


benhoyt:
I haven't looked at the rest of this PR, but the binaries total 1.5MB. I wonder if there's a way we can avoid adding quite this much to the repo (for everyone), especially if this package is going to be moved to `x-go` soon?

atesburak:
We should look to either generate the test files programmatically or decompress from an archive as a test setup step to save space in the repository.

anpep:
> I haven't looked at the rest of this PR, but the binaries total 1.5MB. I wonder if there's a way we can avoid adding quite this much to the repo (for everyone), especially if this package is going to be moved to `x-go` soon?

> We should look to either generate the test files programmatically or decompress from an archive as a test setup step to save space in the repository.

Generating the test files programatically would take a lot of precious build time. What can be done is to include only the first N bytes of the disk images instead of the full images. Do you both think is an acceptable compromise?

I really see no better way of testing this than including some (smaller) test data.

benhoyt:
> What can be done is to include only the first N bytes of the disk images instead of the full images. Do you both think is an acceptable compromise?

Yeah, this seems fine. Now it's only a few KB of binary data, which is great. Thanks for the fix.

internal/partinfo/ext.go
```diff
@@ -0,0 +1,46 @@
...
+	Label [16]byte // Volume name.
+}
+
+func (sb *extSuperblock) isValid() bool {
+	return sb.Magic == extMagic
+}
+
+func newExtSuperblock(r io.Reader) (*extSuperblock, error) {
+	sb := &extSuperblock{}
+	if err := binary.Read(r, binary.LittleEndian, sb); err != nil {
```
benhoyt, line 39: Huh, I didn't know about `encoding/binary.Read` -- that's cool.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+type FSType string
+
+const (
+	FAT32 FSType = "fat32"
+	Ext   FSType = "ext"
+)
+
+type PartInfo interface {
```
benhoyt, line 35: Normally in Go an interface would only be defined where it's used (i.e., where you're depending on the thing), with the subset of methods that you need. So this package would define `struct Partition` and then callers could use that directly (or define an interface and let `Partition` fulfil that if they needed to).

Interestingly, here `partition` doesn't even seem to implement this interface. Is that intentional?

Additionally, a naming thing: normally in Go you don't use the `Get` prefix on method names, so for example these would be `DevNode()` etc.

internal/partinfo/ext.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, fmt.Errorf("cannot read ext2/3/4 superblock: %w", err)
+	}
+	if !sb.isValid() {
+		return nil, errors.New("cannot read ext2/3/4 superblock: not an ext2/3/4 partition")
```
benhoyt, line 43: Normally with errors you don't include the context of this function here (that would be further up). So here you'd just return `errors.New("not an ext2/3/4 partition")` -- and similar for other cases.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+type partition struct {
+	path   string
+	label  string
+	fsType FSType
+}
+
+func (b *partition) parseSuperblock() error {
+	f, err := os.OpenFile(b.path, os.O_RDONLY, 0)
+	if err != nil {
+		return fmt.Errorf("cannot parse superblock: %w", err)
```
benhoyt, line 50: Once again, you don't need to prefix all of these with `cannot parse superblock` -- the is normally done by the caller. However, here you would/could provide the context of the operation you're performing, for example `fmt.Errorf("cannot open superblock: %w", err)`.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	defer f.Close()
+
+	// Attempt to read first sector of the partition.
+	s := make([]byte, 512)
+	n, err := f.Read(s)
```
benhoyt, line 56: This might be a good use case for `io.ReadFull`. Similar below.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot parse superblock: %w", err)
+	}
+	if n != len(s) {
+		return fmt.Errorf("cannot parse superblock: cannot read first sector")
+	}
+
+	// Attempt to parse FAT32 superblock
+	if sb, err := newFat32Superblock(bytes.NewReader(s)); err == nil {
+		b.label = strings.TrimSpace(string(sb.Label[:]))
```
benhoyt, line 66: It's mildly annoying the caller has to do this trimming instead of the specific function. I wonder if it'd be better if `newFat32Superblock` internally parsed into a struct like the one you have, but then returned a "higher level" struct with only the fields you need, and the label as a Go string. Not sure.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+		b.fsType = FAT32
+		return nil
+	}
+
+	// Skip the 2-sector padding of ext2/3/4
+	if _, err := f.Seek(1024, io.SeekStart); err != nil {
+		return fmt.Errorf("cannot parse superblock: %w", err)
```
benhoyt, line 73: Similar to above: `fmt.Errorf("cannot skip superblock padding: %w", err)` or similar.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+		b.label = strings.TrimRight(string(sb.Label[:]), "\x00")
+		b.fsType = Ext
+		return nil
+	}
+
+	return errors.New("cannot parse superblock: unrecognized file system")
+}
+
+var SysfsPath = "/sys"
+var DevfsPath = "/dev"
```
benhoyt, line 96: Do these need to be capitalized/exported? It looks like they're only exported for tests, but the tests are in the same package, so that shouldn't be needed. (And it's better to use the `export_test.go` pattern if that's needed).

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, err
+	}
+	disks := make([]string, 0)
+	for _, d := range dirs {
+		disks = append(disks, path.Join(base, d.Name()))
+	}
+	return disks, nil
+}
+
+func EnumeratePartitions() ([]partition, error) {
```
benhoyt, line 111: Could this just be called `Enumerate` given that it's in the `partinfo` package? The `Partitions` suffix seems redundant.

internal/partinfo/partinfo.go
```diff
@@ -0,0 +1,136 @@
...
+	disks, err := enumerateDisks()
```
benhoyt, line 112: I think you're not gaining much by factoring this out -- it's more complicated and it means creating an intermediate list. IMO just to the ReadDir inline here and avoid the extra function / for loop / intermediate list. You're also creating the joined path in `enumerateDisks` and then recreating the base name with `path.Base` below (I think that could be avoided too).

benhoyt (changes requested):
Structure and approach looks good to me. I've left a few comments about Go coding style: naming, interfaces, errors and such.

internal/partinfo/ext.go
```diff
@@ -0,0 +1,68 @@
...
+func (p extPartition) FSType() string {
+	// TODO: Could this lead to issues when attempting to mount ext2/3 partitions?
+	return "ext4"
+}
+
+func (p extPartition) Label() string {
+	return strings.TrimRight(string(p.superblock.VolName[:]), "\x00")
+}
+
+func newExtPartition(f *os.File) (Partition, error) {
```
benhoyt, line 53: I realize these are internal functions, but it'd be slightly better (for testability etc) if these functions took an [`io.ReadSeeker`](https://pkg.go.dev/io#ReadSeeker) instead of a concrete `*os.File` (which implements `io.ReadSeeker`).

Similarly, it should probably return a concrete `extPartition` (which implements `Partition`). The pattern in Go is very often "accept interfaces, return structs".

internal/partinfo/fat32.go
```diff
@@ -0,0 +1,70 @@
...
+func (p fat32Partition) FSType() string {
+	return "vfat"
+}
+
+func (p fat32Partition) Label() string {
+	if s := strings.TrimSpace(string(p.superblock.Label[:])); s != emptyLabel {
+		return s
+	} else {
+		return ""
+	}
```
benhoyt, line 55: Nit: early returns and avoiding the `else` / putting the happy path up an indentation level is customary in Go. So I'd suggest:

```suggestion
	label := strings.TrimSpace(string(p.superblock.Label[:]))
	if s == emptyLabel {
		return ""
	}
	return label
```

benhoyt (approved):
Nice. Leaving an Approve (just a couple of nit comments).

internal/partinfo/ext.go
anpep: https://github.com/anpep/pebble/blob/05f9184528ed8c905aff6a00c61258fdc0101a94/internal/partinfo/ext.go#L40-L42

benhoyt: I guess you could pass the path in separately. But yeah, this is fine -- concrete is simpler if we can get away with it. Files are easy to create in tests anyway. :-)

atesburak (approved):

internal/partinfo/partition.go
```diff
@@ -0,0 +1,97 @@
...
+			if _, err := f.Seek(0, io.SeekStart); err != nil {
+				return nil, fmt.Errorf("cannot seek: %w", err)
+			}
+			continue
+		} else {
+			return p, nil
+		}
+	}
+
+	return nil, errors.New("unrecognized partition")
```
flotter, line 60: Perhaps "unsupported partition type" ? We know its fine, otherwise the kernel would not show it.

internal/partinfo/partition.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+var (
+	sysfsPath = "/sys"
+	devfsPath = "/dev"
+)
+
+// Enumerate returns a slice of Partition elements representing the disk partitions that can be mounted.
```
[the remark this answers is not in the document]
anpep: That's exactly what I did at first, but after considering feedback from Ben I simplified the partition enumeration code not to include a second helper function for enumerating disks. I think it's cleaner this way

internal/partinfo/partition.go
```diff
@@ -0,0 +1,97 @@
...
+func Enumerate() ([]Partition, error) {
```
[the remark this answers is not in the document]
anpep: Agreed! Thank you!

internal/partinfo/ext_test.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package partinfo
```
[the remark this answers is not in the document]
anpep: As I understand, you are suggesting I put the tests on a separate package (e.g. `partinfo_test`) and only use exported symbols. Am I wrong?

If that's the case, I don't really see the value in this particular example since I'm testing the internals of the superblock parser implementations (and thus doing black-box testing, for which using a separate package seems inconvenient).

I'm open to discuss this matter further (:

niemeyer:

PR #193, opened 2023-02-21
Implement bootloader interface for GRUB

anpep:
This commits implements the new bootloader interface for GRUB.
For simplicity, slot functionality has been integrated into one interface but we might split it into two interfaces so that the `Bootloader` interface remains compatible with the snapd one.

### DIFF
--- a/internal/bootloader/bootloader.go
+++ b/internal/bootloader/bootloader.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2023 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package bootloader
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Bootloader provides primitives to retain state across boots and handle
+// boot slots.
+type Bootloader interface {
+	// Name returns the bootloader name.
+	Name() string
+
+	// GetBootVars populates the specified variables from the bootloader.
+	// These variables are preserved across reboots.
+	GetBootVars(names ...string) (map[string]string, error)
+
+	// SetBootVars saves a set of variables to be persisted across reboots.
+	SetBootVars(values map[string]string) error
+
+	// Present returns whether the bootloader is currently present on the
+	// system--in other words, whether this bootloader has been installed to
+	// the current system. Implementations should only return non-nil error if
+	// they can positively identify that the bootloader is installed, but there
+	// is actually an error with the installation.
+	Present() (bool, error)
+
+	// GetActiveSlot obtains the label of the currently booted slot.
+	GetActiveSlot() (label string, err error)
+
+	// SetActiveSlot instructs the bootloader to select the slot with the
+	// specified label on the next reboot.
+	SetActiveSlot(label string) error
+
+	// GetStatus obtains the status of the slot with the specified label.
+	// If there is no saved status for the slot, or if the saved status is not
+	// any of Unbootable, Try or Fail, Try will be returned.
+	GetStatus(label string) (Status, error)
+}
+
+// Status represents the conditions in which a boot attempt was made.
+type Status string
+
+const (
+	// Unbootable indicates that the slot cannot be booted from in any case.
+	// For example, this slot might be empty.
+	Unbootable Status = "unbootable"
+
+	// Try indicates that the slot can potentially be booted from.
+	Try Status = "try"
+
+	// Fail indicates that there was a problem preventing a slot from being
+	// booted from.
+	Fail Status = "fail"
+)
+
+// BootloaderMountpoint is the path where the root directory for the current
+// bootloader configuration is mounted on bootstrap.
+const BootloaderMountpoint = "/var/termus/boot"
+
+type BootloaderNewFunc func(rootdir string) Bootloader
+
+var (
+	// bootloaders list all possible bootloaders by their constructor
+	// function.
+	Bootloaders = []BootloaderNewFunc{
+		NewGRUB,
+	}
+)
+
+// Find obtains an instance of the first supported bootloader that is available
+// on the system.
+func Find() (*Bootloader, error) {
+	for _, newBl := range Bootloaders {
+		bl := newBl(BootloaderMountpoint)
+		isPresent, err := bl.Present()
+		if err != nil {
+			return nil, fmt.Errorf("bootloader %q found but not usable: %w", bl.Name(), err)
+		}
+		if isPresent {
+			return &bl, nil
+		}
+	}
+	return nil, errors.New("cannot determine bootloader")
+}
--- a/internal/bootloader/grub.go
+++ b/internal/bootloader/grub.go
@@ -1,0 +1,104 @@
+// Copyright (c) 2023 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package bootloader
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/canonical/pebble/internal/bootloader/grubenv"
+	"github.com/canonical/pebble/internal/osutil"
+)
+
+// GRUB implements the Bootloader interface to support bootloader operations
+// on GRUB 2.
+type GRUB struct {
+	// rootdir is the directory where the GRUB prefix is mounted.
+	rootdir string
+}
+
+// newGrub initializes a new instance of the GRUB bootloader.
+func NewGRUB(rootdir string) Bootloader {
+	return &GRUB{
+		rootdir: rootdir,
+	}
+}
+
+func (g *GRUB) envFile() string {
+	return filepath.Join(g.rootdir, "grubenv")
+}
+
+func (g *GRUB) Name() string {
+	return "grub"
+}
+
+func (g *GRUB) GetBootVars(names ...string) (map[string]string, error) {
+	out := make(map[string]string)
+
+	env := grubenv.NewEnv(g.envFile())
+	if err := env.Load(); err != nil {
+		return nil, err
+	}
+
+	for _, name := range names {
+		out[name] = env.Get(name)
+	}
+
+	return out, nil
+}
+
+func (g *GRUB) SetBootVars(values map[string]string) error {
+	env := grubenv.NewEnv(g.envFile())
+	if err := env.Load(); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	for k, v := range values {
+		env.Set(k, v)
+	}
+	return env.Save()
+}
+
+func (g *GRUB) Present() (bool, error) {
+	doesExist, _, err := osutil.ExistsIsDir(filepath.Join(g.rootdir, "grub.cfg"))
+	return doesExist, err
+}
+
+func (g *GRUB) GetActiveSlot() (string, error) {
+	vars, err := g.GetBootVars("boot.slot")
+	if err != nil {
+		return "", err
+	}
+	return vars["boot.slot"], nil
+}
+
+func (g *GRUB) SetActiveSlot(label string) error {
+	return g.SetBootVars(map[string]string{
+		"boot.slot": label,
+	})
+}
+
+func (g *GRUB) GetStatus(label string) (Status, error) {
+	varName := "boot." + label + ".status"
+	vars, err := g.GetBootVars(varName)
+	if err != nil {
+		return "", err
+	}
+	s := Status(vars[varName])
+	switch s {
+	case Unbootable, Try, Fail:
+		return s, nil
+	}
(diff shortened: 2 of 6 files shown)


benhoyt:
@anpep A couple of things:

1) I don't really have enough context on this at the moment to review this. If you'd like me to review, can you please point me to specs/discussions/info to give background? The PR itself doesn't show how it'll be used in Pebble, as it's just the interface and new code, not wired up.
2) This seems quite low-level and specific. I wonder if this kind of code would be better in a "bootloader" or "rocks" repo, or even x-go, and then imported here via Go modules? (Similar with the [FAT/FS PR](https://github.com/canonical/pebble/pull/192).) I know that's a trade-off, but depending on who needs to review/approve/etc, that approach may help you work more quickly? Let me know if this has already been discussed/decided.

anpep:
> @anpep A couple of things:
> 
> 1. I don't really have enough context on this at the moment to review this. If you'd like me to review, can you please point me to specs/discussions/info to give background? The PR itself doesn't show how it'll be used in Pebble, as it's just the interface and new code, not wired up.
>

Sorry about that, should have included a link to the document in the first place. [I've added it now](https://docs.google.com/document/d/1mWzPjGq6ijntlXjLV6ck4AeZyO_TRzWQSdHEGTbhYBw/edit?usp=share_link).

> 
> 2. This seems quite low-level and specific. I wonder if this kind of code would be better in a "bootloader" or "rocks" repo, or even x-go, and then imported here via Go modules? (Similar with the [FAT/FS PR](https://github.com/canonical/pebble/pull/192).) I know that's a trade-off, but depending on who needs to review/approve/etc, that approach may help you work more quickly? Let me know if this has already been discussed/decided.

It is, and in a similar fashion with the `partinfo` PR, I intend it to be included in `x-go` instead of Pebble. However, since the `x-go` patch is not being merged very soon, I wanted to include these changes in Pebble (specially because these PRs *need* to be merged so that we can implement critical functionality for Termus), and move them out to `x-go` when it's ready (which will be trivial since I'm not adding any dependency to other Pebble packages).

Let me know what you think, I greatly appreciate your feedback!

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (g *Env) Get(name string) string {
+	return g.env[name]
+}
+
+func (g *Env) Set(key, value string) {
+	if !strutil.ListContains(g.ordering, key) {
+		g.ordering = append(g.ordering, key)
+	}
```
benhoyt, line 53: To avoid an O(N) lookup here, can we use the map to check for existence instead?
```suggestion
	if _, present := g.env[key]; !present {
		g.ordering = append(g.ordering, key)
	}
```

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+
+	g.env[key] = value
+}
+
+func (g *Env) Load() error {
+	buf, err := ioutil.ReadFile(g.path)
```
benhoyt, line 59: Could consider using `io.ReadAtLeast` with a 1025-byte buffer and min of 1024 to avoid accidentally reading a huge file in case something is corrupt/hacked?

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+	if len(buf) != 1024 {
+		return fmt.Errorf("grubenv %q must be exactly 1024 byte, got %d", g.path, len(buf))
+	}
+	if !bytes.HasPrefix(buf, []byte("# GRUB Environment Block\n")) {
+		return fmt.Errorf("cannot find grubenv header in %q", g.path)
+	}
+	rawEnv := bytes.Split(buf, []byte("\n"))
+	for _, env := range rawEnv[1:] {
+		l := bytes.SplitN(env, []byte("="), 2)
+		// be liberal in what you accept
```
benhoyt, line 72: I think the GRUB file can have other `# comment` lines in it (mine does), and there could be an `=` sign in a comment. So instead of looping over `rawEnv` (which should probably be called `lines`), how about we loop over all the lines, and `if bytes.HasPrefix(line, []byte("#")) { continue }`?

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if len(buf) != 1024 {
+		return fmt.Errorf("grubenv %q must be exactly 1024 byte, got %d", g.path, len(buf))
+	}
+	if !bytes.HasPrefix(buf, []byte("# GRUB Environment Block\n")) {
+		return fmt.Errorf("cannot find grubenv header in %q", g.path)
+	}
+	rawEnv := bytes.Split(buf, []byte("\n"))
+	for _, env := range rawEnv[1:] {
+		l := bytes.SplitN(env, []byte("="), 2)
```
benhoyt, line 71: Perhaps `parts` or `fields` instead of `l`?

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+		// be liberal in what you accept
+		if len(l) < 2 {
+			continue
+		}
+		k := string(l[0])
+		v := string(l[1])
+		g.env[k] = v
+		g.ordering = append(g.ordering, k)
```
benhoyt, line 79: Should we only append to `g.ordering` if it wasn't already in `g.env`? How should we handle duplicate keys? Currently they won't round trip correctly, as Write will write out the duplicates but will write both with the last value read.

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+
+	g.env[key] = value
+}
+
+func (g *Env) Load() error {
+	buf, err := ioutil.ReadFile(g.path)
+	if err != nil {
+		return err
+	}
+	if len(buf) != 1024 {
```
benhoyt, line 63: I'd probably extract this to a non-exported `const envBlockSize` or something as the magic number is used a few times.

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+			return err
+		}
+	}
+	if w.Len() > 1024 {
+		return fmt.Errorf("cannot write grubenv %q: bigger than 1024 bytes (%d)", g.path, w.Len())
+	}
+	content := w.Bytes()[:w.Cap()]
+	for i := w.Len(); i < len(content); i++ {
+		content[i] = '#'
+	}
```
benhoyt, line 101: This logic was a bit confusing to me. What about this?

```suggestion
	for w.Len() < 1024 {
		w.WriteByte('#')
	}
	content := w.Bytes()
```

Alternatively:

```go
	content := w.Bytes()
	padding := bytes.Repeat([]byte("#"), 1024-len(content))
	content = append(content, padding...)
```

internal/bootloader/bootloader.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Bootloader provides primitives to retain state across boots and handle
+// boot slots.
+type Bootloader interface {
+	// Name returns the bootloader name.
+	Name() string
+
+	// GetBootVars populates the specified variables from the bootloader.
+	// These variables are preserved across reboots.
+	GetBootVars(names ...string) (map[string]string, error)
```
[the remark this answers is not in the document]
anpep: In the end, we decided to get rid of these `GetBootVars`/`SetBootVars` methods altogether, since there is no requirement for those anymore (we just need the `Status()`, and `ActiveSlot()`/`SetActiveSlot()` for interacting with the boot environment)

internal/bootloader/bootloader.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// SetBootVars saves a set of variables to be persisted across reboots.
+	SetBootVars(values map[string]string) error
```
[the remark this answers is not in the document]
anpep: As mentioned, we got rid of these (:

internal/bootloader/bootloader.go
```diff
@@ -0,0 +1,99 @@
...
+	// GetActiveSlot obtains the label of the currently booted slot.
+	GetActiveSlot() (label string, err error)
+
+	// SetActiveSlot instructs the bootloader to select the slot with the
+	// specified label on the next reboot.
+	SetActiveSlot(label string) error
+
+	// GetStatus obtains the status of the slot with the specified label.
+	// If there is no saved status for the slot, or if the saved status is not
+	// any of Unbootable, Try or Fail, Try will be returned.
```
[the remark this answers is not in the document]
anpep: `Try` here means "attempt to boot from this slot", so it makes sense for a default -- if the boot can't be completed, either `Fail` or `Unbootable` will be set during boot.

internal/bootloader/grub.go
```diff
@@ -0,0 +1,104 @@
...
+
+// GRUB implements the Bootloader interface to support bootloader operations
+// on GRUB 2.
+type GRUB struct {
+	// rootdir is the directory where the GRUB prefix is mounted.
+	rootdir string
+}
+
+// newGrub initializes a new instance of the GRUB bootloader.
+func NewGRUB(rootdir string) Bootloader {
```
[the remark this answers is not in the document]
anpep: These don't need to be exported anymore. However, we do need to return a `Bootloader` because we need these functions to fit the `func(rootdir string) Bootloader` mold.

internal/bootloader/grubenv/grubenv.go
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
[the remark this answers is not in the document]
anpep: A bit of a disclaimer, but regarding all the `grubenv` code, it's been copied straight from [snapd](https://github.com/snapcore/snapd/blob/master/bootloader/grubenv/grubenv.go), so I'm a bit reluctant to touch it just yet since it's mostly going to end up in `x-go`. I agree some parts of it can be streamlined but I wouldn't go down that rabbithole just yet, what do you think?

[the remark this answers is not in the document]
benhoyt: Ah, fair enough.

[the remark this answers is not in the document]
benhoyt: You might want to mention that in the PR description.

internal/bootloader/bootloader.go
```diff
@@ -0,0 +1,99 @@
...
+	// bootloaders list all possible bootloaders by their constructor
+	// function.
+	Bootloaders = []BootloaderNewFunc{
+		NewGRUB,
+	}
+)
+
+// Find obtains an instance of the first supported bootloader that is available
+// on the system.
+func Find() (*Bootloader, error) {
```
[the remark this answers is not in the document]
benhoyt: This was marked as resolved but hasn't been fixed.

benhoyt (approved):
Just one comment about changing `Find` to return `Bootloader` (an interface type) instead of `*Bootloader`.

My comments on `grubenv.go` remain, but yeah, if we're copying that verbatim from snapd, might be better to leave as is for now.

internal/bootloader/bootloader.go
```diff
@@ -0,0 +1,57 @@
...
+	ActiveSlot() string
+
+	// SetActiveSlot instructs the bootloader to select the slot with the
+	// specified label on the next reboot.
+	SetActiveSlot(label string) error
+
+	// Status obtains the status of the slot with the specified label.
+	// If there is no saved status for the slot, or if the saved status is not
+	// any of Unbootable, Try or Fail, Try will be returned.
+	Status(label string) Status
```
flotter: I do not think we should add Status yet, because we have no design or idea of how closely a Project Crystal recovery flow will match snapd. The logic may turn out to be much simpler. Once we add the PR that adds our first feature that adds step 1 for recovery, we would be in a better position to know what this should look like.

For example, see https://docs.google.com/document/d/1RdFL-caA7XHdYmukcibPL-ZzKmlEjJqBs47JWONs9jA/edit#heading=h.2ahke9g9q0im . You will see there is also a proposal for "try", which at this point has not been discussed with Gustavo. However, this may look different than the way snapd organised it.

internal/bootloader/bootloader.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Present returns whether the bootloader is currently present on the
+	// system--in other words, whether this bootloader has been installed to
+	// the current system. Implementations should only return non-nil error if
+	// they can positively identify that the bootloader is installed, but there
+	// is actually an error with the installation.
+	Present() (bool, error)
+
+	// ActiveSlot obtains the label of the currently booted slot.
+	ActiveSlot() string
```
flotter, line 31: How about GetActiveSlot() for symmetry ?

internal/bootloader/grub.go
```diff
@@ -0,0 +1,73 @@
...
+type grub struct {
+	rootdir string
+	env     *grubenv.Env
+}
+
+// newGrub initializes a new instance of the GRUB bootloader.
+func newGrub(rootdir string) Bootloader {
+	return &grub{
+		rootdir: rootdir,
+		env:     grubenv.NewEnv(filepath.Join(rootdir, "grubenv")),
```
flotter, line 38: spec changed to grub.env if that is still on.

internal/bootloader/grub.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (g *grub) ActiveSlot() string {
+	if err := g.env.Load(); err != nil {
+		return ""
+	}
+	return g.env.Get("boot.slot")
+}
+
+func (g *grub) SetActiveSlot(label string) error {
+	g.env.Set("boot.slot", label)
```
flotter, line 59: Are you sure grubenv supports "." in variable names?

From the examples I can find at least, the common way to define variables are uppercase or lowercase with underscores.

internal/bootloader/grub.go
anpep: Nice catch! Thank you

internal/bootloader/grub.go
anpep: Let's keep it safe and change it to underscores (:

niemeyer:

PR #194, opened 2023-02-22
chore(deps): Update deps due to CVEs

merkata:
Security tools such as `trivy` report known CVEs against the dependencies in the `go.mod` that can be fixed. This PR updates the deps to the suggested patched version. Below is the report of the found vulnerabilities and a successful `go test ./...` afterwards.

```
go.mod (gomod)

Total: 14 (UNKNOWN: 0, LOW: 1, MEDIUM: 3, HIGH: 10, CRITICAL: 0)

┌─────────────────────┬────────────────┬──────────┬───────────────────────────────────┬───────────────────────────────────┬──────────────────────────────────────────────────────────────┐
│       Library       │ Vulnerability  │ Severity │         Installed Version         │           Fixed Version           │                            Title                             │
├─────────────────────┼────────────────┼──────────┼───────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ golang.org/x/crypto │ CVE-2021-43565 │ HIGH     │ 0.0.0-20201221181555-eec23a3978ad │ 0.0.0-20211202192323-5770296d904e │ golang.org/x/crypto: empty plaintext packet causes panic     │
│                     │                │          │                                   │                                   │ https://avd.aquasec.com/nvd/cve-2021-43565                   │
│                     ├────────────────┤          │                                   ├───────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│                     │ CVE-2022-27191 │          │                                   │ 0.0.0-20220314234659-1baeb1ce4c0b │ golang: crash in a golang.org/x/crypto/ssh server            │
│                     │                │          │                                   │                                   │ https://avd.aquasec.com/nvd/cve-2022-27191                   │
├─────────────────────┼────────────────┤          ├───────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ golang.org/x/net    │ CVE-2019-9512  │          │ 0.0.0-20190404232315-eb5bcb51f2a3 │ 0.0.0-20190813141303-74dc4d7220e7 │ HTTP/2: flood using PING frames results in unbounded memory  │
│                     │                │          │                                   │                                   │ growth                                                       │
│                     │                │          │                                   │                                   │ https://avd.aquasec.com/nvd/cve-2019-9512                    │
│                     ├────────────────┤          │                                   │                                   ├──────────────────────────────────────────────────────────────┤
[...]

PR #196, opened 2023-02-22
feat(secscan): Introduce security scanning

### LINKED ISSUE #195: Introduce security scanning in CI/CD
This is a proposal to introduce a security scanning step in CI/CD to check for known vulnerabilities of third party libraries that are used with pebble. Findings that cannot be fixed at the moment can be filtered out in a configuration file, and having this step in CI/CD would pick up new vulnerabilities as they appear.

merkata:
This closes #195 with having trivy doing an fs scan against the code base and reporting known vulnerabilities. Should #194 land, I've filtered the known vulnerabilities that cannot be resolved without dropping support for Go < 1.17.


trivy.yaml
```diff
@@ -0,0 +1,3 @@
+timeout: 20m
+scan:
+  offline-scan: true
```
benhoyt, line 3: Can we put these two files somewhere other than the repo root, to avoid cluttering up the root dir? Looks like we could relatively easily move `trivy.yaml` as that's explicitly specified, and it looks like you can specify `--ignorefile value` to move `.trivyignore`.

We can either move them alongside `scanning.yml`, or add a new "scripts" or "config" dir or something? (But if the latter, then we need to bikeshed the name :-)

merkata: Sounds good! I placed `trivy.yaml` and `.trivyignore` under a new directory - `.github/files/`, reason being I wanted it to be under .github as it relates to CI/CD and I don't want to have any unexpected behavior with .yaml files under .github/workflows - I don't know whether Github would try to parse `trivy.yaml` as a workflow and complain on the syntax. That's just a suggestion, if you're fine with the approach of having that under .github, the new folder name can be discussed as "files" might be too broad. I thought about naming it "x-files" as it does contain an X amount of files and is a funny reference, though I abstained.

jnsgruk: Yeh using the `.github/` directory is nicer for sure. In other projects, I've just used `.github/foobar.yaml` without the extra subdirectory and not had any issues fwiw.

benhoyt: Just using `.github/` per @jnsgruk sounds good. Though if we're going with a subdirectory I'd use "config" as everything is "files".

benhoyt (approved):

trivy.yaml
merkata: I'm fine with using `.github/`, I hadn't considered that and it does sound good. Agree that "files" is way too generic.

jnsgruk (approved):

merkata:
Just a note that I cannot merge this on my own, thanks!

.github/workflows/scanning.yml
```diff
@@ -0,0 +1,22 @@
+name: Vulnerability scanning
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
```
cjdcordeiro, line 10: nitpick: `build` is not the right job name

cjdcordeiro (changes requested):

.github/workflows/scanning.yml
jnsgruk: Fixed.

PR #197, opened 2023-02-22
Replace `/sbin/shutdown` with `reboot(2)` syscall

anpep:
Use a direct `reboot(2)` syscall instead of depending on `/sbin/shutdown` so that we don't rely on systemd for rebooting the system.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -23,20 +23,19 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"os/exec"
 	"os/signal"
 	"runtime"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"gopkg.in/tomb.v2"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/gorilla/mux"
 
 	"github.com/canonical/pebble/internal/logger"
-	"github.com/canonical/pebble/internal/osutil"
 	"github.com/canonical/pebble/internal/osutil/sys"
 	"github.com/canonical/pebble/internal/overlord"
 	"github.com/canonical/pebble/internal/overlord/checkstate"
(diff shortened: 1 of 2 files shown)


internal/daemon/daemon_test.go
```diff
@@ -725,29 +724,39 @@ func (s *daemonSuite) TestRestartSystemWiring(c *check.C) {
...
+	oldTimeSleep := timeSleep
+	oldSyscallReboot := syscallReboot
+	defer func() {
+		timeSleep = oldTimeSleep
+		syscallReboot = oldSyscallReboot
+	}()
 
-		cmd.ForgetCalls()
+	timeSleep = func(d time.Duration) {
+		didSleep = true
```
benhoyt, line 738: Nit, but it's probably slightly better to use `numSleeps` and `numReboots` counters instead of `didX` so we know these are only called once.

benhoyt (approved):
Looks good to me!

internal/daemon/daemon.go
```diff
@@ -721,18 +720,19 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
 
-func rebootImpl(rebootDelay time.Duration) error {
-	if rebootDelay < 0 {
-		rebootDelay = 0
+func rebootImpl(delay time.Duration) error {
+	if delay < 0 {
+		delay = 0
 	}
-	mins := int64(rebootDelay / time.Minute)
-	cmd := exec.Command("shutdown", "-r", fmt.Sprintf("+%d", mins), shutdownMsg)
```
flotter, line 731: This would have been a call to systemd in the past for other systems (/usr/sbin/shutdown -> /bin/systemctl) which do include systemd. Pebble calling this directly under systemd could be bad I would imagine. At the same time I would also imagine having no /sbin/reboot dependency would be perfect for some usescases, maybe even Rocks.

Perhaps we need library offering different reboot implementations, and a Pebble variant specific selection of the reboot implementation. I guess we should try to understand if this is an issue or not for the Pebble use cases.

internal/daemon/daemon.go
```diff
@@ -827,9 +827,9 @@ func getListener(socketPath string, listenerMap map[string]net.Listener) (net.Li
 	}
 
 	runtime.LockOSThread()
-	oldmask := syscall.Umask(0111)
+	oldmask := unix.Umask(0111)
 	listener, err := net.ListenUnix("unix", address)
-	syscall.Umask(oldmask)
+	unix.Umask(oldmask)
```
flotter, line 832: The change from syscall to unix, could you just add some insight for me why this is done ?

anpep: Documentation states the `syscall` package is deprecated! https://pkg.go.dev/syscall

internal/daemon/daemon.go
anpep: The only problem I see here would be syncing filesystems before reboot. However, I'm concerned about `/sbin/shutdown` having more logic than just calling `sync(2)` and `reboot(2)` that current Pebble users have a dependency on. But I don't think this argument is solid enough to justify the effort of maintaining (and polluting the internal Pebble API with) differing implementations of very trivial functionality.

anpep:
Closing this as it's no longer relevant for Termus (we'll be doing our own implementation in our fork).

PR #200, opened 2023-03-01
Changes service status on success ignore.

tlm:
When a service under Pebble's control exits with code 0 and has it's OnSuccess service key configured to ignore Pebble set's the services state to one of error.

This change configures this logic so that when a service layer specifies ignoring a service that has exited with a code of 0 it is now placed into an inactive state.

For this PR I have not enabled any new tests instead modified the ones we have that were already checking this condition.

### DIFF
--- a/internal/overlord/servstate/handlers.go
+++ b/internal/overlord/servstate/handlers.go
@@ -469,6 +469,11 @@ func (s *serviceData) exited(exitCode int) error {
 		switch action {
 		case plan.ActionIgnore:
 			logger.Noticef("Service %q %s action is %q, not doing anything further", s.config.Name, onType, action)
+			// On success we transition to state stopped.
+			if exitCode == 0 {
+				s.transition(stateStopped)
+				break
+			}
 			s.transition(stateExited)
 
 		case plan.ActionShutdown:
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Looks good to me, thanks!

jnsgruk (approved):
Looks representative of our conversation on Tue/Wed, thanks :)

internal/overlord/servstate/handlers.go
```diff
@@ -469,6 +469,11 @@ func (s *serviceData) exited(exitCode int) error {
 		switch action {
 		case plan.ActionIgnore:
 			logger.Noticef("Service %q %s action is %q, not doing anything further", s.config.Name, onType, action)
+			// On success we transition to state stopped.
+			if exitCode == 0 {
```
niemeyer, line 473: We shoud have a success flag here so we're not doing the comparison in multiple places, and it becomes a bit less error prone and easier to refactor this logic in the future. For now it's only being done in two places nearby, so I'm going ahead regardless in the interest of unblocking you.

niemeyer (approved):
Just a trivial comment for the future.

Thanks for the fix.

jnsgruk:
Thanks @tlm @niemeyer!

Issue #202, opened 2023-03-07
Versioned releases


jnsgruk:
Given that this project is now used in:

- KernOS
- Juju (both 2.9 and 3.x)
- Rockcraft
- Several ROCKs

I think it's important that we start to land some versioned releases. As we start to see more usage, it'd be useful to enable people to report the version they're on with some clarity, and we could also begin a "public roadmap" / "release log" similar to `snapd` and `juju`.

benhoyt:
Makes sense. I think we want Pebble to be able to say its version number, too, meaning we have to include a version number in a `.go` file somewhere. Is the best way to do that just good old fashioned `const Version = "1.2.3"` in a `version.go` file that we update by hand? Mildly annoying, but probably simplest.

If it's just a git tag, that's hard to embed in the binary, right? Or it will depend on how Pebble is built.

jnsgruk:
Looks like we already have some [scripts](https://github.com/canonical/pebble/blob/master/cmd/mkversion.sh) and [infra](https://github.com/canonical/pebble/blob/master/cmd/version.go) in place for generating versions, so shouldn't be too hard.

I've also used `-ldfags `-X main.version=$(git describe --tags)` or similar before. Either approach works fine and prevents us from needed to hard-code a version using a human - I think doing it based on tags would be most sensible?

PR #203, opened 2023-03-07
Change the generate directory to 'cmd/'

jameinel:
By default it was populating the 'version_generated.go' file into the top level directory, but we want to populate it in the same directory as the cmd package. (That is where 'pebble version' currently looks for it.)

I did test just running `go generate ./...` from the top level directory, and running in as `cd cmd; go generate`, and it picked up the version string. They are still just git hashes, because we haven't done a version tag, but I can see that mkversion.sh is using 'git describe' as you would expect.

### DIFF
--- a/cmd/mkversion.sh
+++ b/cmd/mkversion.sh
@@ -19,11 +19,11 @@ set -e
 #   that dh-golang creates and that only contains a subset of the
 #   files of the toplevel buildir. 
 PKG_BUILDDIR=$(dirname "$0")/..
-GO_GENERATE_BUILDDIR="$(pwd)"
+GO_GENERATE_BUILDDIR="$(pwd)/cmd"
 
 # run from "go generate" adjust path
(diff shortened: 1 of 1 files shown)


jameinel:
I'm not sure if this is perfect in the case of snap building, the comments in 'mkversion.sh' seem to indicate that it would actually be in _build, which means we might actually need the build directory to be "${pwd}/.../cmd" or something along those lines.
But as we aren't building a snap just yet, we can address that concern when we get there.

benhoyt (approved):

benhoyt:
Thanks for this. There's obviously more to do for CI and tagging (https://github.com/canonical/pebble/issues/202), but this fixes the immediate issues with `mkversion.sh`, so merging without further review.

```
# Before:
$ go run ./cmd/pebble version
client  unknown
server  unknown
# After:
$ go run ./cmd/pebble version
client  73eacaf
server  73eacaf
```

PR #204, opened 2023-03-09
Add a GitHub Action for building and uploading binaries

### LINKED ISSUE #202: Versioned releases

benhoyt:
DRAFT: Still playing with this on my fork.

This uses GoReleaser, which is widely used and well-maintained. Wouldn't be hard to do it ourselves just using "go build" either if we wanted to do that.

Fixes #202.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,28 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  binaries:
+    name: GoReleaser release
+    runs-on: ubuntu-latest
+    container:
+      image: goreleaser/goreleaser-cross:v1.20.2
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v3
+
+      - name: Run GoReleaser
+        run: goreleaser release --clean
+
+      - name: Upload binaries
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            dist/*_checksums.txt
+            dist/*.tar.gz
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ cmd/VERSION
 .idea
 .vscode
 /pebble
+
+dist/
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,28 @@
+before:
+  hooks:
+    - go generate ./cmd
+
+builds:
+  - main: ./cmd/pebble
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    flags:
+      - "-trimpath"
+
+archives:
+  - format: tar.gz
+    name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+    files:
+      - src: COPYING
+        dst: ./COPYING
+      - src: README.md
+        dst: ./README.md
+    rlcp: true  # See https://goreleaser.com/deprecations/#archivesrlcp
+
+snapshot:
+  name_template: '{{ if ne .Version "0.0.0" }}{{ .Version }}{{ else }}{{ .ShortCommit }}-devel{{ end }}'


jnsgruk (approved):
LGTM, and adding goreleaser seems like a good choice -- I can see how we'd use some its other features down the line.

simskij:
Goreleaser has served me very well for years. Great choice, and happy to see this change!

cjdcordeiro (approved):
This is great. Thanks! I'm sure we'll reuse some of it for Chisel at some point.

> Linux amd64 and arm64 binaries when a tag/release is created (do we need i386 as well? easy to add if so)

(FYI) Thinking about containers and ROCKs, we do build Ubuntu base container images for amd64, arm64, s390x, ppc64el, armhf, riscv64 and i386 (Bionic only), so in theory, Rockcraft could also have the ability to build ROCKs for those architectures. It isn't an immediate need though, so `amd64` and `arm64` are good enough to start with.

This LGTM and I agree with the choice of compilation flags, as that's also what's being proposed for the Snap in https://github.com/canonical/pebble/pull/179/files

jnsgruk:
@flotter - this is interesting, and my sense is that Goreleaser could make it easier to automate/define how we do builds for different "flavours" of Pebble down the line, should we need to.

niemeyer (changes requested):
Folks, I truly hate to be a buzzkill after so much excitement for the goreleaser route, but this has a few drawbacks which seem concerning in general, and even more so for this project in particular. These drawbacks stem from the fact that we're layering a matrix build system on top of a matrix build system. In other words, we're adopting GitHub Actions which has its own pipeline and very nice integartion with GitHub, to call into Go Releaser, which is again a matrix build system specialized in Go, to then call the Go build system which tends to be nice to work with, to then release the artifacts back to GitHub itself.

This may even look nice when you look into the trivial cases, but as soon as you step outside of these you'll start to find yourself working against two different matrix build systems to satisfy one another, when we could literally just setup our GitHub Actions to build exactly on the architectures, as a nice matrix, and trivially upload the tarballs.

Just as a couple of examples of real issues we'd need to worry about for Pebble:

https://goreleaser.com/limitations/cgo/
https://goreleaser.com/cookbooks/cgo-and-crosscompiling/

Can we please try to give a shot at creating a very simple and neat build/upload pipeline with GitHub Actions and standard tooling alone?

benhoyt:
@niemeyer I'm not sure I quite understand the drawbacks. Is it just the complexity of having another tool / config file you're concerned about? We're not using GitHub Actions matrix functionality here, and using GoReleaser's instead, so they don't conflict or anything.

That said, I was on the fence about GoReleaser initially as it should be very easy to use `go build` directly, so I'll whip up a version that uses that with GitHub Actions directly instead, and we can see what that looks like.

niemeyer:
> We're not using GitHub Actions matrix functionality here,

This idea that we're not using it is precisely the point being raised. GitHub Actions sets up the machine, with a selected architecture, with a selected container image, offers us good visualization of failure and success, with logs, and the whole history over time. We're absolutely using it, but we are just not using it as much as we could, and are opting to use a third-party tool with an open core model to do that instead, with known shortcomings in areas that we'll want to focus on.

Thanks for looking into it.

cmd/pebble/cmd_warnings.go
```diff
@@ -56,7 +56,7 @@ var shortOkayHelp = "Acknowledge warnings"
 var longOkayHelp = `
 The okay command acknowledges the warnings listed with 'pebble warnings'.
 
-Once acknowledged a warning won't appear again unless it re-occurrs and
+Once acknowledged a warning won't appear again unless it reoccurs and
```
benhoyt, line 59: Drive-by fix: "reoccur" is a word in its own right without a hyphen ([ref](https://www.merriam-webster.com/dictionary/reoccur)).

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Set up Go
+      uses: actions/setup-go@v3
```
barrettj12, line 24: Should we fix a Go version here? You can use `go-version: 1.x`, or `go-version-file: go.mod` to read it from `go.mod`.

cmd/pebble/cmd_warnings.go
barrettj12: While you're here, I'd put a comma after "acknowledged".

barrettj12 (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+        go build -trimpath -ldflags='-s -w' -o dist/build/pebble ./cmd/pebble
+
+        ARCHIVE_FILE=pebble_${GITHUB_REF_NAME}_${GOOS}_${GOARCH}.tar.gz
+        echo Creating archive $ARCHIVE_FILE
+        cp COPYING dist/build
+        cp README.md dist/build
+        tar -czf dist/$ARCHIVE_FILE -C dist/build .
+
+        echo Uploading $ARCHIVE_FILE to release $GITHUB_REF_NAME
+        gh release upload $GITHUB_REF_NAME dist/$ARCHIVE_FILE
```
jnsgruk, line 46: I'm not an export on `gh release upload`, but what does the resulting release look like here?

I quite liked what @SimonRichardson proposed in the `juju-qa-jenkins` where the release notes can be generated too? https://github.com/canonical/juju-release-jenkins/pull/67/files#diff-c759a587ce8b6c2c352ae5e2e5c4ca62cad308c186d9ed844b4d8d823d1eb59bR65

jnsgruk: Looking at your [fork](https://github.com/benhoyt/pebble/releases/tag/v0.2.1), I think using `gh release create` might allow us to set nicer looking titles etc. too

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+  push:
+    tags:
+      - v*
+
+jobs:
+  binaries:
+    strategy:
+      matrix:
+        goos: [linux]
+        goarch: [amd64, arm64, ppc64le, s390x]
```
flotter, line 13: Would it perhaps be proactive to create a build tags dimension here with a single entry for now? Or, would it be simpler to have the vanilla pebble without any additional information:

For example:

`ARCHIVE_FILE=pebble_${GITHUB_REF_NAME}_${VARIANT}_${GOOS}_${GOARCH}.tar.gz`

**# Case always including a variant:**

```
pebble_v1.0_default_linux_amd64.tar.gz
pebble_v1.0_termus_linux_amd64.tar.gz
:
```

or simply:

```
pebble_v1.0_linux_amd64.tar.gz
pebble_v1.0_termus_linux_amd64.tar.gz
:
```

In the former case the file names have the same information at the same delimited offsets.

.github/workflows/release.yml
benhoyt: Well, the idea is that this job will kick in *after* you manually create a GitHub release (which will create the tag). Then this job kicks off with that release's tag, so you want to `gh release upload` to the just-created, release, rather than `gh release create`. I'm old-school and like to nicely curate release notes rather than just "here's a log of all the commits" (which is easy to link to in the curated release notes if we want).

So we'll set the release title manually too (I just typed in that title you linked to manually).

.github/workflows/release.yml
benhoyt: I'd prefer we keep the names simple for the common case (just using goos_goarch is very common for naming Go binaries), and add the variant to the end as needed. I'd prefer it was on the end anyway, like `pebble_v1.0.0_linux_amd64_termus.tar.gz`.

.github/workflows/release.yml
jnsgruk: Aha that makes sense. Maybe we should add a section in the HACKING doc covering how to make a release, then? :)

benhoyt: Good call -- done.

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  binaries:
```
benhoyt, line 9: verb: create-binary

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,72 @@
...
+        CGO_ENABLED: "0"
+      run: |
+        echo Generating version file
+        go generate ./cmd
+
+        echo Building for $GOOS $GOARCH
+        go build -trimpath -ldflags='-s -w' -o dist/build/pebble ./cmd/pebble
+
+        # Get version via "pebble version" to ensure it matches that exactly
+        PEBBLE_VERSION=$(GOOS=linux GOARCH=amd64 go run ./cmd/pebble version | awk '/client/ { print $2 }')
```
benhoyt, line 46: We could get the version via GitHub environment variables or git commands, but doing this ensures that it's the exact same version `cmd/mkversion.sh` has produced and embedded in the binary to use in `pebble version` output.

The `pebble version` command currently tries to fetch the server version as well, falling back to `-` if it fails with the timeout of 5s. I might put up a second PR to add a `--client` arg to `pebble version` so to avoid the 5s delay and the `awk`ward parsing.

internal/osutil/sys/sysnum_32_linux.go
```diff
@@ -1,6 +1,6 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
-//go:build arm64 || amd64 || ppc64le || s390x || ppc
-// +build arm64 amd64 ppc64le s390x ppc
+//go:build arm64 || amd64 || ppc64le || s390x || ppc || riscv64
+// +build arm64 amd64 ppc64le s390x ppc riscv64
```
benhoyt, line 3: The riscv64 build didn't compile without this addition.

cmd/pebble/cmd_warnings.go
```diff
@@ -56,7 +56,7 @@ var shortOkayHelp = "Acknowledge warnings"
 var longOkayHelp = `
 The okay command acknowledges the warnings listed with 'pebble warnings'.
 
-Once acknowledged a warning won't appear again unless it re-occurrs and
+Once acknowledged, a warning won't appear again unless it reoccurs and
```
benhoyt, line 59: Drive-by grammar fixes.

.github/workflows/binaries.yml
benhoyt: I've created https://github.com/canonical/pebble/pull/208 so we could simplify the above to:

```
PEBBLE_VERSION=$(GOOS=linux GOARCH=amd64 go run ./cmd/pebble version --client)
```

niemeyer (approved):
Thanks for working on this, Ben.

One trivial question below, not a blocker. Merging it so we can use it.

PR #207, opened 2023-03-27
Add firmware metadata load support

flotter:
Firmware metadata are accessible in the root filesystem of the running firmware at the following location, encoded as JSON:

/termus/meta/termus.json

Information obtained from this file represents build-time attributes of the firmware that is required by the firmware state manager.

Revision is taken verbatim from the snapd project while the firmware info logic takes inspiration from the snap info code.

### DIFF
--- a/internal/firmware/export_test.go
+++ b/internal/firmware/export_test.go
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2023 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package firmware
+
+// UpdateMetaInfoPath provides a hook to allow temporary override of the location
+// of the firmware metadata file
+func UpdateMetaInfoPath(path string) func() {
+	origMetaPath := metaInfoPath
+	metaInfoPath = path
+	return func() {
+		metaInfoPath = origMetaPath
+	}
+}
+
+func MetaInfoFromRunning() (*Info, error) {
+	return metaInfoFromRunning()
+}
--- a/internal/firmware/info.go
+++ b/internal/firmware/info.go
@@ -1,0 +1,109 @@
+/*
+ * Copyright (C) 2023 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/*
+ * Package firmware extracts information from the running system by
+ * inspecting metadata files embedded in the root filesystem.
+ */
+package firmware
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+)
+
+// Info provides information about firmware.
+type Info struct {
+	Version         string
+	OriginalName    string
+	OriginalSummary string
+
+	StoreInfo
+}
+
+// Rev returns the revision of the firmware.
+func (i *Info) Rev() Revision {
+	return i.Revision
+}
+
+// Name returns the store approved name of the firmware, otherwise the
+// firmware supplied name (if the store name is not available).
+func (i *Info) Name() string {
+	if i.ApprovedName != "" {
+		return i.ApprovedName
+	}
+	return i.OriginalName
+}
+
+// Summary returns the store approved summary of the firmware, otherwise the
+// firmware supplied summary (if the store summary is not available).
+func (i *Info) Summary() string {
+	if i.ApprovedSummary != "" {
+		return i.ApprovedSummary
+	}
+	return i.OriginalSummary
+}
+
+// GetInfo returns a populated Info structure describing the running firmware.
+func GetInfo() (*Info, error) {
+	// Information from the firmware will later be supplemented by approved
+	// metadata coming from the store.
+	return metaInfoFromRunning()
+}
+
+// StoreInfo holds firmware metadata for which the store is the
+// canonical source.
+//
+// It can be marshalled and will be stored in the system state for
+// each installed firmware so it needs to be evolved carefully.
+type StoreInfo struct {
+	Revision        Revision `json:"revision"`
+	ApprovedName    string   `json:"name,omitempty"`
+	ApprovedSummary string   `json:"summary,omitempty"`
+}
+
+// metaInfoPath points to firmware metadata located in the
+// rootfs of a running system.
+var metaInfoPath string = "/termus/meta/termus.json"
+
+// metaInfo represents firmware metadata created during the firmware
+// build process.
+type metaInfo struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+	Summary string `json:"summary"`
+}
+
+// metaInfoFromRunning returns the firmware metadata of the running system
+// exported as an Info struct.
+func metaInfoFromRunning() (*Info, error) {
+	data, err := ioutil.ReadFile(metaInfoPath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot open firmware metadata file %s: %w", metaInfoPath, err)
+	}
+
+	var m metaInfo
+	if err := json.Unmarshal(data, &m); err != nil {
+		return nil, fmt.Errorf("cannot parse firmware metadata json: %w", err)
+	}
+	return &Info{
+		Version:         m.Version,
+		OriginalName:    m.Name,
+		OriginalSummary: m.Summary,
+	}, nil
+}
--- a/internal/firmware/info_test.go
+++ b/internal/firmware/info_test.go
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2023 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package firmware_test
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/pebble/internal/firmware"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type S struct{}
+
+var _ = Suite(&S{})
+
+const metaInfoTestPath = "test/path/fwinfo/json"
+
+func (s *S) TestFwInfoFileMissing(c *C) {
+	path := filepath.Join(c.MkDir(), metaInfoTestPath)
+	defer firmware.UpdateMetaInfoPath(path)()
+
+	_, err := firmware.MetaInfoFromRunning()
+	c.Assert(err, ErrorMatches, `cannot open firmware metadata file.*`)
+}
+
+func (s *S) TestFwInfoFileContentInvalid(c *C) {
+	path := filepath.Join(c.MkDir(), metaInfoTestPath)
+	if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
+		c.Errorf("failed to create directory tree %s: %v", filepath.Dir(path), err)
+	}
+	if err := ioutil.WriteFile(path, []byte(``), 0755); err != nil {
+		c.Errorf("failed to create file %s", path)
+	}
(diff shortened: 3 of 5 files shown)


anpep (approved):
Lots of nitpicking :) on a good PR :+1:

internal/firmware/info_fw.go
```diff
@@ -0,0 +1,59 @@
...
+	data, err := ioutil.ReadFile(fwInfoPath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot open firmware metadata file %s: %w", fwInfoPath, err)
+	}
+
+	var f fwInfo
+	err = json.Unmarshal(data, &f)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse firmware metadata json: %w", err)
+	}
```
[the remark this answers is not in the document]
flotter: Fixed. Also in info_test.go.

internal/firmware/info_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/canonical/pebble/internal/firmware"
+	. "gopkg.in/check.v1"
```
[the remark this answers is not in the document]
flotter: Fixed.

internal/firmware/info_test.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type S struct{}
+
+var _ = Suite(&S{})
+
+const FwInfoTestPath = "test/path/fwinfo.json"
```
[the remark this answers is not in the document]
flotter: Fixed.

internal/firmware/info_test.go
```diff
@@ -0,0 +1,148 @@
...
+	path := filepath.Join(c.MkDir(), FwInfoTestPath)
+	defer firmware.UpdateFwInfoPath(path)()
+
+	_, err := firmware.FwInfoFromRunning()
+	c.Assert(err, ErrorMatches, `cannot open firmware metadata file.*`)
+}
+
+func (s *S) TestFwInfoFileContentInvalid(c *C) {
+	path := filepath.Join(c.MkDir(), FwInfoTestPath)
+	fmt.Println(path)
```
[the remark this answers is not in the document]
flotter: Removed in two places. Done.

internal/firmware/info_test.go
```diff
@@ -0,0 +1,148 @@
...
+			[]byte(`{"name":"foo"}`),
+			&firmware.Info{OriginalName: "foo"},
+		},
+		{
+			[]byte(`{"name":"foo", "version":"1.0~dev", "summary":"Foo firmware dev build"}`),
+			&firmware.Info{OriginalName: "foo", Version: "1.0~dev", OriginalSummary: "Foo firmware dev build"},
+		},
+	}
+	for _, t := range tests {
+		err = ioutil.WriteFile(path, []byte(t.json), 0755)
```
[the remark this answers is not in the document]
flotter: Done.

internal/firmware/revision.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (r Revision) String() string {
+	if r.N == 0 {
+		return "unset"
+	}
+	if r.N < 0 {
+		return fmt.Sprintf("x%d", -r.N)
+	}
+	return strconv.Itoa(int(r.N))
```
benhoyt, line 38: Seems like the `int(...)` should be unnecessary here? Though if this is taken verbatim from the snapd project, maybe we want to keep it in sync?

internal/firmware/export_test.go
```diff
@@ -0,0 +1,34 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
```
benhoyt, line 1: A few files in Pebble have this editor cruft, but most don't. IMO we should drop it. Surely editors a smart enough to use the `.go` file extension these days...

internal/firmware/export_test.go
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package firmware
+
+// UpdateFwInfoPath provides a hook to allow temporary override of the location
+// of the firmware metadata file
+func UpdateFwInfoPath(path string) func() {
```
benhoyt, line 24: I know this is a testing/internal function, but it's in the `firmware` package, so I think we can drop the `Fw` from the name. Similar below.

internal/firmware/info_store.go
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package firmware
+
+// StoreInfo holds firmware metadata for which the store is the
+// canonical source.
+//
+// It can be marshalled and will be stored in the system state for
+// each installed firmware so it needs to be evolved carefully.
+type StoreInfo struct {
```
benhoyt, line 27: Any reason this needs to go into its own file? It seems to me this would be clearer right alongside `type Info`.

internal/firmware/info.go
```diff
@@ -0,0 +1,63 @@
...
+ */
+package firmware
+
+// Info provides information about firmware.
+type Info struct {
+	Version         string
+	OriginalName    string
+	OriginalSummary string
+
+	StoreInfo
```
benhoyt, line 32: I don't really understand why `StoreInfo` is its own type, rather than the fields just listed here. It seems it's not used elsewhere (except in one test).

benhoyt (approved):
Just a couple of minor comments, but looks good to me.

internal/firmware/info.go
flotter: I will update the comments to explain this, but the historical reason is that the store attributes `StoreInfo` are used as a transactional handle to locate a specific firmware, and is used everywhere in the code in different packages. The wider `Info` structure includes lots of additional information which are not replicated in the state, but rather looked up if needed (some of the `Info` attributes are overridden by `StoreInfo` store approved versions).

internal/firmware/info_fw.go
```diff
@@ -0,0 +1,57 @@
...
+// build process.
+type fwInfo struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+	Summary string `json:"summary"`
+}
+
+// fwInfoFromRunning returns the firmware metadata of the running system
+// exported as an Info struct.
+func fwInfoFromRunning() (*Info, error) {
```
atesburak, line 42: Even though the "running firmware" is a good statement I don't think "Running" is the right term to use as if/when we introduce the non-running firmware the naming may not be as clean. May be "active" vs "inactive"?

atesburak (approved):

internal/firmware/revision.go
flotter: I will propagate this change to snapd.

internal/firmware/export_test.go
flotter: Fixed.

internal/firmware/info_store.go
flotter: Fixed.

internal/firmware/export_test.go
flotter: I renamed this to "meta" which is currently accepted in the spec for its path.

internal/firmware/info.go
flotter: Had a discussion. Fixed.

internal/firmware/info_fw.go
flotter: The challenge is that a firmware may be marked as active, but the reboot has not yet taken place, so its not running yet.

The other variant of this function will be called "metaInfoFromSlot()" to indicate indirect metadata extraction. So we will not need to have a "NotRunning...()".

I am going to keep this for now, until more complaints pile up.

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+package firmware
+
+import (
+	"fmt"
+	"strconv"
+)
+
+type Revision struct {
```
hpidcock, line 25: Comment

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
+/*
```
hpidcock, line 1: Use // comments for consistency

internal/firmware/revision_test.go
```diff
@@ -0,0 +1,196 @@
+/*
```
hpidcock, line 1: Use // comments for consistency

internal/firmware/info_test.go
```diff
@@ -0,0 +1,140 @@
+/*
```
hpidcock, line 1: Use // comments for consistency

internal/firmware/info.go
```diff
@@ -0,0 +1,109 @@
+/*
```
hpidcock, line 1: Use // comments for consistency

internal/firmware/export_test.go
```diff
@@ -0,0 +1,32 @@
+/*
```
hpidcock, line 1: Use // comments for consistency

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+	if r.N == 0 {
+		return "unset"
+	}
+	if r.N < 0 {
+		return fmt.Sprintf("x%d", -r.N)
+	}
+	return strconv.Itoa(r.N)
+}
+
+func (r Revision) Unset() bool {
```
hpidcock, line 39: `Unset` is ambiguous, please comment.

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+	return r.N == 0
+}
+
+func (r Revision) Local() bool {
```
hpidcock, line 43: `Local` is ambiguous, please comment.

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+	return r.N < 0
+}
+
+func (r Revision) Store() bool {
```
hpidcock, line 47: `Store` is ambiguous, please comment.

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+		i, err := strconv.Atoi(s[1:])
+		if err == nil && i > 0 {
+			return Revision{-i}, nil
+		}
+	}
+	i, err := strconv.Atoi(s)
+	if err == nil && i > 0 {
+		return Revision{i}, nil
+	}
+	return Revision{}, fmt.Errorf("invalid firmware revision: %#v", s)
```
hpidcock, line 100: err is not fully handled here, nor is "why" the firmware revision is invalid (i.e. -ve revision number)

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+// See R for a function more suitable for hardcoded revisions.
+func ParseRevision(s string) (Revision, error) {
+	if s == "unset" {
+		return Revision{}, nil
+	}
+	if s != "" && s[0] == 'x' {
+		i, err := strconv.Atoi(s[1:])
+		if err == nil && i > 0 {
+			return Revision{-i}, nil
+		}
```
hpidcock, line 94: so if s[0] is 'x' then we will never parse the revision number below. So we should just return a specific error message here.

internal/firmware/revision.go
```diff
@@ -0,0 +1,119 @@
...
+	if err == nil && i > 0 {
+		return Revision{i}, nil
+	}
+	return Revision{}, fmt.Errorf("invalid firmware revision: %#v", s)
+}
+
+// R returns a Revision given an int or a string.
+// Providing an invalid revision type or value causes a runtime panic.
+// See ParseRevision for a polite function that does not panic.
+func R(r interface{}) Revision {
```
hpidcock, line 106: What is the purpose of this function?

hpidcock: Like if it is just for tests then it shouldn't be exported here.

flotter:
@hpidcock Thanks for taking the time to review. I will resolve and let you know. This PR is scheduled to migrate to the spinoff Pebble repo for KernOS, which we are in the process of finalising. I will pause work here until I can migrate it over, and then update this PR. I hope this is OK with you.

niemeyer:

internal/firmware/revision.go
flotter: Fixed.

internal/firmware/revision.go
flotter: Fixed in all files, thanks.

internal/firmware/revision_test.go
flotter: Fixed in all files.

internal/firmware/info_test.go
flotter: Fixed in all files.

internal/firmware/info.go
flotter: Fixed in all files.

internal/firmware/revision.go
flotter: Hi Harry. You are right this does appear ambiguous, but in the context of store-backed products, Local and Store revisions are well understood concepts. I took this file verbatim from the snapd codebase, and again in that context the idea of a "local revision" is very well understood. I will add comments as you propose.

internal/firmware/revision.go
flotter: Hi Harry. You are right this does appear ambiguous, but in the context of store-backed products, Local and Store revisions are well understood concepts. I took this file verbatim from the snapd codebase, and again in that context the idea of a "store revision" is very well understood. I will add comments as you propose.

internal/firmware/revision.go
flotter: Hi Harry. I am going to try to cover the last 3 comments in one here:

Just a reminder that the revision code is taken verbatim from snapd, and has been in use forever. I keep the two versions in sync as both Project Crystal and snapd needs it for very similar reasons.

Some details to understand how the system works. 

```revision > 0``` => positive revision number which indicates the revision was assigned by a product store authority. 
```revision < 0``` => negative revision, but the string representation replaces the minus with ```x```.

The CLI interface and other sources can use R() to convert either a int or string format of a revision.

For the string case, if its not:

local revision: x1, x5, x122 etc..
store revision: 1, 5, 122 etc...

Then its invalid.

internal/firmware/revision.go
flotter: I hope this now makes more sense  - its a slightly confusing system if you have not seen this in action before :)

internal/firmware/revision.go
flotter: As explained above, this file is like a library which I keep in sync between snapd and Project Crystal. We will use R() soon.

internal/firmware/export_test.go
flotter: Fixed.

internal/firmware/revision.go
flotter: General Discussion with Harry to explain source of this file. Will add comment here.

PR #208, opened 2023-03-28
Add --client arg to "pebble version"

benhoyt:
This is a small change that allows you to type `pebble version --client` if you just want the client version, without "pebble version" trying to hit the server (and taking 5 seconds to respond). For example, to avoid [this kind of awkwardness](https://github.com/canonical/pebble/pull/204/files#diff-bdfe89ee8491e51e3f81ada685c51d24188416d8b529b443ee6085a0e0fd4a4aR46):

```
# Get version via "pebble version" to ensure it matches that exactly
PEBBLE_VERSION=$(GOOS=linux GOARCH=amd64 go run ./cmd/pebble version | awk '/client/ { print $2 }')
```

If this PR is merged, we can simplify the above to the following (and it will respond instantly instead of taking 5s):

```
# Get version via "pebble version" to ensure it matches that exactly
PEBBLE_VERSION=$(GOOS=linux GOARCH=amd64 go run ./cmd/pebble version --client)
```

### DIFF
--- a/cmd/pebble/cmd_version.go
+++ b/cmd/pebble/cmd_version.go
@@ -20,7 +20,7 @@ import (
 	"github.com/jessevdk/go-flags"
 
 	"github.com/canonical/pebble/client"
-	"github.com/canonical/pebble/cmd"
+	cmdpkg "github.com/canonical/pebble/cmd"
 )
 
 var shortVersionHelp = "Show version details"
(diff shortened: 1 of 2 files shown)


jnsgruk (approved):
Nice simple change. LGTM.

niemeyer (approved):
Thanks!

PR #165, opened 2022-11-30
[JUJU-1889] Log forwarding, take two

barrettj12:
This PR enables forwarding of Pebble service logs to remote Loki servers and UDP/TCP syslog receivers, following the spec [here](https://docs.google.com/document/d/12tJFhIbVIlOcblG6l9_u-K5bCib78lojRJ-hdGBxSsw/edit?usp=sharing).

## Design

Each (service, log target) pair has an associated "forwarder", which pulls logs from the service ringbuffer (using the iterators introduced in #29), and sends them to the log target. Each forwarder runs two goroutines:
- an `iterate` goroutine, which pulls logs from the iterator/parser and sends them on a `messages` channel
- a `forward` goroutine, which reads logs from the `messages` channel, buffers them, and flushes periodically to the remote log target (after 100 logs have been buffered, or 1 second after the first write)

The forwarders are controlled by a "log manager", which is notified when a new service starts, or when the plan changes. On a plan change, the log manager tears down all forwarders, and starts new ones as determined by the plan.

## Plan changes

The Pebble layer specification has changed as follows:
- A new top-level section `log-targets`, where users specify the log targets they want Pebble to interact with.
- A new (optional) field `log-targets` in the service spec. This takes a list of names, which are matched against the targets defined in the `log-targets` top-level section.

The default behaviour is to forward all services' logs to all targets. This makes it easy to get started with log forwarding by just defining a new layer with the log targets - no need to change the service definitions.

Users can have more granular control over log forwarding by:
- Setting the optional `selection` field for a log target;
- In the definition for a service, explicitly listing log targets they want to forward this service's logs to.

More details are in the README.

## Other changes

The `plan.Service.Equal` method was being used by the service manager to determine whether or not to restart a service on a replan. However, not all changes to the service config require a service restart. In particular, a change to the new `log-targets` field should **not** invoke a service restart. Hence, we've defined a `servstate.NeedsRestart(old, new *plan.Service)` function, which compares the old and new config structs for a service and determines if the service needs to be restarted. At the moment, `NeedsRestart` only ignores changes to `log-targets` - future work should also ignore changes to e.g. metadata, so we can be more conservative with restarting services.

This was the only use of `plan.Service.Equal`, hence we've also removed that method.

## Future work

- Add support for custom Loki labels ([spec](https://docs.google.com/document/d/1HUHH9OQW-czPxn1mXr6AVjhwRMM-DRYp02ATllFSxjE/edit)).
[...]

PR #209, opened 2023-03-29
[JUJU-3388] Plan changes for log forwarding

barrettj12:
This PR contains the first part of #165, which includes all necessary changes to the plan to allow log forwarding. It does NOT include the actual implementation of log forwarding - this will come in a future PR.

## Plan changes

The Pebble layer specification has changed as follows:
- A new top-level section `log-targets`, where users specify the log targets they want Pebble to interact with.
- A new (optional) field `log-targets` in the service spec. This takes a list of names, which are matched against the targets defined in the `log-targets` top-level section.

The default behaviour is to forward all services' logs to all targets. This makes it easy to get started with log forwarding by just defining a new layer with the log targets - no need to change the service definitions.

Users can have more granular control over log forwarding by:
- Setting the optional `selection` field for a log target;
- In the definition for a service, explicitly listing log targets they want to forward this service's logs to.

More details can be found in the proposed changes to the README [here](https://github.com/canonical/pebble/pull/165/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

## QA steps

<details>
Create a simple service:

```bash
#!/bin/bash
while true; do
  echo "hello"
  sleep 1
done
```

Create a simple layer:

```yaml
services:
  svc1:
    override: merge
    command: <path/to/simple/service>
    startup: enabled
  svc2:
    log-targets: ['tgtA', 'tgtC']
    override: merge
    command: <path/to/simple/service>
    startup: enabled
  svc3:
    log-targets: ['tgtD']
    override: merge
    command: <path/to/simple/service>
    startup: enabled

log-targets:
  tgtA:
    type: loki
    location: foo
    override: merge
  tgtB:
    selection: opt-out
    type: loki
    location: foo
    override: merge
  tgtC:
    selection: opt-in
    type: loki
    location: foo
    override: merge
  tgtD:
    selection: disabled
    type: loki
    location: foo
    override: merge
```

Run Pebble with `PEBBLE_DEBUG=1`:

```
PEBBLE_DEBUG=1 go run ./cmd/pebble run
```

and check the log output to ensure that the log manager is receiving the correct events.

```
...
2023-03-29T00:32:09.847Z [pebble] DEBUG LogManager.PlanChanged called
2023-03-29T00:32:09.847Z [pebble] DEBUG LogManager: selected the following targets:
2023-03-29T00:32:09.847Z [pebble] DEBUG - svc1 -> [tgtA tgtB]
2023-03-29T00:32:09.847Z [pebble] DEBUG - svc2 -> [tgtA tgtC]
2023-03-29T00:32:09.847Z [pebble] DEBUG - svc3 -> []
...
2023-03-29T00:32:09.852Z [pebble] DEBUG LogManager.ServiceStarted called for service "svc1"
...
2023-03-29T00:32:10.865Z [pebble] DEBUG LogManager.ServiceStarted called for service "svc2"
...
2023-03-29T00:32:11.878Z [pebble] DEBUG LogManager.ServiceStarted called for service "svc3"
...
^C
2023-03-29T00:32:15.898Z [pebble] Exiting on interrupt signal.
...
2023-03-29T00:32:15.948Z [pebble] DEBUG LogManager.Stop called
```
</details>

## See also

- The [spec](https://docs.google.com/document/d/12tJFhIbVIlOcblG6l9_u-K5bCib78lojRJ-hdGBxSsw/edit?usp=sharing).


internal/plan/plan.go
```diff
@@ -178,6 +184,27 @@ func (s *Service) Merge(other *Service) {
 	if other.BackoffLimit.IsSet {
 		s.BackoffLimit = other.BackoffLimit
 	}
+	s.LogTargets = dedupStrings(s.LogTargets, other.LogTargets)
+}
+
+// dedupStrings merges arr1 and arr2 with deduplication (removing any duplicate
+// entries in the merged slice).
+func dedupStrings(arr1, arr2 []string) []string {
```
benhoyt, line 192: Minor thing: I was originally going to suggest a different name for this, but then I realised this is doing two things: concatenating the arrays and then de-duping (but it's name only mentions the de-duping). I think it'd be simpler and easier to name to separate the two. Something like this:

```go
s.LogTargets = dedupStrings(append(s.LogTargets, other.LogTargets...))

...
// dedupStrings removes duplicates in slice, maintaining the order.
func dedupStrings(slice []string) []string { ... }
```

internal/plan/plan.go
```diff
@@ -711,8 +855,9 @@ func (l *Layer) checkCycles() error {
 
 func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 	layer := Layer{
-		Services: map[string]*Service{},
-		Checks:   map[string]*Check{},
+		Services:   map[string]*Service{},
```
benhoyt, line 858: Should we also check for duplicates in Services.LogTargets after parsing it, and error if there are dupes in a single layer? (I think for merging layers it shouldn't be an error be de-duping, as you've done.)

benhoyt (approved):
Looks good to me, thanks! Just a minor comment, and a question.

internal/overlord/servstate/manager.go
```diff
@@ -34,6 +34,12 @@ type ServiceManager struct {
 
 	randLock sync.Mutex
 	rand     *rand.Rand
+
+	logMgr LogManager
+}
+
+type LogManager interface {
+	ServiceStarted(serviceName string, buffer *servicelog.RingBuffer)
```
niemeyer, line 42: We call the log ring buffer as `logs` everywhere in the existing code.  I don't mind which name we pick too much so please feel free to suggest something else, but I would prefer to pick one single name for this value across the implementation, instead of changing depending on context (`buffer` is probably not it either, though).

internal/overlord/servstate/manager.go
```diff
@@ -431,11 +439,14 @@ func (m *ServiceManager) Replan() ([]string, []string, error) {
...
-			if config.Equal(s.config) {
+		if newConfig, ok := m.plan.Services[name]; ok {
+			// TODO: be more conservative with restarts. Changes to metadata, log
+			// targets, dependencies, etc shouldn't require a restart.
+			// https://github.com/canonical/pebble/pull/165#discussion_r1142800922
+			if newConfig.Equal(s.config) {
 				continue
 			}
-			s.config = config.Copy() // update service config from plan
+			s.config = newConfig.Copy() // update service config from plan
```
niemeyer, line 449: The changes to this for loop need to be reverted as they're applying a personal preference to code that is not otherwise changing or improving meaningfully. If we allow this as a practice, the next person can come in and change it to a third thing and we'll never stop.

The TODO can be dropped for now too, as it's a bit vague ("metadata" may cause restarts) and debatable (dependencies should probably cause restarts in some situations too). Please feel free to open an issue if you'd like to track specific concerns separate from the on going work that you already have coming in the next couple of PRs, though.

internal/plan/plan.go
```diff
@@ -178,6 +184,27 @@ func (s *Service) Merge(other *Service) {
...
+	for _, s := range arr1 {
+		if _, ok := seen[s]; !ok {
+			merged = append(merged, s)
+		}
+	}
+	for _, s := range arr2 {
+		if _, ok := seen[s]; !ok {
+			merged = append(merged, s)
+		}
+	}
```
niemeyer, line 205: Can we please change this function so it looks like this:
```
// appendUnique appends into a the elements from b which are not yet present
// and returns the modified slice.
func appendUnique(a []string, b ...string) []string {
Outer:
	for _, bn := range b {
		for _, an := range a {
			if an == bn {
				continue Outer
			}
		}
		a = append(a, bn)
	}
	return a
}
```

Besides being shorter, this is significantly faster for cases we care about, uses only the minimum required memory, and does not allocate at all when there's nothing to do. The signature also matches that of `append`, which makes adding single strings slightly nicer and the behavior more expectable.

After this PR series is done, it'd be nice to move this into x-go/strutil.

internal/plan/plan.go
```diff
@@ -404,6 +431,53 @@ func (c *ExecCheck) Merge(other *ExecCheck) {
...
+	Override  Override      `yaml:"override,omitempty"`
+}
+
+// LogTargetType defines the protocol to use to forward logs.
+type LogTargetType string
+
+const (
+	LokiTarget   LogTargetType = "loki"
+	SyslogTarget LogTargetType = "syslog"
+	UnsetTarget  LogTargetType = ""
```
niemeyer, line 449: UnsetTarget for what?  `UnsetLogTarget`, or maybe `NoLogTarget`.

internal/plan/plan.go
```diff
@@ -404,6 +431,53 @@ func (c *ExecCheck) Merge(other *ExecCheck) {
...
+)
+
+// Selection describes which services' logs will be forwarded to this target.
+type Selection string
+
+const (
+	OptOutSelection   Selection = "opt-out"
+	OptInSelection    Selection = "opt-in"
+	DisabledSelection Selection = "disabled"
+	UnsetSelection    Selection = ""
```
niemeyer, line 459: This is fine to miss the log context, though. I can see ourselves using this same idea elsewhere.

internal/plan/plan_test.go
```diff
@@ -889,6 +1190,7 @@ func (s *S) TestReadDir(c *C) {
 	tempDir := c.MkDir()
 
 	for testIndex, test := range planTests {
+		fmt.Println(test.summary)
```
niemeyer, line 1193: c.Logf

internal/plan/plan_test.go
```diff
@@ -762,10 +768,305 @@ var planTests = []planTest{{
...
+			tgt1:
+				type: loki
+				location: http://10.1.77.196:3100/loki/api/v1/push
+				override: merge
+`},
 }}
 
 func (s *S) TestParseLayer(c *C) {
 	for _, test := range planTests {
+		fmt.Println(test.summary)
```
niemeyer, line 1069: c.Logf

internal/servicelog/iterator.go
```diff
@@ -69,6 +69,10 @@ func (it *iterator) Close() error {
 	return nil
 }
 
+// Next returns true if there is more data to read from the RingBuffer.
+// If a non-nil cancel channel is passed in, Next will wait for more data to
+// become available. Sending on this channel, or closing it, will cause Next to
+// return immediately.
```
niemeyer, line 75: The documentation for all the Iterator methods, including this one, is in the public interface above. Let's please address any necessary changes there instead of here.

niemeyer (changes requested):
Thanks, Jordan!

A few minor comments. Overall logic is great, thank you.

internal/overlord/logstate/manager.go
```diff
@@ -0,0 +1,92 @@
...
+			targetNames = append(targetNames, targetName)
+		}
+		logger.Debugf("- %s -> %v", serviceName, targetNames)
+	}
+}
+
+// selectTargets goes through the given plan and determines which service's
+// logs should be forwarded to which log targets.
+func selectTargets(p *plan.Plan) map[string]map[string]*plan.LogTarget {
+	planTargets := make(map[string]map[string]*plan.LogTarget, len(p.Services))
```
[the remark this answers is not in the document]
niemeyer: Update: I've updated the proposed signature.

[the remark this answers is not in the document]
barrettj12: Good idea. I've called the method `LogsTo` (seemed a bit more grammatical), and moved the target selection test to `internal/plan_test.go`.

internal/overlord/servstate/manager.go
barrettj12: Ok, I changed the argument name to `logs` everywhere inside the `servstate` package. I think inside the `logstate` package, `buffer` makes more sense as `logs` could refer to anything. I interpreted "across the implementation" to mean just the `servstate` package, as the buffer is called `rb` elsewhere.

jnsgruk:
Nice work, Jordan! :rocket:

internal/plan/plan.go
```diff
@@ -215,6 +211,27 @@ func (s *Service) Equal(other *Service) bool {
 	return reflect.DeepEqual(s, other)
 }
 
+// LogsTo returns true if the logs from s should be forwarded to target t.
+// This happens if:
+//   - t.Selection is "opt-out" or empty, and s.LogTargets is empty; or
```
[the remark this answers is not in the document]
barrettj12: Good point, and I did discuss this with Ben, but we opted to defer the discussion at the time in favour of more pressing things. I did a [quick test](https://go.dev/play/p/CR6teO5VuRG) and it seems that an unspecified `log-targets` field unmarshals to `[]string(nil)`, while `log-targets: []` unmarshals to `[]string{}`. So it would be possible to make the distinction - in the LogsTo function for example, we would check if `s.LogTargets == nil` instead of checking if `len(s.LogTargets) == 0`. We would just need to be careful about maintaining this distinction, especially when merging - I can see that when `append`ing, it would be easy for a nil slice to turn into an empty slice or vice versa. :)

PR #213, opened 2023-04-07
Fix checks timeout definition in README.md

NohaIhab:
small correction in Layer specification section in README.md. updated timeout definition: `timeout must not be more than the period` is the correct description NOT `timeout must not be less than the period`. defaults of timeout and period confirm that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,7 +541,7 @@ checks:
         period: <duration>
 
         # (Optional) If this time elapses before a single check operation has
-        # finished, it is cancelled and considered an error. Must not be less
+        # finished, it is cancelled and considered an error. Must not be more
         # than the period, and must not be zero. Default is "3s".
         timeout: <duration>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -541,7 +541,7 @@ checks:
         period: <duration>
 
         # (Optional) If this time elapses before a single check operation has
-        # finished, it is cancelled and considered an error. Must not be less
+        # finished, it is cancelled and considered an error. Must not be more
```
benhoyt, line 544: Let's be direct instead of using the negative:

```suggestion
        # finished, it is cancelled and considered an error. Must be less
```

benhoyt: Additionally the code / error message uses the direct version: https://github.com/canonical/pebble/blob/0c8bb88da360012178b23329cd4a0b395596c467/internal/plan/plan.go#L666

benhoyt:
Trivial doc fix, merging without further review. Thanks @NohaIhab!

PR #214, opened 2023-04-18
feat(cli): add `--dry` option to `pebble run`

anpep:
This PR adds dry start functionality to managers through the `DryStart()` method on the `StateManager` interface (KO026) and adds the `--dry` option to the `pebble run` command.

The `--dry` option is useful for validating the Pebble layers without having to actually start the Pebble daemon.
Note that, in any case, pebble will exit. If the plan was valid, pebble will exit with an error code of 0; otherwise, with a non-zero exit code.

The previous iteration of this was done in [PR #175](https://github.com/canonical/pebble/pull/175).


internal/overlord/backend.go
```diff
@@ -34,3 +34,15 @@ func (osb *overlordStateBackend) Checkpoint(data []byte) error {
 func (osb *overlordStateBackend) EnsureBefore(d time.Duration) {
 	osb.ensureBefore(d)
 }
+
+type nilStateBackend struct {
```
benhoyt, line 38: `nil` has a special meaning in Go, and a struct isn't nil, so this name is a bit confusing. How about `noopStateBackend`?

internal/overlord/stateengine.go
```diff
@@ -72,6 +80,36 @@ func (se *StateEngine) State() *state.State {
 	return se.state
 }
 
+type dryStartError struct {
+	errs []error
+}
+
+func (e *dryStartError) Error() string {
+	return fmt.Sprintf("dry start errors: %v", e.errs)
```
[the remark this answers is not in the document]
benhoyt: Additionally, this isn't really specific to dry-start, and probably shouldn't have the "dry start errors:" prefix. Maybe call it `multiError` and drop the prefix?

benhoyt (commented):
Looks good overall (though I don't think I was part of the discussions about the new way to approach this -- but this seems reasonable). A few minor comments.

anpep:
> Looks good overall (though I don't think I was part of the discussions about the new way to approach this -- but this seems reasonable). A few minor comments.

Thanks for your review Ben! Please refer to KO026 where I explain the new approach. I will be fixing this stuff starting tomorrow, and will let you know when it's ready for another review. Thanks again!

internal/overlord/stateengine.go
```diff
@@ -27,7 +27,7 @@ import (
 // the system and manipulating it to reflect the desired state.
 type StateManager interface {
 	// DryStart prepares the manager to run its activities without
-	// inccurring in side effects that might affect the system
+	// incurring in side effects that might affect the system
```
benhoyt, line 30: Grammar nit (sorry, missed this previously).

```suggestion
	// incurring side effects that might affect the system
```

hpidcock:
@anpep was this ready for re-review? It's marked as Priority, but doesn't appear to be ready for review?

anpep:
> @anpep was this ready for re-review? It's marked as Priority, but doesn't appear to be ready for review?

Sorry for the confusion, yes, this PR is ready for review.

internal/daemon/daemon.go
```diff
@@ -356,13 +360,13 @@ func logit(handler http.Handler) http.Handler {
 func (d *Daemon) Init() error {
 	listenerMap := make(map[string]net.Listener)
 
-	if listener, err := getListener(d.normalSocketPath, listenerMap); err == nil {
+	if listener, err := getListener(d.normalSocketPath, listenerMap, d.dry); err == nil {
```
flotter, line 363: I could be completely wrong here, but looking at the changes here for getListener makes me uncomfortable. I would have classed ```net.Dail()``` as a system state changing action, which is outside the scope of a dry run. In fact, it feels like a very slippery slope here, trying to plumb in new code paths for dry running at this level. For example, getListener() now has a completely new return state ```nil, nil`` that did not exist before, and which requires tests and subsequent code changes to accommodate this. Again, I stand to be corrected.

flotter: @anpep I feel there is a danger I lost sync on where this is going, so I will catch up with you when you are back before I continue my review. It feels like I may be missing something.

cmd/pebble/cmd_run.go
```diff
@@ -52,6 +53,7 @@ func init() {
 			"hold":        "Do not start default services automatically",
 			"http":        `Start HTTP API listening on this address (e.g., ":4000")`,
 			"verbose":     "Log all output from services to stdout",
+			"dry":         "Attempt to run without actually running",
```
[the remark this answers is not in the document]
flotter: I noted a previous comment the current version is preferred, so please ignore my proposal.

internals/overlord/overlord_test.go
```diff
@@ -502,7 +506,8 @@ func (ovs *overlordSuite) TestCheckpoint(c *C) {
 }
 
 type sampleManager struct {
-	ensureCallback func()
+	dryStartCallback func()
```
benhoyt, line 509: It seems like this field is never set. Is it needed?

internals/overlord/stateengine.go
```diff
@@ -15,16 +15,24 @@
...
-	"fmt"
+	"errors"
+	"strings"
 	"sync"
 
 	"github.com/canonical/pebble/internals/logger"
 	"github.com/canonical/pebble/internals/overlord/state"
 )
 
+var ErrStateEngineStopped = errors.New("state engine already stopped")
```
benhoyt, line 26: It seems like this doesn't need to be exported. Could it be `errStateEngineStopped`?

internals/overlord/stateengine.go
```diff
@@ -72,12 +80,38 @@ func (se *StateEngine) State() *state.State {
...
+	se.mgrLock.Lock()
+	defer se.mgrLock.Unlock()
+	if se.stopped {
+		return ErrStateEngineStopped
+	}
+	var errs []error
+	for _, m := range se.managers {
+		err := m.DryStart()
+		if err != nil {
+			logger.Noticef("dry start error: %v", err)
```
benhoyt, line 107: Do we need to log this in addition to to returning the errors? It seems like whoever's handling the dry start should already check that (multi)error.

internals/overlord/stateengine.go
```diff
@@ -92,18 +126,17 @@ func (se *StateEngine) Ensure() error {
...
 	defer se.mgrLock.Unlock()
 	if se.stopped {
-		return fmt.Errorf("state engine already stopped")
+		return ErrStateEngineStopped
 	}
 	var errs []error
 	for _, m := range se.managers {
 		err := m.Ensure()
 		if err != nil {
-			logger.Noticef("state ensure error: %v", err)
```
benhoyt, line 101: Kind of the opposite of above: I notice you removed the logging here -- why? I just want *ensure* the callers of Ensure always log/handle the error?

internals/overlord/overlord.go
```diff
@@ -133,6 +141,11 @@ func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writ
 	// the shared task runner should be added last!
 	o.stateEng.AddManager(o.runner)
 
+	// Dry start all managers.
+	if err := o.stateEng.DryStart(); err != nil {
```
benhoyt, line 145: Should this only happen if `dry` is true?

anpep: Not really; the dry start mechanism, as discussed on KO026, always runs in order to run initialization tasks that do not incur in unwanted side-effects.

internals/overlord/servstate/manager.go
```diff
@@ -237,6 +237,12 @@ func (m *ServiceManager) acquirePlan() (release func(), err error) {
 	return release, nil
 }
 
+// DryStart implements StateManager.DryStart.
+func (m *ServiceManager) DryStart() error {
+	_, err := plan.ReadDir(m.pebbleDir)
```
flotter, line 242: If the intention is that ```DryStart()``` should be the place where early unwanted-side-effect-free init happens (which is what I understand), I think we can use this is an example:

```
        releasePlan, err := m.acquirePlan()
        if err != nil {
                return err
        }
        defer releasePlan()
```

This will initialize ```m.plan```, which will prevent any future acquire from re-doing it.

internals/overlord/stateengine.go
```diff
@@ -15,16 +15,24 @@
...
 	"github.com/canonical/pebble/internals/overlord/state"
 )
 
+var errStateEngineStopped = errors.New("state engine already stopped")
+
 // StateManager is implemented by types responsible for observing
 // the system and manipulating it to reflect the desired state.
 type StateManager interface {
+	// DryStart prepares the manager to run its activities without
+	// incurring side effects that might affect the system
```
flotter, line 32: align wording: "without incurring unwanted side-effects"

internals/overlord/stateengine.go
```diff
@@ -92,7 +129,7 @@ func (se *StateEngine) Ensure() error {
 	se.mgrLock.Lock()
 	defer se.mgrLock.Unlock()
 	if se.stopped {
```
flotter, line 131: If you look at the ```snapd``` version of ```StartUp``` you will see they added a check inside ```Ensure``` to enforce running ```StartUp``` before ```Ensure```. Since our ```DryStart``` is very similar, in that it may initialise critical state before the manager may be able to run, should we do the same? Since ```DryStart``` is in ```New()``` while ```Ensure``` happens later, it seems like this will not be something likely to go wrong.

flotter (approved):

internals/overlord/backend.go
```diff
@@ -34,3 +34,4 @@ func (osb *overlordStateBackend) Checkpoint(data []byte) error {
 func (osb *overlordStateBackend) EnsureBefore(d time.Duration) {
 	osb.ensureBefore(d)
 }
+
```
flotter, line 37: Is this a deliberate addition of a newline?

flotter (approved):

internals/overlord/backend.go
anpep: It's accidental, good catch!

benhoyt (approved):

atesburak (approved):

internals/cli/cmd_run.go
```diff
@@ -154,6 +157,9 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 	if err != nil {
 		return err
 	}
+	if rcmd.Dry {
```
niemeyer, line 160: Shouldn't this go all the way to the sanityCheck below, right before Start?

internals/overlord/overlord.go
niemeyer: Indeed, but it's a bit awkward to see this here at the end of Init.

I think this shoud be run when starting, or on a specific `DryStart` method that only does the dry-starting. This way we also don't need to provide further arguments into the constructor here, and no bools either.

internals/overlord/managers_test.go
```diff
@@ -42,7 +42,7 @@ func (s *mgrsSuite) SetUpTest(c *C) {
 
 	s.dir = c.MkDir()
 
-	o, err := overlord.New(s.dir, nil, nil)
+	o, err := overlord.New(s.dir, nil, nil, false)
```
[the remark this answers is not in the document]
niemeyer: This method has remained quite simple, with those couple of arguments, for several years now. But every few weeks we have a PR that tries to add more parameters to it, and if we had allowed these to go on we'd have a struct with 50 different options by now. Not everything that needs to be passed down a chain of objects needs to be in a constructor. In the case at hand, per the other comment, we should not do DryStart on the Init, but rather inside Start itself and only if it has not yet run externally. On the Daemon, we can call a DryStart on the overlord itself when we're doing a dry run.

internals/overlord/stateengine.go
```diff
@@ -72,12 +80,42 @@ func (se *StateEngine) State() *state.State {
 	return se.state
 }
 
-type ensureError struct {
+type multiError struct {
```
niemeyer, line 83: Related to the other point: this was a semantic error (an error in ensure, specifically), and it's now just a grouping. If we want to generalize this, should we be using the "new" errors package (new for the lifecycle of this system, not as in recent)?  Are there any places catching this error that are actually trying to handle ensure errors, or is it always just grouping?  Let's please check this in snapd as well.

internals/overlord/overlord.go
```diff
@@ -318,15 +320,15 @@ func (o *Overlord) settle(timeout time.Duration, beforeCleanups func()) error {
...
+				return &multiError{append(errs, err)}
 			}
 			return err
 		}
 		next := o.ensureTimerReset()
 		err := o.stateEng.Ensure()
 		switch ee := err.(type) {
 		case nil:
-		case *ensureError:
+		case *multiError:
```
niemeyer, line 331: Major difference here: "is it an error in ensure" vs. "is this any sort of error that happens to group multiple errors". Doesn't matter for the use case here, but is there any other where it actually does?

internals/overlord/stateengine.go
```diff
@@ -15,16 +15,23 @@
...
-	"fmt"
+	"errors"
+	"strings"
 	"sync"
 
 	"github.com/canonical/pebble/internals/logger"
 	"github.com/canonical/pebble/internals/overlord/state"
 )
 
+var errStateEngineStopped = errors.New("state engine already stopped")
```
niemeyer, line 26: This should be defined close to where it's relevant.

niemeyer (changes requested):
Thanks Angel. Probably the last pass.

internals/overlord/stateengine.go
anpep: This indeed is not a good solution. However, I think we could replace `ensureError`/`dryStartError` with snapd's `multiError`, since the `newMultiError(header, errs string)` function forces you to provide a "primary" error message.

In our case, I haven't seen anywhere in the code where we handle `multiError`s meaningfully, we either discard the function return value or discriminate `multiError`s only to construct a new one.

In snapd, only two functions create new `multiError`s: `CleanupAbandonedImports()` and `(t *importTransaction) Cancel()`. The `multiError` of the former is logged, while in the latter errors are not handled in any case.

internals/overlord/overlord.go
anpep: As I indicated in the comment above, I haven't seen any logic in Pebble that handles the various error types in a different way.

internals/cli/cmd_run.go
anpep: I've extensively discussed this with @flotter, and one of his (rightful) concerns where around system-wide side effects caused by e.g. listening on a port, since it's an "external" side effect (i.e. contributes to Pebble initialization by persisting some state externally, be it listening on a socket or calling `.Checkpoint()` on the state backend).

The idea is for the validation to be contained in the `.DryStart()` methods, and also `sanityCheck()` is a no-op for now anyway.

niemeyer:
> [anpep](https://github.com/anpep) [force-pushed](https://github.com/canonical/pebble/compare/73f1670e34bc4d58a09a7b2db9fd6247e467da79..782783f01cafcea839ba1298e5f8f59598b04794) the overlord-prevalidation branch from [73f1670 ](https://github.com/canonical/pebble/commit/73f1670e34bc4d58a09a7b2db9fd6247e467da79)to [782783f ](https://github.com/canonical/pebble/commit/782783f01cafcea839ba1298e5f8f59598b04794)
[4 hours ago](https://github.com/canonical/pebble/pull/214#event-9635595715)

Not sure if we discussed this before, but as a rule let's please not force-push after the reviews have initiated, as it makes incremental reviews uncomfortatable given the current GitHub features.

internals/overlord/stateengine.go
niemeyer: Okay, so here is a suggested algorithm for the Error method on multiError:

1) If it's a single error, return `<prefix>+<errs[0]>`
2) With more than one, return: `Join([<prefix>, errs...], "\n<two spaces>-<one space>")`

That is, with len(errs) == 1, we get:
```
cannot do foo bar: something bad
```
while len(errs) > 1 become:
```
cannot do foo bar:
  - something bad
  - something else also bad
```

jnsgruk:
@anpep looks like we could do with a merge conflict fixing here, and still some outstanding feedback.

anpep:
> @anpep looks like we could do with a merge conflict fixing here, and still some outstanding feedback.

This PR still needs some discussion before getting in—will try to catch up rather soon.

internals/overlord/overlord.go
anpep: The issue of running this when starting (e.g. at the top of `overlord.Loop()`) is that `Loop()` might be called too late and other components (e.g. the daemon) might have already done some state-mutating operations.

The issue of requiring callers to call an `overlord.DryStart()` method manually is that it's easy to forget, and thus misuse the API. Also, `DryStart()` is not an operation you do on the overlord, but rather on the each of the managers. Holding the caller responsible for initializing the managers doesn't seem right to me: the overlord should be the one doing this very internal operation on the managers.

I think running this at the bottom of `overlord.New()` is the sweet spot for these reasons. Also, we don't need any kind of arguments (that argument from previous commits was clearly a bug and a misunderstanding of the problem on my side), because `DryStart()` is meant to be called always.

paul-rodriguez (approved):

thp-canonical:
From the PR description:

> The --dry option is useful for validating the Pebble layers without having to actually start the Pebble daemon.

From what I can tell looking at the code, Rockcraft currently uses [Pydantic](https://docs.pydantic.dev/latest/) for validating the Pebble configuration it creates (implemented in https://github.com/canonical/rockcraft/pull/331, in `rockcraft/pebble.py`).

Is this `--dry` feature something that Rockcraft could eventually use instead of Pydantic, or do the two implementations (`--dry` vs Pydantic) solve different problems and can/should coexist?

Or alternatively, would using Pydantic in *craft tools be a good alternative to running Pebble with `--dry`?

cc @cjdcordeiro and @tigarmo for the Rockcraft/*craft view on this, so we don't duplicate (and maintain) Pebble layer validation as two different methods in various craft tools.

cjdcordeiro:
@thp-canonical Ideally, the two would coexist. Rockcraft would still use Pydantic to declare the `services` and `checks` top-level fields, but then possibly rely on `--dry` to validate the user's YAML input.

tigarmo:
I sometimes worry about the two validations "drifting". Since Rockcraft provisions `latest/stable` Pebble, we kind of have to evolve in lockstep. For example: if the Pebble service layer format gets a new field, users of Rockcraft won't be able to use it until we update the Pydantic validation and release a whole new version of the tool.

benhoyt:
Per discussion with @anpep, this is likely to end up in a different form later, so closing this for now.

PR #215, opened 2023-04-25
Use "pebble version --client" in GitHub Action now that that exists

benhoyt:
This simplifies, and avoids the 5-second delay from trying to look up the server version. Note that the `--client` arg for `pebble version` was added very recently in https://github.com/canonical/pebble/pull/208.


jnsgruk (approved):

benhoyt:
Trivial 1-line GitHub Actions update following on from #208 -- merging without further review.

PR #218, opened 2023-05-09
[JUJU-3776] Log forwarding implementation

barrettj12:
This PR contains the second part of #165, including the actual mechanics of log forwarding in Pebble. It builds upon #209 and #252. This includes an abstract `logClient` interface but doesn't include the specific Loki/syslog implementations - this will come in later PRs.

### Current design

For each service, there is a single "forwarder" which continually pulls logs from the service ringbuffer, and writes them to the relevant "gatherers". Each forwarder runs in a separate goroutine.

For each log target, there is a single "gatherer" which receives logs from forwarders, serialises them, prepares a request to the remote, and flushes periodically. Each gatherer runs in a separate goroutine - although the gatherer's `addLog` method is called inside a forwarder's goroutine.

The gatherer uses the abstract interfaces
- `logBuffer`, which is responsible for serialising/storing logs in a backend-specific format
- `logClient`, which is responsible for sending logs to a remote target via a backend-specific protocol.

### DIFF
--- a/internals/overlord/logstate/forwarder.go
+++ b/internals/overlord/logstate/forwarder.go
@@ -1,0 +1,71 @@
+// Copyright (c) 2023 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package logstate
+
+import (
+	"sync"
+
+	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/servicelog"
+)
+
+// logForwarder is responsible for pulling logs from a service's ringbuffer,
+// and distributing each log message to its logGatherers. Its gatherers field
+// holds a reference to the gatherer for each log target that the service is
+// sending logs to.
+// One logForwarder will run per service. Its forward() method should be run
+// in its own goroutine.
+type logForwarder struct {
+	serviceName string
+
+	mu        sync.Mutex // mutex for gatherers
+	gatherers []*logGatherer
+
+	cancel chan struct{}
+}
+
+func newLogForwarder(serviceName string) *logForwarder {
+	f := &logForwarder{
+		serviceName: serviceName,
+		cancel:      make(chan struct{}),
+	}
+
+	return f
+}
+
+func (f *logForwarder) forward(buffer *servicelog.RingBuffer) {
+	iterator := buffer.TailIterator()
+	// TODO: don't use the parser, just pull/write bytes from iterator
+	parser := servicelog.NewParser(iterator, 1024 /* TODO*/)
+
+	for iterator.Next(f.cancel) {
+		for parser.Next() {
+			entry := parser.Entry()
+			f.mu.Lock()
+			gatherers := f.gatherers
+			f.mu.Unlock()
+			for _, c := range gatherers {
+				c.addLog(entry)
+			}
+		}
+		if err := parser.Err(); err != nil {
+			logger.Noticef("Cannot read logs from service %q: %v", f.serviceName, err)
+		}
+	}
+}
+
+func (f *logForwarder) stop() {
+	close(f.cancel)
+}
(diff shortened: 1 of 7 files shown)


internals/overlord/logstate/forwarder.go
```diff
@@ -0,0 +1,71 @@
...
+			entry := parser.Entry()
+			f.mu.Lock()
+			gatherers := f.gatherers
+			f.mu.Unlock()
+			for _, c := range gatherers {
+				c.addLog(entry)
+			}
+		}
+		if err := parser.Err(); err != nil {
+			logger.Noticef("Cannot read logs from service %q: %v", f.serviceName, err)
```
hpidcock, line 64: So from what I can tell, if we get a parser.Err() != nil here, that means we had a failure with the RingBuffer iterator, so it is probably safe to either:
1. exit after printing an error
2. reset the parser + get the buffer.TailIterator() and start again.
3. this code is already fine as iterator.Next() should return false if the RingBuffer has reached the end, and we will never actually have an error with the parser.

hpidcock: All I'm asking for here is clarity on error handling, either by adding a comment, or handling the error in the way it needs to.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,170 @@
...
+	req, err := g.buffer.Request()
+	if err != nil {
+		logger.Noticef("couldn't generate request for target %q: %v", g.target.Name, err)
+		return
+	}
+
+	err = g.client.Send(req)
+	if err != nil {
+		logger.Noticef("couldn't send logs to target %q: %v", g.target.Name, err)
+		// TODO: early return here? should we reset buffer if send fails?
```
hpidcock, line 125: I think this is fine to reset the buffer, it should be up to the client to retry if it knows it can do so safely without holding up the forwarders.

internals/overlord/logstate/manager.go
```diff
@@ -15,26 +15,98 @@
...
+			delete(m.forwarders, serviceName)
+		}
+
+		// update clients
+		forwarder.mu.Lock()
+		forwarder.gatherers = []*logGatherer{}
+
+		for _, target := range pl.LogTargets {
+			// Only create the gatherer if there is a service logging to it.
+			if service.LogsTo(target) {
```
hpidcock, line 68: invert this if statement and continue to avoid indentation.
```
if !service.LogsTo(target) {
    continue
}
gatherer := m.gatherers[serviceName]
...

File: internals/overlord/logstate/manager.go [hunk 4]
```diff
@@ -44,5 +116,13 @@ func (m *LogManager) Ensure() error {
 
 // Stop implements overlord.StateStopper and stops all log forwarding.
 func (m *LogManager) Stop() {
-	// TODO: implement
+	wg := sync.WaitGroup{}
+	for _, f := range m.forwarders {
+		wg.Add(1)
+		go func(f *logForwarder) {
```
hpidcock, line 122: why do we need the goroutine here to stop the forwarder? the forwarder stop just closes a channel.

hpidcock (commented):
Looks like good progress here.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,170 @@
...
+func (g *logGatherer) stop() {
+	close(g.cancel)
+}
+
+// logBuffer is an interface encapsulating format-specific buffering of log
+// messages. E.g. a logBuffer for Loki would encode the log messages in the
+// JSON format expected by Loki.
+// A logBuffer's methods may not be concurrency-safe. Callers should protect
+// the logBuffer using a sync.Mutex.
+type logBuffer interface {
```
benhoyt, line 141: Per discussion, this is an awkward interface (and possibly inefficient and buffer-requiring with `Request` returning an `io.Reader` for later) and has to be implemented per client anyway, so the plan is to just move any buffering needed for sending into the client implementation.

benhoyt (commented):
Per voice discussion, this approach is still problematic due to the buffer locking in `addLog`, so `addLog` is essentially blocking if the sends are slow anyway. We do want to protect against a slow target meaning we drop logs for other targets on the floor, so the updated plan is:

* Get rid of the forwarder.
* Have the gatherer create a list of ring buffer iterators, and start a goroutine with a `for iterator.Next(cancel)` loop for each (it's currently not really possible to watch multiple iterators in a single goroutine, and goroutines are cheap).
* This will avoid additional buffering, as well as preventing one slow target meaning that logs for other services are lost.

hpidcock:
> it's currently not really possible to watch multiple iterators in a single goroutine, and goroutines are cheap

Yes we can, we can share one Notify channel on all the iterators. Then you just call `Next(nil)` on each.

barrettj12:
Closing this as we've identified problems with this approach, and will be taking a slightly different tack, as described in Ben's comment [here](https://github.com/canonical/pebble/pull/218#pullrequestreview-1515605492).

PR #221, opened 2023-05-18
Allow `!replace`-ing service `log-targets` in merge

barrettj12:
Alternative to #223.

### Context

#209 and #217 have included extensive discussion of the configuration and merging of log targets specified in a service definition. Two issues have been identified when it comes to merging the `log-targets` field:

1. It is hard to restore the default behaviour provided when `log-targets` is not specified. The issue is that specifying a null value
   ```yaml
   svc1:
     command: foo
     log-targets: ~ # or null
   ```
   parses identically to not specifying a value at all:
   ```yaml
   svc1:
     command: foo
   ```

2. It is hard to remove a log target which was previously specified. If your base layer contains
   ```yaml
   svc1:
     log-targets: [a,b]
   ```
   and you merge the following layer on top of it:
   ```yaml
   svc1:
     log-targets: [c]
     override: merge
   ```
   it will yield `log-targets: [a,b,c]`.

As it stands, the only way to get around either issue is to replace the entire service - which is a real usability issue. This PR proposes one solution to the problem.

### The solution

To solve (1), we need an **explicit value** to represent the default behaviour. We can't use `null` in a merge, as when this is parsed, it is indistinguishable from an unspecified value. To this end, I have added a `default` keyword which can be specified as the value of `log-targets` to restore the default behaviour:
```yaml
svc1:
  log-targets: default
```

To solve the issue raised by Gustavo [here](https://github.com/canonical/pebble/pull/209#discussion_r1155712205), I have also added a `none` keyword which a service can use to entirely opt-out of log forwarding. Personally I feel this is a safer approach than what I've proposed in #217. Distinguishing between `[]string(nil)` and `[]string{}` in Go is fiddly and error-prone, and it's bound to cause issues down the road when merging and passing structs around.

Finally, I've added an `all` keyword - this is not strictly necessary, but could be convenient if a user has a large number of log targets.

To solve (2), I've added a `!replace` tag, which a service can use during a merge to indicate that the value of `log-targets` should be replaced, not merged. i.e. merging
```yaml
svc1:
  log-targets: !replace [c]
  override: merge
```
on top of another layer will overwrite the value of `log-targets`, yielding just `log-targets: [c]`.

So, the full schema for the `log-targets` field is now:
```
log-targets: [!replace] default | none | all | <list of log targets>
```

I've implemented a custom type `LogTargets` with a custom `UnmarshalYAML` method to handle all this.

### Merging rules

These are the rules I decided on for merging `log-targets` - see the `MergeLogTargets` function. Open to discussing the details here.

1. Merging `anything1` <- `!replace anything2` yields `anything2`. In particular, merging `list1` <- `!replace list2` yields `list2`.
2. Merging `list1` <- `list2` will combine the two lists.
3. Merging `list` <- `keyword` yields `keyword`.
4. Merging `keyword1` <- `keyword2` yields `keyword2`
5. Merging `keyword` <- `list` yields `list` (if the list is nonempty).

Essentially, the `!replace` directive is only needed when wanting to replace a list with another list.

### DIFF
--- a/internal/plan/plan.go
+++ b/internal/plan/plan.go
@@ -90,7 +90,7 @@ type Service struct {
 	KillDelay      OptionalDuration         `yaml:"kill-delay,omitempty"`
 
 	// Log forwarding
-	LogTargets LogTargets `yaml:"log-targets,omitempty"`
+	LogTargets *LogTargets `yaml:"log-targets,omitempty"`
 }
 
 // Copy returns a deep copy of the service.
@@ -119,7 +119,9 @@ func (s *Service) Copy() *Service {
 			copied.OnCheckFailure[k] = v
 		}
 	}
-	copied.LogTargets = s.LogTargets.Copy()
+	if s.LogTargets != nil {
+		copied.LogTargets = s.LogTargets.Copy()
+	}
 	return &copied
 }
 
@@ -184,7 +186,9 @@ func (s *Service) Merge(other *Service) {
 	if other.BackoffLimit.IsSet {
 		s.BackoffLimit = other.BackoffLimit
 	}
-	s.LogTargets = MergeLogTargets(s.LogTargets, other.LogTargets)
+	if s.LogTargets != nil {
+		s.LogTargets.Merge(other.LogTargets)
+	}
 }
 
 // appendUnique appends into a the elements from b which are not yet present
@@ -280,27 +284,29 @@ func (t *LogTargets) UnmarshalYAML(value *yaml.Node) error {
 }
 
 // Copy returns a deep copy of this LogTargets struct.
-func (t *LogTargets) Copy() LogTargets {
-	return LogTargets{
+func (t *LogTargets) Copy() *LogTargets {
+	return &LogTargets{
 		Targets: append([]string(nil), t.Targets...),
 		Keyword: t.Keyword,
 		Replace: t.Replace,
 	}
 }
 
-// MergeLogTargets returns the result of merging other into t.
-func MergeLogTargets(t, other LogTargets) (res LogTargets) {
-	if other.Replace || other.Keyword != "" {
-		return other
-	}
+// Merge merges the fields set in other into t.
+func (t *LogTargets) Merge(other *LogTargets) {
+	if other.Replace ||
+		// anything <- keyword yields keyword
+		other.Keyword != "" ||
+		// keyword <- [list] yields [list]
+		t.Keyword != "" && len(other.Targets) > 0 {
 
-	// keyword <- [list] yields [list]
-	if t.Keyword != "" && len(other.Targets) > 0 {
-		return other
+		// Replace t with other
+		t.Keyword = other.Keyword
+		t.Targets = append([]string(nil), other.Targets...)
 	}
 
 	// [list] <- [list] - merge lists
-	res.Targets = appendUnique(t.Targets, other.Targets...)
+	t.Targets = appendUnique(t.Targets, other.Targets...)
 	return
 }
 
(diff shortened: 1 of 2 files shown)


internal/plan/plan.go
```diff
@@ -184,7 +184,7 @@ func (s *Service) Merge(other *Service) {
 	if other.BackoffLimit.IsSet {
 		s.BackoffLimit = other.BackoffLimit
 	}
-	s.LogTargets = appendUnique(s.LogTargets, other.LogTargets...)
+	s.LogTargets = MergeLogTargets(s.LogTargets, other.LogTargets)
```
benhoyt, line 187: Could this should be `s.LogTargets.Merge(other.LogTargets)`, rather than a stand-alone function, for consistency with the other `Merge` methods? `Merge` is a method on the pointer receiver, so it can modify the object as needed.

internal/plan/plan.go
```diff
@@ -258,6 +245,93 @@ const (
...
+			return fmt.Errorf("invalid value %q for LogTargets", value.Value)
+		}
+	}
+
+	// unmarshal to []string
+	return value.Decode(&t.Targets)
+}
+
+// Copy returns a deep copy of this LogTargets struct.
+func (t *LogTargets) Copy() LogTargets {
```
benhoyt, line 283: I think this should probably return a `*LogTargets` like the other `Copy` methods.

internal/plan/plan.go
```diff
@@ -258,6 +245,93 @@ const (
...
+	AllLogTargets     = "all"
+	NoLogTargets      = "none"
+
+	// Tag used to replace value (rather than append)
+	logTargetsReplaceTag = "!replace"
+)
+
+func (t *LogTargets) UnmarshalYAML(value *yaml.Node) error {
+	// Check for !replace tag
+	t.Replace = (value.Tag == logTargetsReplaceTag)
```
benhoyt, line 266: I think we should disallow `!replace` when using a keyword, because it doesn't really add any meaning over `none` when using something like `!replace none`, right? The user might be confused and try the latter, so I think we should disallow it to avoid that confusion. So the field spec would be:

```
log-targets: default | all | none | [!replace] <list of targets>
```

benhoyt (approved):
I like this overall, and "all" and "none" seem potentially useful. Small suggestion about the `Merge` and `Copy` signatures, and recommendation that we only allow `!replace` with a list.

internal/plan/plan.go
barrettj12: Originally I had `s.LogTargets` as a pointer, but kept running into nil pointer dereference issues. Easy fix was to make it a value field instead. I could change it back to a pointer and just be more vigilant about checking for nil before every operation.

benhoyt: I don't think we need the `LogTargets` field to be a pointer to do this ([example](https://go.dev/play/p/JKGF0xTslLv)). I'm just referring to `Merge` being a method and having a similar signature to the other `Merge` methods.

internal/plan/plan.go
barrettj12: "Disallow" is a strong word. Do you mean the following should result in a parsing error?
```yaml
services:
  svc1:
    command: foo
    override: merge
    log-targets: !replace none
```

Because currently, you are allowed to put whatever tags you want on any field, and Pebble doesn't care. This parses just fine:
```yaml
services:
  svc1:
    command: !mycustomtag foo
    override: !iliketagging merge
```

internal/plan/plan.go
barrettj12: yep, which means we should be using `*LogTargets` instead of `LogTargets` everywhere

internal/plan/plan.go
benhoyt: Ah yes, parsing error is what I meant. However, in light of what you've said about arbitrary tags being allowed in general, I retract that -- resolving.

PR #222, opened 2023-05-18
randutil: allow pseudo rng seeding by caller

flotter:
The randutil package was promoted to the x-go library because it provides helper functions needed by multiple projects.

Feedback from https://github.com/canonical/x-go/pull/17 highlighted the need for seeding of the pseudo random dependent functions to be left to the user of randutil (the crypto/rand functions are unaffected). The reason is simply that the seeding requirements differ between use cases (and projects), and it should be the responsibility of the caller to supply the seed.

Example Usage:

prng := randutil.NewPseudoRand(nil)
:
tmpfile := prng.RandomString(12)
:
wait := prng.RandomDuration(time.Hour)

Apply the changes required to make Pebble use the new randutil api.

### DIFF
--- a/internal/osutil/export_test.go
+++ b/internal/osutil/export_test.go
@@ -27,6 +27,8 @@ import (
 	"github.com/canonical/pebble/internal/osutil/sys"
 )
 
+var AtomicFilePrng = atomicFilePrng
+
 func FakeUserCurrent(f func() (*user.User, error)) func() {
 	realUserCurrent := userCurrent
 	userCurrent = f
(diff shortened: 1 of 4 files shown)


flotter:
Dear reviewers, for details on this PR please refer to: https://github.com/canonical/x-go/pull/22

This PR requires an update to pull in the required x-go commit once an agreement is in place. In the meantime, the impact of the changes can be seen in this PR.

benhoyt (approved):
Seems reasonable to me, dependent on https://github.com/canonical/x-go/pull/22 changes.

hpidcock (approved):

niemeyer (changes requested):
This doesn't seem to agree with prior conversations we had on the topic or the review that was discussed on x-go. I have a call with Fred tomorrow, so I believe this will be addressed there.

flotter:
@niemeyer Yes, this will get an update after our chat tomorrow.

flotter:
I am going to close this PR for now. Some rework is needed on a proposal I made on the snapd side, which will allow change proposed in Pebble, x-go and Snapd to align. Until this happens, this PR is stuck, and for now I would rather remove this from the Pebble backlog.

PR #223, opened 2023-05-19
[JUJU-3777] Allow replacing in merge using `^log-targets`

barrettj12:
Replaces #217 and #221.

### Context

#209 and #217 have included extensive discussion of the configuration and merging of log targets specified in a service definition. Two issues have been identified when it comes to merging the `log-targets` field:

1. It is hard to restore the default behaviour provided when `log-targets` is not specified. The issue is that specifying a null value
   ```yaml
   svc1:
     command: foo
     log-targets: ~ # or null
   ```
   parses identically to not specifying a value at all:
   ```yaml
   svc1:
     command: foo
   ```

2. It is hard to remove a log target which was previously specified. If your base layer contains
   ```yaml
   svc1:
     log-targets: [a,b]
   ```
   and you merge the following layer on top of it:
   ```yaml
   svc1:
     log-targets: [c]
     override: merge
   ```
   it will yield `log-targets: [a,b,c]`.

As it stands, the only way to get around either issue is to replace the entire service - which is a real usability issue. This PR proposes one solution to the problem.

### The solution

To solve (2), I've added a new key `^log-targets`, which a service can use during a merge to indicate that the value of `log-targets` should be replaced, not merged. So merging
```yaml
svc1:
  ^log-targets: [c]
  override: merge
```
on top of another layer will overwrite the value of `log-targets`, yielding just `log-targets: [c]`.

To solve (1), we need an **explicit value** to represent the default behaviour. We can't use `null` in a merge, as when this is parsed, it is indistinguishable from an unspecified value. To enable this, the `^log-targets` key also accepts a `default` keyword which restores the default behaviour. So merging
```yaml
svc1:
  ^log-targets: default
```
on top of another layer will result in `log-targets: ~`, i.e. equivalent to if `log-targets` had never been specified at all.

I've also copied over the relevant code from #217 so that we correctly distinguish between
- `log-targets: null` (= unspecified = default behaviour = forward to all `opt-out` targets)
- `log-targets: []` (= explicitly specify empty list = opt-out = don't forward to any targets)

So, the relevant part of the service spec now looks like:
```yaml
<service name>:
  ...
  log-targets: <list of log targets>
  ^log-targets: <list of log targets> | default
```

I've implemented a custom type `LogTargetsReplace` with a custom `UnmarshalYAML` method to handle the new `^log-targets` field. The `Service` struct now has fields
```go
type Service struct {
  ...
  LogTargets        []string           `yaml:"log-targets,omitempty"`
  LogTargetsReplace *LogTargetsReplace `yaml:"^log-targets,omitempty"`
}
```
and the `LogTargetsReplace` field takes precedence in a merge.

I've also added documentation about all this to the README - which will remain commented out until log forwarding is fully implemented.

### Merging rules

The merge rules for `log-targets` and `^log-targets` are documented in `TestParseAndMergeLogTargets`. In short:
1. Merging `log-targets: [list1]` <- `log-targets: [list2]` yields `log-targets: [list1+list2]` (with any duplicates removed).
2. Merging `log-targets: [list1]` <- `^log-targets: [list2]` yields `log-targets: [list2]`.
3. Merging `log-targets: [list1]` <- `^log-targets: default` yields `log-targets: ~`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,8 @@ services:
         log-targets: ['tgtA', 'tgtC']
     svc3:
         log-targets: ['tgtD']
+    svc4:
+        log-targets: []
 log-targets:
     tgtA:
         ...
@@ -450,9 +452,12 @@ log-targets:
         selection: disabled
 ```
 
-- `svc1` doesn't explicitly specify log-targets, so its logs will be forwarded to `tgtA` and `tgtB`.
+- `svc1` doesn't explicitly specify log-targets, so its logs will be forwarded to the opt-out targets: `tgtA` and `tgtB`.
 - `svc2` explicitly specifies `tgtA` and `tgtC`, so its logs will be sent to those targets. `svc2`'s logs will not be sent to `tgtB` or `tgtD`.
 - `svc3` explicitly specifies `tgtD`, but since this target is disabled, it will not receive any logs. Hence, `svc3`'s logs will not be sent anywhere.
+- `svc4` explicitly specifies an *empty list* of log targets. Hence, its logs will not be sent anywhere.
+
+TODO: document the replace key ^log-targets and default value
 -->
 
 ## Container usage
--- a/README.md
+++ b/README.md
@@ -403,6 +403,58 @@ $ pebble run --verbose
 ...
 ```
 
+<!--
+TODO: uncomment this section once log forwarding is fully implemented
+TODO: add log targets to the Pebble layer spec below
+
+#### Log forwarding
+
+Pebble supports forwarding its services' logs to a remote Loki server or syslog receiver (via UDP/TCP). In the `log-targets` section of the plan, you can specify destinations for log forwarding, for example:
+```yaml
+log-targets:
+    loki-example:
+        override: merge
+        type: loki
+        location: http://10.1.77.205:3100/loki/api/v1/push
+    syslog-example:
+        override: merge
+        type: syslog
+        location: tcp://192.168.10.241:1514
+```
+
+By default, all services' logs will be forwarded to these targets. For each log target, you can optionally specify the `selection` key, with the following possible values:
+
+* `opt-out`: each service which doesn't *explicitly specify* log targets will have its logs forwarded to this target. This is the default behaviour.
+* `opt-in`: a service must explicitly specify this log target to have its logs forwarded here.
+* `disabled`: no logs will be sent to this target. This can be used in combination with `override: replace` to disable a log target defined in an earlier layer.
+
+In the definition for each service, you can also specify a list `log-targets` of log targets to forward this service's logs to. The names in the list will be matched against targets defined in the `log-targets` section of the plan. If no targets are specified, the service's logs will be sent to all "opt-out" targets.
+
+For example, in the following (incomplete) plan:
+```yaml
+services:
+    svc1:
+        ...
+    svc2:
+        log-targets: ['tgtA', 'tgtC']
+    svc3:
+        log-targets: ['tgtD']
+log-targets:
+    tgtA:
+        ...
+    tgtB:
+        selection: opt-out
+    tgtC:
+        selection: opt-in
+    tgtD:
+        selection: disabled
+```
+
+- `svc1` doesn't explicitly specify log-targets, so its logs will be forwarded to `tgtA` and `tgtB`.
+- `svc2` explicitly specifies `tgtA` and `tgtC`, so its logs will be sent to those targets. `svc2`'s logs will not be sent to `tgtB` or `tgtD`.
+- `svc3` explicitly specifies `tgtD`, but since this target is disabled, it will not receive any logs. Hence, `svc3`'s logs will not be sent anywhere.
+-->
+
 ## Container usage
 
 Pebble works well as a local service manager, but if running Pebble in a separate container, you can use the exec and file management APIs to coordinate with the remote system over the shared unix socket.
--- a/README.md
+++ b/README.md
@@ -457,7 +457,47 @@ log-targets:
 - `svc3` explicitly specifies `tgtD`, but since this target is disabled, it will not receive any logs. Hence, `svc3`'s logs will not be sent anywhere.
 - `svc4` explicitly specifies an *empty list* of log targets. Hence, its logs will not be sent anywhere.
 
-TODO: document the replace key ^log-targets and default value
+When merging services, new `log-targets` specified in the override layer will be *appended* to the existing `log-targets` specified in the base layer. For example, with the (partial) base layer:
+```yaml
+svc1:
+    log-targets: [tgt1]
+```
+and override layer
+```yaml
+svc1:
+    log-targets: [tgt2]
+    override: merge
+```
+the resulting merged layer will be
+```yaml
+svc1:
+    log-targets: [tgt1, tgt2]
+```
+
+To remove existing `log-targets` specified by a service, use the special `^log-targets` key. This behaves like `log-targets` but has "replace" semantics, even in a merge. Taking the previous example, if we instead use the override layer
+```yaml
+svc1:
+    ^log-targets: [tgt2]
+    override: merge
+```
+the resulting merged layer will be
+```yaml
+svc1:
+  log-targets: [tgt2]
+```
+
+To restore the default unspecified behaviour for `log-targets` (i.e. forwarding logs to all opt-out targets), we can use the special `default` keyword for `^log-targets`. For example, using the following override layer
+```yaml
+svc1:
+    ^log-targets: default
+    override: merge
+```
+will result in
+```yaml
+svc1:
+    log-targets: ~
+```
+equivalent to if `svc1.log-targets` had never been specified at all.
 -->
 
 ## Container usage
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -919,7 +919,7 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 		}
 
 		if logTargets != nil {
-			for _, targetName := range service.LogTargets.Targets {
+			for _, targetName := range logTargets.Targets {
 				_, ok := combined.LogTargets[targetName]
 				if !ok {
 					return nil, &FormatError{
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -355,6 +355,12 @@ func (t *LogTargets) Copy() *LogTargets {
 
 // Merge merges the fields from other into t.
 func (t *LogTargets) Merge(other *LogTargets) {
+	// anything <- keyword yields keyword
+	if other.Keyword != "" {
+		t.Keyword = other.Keyword
+		t.Targets = nil
+	}
+
 	// keyword <- [list] yields [list]
 	if t.Keyword != "" && len(other.Targets) > 0 {
 		t.Keyword = ""
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -90,7 +90,8 @@ type Service struct {
 	KillDelay      OptionalDuration         `yaml:"kill-delay,omitempty"`
 
 	// Log forwarding
-	LogTargets LogTargets `yaml:"log-targets,omitempty"`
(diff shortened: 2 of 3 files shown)


benhoyt:
Nice work. I guess I mildly prefer the `!replace` version now that I've seen both.

barrettj12:
John suggested [JSON Patch](https://jsonpatch.com/) as one approach to look at - solving this general problem of "how do add/remove/replace/etc fields in a document".

internal/plan/plan.go
```diff
@@ -90,7 +90,8 @@ type Service struct {
 	KillDelay      OptionalDuration         `yaml:"kill-delay,omitempty"`
 
 	// Log forwarding
-	LogTargets []string `yaml:"log-targets,omitempty"`
+	LogTargets        *LogTargets `yaml:"log-targets,omitempty"`
+	LogTargetsReplace *LogTargets `yaml:"^log-targets,omitempty"`
```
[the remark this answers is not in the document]
flotter: It feels to me from the points above (if I am correct), that it would be better to shift the ```replace``` operand to the right-hand side of one a single ```LogTargets``` attribute, and make it only affect the log target list, not the keywords.

I also cannot help to observe that this could be a very common theme for attributes ... on service merge, we want to be able to allow attributes to opt in to replace.

internal/plan/plan.go
```diff
@@ -258,6 +259,88 @@ const (
...
+// Merge merges the fields from other into t.
+func (t *LogTargets) Merge(other *LogTargets) {
+	// anything <- keyword yields keyword
+	if other.Keyword != "" {
+		t.Keyword = other.Keyword
+		t.Targets = nil
+	}
+
+	// keyword <- [list] yields [list]
+	if t.Keyword != "" && len(other.Targets) > 0 {
```
flotter, line 306: I think ```UnmarshalYAML``` can only ever result in either a keyword, or a log target list, so the logic seems to work. However, it feels to me that these cases should be mutually exclusive, so would it not be clearer to explicitly enforce that using ```if {} else if ... ``` ? Maybe I am missing a case.

internal/plan/plan.go
```diff
@@ -258,6 +259,88 @@ const (
...
+	case AllLogTargets:
+		return tgt.Selection != DisabledSelection
+	case DefaultLogTargets:
+		return tgt.Selection == OptOutSelection || tgt.Selection == UnsetSelection
+	case NoLogTargets:
+		return false
+	}
+
+	// No keyword
+	if tgt.Selection == DisabledSelection {
```
flotter, line 328: I've not spent too long looking at this, but if feels like some code duplication can be removed if we merge some of the checks?

```
if tgt.Selection == DisabledSelection ||  t.Keyword == NoLogTargets {
    return false
}
```
As the first check?

internal/plan/plan.go
```diff
@@ -752,11 +835,23 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 
 	// Validate service log targets
 	for serviceName, service := range combined.Services {
-		for _, targetName := range service.LogTargets {
-			_, ok := combined.LogTargets[targetName]
-			if !ok {
-				return nil, &FormatError{
-					Message: fmt.Sprintf(`unknown log target %q for service %q`, targetName, serviceName),
+		if service.LogTargets != nil {
```
flotter, line 838: Should we enforce only have one or the other? (```LogTargets``` vs. ```LogTargetsReplace```)

internal/plan/plan.go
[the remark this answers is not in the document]
barrettj12: @flotter 1 is correct. 2, 3, 4 are open questions as to how it should behave here.

I think you're suggesting an approach similar to #221. We discussed the pros and cons of this. It is nicer having a single key, but the issue is that it makes it hard to programmatically write a Pebble layer e.g. inside a charm. You'd have to use a YAML library to write the `!replace` tag - you can't just pass in the Python list as you can in this approach.

internal/plan/plan.go
barrettj12: you're right. I'll try to make this more clear

internal/plan/plan.go
flotter: I was thinking of merging all the case statements with the if/else cases below, similar to the one example I provided. But to be honest I have not thought it through completely, so maybe the outcome does not look much nicer.

internal/plan/plan.go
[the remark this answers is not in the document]
flotter: Sorry If I am repeating some decisions that was made - just ignore them :)

internal/plan/plan.go
barrettj12: @flotter I updated the method, let me know what you think.

internals/plan/plan.go
```diff
@@ -317,6 +318,90 @@ const (
...
+	}
+}
+
+// Merge merges the fields from other into t.
+func (t *LogTargets) Merge(other *LogTargets) {
+	if other.Keyword != "" {
+		// anything <- keyword yields keyword
+		t.Keyword = other.Keyword
+		t.Targets = nil
+
```
benhoyt, line 365: Nits: extraneous blank line (here and below). Also might be cleaner to do an early `return` here and below to avoid the additional `else` and indentation (early returns to avoid nesting are typical in Go).

benhoyt (approved):
Looks really good to me, thanks. Just left one ultra-nit comment.

internals/plan/plan.go
```diff
@@ -317,6 +318,90 @@ const (
...
+// Copy returns a deep copy of this LogTargets struct.
+func (t *LogTargets) Copy() *LogTargets {
+	return &LogTargets{
+		Targets: append([]string(nil), t.Targets...),
+		Keyword: t.Keyword,
+	}
+}
+
+// Merge merges the fields from other into t.
```
hpidcock, line 359: `// Merge merges` stutters. Try rephrasing.

internals/plan/plan.go
```diff
@@ -317,6 +318,90 @@ const (
...
+func (t *LogTargets) Merge(other *LogTargets) {
+	if other.Keyword != "" {
+		// anything <- keyword yields keyword
+		t.Keyword = other.Keyword
+		t.Targets = nil
+
+	} else {
+		if t.Keyword != "" {
+			// keyword <- [list] yields [list]
+			t.Keyword = ""
```
hpidcock, line 369: With this logic `t.Keyword == "all"` is no longer `all`.

`Targets` and `Keyword` are not mutually exclusive.

The merge logic here should be:
```
if other.Keyword is less inclusive than t.Keyword {
    other.Keyword = t.Keyword
}
other.Targets = other.Targets + t.Targets
```

`""` < `"None"` < `"Default"` < `"All"`

internals/plan/plan.go
```diff
@@ -317,6 +318,90 @@ const (
...
+}
+
+const (
+	// Valid keywords for LogTargets
+	DefaultLogTargets = "default"
+	AllLogTargets     = "all"
+	NoLogTargets      = "none"
+)
+
+func (t *LogTargets) UnmarshalYAML(value *yaml.Node) error {
```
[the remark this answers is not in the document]
benhoyt: Ah, good point. We'll have to ensure this round trips correctly when they run `pebble plan`.

internals/plan/plan.go
```diff
@@ -317,6 +318,90 @@ const (
...
+// LogTargets specifies which log targets should receive a service's logs.
+// It is either an explicit list of targets, or one of the below keywords.
+type LogTargets struct {
+	Targets []string
+	Keyword string
+}
+
+const (
+	// Valid keywords for LogTargets
+	DefaultLogTargets = "default"
```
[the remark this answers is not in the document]
benhoyt: @hpidcock Just to clarify, do you mean something like `type LogTargetKeyword string`? Yeah, that's a good call. We've done this for `ServiceStartup`, `Override`, and `ServiceAction` too.

internals/plan/plan.go
benhoyt: True, though this follows the pattern of the other `Merge` methods, so I think we should keep it as is for consistency.

internals/plan/plan.go
barrettj12: Not really sure what you mean here. `Targets` and `Keyword` *are* mutually exclusive, in that only one should be defined at a time.

Maybe you're debating the current behaviour that merging `all <- [list]` yields the `[list]`? Admittedly this is a bit quirky and I'm willing to change this if we think it's confusing / counterintuitive.

flotter (approved):
Looks great!

internals/plan/plan.go
```diff
@@ -317,6 +347,40 @@ const (
 	ActionIgnore   ServiceAction = "ignore"
 )
 
+// LogTargetsReplace represents a possible value for a service's `^log-targets`
+// field - either an explicit list of targets, or the special keyword "default".
+type LogTargetsReplace struct {
+	Targets []string
+	Default bool
```
barrettj12, line 354: IMO these fields shouldn't be exported - makes testing a little harder though. Maybe the `LogTargetsReplace` struct shouldn't even be exported?

README.md
```diff
@@ -403,6 +403,103 @@ $ pebble run --verbose
...
+svc1:
+    ^log-targets: default
+    override: merge
+```
+will result in
+```yaml
+svc1:
+    log-targets: ~
+```
+equivalent to if `svc1.log-targets` had never been specified at all.
```
benhoyt, line 500: Nice docs -- thanks!

internals/plan/plan.go
```diff
@@ -184,7 +188,33 @@ func (s *Service) Merge(other *Service) {
...
+	} else {
+		s.LogTargets = appendUnique(s.LogTargets, other.LogTargets...)
+	}
+}
+
+// copyStrings returns a copy of the provided []string, in such a way that:
+//   - copyStrings(nil) returns nil;
+//   - copyStrings([]string{}) returns []string{}.
+//
+// TODO: move this function into canonical/x-go/strutil
```
benhoyt, line 210: I think TODOs that are just "maybe we'll do this sometime" just add noise. When we want to do it, let's just do it, but otherwise it's fine as is.

internals/plan/plan.go
benhoyt: I think it's probably fine as is. It keeps the plan structs simple and consistent. And besides, this is in "internals" anyway, not something that should be imported elsewhere.

internals/plan/plan.go
```diff
@@ -317,6 +347,40 @@ const (
...
 )
 
+// LogTargetsReplace represents a possible value for a service's `^log-targets`
+// field - either an explicit list of targets, or the special keyword "default".
+type LogTargetsReplace struct {
+	Targets []string
+	Default bool
+}
+
+const LogTargetsReplaceDefaultKeyword = "default"
```
benhoyt, line 357: Naming nit: this is rather long. What about dropping the `Keyword` suffix?

internals/plan/plan.go
```diff
@@ -811,11 +875,24 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+				Message: fmt.Sprintf(`service %q cannot define both "log-targets" and "^log-targets" keys`, serviceName),
+			}
+		}
+
+		logTargets := service.LogTargets
+		if logTargets == nil && service.LogTargetsReplace != nil {
+			logTargets = service.LogTargetsReplace.Targets
+		}
+
+		if logTargets != nil {
```
benhoyt, line 889: Nit: probably don't need this `if` statement, as the range loop will do no iterations if `logTargets` is nil.

benhoyt (approved):
Thanks! This looks good to me. Just a couple of very minor nit comments.

internals/plan/plan.go
barrettj12: nice catch!

hpidcock:
I have one unanswered question:

Considering these log targets:
```
log-targets:
  target1:
    selection: opt-out
  target2:
    selection: opt-in
```
What does merging `^log-targets: default` and `log-targets: [target2]` yield?

PR #229, opened 2023-05-31
Public CLI API

anpep:
This PR evolves #226 and introduces:
 * a public `cli.AddCommand()` function so well-known Pebble-based applications can implement new commands on top of the basic ones in Pebble,
 * a public `cli.AddHelpCategory()` function that can append a category to the help manual, and
 * a public variable, `Personality`, which holds
   * one value that represents the name of the binary (e.g. *pebble*), and
   * one value that represents the display name of the application (e.g. *Pebble*).

This PR also introduces many changes to the help manual and hard-coded references to *pebble* and *Pebble* in order to replace them with the correct references to the `Personality` struct.

### DIFF
--- a/internals/cli/cmd_enter.go
+++ b/internals/cli/cmd_enter.go
@@ -58,7 +58,7 @@ func init() {
 	for k, v := range sharedRunEnterOptsHelp {
 		optsHelp[k] = v
 	}
-	cmdInfo := addCommand("enter", shortEnterHelp, longEnterHelp, func() flags.Commander { return &cmdEnter{} }, optsHelp, nil)
+	cmdInfo := AddCommand("enter", shortEnterHelp, longEnterHelp, func() flags.Commander { return &cmdEnter{} }, optsHelp, nil)
 	cmdInfo.extra = func(cmd *flags.Command) {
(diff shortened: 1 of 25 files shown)


internals/cli/cmd_enter.go
```diff
@@ -29,7 +30,8 @@ These subcommands are currently supported:
 (1) Services are not started.
 (2) No logs on stdout unless -v is used.
 (3) Services continue running after the subcommand succeeds.
-`
+`,
+	cmd.Personality.ProgramName)
```
paul-rodriguez, line 34: If my understanding is correct, a dynamic change of the `Personality` struct will not change the content of `longEnterHelp` if `longEnterHelp` has already been initialized. And that initialization is going to happen very early, when the file level content of `cmd_enter.go` is executed. I assume that happens when the package is imported, though I don't know Go well enough to be certain.

If all of that is true, then you effectively can't change the `Personality` after importing this one package (and any other with the same kind of definition). I think this data flow dependency would be really obscure.

anpep:
Closing this pull request. Internal discussions revolving around the point made by @paul-rodriguez concluded with the proposal of a new PR that is narrower in scope.

PR #232, opened 2023-06-05
cli: public commands API

anpep:
Introduce a public API for well-known Pebble-based applications to use.
This PR introduces two basic changes on top of those included in #226:
 * **Export the `cli.AddCommand()` function**, so that applications can add new commands.
 * **Export the `cli.HelpCategories` variable**, so that applications can add new sections to the help manual.

```go
// Example: adding a help category
package main

import "github.com/canonical/pebble/internals/cli"

var myCategory = cli.HelpCategory {
    Label: "My category",
    Description: "Description for my help category",
    Commands: []string{"new-command"}
}

func main() {
    cli.HelpCategories = append(cli.HelpCategories, myCategory)
    cli.Run()
}
```

```go
// Example: adding a new command
package main

import (
    "fmt"

    "github.com/canonical/go-flags"
    "github.com/canonical/pebble/internals/cli"
)

type cmdNewCommand struct {
    Force bool `short:"f" long:"force"`
}

func init() {
    cli.AddCommand(&cli.CmdInfo{
        Name:        "new-command",
        Summary:     "Lorem ipsum. Foo bar.",
        Description: "Dolor sit amet.",
        Builder:     func() flags.Commander { return &cmdNewCommand{} },
        ArgsHelp: map[string]string{
            "--force": "Forces the operation",
        },
    })
}

func (cmd *cmdNewCommand) Execute(args []string) error {
    if len(args) > 0 {
        return cli.ErrExtraArgs
    }
    fmt.Fprintf(cli.Stdout, "Hello, world!\n")
    return nil
}

```


internals/cli/cli.go
```diff
@@ -80,9 +80,9 @@ var commands []*cmdInfo
 // debugCommands holds information about all debug commands.
 var debugCommands []*cmdInfo
 
-// addCommand replaces parser.addCommand() in a way that is compatible with
+// AddCommand replaces parser.AddCommand() in a way that is compatible with
```
atesburak, line 83: Extremely minor point but parser.addCommand vs parser.AddCommand for the original one?

paul-rodriguez (approved):

internals/cli/cli.go
```diff
@@ -80,9 +80,9 @@ var commands []*cmdInfo
...
 // re-constructing a pristine parser.
-func addCommand(name, shortHelp, longHelp string, builder func() flags.Commander, optDescs map[string]string, argDescs []argDesc) *cmdInfo {
+func AddCommand(name, shortHelp, longHelp string, builder func() flags.Commander, optDescs map[string]string, argDescs []argDesc) *cmdInfo {
```
benhoyt, line 85: Now that this is exported, maybe we can reconsider the signature. It has 6 args, which is a bit unwieldy. What about exporting a `CommandInfo` struct so these can become "named args" (or even just exporting the relevant fields from `CmdInfo`).

```
func AddCommand(info *CommandInfo) *cmdInfo { ... }

...

cli.AddCommand(&cli.CommandInfo{
    Name: "foo",
    ShortHelp: "short help",
    ...
})
```

Also, does it need to return anything? Only cmd_enter and cmd_exec use the result value to add "extra", but maybe they can still call the internal one? Or we expose `Extra` in the passed-in struct too?

internals/cli/cmd_help.go
```diff
@@ -191,6 +191,11 @@ var helpCategories = []helpCategory{{
 	Commands:    []string{"warnings", "okay"},
 }}
 
+// AddHelpCategory appends an existing help category to the Pebble help manual.
+func AddHelpCategory(categ HelpCategory) {
```
benhoyt, line 195: Naming nit: `categ` seems like an unusual and unnecessary abbreviation. [Wiktionary lists cat](https://en.wiktionary.org/wiki/cat) as an abbrevation for category, but there's no "categ". Should probably just be `category`.

flotter (approved):
This is the last change required on the Pebble side before we can create a different variant of Pebble in a separate repository.

Some work remains to allow for personality changes to update string references to "pebble" in the help, but that will be a followup PR.

internals/cli/cli.go
```diff
@@ -80,9 +80,9 @@ var commands []*cmdInfo
 // debugCommands holds information about all debug commands.
 var debugCommands []*cmdInfo
 
-// addCommand replaces parser.addCommand() in a way that is compatible with
+// AddCommand replaces parser.addCommand() in a way that is compatible with
 // re-constructing a pristine parser.
-func addCommand(name, shortHelp, longHelp string, builder func() flags.Commander, optDescs map[string]string, argDescs []argDesc) *cmdInfo {
```
niemeyer, line 85: This function was not made for being exported. It's already not a great API, but we forgive as it's an internal detail that we can easily fix.  The adoption on Termus is still somewhat of a private use, but still, if we change the shape of this function we'll break every copy of it that is out there. We should probably turn it into a struct with nicely named fields (unlike the parameters) and docs.

niemeyer (changes requested):
Looks mostly okay since it's a private use, but I still hope we can improve it slightly as outlined below. Bonus points if there's a test for it.

internals/cli/cmd_help.go
```diff
@@ -164,8 +169,8 @@ type HelpCategory struct {
 	Commands    []string
 }
 
-// helpCategories helps us by grouping commands
-var helpCategories = []HelpCategory{{
+// HelpCategories helps us by grouping commands
+var HelpCategories = []HelpCategory{{
```
paul-rodriguez, line 173: Is there a reason to expose the whole thing as opposed to only allowing add ?

anpep: It is a common idiom in Go, and allows us to tinker with the categories in different ways (e.g. removing some commands from the help manual on certain applications); besides, that was the behavior that was agreed upon last meeting (see 2023/06/13 meeting notes).

internals/cli/cli.go
```diff
@@ -63,52 +63,34 @@ var optionsData options
...
-func AddCommand(name, shortHelp, longHelp string, builder func() flags.Commander, optDescs map[string]string, argDescs []argDesc) *cmdInfo {
-	info := &cmdInfo{
-		name:      name,
-		shortHelp: shortHelp,
-		longHelp:  longHelp,
-		builder:   builder,
-		optDescs:  optDescs,
-		argDescs:  argDescs,
-	}
+func AddCommand(info *CmdInfo) {
```
paul-rodriguez, line 85: This is a much better API, but what if the caller changes `info` later on?

anpep: Agreed, we should instead pass in `info` by value.

internals/cli/cli.go
```diff
@@ -63,52 +63,34 @@ var optionsData options
...
-type cmdInfo struct {
-	name, shortHelp, longHelp string
-	builder                   func() flags.Commander
-	hidden                    bool
-	optDescs                  map[string]string
-	argDescs                  []argDesc
-	alias                     string
-	extra                     func(*flags.Command)
+// CmdInfo holds information needed to call parser.AddCommand(...).
+type CmdInfo struct {
```
atesburak, line 67: The field names here are a bit ambiguous, which would normally be fine but since we are making it public may be a short description to help people out.

internals/cli/cli.go
```diff
@@ -63,52 +63,34 @@ var optionsData options
...
+	Name, ShortHelp, LongHelp string
+	Builder                   func() flags.Commander
+	Hidden                    bool
+	OptDescs                  map[string]string
+	ArgDescs                  []argDesc
```
niemeyer, line 72: Per preview review, the goal here is not to just make things public, but to actually polish them enough so that making it public is reasonable. OptDesc and ArgDesc are not good names, and are close enough together that it's hard to see why they both exist. Even more when they are undocumented. Also, we should pretty much never have public types exposing public types as is the case of ArgDescs above, since clients won't be able to define those values.

internals/cli/cli.go
```diff
@@ -63,52 +63,34 @@ var optionsData options
...
-		name:      name,
-		shortHelp: shortHelp,
-		longHelp:  longHelp,
-		builder:   builder,
-		optDescs:  optDescs,
-		argDescs:  argDescs,
-	}
-	commands = append(commands, info)
-	return info
+func AddCommand(info CmdInfo) {
```
niemeyer, line 85: The argument here shoud be a pointer.

niemeyer (changes requested):
Thanks for moving it forward. There are still remaining issues related to the points made in the prior review it seems.

internals/cli/cli.go
```diff
@@ -53,62 +53,71 @@ type options struct {
...
+	// LongHelp contains exhaustive documentation about the command,
+	// that will be reflected in the specific help manual for the
+	// command, and in the Pebble man page.
+	LongHelp string
+
+	// Builder is a function that creates a new instance of the command
+	// struct containing an Execute(args []string) implementation.
+	Builder func() flags.Commander
+
+	// OptionsHelp (optional) maps long option names (e.g. "foo" for --foo),
```
flotter, line 90: Help supplied for all options, or not help supplied at all.

internals/cli/last.go
```diff
@@ -33,9 +33,9 @@ var changeIDMixinOptDesc = map[string]string{
 	"last": "Select last change of given type (install, refresh, remove, try, auto-refresh, etc.). A question mark at the end of the type means to do nothing (instead of returning an error) if no change of the given type is found. Note the question mark could need protecting from the shell.",
 }
 
-var changeIDMixinArgDesc = []argDesc{{
-	name: "<change-id>",
-	desc: "Change ID",
+var changeIDMixinArgDesc = []PositionalArg{{
```
flotter, line 36: If you do plan to change the struct name on the right, would it make sense to change the variable name on the left, in the same way you've aligned the names for ```OptionsHelp``` variables?

flotter (approved):
Some subjective preferences, but otherwise looks great!

internals/cli/cli.go
```diff
@@ -53,62 +53,71 @@ type options struct {
...
+	Name string
+
+	// ShortHelp is a single-line help string that will be displayed
+	// in the full Pebble help manual (i.e. help --all)
+	ShortHelp string
+
+	// LongHelp contains exhaustive documentation about the command,
+	// that will be reflected in the specific help manual for the
+	// command, and in the Pebble man page.
+	LongHelp string
```
benhoyt, line 84: Potentially `Summary` and `Description`? These are nice single-word descriptions of what each field represents, and in line with terminology we use elsewhere, for example in the Pebble layer config.

internals/cli/cli.go
```diff
@@ -53,62 +53,71 @@ type options struct {
...
-	desc string
+// PositionalArg contains information about the positional arguments accepted
+// by a command.
+type PositionalArg struct {
+	// Name of the positional argument as displayed in the command usage string.
+	Name string
+	// ShortHelp is a single-line help string that will be displayed in the
+	// command's manual alongside the argument name.
+	ShortHelp string
 }
```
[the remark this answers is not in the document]
benhoyt: I like the term "placeholder", given what it is for, as these arguments don't really have a "name". Though this problem might go away if we use a map, per my comment below.

internals/cli/cmd_add.go
```diff
@@ -67,5 +67,11 @@ func (cmd *cmdAdd) Execute(args []string) error {
 }
 
 func init() {
-	addCommand("add", shortAddHelp, longAddHelp, func() flags.Commander { return &cmdAdd{} }, addDescs, nil)
+	AddCommand(&CmdInfo{
```
benhoyt, line 70: These are a lot more readable on multiple lines and with the field names now, thanks!

internals/cli/cli.go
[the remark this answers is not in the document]
anpep: Great suggestion, thank you!

internals/cli/last.go
anpep: Definitely, but as I knew this would be a set of controversial changes, I wanted to come to an agreement on naming the structs first before I went ahead and created a bunch of noise by renaming everything under the sun (:

internals/cli/cli.go
anpep: I don't fully get this comment. Do you mean I should state that either no help should be supplied (`nil`) or all the options should be documented?

internals/cli/cli.go
anpep: Lovely! Thank you

internals/cli/cli.go
flotter: Sorry this is vague indeed, and nothing related to this PR. The optional hint feels like it is not making it clear that it does not apply to an individual option, but rather the map as a whole is optional.

flotter (approved):

internals/cli/cli.go
```diff
@@ -53,62 +53,71 @@ type options struct {
...
+	Builder func() flags.Commander
+
+	// OptionsHelp (optional) maps long option names (e.g. "foo" for --foo),
+	// to the help strings that will be shown in the command's manual
+	// besides every option.
+	OptionsHelp map[string]string
+
+	// PositionalArgs (optional) contains help information about the
+	// positional arguments accepted by the command.
+	PositionalArgs []PositionalArg
```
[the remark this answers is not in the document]
benhoyt: > I think it's still better than having it match the positional arguments order, which is not explicit at all.

Yeah, I agree. Having to keep two lists (defined in different places in the code) in the same order seems error-prone. Then again, I guess when you change one you probably want to change the other. Still, I prefer the named approach.

internals/cli/cli.go
```diff
@@ -165,37 +146,50 @@ func fixupArg(optName string) string {
...
+}
+
 // Parser creates and populates a fresh parser.
 // Since commands have local state a fresh parser is required to isolate tests
 // from each other.
 func Parser(cli *client.Client) *flags.Parser {
-	optionsData.Version = func() {
-		printVersions(cli)
-		panic(&exitStatus{0})
+	// Implement --version by default on every command
```
paul-rodriguez, line 173: Shouldn't `--version` only work when used on the executable without specifying a command?

internals/cli/cli.go
```diff
@@ -214,92 +208,57 @@ func Parser(cli *client.Client) *flags.Parser {
...
-			if desc != "" {
-				opt.Description = desc
+		optionsHelp := map[string]string{}
+		positionalArgsHelp := map[string]string{}
+
+		for specifier, help := range c.ArgsHelp {
+			if strings.HasPrefix(specifier, "--") {
+				optionsHelp[specifier] = help
+			} else if utf8.RuneCountInString(specifier) == 2 && strings.HasPrefix(specifier, "-") {
+				optionsHelp[specifier] = help
```
paul-rodriguez, line 218: How do you handle options that have a `--longName` and shorthand `-l` with the same effect?

internals/cli/cli.go
```diff
@@ -214,92 +208,57 @@ func Parser(cli *client.Client) *flags.Parser {
...
-			if !(c.OptionsHelp == nil || ok) {
-				logger.Panicf("%s missing description for %s", c.Name, name)
-			}
-			lintDesc(c.Name, name, desc, opt.Description)
-			if desc != "" {
-				opt.Description = desc
+			if description, ok := optionsHelp["--"+opt.LongName]; ok {
+				lintDesc(c.Name, opt.LongName, description, opt.Description)
+				opt.Description = description
+			} else if description, ok := optionsHelp["-"+string(opt.ShortName)]; ok {
```
paul-rodriguez, line 243: Why the string conversion here?

internals/cli/cmd_logs.go
```diff
@@ -41,17 +36,26 @@ type cmdLogs struct {
...
+	AddCommand(&CmdInfo{
+		Name:    "logs",
+		Summary: "Fetch service logs",
+		Description: `
 The logs command fetches buffered logs from the given services (or all services
 if none are specified) and displays them in chronological order.
-`
+`,
+		ArgsHelp: map[string]string{
+			"--follow": "Follow (tail) logs for given services until Ctrl-C is\npressed. If no services are specified, show logs from\nall services running when the command starts.",
```
paul-rodriguez, line 48: This should be broken down for readability

internals/cli/cmd_restart.go
```diff
@@ -34,11 +28,13 @@ type cmdRestart struct {
...
-		Summary:     shortRestartHelp,
-		Description: longRestartHelp,
-		Builder:     func() flags.Commander { return &cmdRestart{} },
-		OptionsHelp: waitOptionsHelp,
+		Name:    "restart",
+		Summary: "Restart a service",
+		Description: `
+The restart command restarts the named service(s) in the correct order.
+`,
+		ArgsHelp: waitArgsHelp,
```
paul-rodriguez, line 36: Considering there is re-use of these argsHelp maps, I think it would make more sense for them to be sets of a specific type describing one option, rather than maps. Doing it like this would also allow you to put more complex information into it: options with (potentially multiple and/or optional) arguments, shorthand variations, repeatability etc. and it would also clarify the way that the options are shared in-between commands.

internals/cli/last.go
```diff
@@ -25,35 +25,29 @@ type changeIDMixin struct {
...
-}
-
-var changeIDMixinArgumentsHelp = map[string]ArgumentHelp{
-	"<id>": ArgumentHelp{
-		Placeholder: "<change-id>",
-		Help:        "Change ID",
-	},
+var changeIDMixinArgsHelp = map[string]string{
+	"<change-id>": "Change ID",
+	"--last":      "Select last change of given type (install, refresh, remove, try, auto-refresh, etc.). A question mark at the end of the type means to do nothing (instead of returning an error) if no change of the given type is found. Note the question mark could need protecting from the shell.",
```
paul-rodriguez, line 34: This too should be broken up

paul-rodriguez (commented):
I feel that the option help stuff could use more structure

internals/cli/cli.go
anpep: `opt.ShortName` is a `rune`, not a `string`.

internals/cli/cli.go
```diff
@@ -86,14 +80,15 @@ type CmdInfo struct {
...
-	ArgumentsHelp map[string]ArgumentHelp
+	// ArgsHelp (optional) contains help about the command-line arguments
+	// (including options) supported by the command.
+	//
+	//  map[string]string{
+	//	    "--long-option": "my very long option",
+	//      "-v": "verbose output",
+	//      "<change-id>": "named positional argument"
+	//  }
+	ArgsHelp map[string]string
```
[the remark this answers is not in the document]
anpep: It intentionally is. The previous approach discriminated between opts and args, but is more ergonomic and readable to have everything in a single place. In a more type-safe environment, I'd rather have multiple variants for `ArgsHelp`:
```
ArgsHelp: {
    Option("v", "verbose"): "verbose output",
    Positional("change-id"): "change ID"
}
```

But when in Rome, write idiomatic Golang!

internals/cli/cli.go
```diff
@@ -49,70 +49,59 @@ var (
 // the pebble client.
 const defaultPebbleDir = "/var/lib/pebble/default"
 
-type options struct {
-	Version func() `long:"version"`
+// ArgumentHelp contains help information about the positional arguments accepted
+// by a command.
+type ArgumentHelp struct {
+	// Placeholer supplies a string representation of the argument.
```
benhoyt, line 55: Typo: Placeholer -> Placeholder

internals/cli/cli.go
```diff
@@ -49,70 +49,59 @@ var (
...
-		argDescs:  argDescs,
-	}
-	debugCommands = append(debugCommands, info)
-	return info
+	// ArgsHelp (optional) contains help about the command-line arguments
+	// (including options) supported by the command.
+	//
+	//  map[string]string{
+	//	    "--long-option": "my very long option",
+	//      "-v": "verbose output",
```
benhoyt, line 88: Nit: Looks like there's an indentation tabs v spaces thing going on above.

internals/cli/times.go
```diff
@@ -44,7 +44,7 @@ type durationMixin struct {
 	AbsTime bool `long:"abs-time"`
 }
 
-var durationDescs = map[string]string{
+var durationOptionsHelp = map[string]string{
```
benhoyt, line 47: Is this still used?

benhoyt (approved):
I like the updated `CmdInfo` struct, thanks.

internals/cli/times.go
anpep: Good catch!

internals/cli/cli.go
```diff
@@ -49,70 +49,59 @@ var (
 // the pebble client.
 const defaultPebbleDir = "/var/lib/pebble/default"
 
-type options struct {
-	Version func() `long:"version"`
+// ArgumentHelp contains help information about the positional arguments accepted
+// by a command.
+type ArgumentHelp struct {
```
niemeyer, line 54: This looks unused. It'd be good to perform a full self-review to make sure there are no left overs from previous iterations.

niemeyer (changes requested):

benhoyt:
@anpep The tests are failing due to a go.mod issue: x/term is imported but not an explicitly dependency. Is that change intentional? In this PR we should probably switch back to using the crypto one (but perhaps explicitly add x/term to go.mod in a future PR?)

flotter (approved):
Much much simpler to read and easier to follow. Awesome.

internals/cli/cli.go
```diff
@@ -157,154 +137,126 @@ func fixupArg(optName string) string {
...
-			}
-			desc, ok := c.optDescs[name]
-			if !(c.optDescs == nil || ok) {
-				logger.Panicf("%s missing description for %s", c.name, name)
-			}
-			lintDesc(c.name, name, desc, opt.Description)
-			if desc != "" {
-				opt.Description = desc
-			}
+		if hasAnyOptionHelp && len(opts) != len(optionsHelp) {
```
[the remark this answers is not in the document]
anpep: You are absolutely right; however, even during the development of this PR I accidentally removed some of the descriptions for testing, and only when I added this check I realized some commands were missing descriptions for args/opts.

Truth is, it is very easy to forget adding these descriptions and this check is a great way to detect it during testing. Let's prefer being explicit here and force the developer write an empty description. After all, I've seen no option/argument without a description yet.

internals/cli/cli.go
```diff
@@ -157,154 +137,126 @@ func fixupArg(optName string) string {
...
-		}
-		cmd, err := debugCommand.AddCommand(c.name, c.shortHelp, strings.TrimSpace(c.longHelp), obj)
-		if err != nil {
-			logger.Panicf("cannot add debug command %q: %v", c.name, err)
-		}
-		cmd.Hidden = c.hidden
-		opts := cmd.Options()
-		if c.optDescs != nil && len(opts) != len(c.optDescs) {
-			logger.Panicf("wrong number of option descriptions for %s: expected %d, got %d", c.name, len(opts), len(c.optDescs))
+		if hasAnyPositionalHelp && len(args) != len(positionalArgsHelp) {
```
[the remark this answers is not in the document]
anpep: See the above comment.

internals/cli/cli.go
```diff
@@ -157,154 +137,126 @@ func fixupArg(optName string) string {
...
+		for specifier, help := range c.ArgsHelp {
+			isLongOption := strings.HasPrefix(specifier, "--")
+			isShortOption := utf8.RuneCountInString(specifier) == 2 && strings.HasPrefix(specifier, "-")
+			if isShortOption || isLongOption {
+				optionsHelp[specifier] = help
+			} else if strings.HasPrefix(specifier, "<") && strings.HasSuffix(specifier, ">") {
+				// This is a positional argument
+				positionalArgsHelp[specifier] = help
+			} else {
+				logger.Panicf("invalid help specifier: %#v %#v", c.Name, strings.HasPrefix(specifier, "-"))
```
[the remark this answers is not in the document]
anpep: Good catch--left this here accidentally for debugging.

internals/cli/cli.go
```diff
@@ -157,154 +137,126 @@ func fixupArg(optName string) string {
...
-			lintDesc(c.name, name, desc, opt.Description)
-			if desc != "" {
-				opt.Description = desc
+			if description, ok := optionsHelp["--"+opt.LongName]; ok {
+				lintDesc(c.Name, opt.LongName, description, opt.Description)
+				opt.Description = description
+			} else if description, ok := optionsHelp["-"+string(opt.ShortName)]; ok {
+				lintDesc(c.Name, string(opt.ShortName), description, opt.Description)
+				opt.Description = description
+			} else if hasAnyOptionHelp {
```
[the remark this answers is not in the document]
anpep: Is it? Please help me verify my own logic here:
- If we found `--<long name>`,
  - lint long option and set option description;
- otherwise, if we found `-<short name>`,
  - lint short option and set option description;
- otherwise, this option doesn't have a description, so:
  - if we supplied any description for any option at all (`hasAnyOptionHelp`),
    - panic
  - otherwise, we have not supplied a description for any option (`!hasAnyOptionHelp`),
    - don't panic and leave every option undocumented

This follows the semantics of `ArgsHelp`, which says that either all options are documented or none is. I guess the question would be: is this a good semantic?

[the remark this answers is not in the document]
flotter: You are right! Sorry for having made you write out the logic ...

flotter (approved):

internals/cli/cli.go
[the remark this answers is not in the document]
flotter: Here is another angle: We only validate the length, not that the keys match. So on key mismatch, we could end up in "otherwise", not because its not supplied, but because there is a mismatch.

So, should the early validation not be combined with the second part that is more key aware?

flotter (approved):

internals/cli/cli.go
```diff
@@ -49,70 +49,50 @@ var (
...
+	// (including options) supported by the command.
+	//
+	//  map[string]string{
+	//      "--long-option": "my very long option",
+	//      "-v": "verbose output",
+	//      "<change-id>": "named positional argument"
+	//  }
+	ArgsHelp map[string]string
+
+	// Whether to pass all arguments after the first non-option as remaining
```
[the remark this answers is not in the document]
flotter: I see now this related to separating options from a subcommand from the main command.

internals/cli/cmd_enter.go
```diff
@@ -29,7 +42,14 @@ These subcommands are currently supported:
 (1) Services are not started.
 (2) No logs on stdout unless -v is used.
 (3) Services continue running after the subcommand succeeds.
-`
```
niemeyer, line 32: Can we please preserve the existing convention of declaring the short and long help texts at the top of the file using variables?   Not only this makes it easy to have a glimpse of what the command does upfront, but it avoids what is happening here, which is having a type construction being broken down in the middle in an ugly way, making it harder to read both the help and the code. This was the original reason for this convention, and it still rings true seeing the code now.

internals/cli/cli.go
```diff
@@ -49,70 +50,50 @@ var (
...
-	}
+func AddCommand(info *CmdInfo) {
 	commands = append(commands, info)
-	return info
-}
-
-// addDebugCommand replaces parser.addCommand() in a way that is
-// compatible with re-constructing a pristine parser. It is meant for
-// adding debug commands.
-func addDebugCommand(name, shortHelp, longHelp string, builder func() flags.Commander, optDescs map[string]string, argDescs []argDesc) *cmdInfo {
```
niemeyer, line 101: We discussed this a few times by now, and it looks like the code related to debug commands is still being dropped?

niemeyer (changes requested):
Here are a few extra details, but it looks like the points discussed last time about debug commands still haven't been touched?

atesburak (approved):

internals/cli/cli.go
```diff
@@ -174,15 +175,13 @@ func Parser(cli *client.Client) *flags.Parser {
...
-		version.Description = "Print the version and exit"
-		version.Hidden = true
-	}
-
 	// Add --help like what go-flags would do for us, but hidden
 	addHelp(parser)
 
+	// Regular expressions for positional and flag arguments
+	positionalRegexp := regexp.MustCompile(`^<[\w-]+>$`)
+	flagRegexp := regexp.MustCompile(`^-(\w|-[\w-]+)$`)
```
flotter, line 195: This may not be an issue, but just pointing out that for flag validation, the following will pass:

"---"
"---------"
"-------gtrfffg---hhhh"

flotter (approved):

paul-rodriguez (approved):

internals/cli/cli.go
```diff
@@ -49,70 +50,62 @@ var (
...
+	//  }
+	ArgsHelp map[string]string
+
+	// Whether to pass all arguments after the first non-option as remaining
+	// command line arguments. This is equivalent to strict POSIX processing.
+	PassAfterNonOption bool
+
+	// When set, the command will be a subcommand of the `debug` command.
+	// Do not set this flag manually, use AddDebugCommand instead.
+	debug bool
```
niemeyer, line 91: Why not having a single public function then, and having this field exposed and using it to decide in which slice to add it?  Probably reduces surface area and avoids the curious private field here.

internals/cli/cli.go
```diff
@@ -157,148 +138,106 @@ func fixupArg(optName string) string {
...
-	flagopts := flags.Options(flags.PassDoubleDash)
-	parser := flags.NewParser(&optionsData, flagopts)
+
+	flagOpts := flags.Options(flags.PassDoubleDash)
+	parser := flags.NewParser(&defaultOptions, flagOpts)
 	parser.ShortDescription = "Tool to interact with pebble"
 	parser.LongDescription = longPebbleDescription
-	// hide the unhelpful "[OPTIONS]" from help output
-	parser.Usage = ""
-	if version := parser.FindOptionByLongName("version"); version != nil {
```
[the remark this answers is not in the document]
niemeyer: This was marked as resolved without comments and the logic is still the same.

internals/cli/cli.go
```diff
@@ -169,134 +162,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
+
+	// Add --help like what go-flags would do for us, but hidden
 	addHelp(parser)
 
-	// Add all regular commands
-	for _, c := range commands {
-		obj := c.builder()
+	// Regular expressions for positional and flag arguments
+	positionalRegexp := regexp.MustCompile(`^<[\w-]+>$`)
+	flagRegexp := regexp.MustCompile(`^-(\w|-[\w]+(-?[\w]+)*)$`)
```
niemeyer, line 195: `[\w]` is a group with a single element, and thus equivalent to `\w`. This is also an expression that is harder to match because the nested optional repetitions cause unnecessary backtracking. That issue is completely irrelevant for the context, but worth considering in the future in trickier environments.

For example, this is the same expression, but simpler and easier to match: `^-\w|--\w+(-\w+)*$`

Note how the repetitions start with a required prefix.

internals/cli/cli.go
```diff
@@ -169,134 +162,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
 
-		cmd, err := parser.AddCommand(c.name, c.shortHelp, strings.TrimSpace(c.longHelp), obj)
-		if err != nil {
-			logger.Panicf("cannot add command %q: %v", c.name, err)
-		}
-		cmd.Hidden = c.hidden
-		if c.alias != "" {
-			cmd.Aliases = append(cmd.Aliases, c.alias)
+		// Target either the `debug` command or the top-level parser depending on this command
+		// being or not marked as debug.
```
niemeyer, line 215: This is just reading the code below in English.  It's easier and faster to read the code rather than the comment. Good comments talk about details that you can't get by simply reading the code at hand.

Note that I'm being a bit more nitpicky in these reviews so we can create some shared agreements for the upcoming work that you'll be doing with the team.

internals/cli/cli.go
```diff
@@ -169,134 +162,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
+		flagHelp := map[string]string{}
+		positionalHelp := map[string]string{}
+
+		for specifier, help := range c.ArgsHelp {
+			if flagRegexp.MatchString(specifier) {
+				flagHelp[specifier] = help
+			} else if positionalRegexp.MatchString(specifier) {
+				positionalHelp[specifier] = help
+			} else {
+				logger.Panicf("invalid help specifier from %q: %q", c.Name, specifier)
```
niemeyer, line 239: Can we please prefix this by "internal error: ".

internals/cli/cli.go
```diff
@@ -169,134 +162,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
-		if err != nil {
-			logger.Panicf("cannot add debug command %q: %v", c.name, err)
-		}
-		cmd.Hidden = c.hidden
+		// Make sure all argument descriptions are set
 		opts := cmd.Options()
-		if c.optDescs != nil && len(opts) != len(c.optDescs) {
-			logger.Panicf("wrong number of option descriptions for %s: expected %d, got %d", c.name, len(opts), len(c.optDescs))
+		if len(opts) != len(flagHelp) {
+			logger.Panicf("wrong number of flag descriptions for %q: expected %d, got %d", c.Name, len(opts), len(flagHelp))
```
niemeyer, line 246: Same here.

internals/cli/cli.go
```diff
@@ -169,134 +162,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
+
+	// Create debug command
+	debugCmd, err := parser.AddCommand("debug", cmdDebugSummary, cmdDebugDescription, &cmdDebug{})
+	debugCmd.Hidden = true
+	if err != nil {
+		logger.Panicf("cannot add command %q: %v", "debug", err)
+	}
+
+	// Add all commands
+	for _, c := range append(commands, debugCommands...) {
```
niemeyer, line 205: This kind of operation need a bit of care. Here it seems irrelevant, but it's worth noting that this kind of operation needs a bit of care as this is potentially mutating the array backing the `command` slice, which does not belong to this function. Again, probably not a big deal, but keep an eye on this apparently innocuous construct.

internals/cli/cli.go
```diff
@@ -169,134 +162,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
-		}
-	}
-	// Add the debug command
-	debugCommand, err := parser.AddCommand("debug", shortDebugHelp, longDebugHelp, &cmdDebug{})
-	debugCommand.Hidden = true
-	if err != nil {
-		logger.Panicf("cannot add command %q: %v", "debug", err)
-	}
-	// Add all the sub-commands of the debug command
-	for _, c := range debugCommands {
```
niemeyer, line 257: The unification of these loops is quite nice, thanks.

internals/cli/cli.go
```diff
@@ -362,7 +329,7 @@ func Run() error {
 						sug = "pebble help " + x.Name
 					}
 				}
-				return fmt.Errorf("unknown command %q, see '%s'.", sub, sug)
+				return fmt.Errorf("unknown command %q, see '%s'", sub, sug)
```
niemeyer, line 332: Thanks.

internals/cli/cmd_changes.go
```diff
@@ -24,15 +24,8 @@ import (
...
-var shortTasksHelp = "List a change's tasks"
-var longChangesHelp = `
-The changes command displays a summary of system changes performed recently.
-`
-var longTasksHelp = `
-The tasks command displays a summary of tasks associated with an individual
-change that happened recently.
-`
+const cmdChangesSummary = "List system changes"
+const cmdChangesDescription = "The changes command displays a summary of system changes performed recently."
```
niemeyer, line 28: Can we please consistently break the description into the next line with backticks as we (mostly?) had before.

internals/cli/cmd_exec.go
```diff
@@ -48,22 +48,11 @@ type cmdExec struct {
...
-	"I":       "Disable interactive mode and use a pipe for stdin",
-}
-
-var shortExecHelp = "Execute a remote command and wait for it to finish"
-var longExecHelp = `
+func init() {
+	AddCommand(&CmdInfo{
+		Name:    "exec",
+		Summary: "Execute a remote command and wait for it to finish",
+		Description: `
```
niemeyer, line 55: Let's please move this up as well.

internals/cli/cmd_help.go
```diff
@@ -25,10 +25,30 @@ import (
 	"github.com/canonical/go-flags"
 )
 
-var shortHelpHelp = "Show help about a command"
-var longHelpHelp = `
-The help command displays information about commands.
-`
+const cmdHelpSummary = "Show help about a command"
+const cmdHelpDescription = "The help command displays information about commands."
```
niemeyer, line 29: Same, let's please preserve the breaking consistently as we had before (see diff above).

internals/cli/cmd_mkdir.go
```diff
@@ -24,6 +24,9 @@ import (
 	"github.com/canonical/pebble/client"
 )
 
+const cmdMkdirSummary = "Create a directory"
+const cmdMkdirDescription = "The mkdir command creates the specified directory."
```
niemeyer, line 28: Same here.

internals/cli/cmd_restart.go
```diff
@@ -16,14 +16,11 @@ package cli
...
-
 	"github.com/canonical/pebble/client"
 )
 
-var shortRestartHelp = "Restart a service"
-var longRestartHelp = `
-The restart command restarts the named service(s) in the correct order.
-`
+const cmdRestartSummary = "Restart a service"
+const cmdRestartDescription = "The restart command restarts the named service(s) in the correct order."
```
niemeyer, line 23: Same here.

internals/cli/cmd_rm.go
```diff
@@ -16,29 +16,32 @@ package cli
 
 import (
 	"github.com/canonical/go-flags"
-
 	"github.com/canonical/pebble/client"
 )
 
+const cmdRmSummary = "Remove a file or directory"
+const cmdRmDescription = "The rm command removes a file or directory."
```
niemeyer, line 23: Same here.

internals/cli/cmd_run.go
```diff
@@ -31,18 +31,6 @@ import (
 	"github.com/canonical/pebble/internals/systemd"
 )
 
-var shortRunHelp = "Run the pebble environment"
-var longRunHelp = `
-The run command starts pebble and runs the configured environment.
```
niemeyer, line 36: Curious that most are being moved up while this one is being moved down.

internals/cli/cmd_version.go
```diff
@@ -20,25 +20,27 @@ import (
...
-	cmdpkg "github.com/canonical/pebble/cmd"
+	version "github.com/canonical/pebble/cmd"
 )
 
-var shortVersionHelp = "Show version details"
-var longVersionHelp = `
-The version command displays the versions of the running client and server.
-`
+const cmdVersionSummary = "Show version details"
+const cmdVersionDescription = "The version command displays the versions of the running client and server."
```
niemeyer, line 27: Same here about the style.

niemeyer (changes requested):
Thanks Angel. One more pass.

internals/cli/cmd_run.go
anpep: That's because I was trying to keep the help strings on top of the `cmdFoo` struct, not necessarily on the absolute top of the file. But I agree that it should be, thanks.

internals/cli/cli.go
```diff
@@ -49,70 +50,55 @@ var (
...
+var commands []*CmdInfo
 
-type parserSetter interface {
-	setParser(*flags.Parser)
+// debugCommands holds information about all the subcommands of the `pebble debug` command.
+var debugCommands []*CmdInfo
+
+// AddCommand adds a command to the top-level parser.
+func AddCommand(info *CmdInfo) {
+	commands = append(commands, info)
```
niemeyer, line 101: Haven't read the rest yet, but something looks suspect here: either we don't need the two slices above, or we should have a branch here that splits them up. If we have a single loop, then not having both slices seems good.

internals/cli/cli.go
```diff
@@ -169,134 +155,108 @@ func (ch *clientMixin) setClient(cli *client.Client) {
...
+	if err != nil {
+		logger.Panicf("internal error: cannot add command %q: %v", "debug", err)
+	}
+
+	// Add all commands
+	ncmds := len(commands)
+	allCmds := make([]*CmdInfo, ncmds, ncmds+len(debugCommands))
+	copy(allCmds, commands)
+	copy(allCmds[ncmds:], debugCommands)
+	for _, c := range allCmds {
```
niemeyer, line 202: Again, there's nothing in debugCommands, so we don't need the extra logic here.

niemeyer (changes requested):
Almost there!

internals/cli/cli.go
anpep: That's right, thank you.

internals/cli/cli.go
anpep: Major doh! here. Thank's for your review (:

niemeyer (approved):
Thanks for all the work, Angel.  Hopefully this is ready to go in now.

PR #233, opened 2023-06-05
Fix comment about channel buffer size in cmd_logs.go

### LINKED ISSUE #230: Misleading comment about buffered channel for signal.Notify
According to [documentation](https://pkg.go.dev/os/signal#Notify) this channel needs to be buffered.
>Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

Because `go func()` won't be scheduled right away, it is possible that signal happened after `signal.Notify` but before `<-ch` will be discarded.

https://github.com/canonical/pebble/blob/0c64c2186ce4dcb72035e9442a67df54901fbaa5/internals/cli/cmd_logs.go#L110-L119

Also buffered channel with capacity 2 might not work well when there will be more than 2 signals.

benhoyt:
Fixes #230.

### DIFF
--- a/internals/cli/cmd_logs.go
+++ b/internals/cli/cmd_logs.go
@@ -107,9 +107,9 @@ func (cmd *cmdLogs) Execute(args []string) error {
 // Needed because signal.NotifyContext is Go 1.16+
 func notifyContext(parent context.Context, signals ...os.Signal) context.Context {
 	ctx, cancel := context.WithCancel(parent)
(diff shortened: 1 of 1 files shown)


hpidcock (approved):

benhoyt:
Trivial comment fix, merging directly.

Issue #172, opened 2022-12-12
Add option in pebble exec to inherit environment variables from caller's environment


averma-canonical:
Request to add an option in pebble exec command to inherit environment variables from the caller's environment, much like the preserve-env option in sudo.

The use case is for pebble to execute a one-shot command as a non-root user inheriting all environment variables the rock has defined for the root user.

niemeyer:
Unlike sudo, the thing being started is not a plain shell or a plain command, but rather a program that will carry its own environment and configuration, and in general will have an expectation of it being preserved. Blindly fiddling with this environment is likely to result in unexpected behavior.  For that kind of reason, even in applications such as ssh the client environment's won't be sent over transparently.

Can you provide some further details about what you intent to do with it?  Maybe there is some alternative.

benhoyt:
See also the planned feature for exec (and exec-type health checks) we've called "context", which would allow you to run commands in the context of another service, or of Pebble itself:

* https://github.com/canonical/pebble/issues/159 -- feature issue
* https://github.com/canonical/pebble/issues/125 -- previous discussion

Would `--context=pebble`, or similar, work for you? As Gustavo mentioned, it would also be good to know more about your specific use case.

benhoyt:
@averma-canonical Thanks for that description. It seems to me that the addition of `--context=pebble` (or `--context=-` or whatever syntax we give it) would solve your problem, right? That would mean run with the same environment variables as Pebble was started with.

averma-canonical:
Thanks @benhoyt. Yes that is exactly what we need.

benhoyt:
Thanks. In that case I'll close this issue in favour of the existing #159.

PR #234, opened 2023-06-12
fix: pebble exec and "exec" health checks inherit environment from daemon

rebornplusplus:
Currently, the environment variables from the pebble daemon are not inherited while executing ``pebble exec`` commands.

Change this behavior and inherit all environment variables available to the pebble daemon, in pebble exec.

Additionally, per discussion, this changes the health checks of "exec" type to inherit the daemon's environment as well.

### DIFF
--- a/internals/daemon/api_exec.go
+++ b/internals/daemon/api_exec.go
@@ -18,7 +18,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"os/exec"
+	"strings"
 	"time"
 
 	"github.com/canonical/pebble/internals/osutil"
@@ -73,13 +75,29 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
 		return statusBadRequest("%v", err)
 	}
 
+	// Inherit the pebble daemon environment.
+	environment := make(map[string]string)
+	for _, kv := range os.Environ() {
+		parts := strings.SplitN(kv, "=", 2)
+		key := parts[0]
+		val := ""
+		if len(parts) == 2 {
+			val = parts[1]
+		}
+		environment[key] = val
+	}
+	for key, val := range payload.Environment {
+		// Client env will overwrite existing daemon env, if exists.
+		environment[key] = val
+	}
+
 	st := c.d.overlord.State()
 	st.Lock()
 	defer st.Unlock()
 
 	args := &cmdstate.ExecArgs{
 		Command:     payload.Command,
-		Environment: payload.Environment,
+		Environment: environment,
 		WorkingDir:  payload.WorkingDir,
 		Timeout:     timeout,
 		UserID:      uid,
--- a/internals/daemon/api_exec.go
+++ b/internals/daemon/api_exec.go
@@ -18,9 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"os"
 	"os/exec"
-	"strings"
 	"time"
 
 	"github.com/canonical/pebble/internals/osutil"
@@ -75,29 +73,13 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
 		return statusBadRequest("%v", err)
 	}
 
-	// Inherit the pebble daemon environment.
-	environment := make(map[string]string)
-	for _, kv := range os.Environ() {
-		parts := strings.SplitN(kv, "=", 2)
-		key := parts[0]
-		val := ""
-		if len(parts) == 2 {
-			val = parts[1]
-		}
-		environment[key] = val
-	}
-	for key, val := range payload.Environment {
-		// Client env will overwrite existing daemon env, if exists.
-		environment[key] = val
-	}
-
 	st := c.d.overlord.State()
 	st.Lock()
 	defer st.Unlock()
 
 	args := &cmdstate.ExecArgs{
 		Command:     payload.Command,
-		Environment: environment,
+		Environment: payload.Environment,
 		WorkingDir:  payload.WorkingDir,
 		Timeout:     timeout,
 		UserID:      uid,
--- a/internals/daemon/api_exec_test.go
+++ b/internals/daemon/api_exec_test.go
@@ -119,7 +119,7 @@ func (s *execSuite) TestEnvironmentInheritedFromDaemon(c *C) {
 	c.Check(stdout, Equals, "FOO=bar\n")
 	c.Check(stderr, Equals, "")
 
-	// Check that client env variables overwrite daemon env.
+	// Check that requested environment takes precedence.
 	stdout, stderr, waitErr = s.exec(c, "", &client.ExecOptions{
 		Command:     []string{"/bin/sh", "-c", "echo FOO=$FOO"},
 		Environment: map[string]string{"FOO": "foo"},
--- a/internals/daemon/api_exec_test.go
+++ b/internals/daemon/api_exec_test.go
@@ -109,18 +109,8 @@ func (s *execSuite) TestEnvironment(c *C) {
 }
 
 func (s *execSuite) TestEnvironmentInheritedFromDaemon(c *C) {
-	oldEnv, envWasSet := os.LookupEnv("FOO")
-	err := os.Setenv("FOO", "bar")
-	c.Check(err, IsNil)
-	defer func() {
-		var err error
-		if envWasSet {
-			err = os.Setenv("FOO", oldEnv)
-		} else {
-			err = os.Unsetenv("FOO")
-		}
-		c.Check(err, IsNil)
-	}()
+	restore := fakeEnv("FOO", "bar")
+	defer restore()
 
 	stdout, stderr, waitErr := s.exec(c, "", &client.ExecOptions{
 		Command: []string{"/bin/sh", "-c", "echo FOO=$FOO"},
--- a/internals/daemon/api_exec_test.go
+++ b/internals/daemon/api_exec_test.go
@@ -108,6 +108,28 @@ func (s *execSuite) TestEnvironment(c *C) {
 	c.Check(stderr, Equals, "")
 }
 
+func (s *execSuite) TestEnvironmentInheritedFromDaemon(c *C) {
+	oldEnv, envWasSet := os.LookupEnv("FOO")
+	err := os.Setenv("FOO", "bar")
+	c.Check(err, IsNil)
+	defer func() {
+		var err error
+		if envWasSet {
+			err = os.Setenv("FOO", oldEnv)
+		} else {
+			err = os.Unsetenv("FOO")
+		}
+		c.Check(err, IsNil)
+	}()
+
+	stdout, stderr, waitErr := s.exec(c, "", &client.ExecOptions{
+		Command: []string{"/bin/sh", "-c", "echo FOO=$FOO"},
+	})
+	c.Check(waitErr, IsNil)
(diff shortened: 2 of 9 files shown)


internals/daemon/api_exec.go
```diff
@@ -73,13 +75,29 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
...
+	for _, kv := range os.Environ() {
+		parts := strings.SplitN(kv, "=", 2)
+		key := parts[0]
+		val := ""
+		if len(parts) == 2 {
+			val = parts[1]
+		}
+		environment[key] = val
+	}
+	for key, val := range payload.Environment {
```
cjdcordeiro, line 89: wouldn't it be better to just expand payload.Environment whenever a key from os.Environ doesn't exist, instead of iterating over two maps?

internals/daemon/api_exec_test.go
```diff
@@ -108,6 +108,28 @@ func (s *execSuite) TestEnvironment(c *C) {
 	c.Check(stderr, Equals, "")
 }
 
+func (s *execSuite) TestEnvironmentInheritedFromDaemon(c *C) {
+	oldEnv, envWasSet := os.LookupEnv("FOO")
+	err := os.Setenv("FOO", "bar")
```
cjdcordeiro, line 113: is this safe? other tests simply set the env var at the test execution time

internals/daemon/api_exec_test.go
```diff
@@ -108,6 +108,28 @@ func (s *execSuite) TestEnvironment(c *C) {
...
+		if envWasSet {
+			err = os.Setenv("FOO", oldEnv)
+		} else {
+			err = os.Unsetenv("FOO")
+		}
+		c.Check(err, IsNil)
+	}()
+
+	stdout, stderr, waitErr := s.exec(c, "", &client.ExecOptions{
+		Command: []string{"/bin/sh", "-c", "echo FOO=$FOO"},
```
cjdcordeiro, line 126: you should also test when `exec` has its own Environment

cjdcordeiro (changes requested):

internals/daemon/api_exec.go
rebornplusplus: `payload.Environment` may be `nil`. In that case, it will error out; it's safer to just iterate over it.

internals/daemon/api_exec_test.go
rebornplusplus: It should be safe. The `Environment` field other tests are using in this file seems to be the client env.

internals/daemon/api_exec_test.go
rebornplusplus: Added in https://github.com/canonical/pebble/pull/234/commits/24bb4d2db1ac721b907b4c30c4bd1bd71ff199f4. Thanks!

internals/daemon/api_exec_test.go
woky: I guess it's not thread-safe or rather goroutine-safe, but we do this with `os.Args` already.

woky (approved):
Nice!

internals/daemon/api_exec_test.go
woky: However, I'd put this into a new `fakeEnv()` function. It's likely other tests may want it too in future.

cjdcordeiro (approved):
thanks!

internals/daemon/api_exec_test.go
rebornplusplus: Added in https://github.com/canonical/pebble/pull/234/commits/8edd665543744e57047b848640031424bbb0a25e. Thank you.

rebornplusplus:
Hiya @niemeyer, will you please take a look at this PR? Thank you.

internals/daemon/api_exec.go
```diff
@@ -73,13 +75,29 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
...
+		// Client env will overwrite existing daemon env, if exists.
```
niemeyer, line 90:
```
// Requested environment takes precedence.
```

This makes it slightly more clear that the environment that takes precedence was explicitly provided in the request. This should be obvious from context, but there were early conversations about making the actual client environment be sent to the daemon transparently, and that's not the case here, so let's avoid that interpretation.

internals/daemon/api_exec.go
```diff
@@ -73,13 +75,29 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
...
+	// Inherit the pebble daemon environment.
+	environment := make(map[string]string)
+	for _, kv := range os.Environ() {
+		parts := strings.SplitN(kv, "=", 2)
+		key := parts[0]
+		val := ""
+		if len(parts) == 2 {
+			val = parts[1]
+		}
+		environment[key] = val
```
niemeyer, line 87: Can we please move this too its own function in osutil:
```
func Environ() map[string]string
```
I believe we use similar logic in at least one other place, while running the service command.

niemeyer (changes requested):
This looks nice. Two trivials only:

flotter (approved):

internals/daemon/api_exec.go
```diff
@@ -73,13 +75,29 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
 		return statusBadRequest("%v", err)
 	}
 
+	// Inherit the pebble daemon environment.
+	environment := make(map[string]string)
```
benhoyt, line 79: This will work here, but I think we should do this one level down, in [`cmdstate.Exec`](https://github.com/canonical/pebble/blob/48d20dc71efb14ed52fdf4430c50f2b39a7676af/internals/overlord/cmdstate/request.go#L72), so that any caller that creates "exec" tasks gets the right behaviour, rather than just the API. I realize the API is the only user right now, but it seems the right place to put "business logic" there rather than in the API handler.

internals/daemon/api_exec.go
benhoyt: Yeah, `cmd_exec.go` does it too.

internals/daemon/api_exec_test.go
```diff
@@ -108,6 +108,27 @@ func (s *execSuite) TestEnvironment(c *C) {
 	c.Check(stderr, Equals, "")
 }
 
+func (s *execSuite) TestEnvironmentInheritedFromDaemon(c *C) {
```
benhoyt, line 111: Thanks for adding a test!

benhoyt:
@rebornplusplus @cjdcordeiro Per above, I'm happy with this change (though I think it should go in `cmdstate/request.go:Exec`), and it sounds like you guys have agreed on it with Gustavo already. I previously thought we didn't want to do this based on [this comment](https://github.com/canonical/pebble/issues/172#issuecomment-1346828553), but on re-read of that I guess that's referring to the `pebble exec` client environment, not the daemon environment. Is that correct?

rebornplusplus:
Hi @benhoyt, I will move the changes to `cmdstate/request.go:Exec` soon. Thanks!

> On re-read of that I guess that's referring to the pebble exec client environment, not the daemon environment.

It seems so to me as well. I think the comment was referring to the downsides of passing all client env, which the OP seemed to ask for.

internals/daemon/api_exec.go
rebornplusplus: Moved in https://github.com/canonical/pebble/pull/234/commits/00d7495b57e5532de07ecdfd9a49a3c53466f3d0. Thanks!

internals/daemon/api_exec.go
rebornplusplus: That's better, thank you. Added in https://github.com/canonical/pebble/pull/234/commits/00d7495b57e5532de07ecdfd9a49a3c53466f3d0.

internals/daemon/api_exec.go
rebornplusplus: I added `osutil.Environ()` in https://github.com/canonical/pebble/pull/234/commits/4487bade960ba4970ee079ea5e77b1c2578bf914, and used it in `cmdstate.Exec` in https://github.com/canonical/pebble/pull/234/commits/e5d80308379917cc14ccf078cc486fcbffff810e.

We are doing a similar thing in [`cmd_exec.go`](https://github.com/canonical/pebble/blob/48d20dc71efb14ed52fdf4430c50f2b39a7676af/internals/cli/cmd_exec.go#L94-L102), but that is on a variable called `cmd.Env` not on `os.Environ()`. To support that, I think I will need to change the signature of `osutil.Environ` to the following:
```
func Environ(env []string) map[string]string
```

I will keep this open. Let me know what you think I should do.

benhoyt:
@rebornplusplus @cjdcordeiro Have you thought about whether we should apply the same change to "exec"-type health checks? Currently they explicitly don't inherit from the daemon's environment either. I think we should probably change health checks to match, however, I plan to discuss this with Gustavo in our spec review tonight. See "Open questions" in [OP034](https://docs.google.com/document/d/1hdzrbvrsQomK04FCUTW-pueAG9F3dvrxyISGL5rzM0g/edit).

internals/daemon/api_exec.go
benhoyt: Yeah, looking more closely there I don't think we should use it there. I thought it was coming from `os.Environ` but it's for the command-line args (whose syntax just happens to be the same).

benhoyt (approved):
Looks good to me, thanks!

cjdcordeiro:
> @rebornplusplus @cjdcordeiro Have you thought about whether we should apply the same change to "exec"-type health checks? Currently they explicitly don't inherit from the daemon's environment either. I think we should probably change health checks to match, however, I plan to discuss this with Gustavo in our spec review tonight. See "Open questions" in [OP034](https://docs.google.com/document/d/1hdzrbvrsQomK04FCUTW-pueAG9F3dvrxyISGL5rzM0g/edit).

Thanks for the review @benhoyt . I've read your spec and provided some comments, and yes, I think the `exec`-type health checks should behave in a similar manner. Let us know how that discussion goes.

benhoyt:
I've created https://github.com/canonical/pebble/pull/237 to make the similar change to exec-type health checks.

cjdcordeiro (approved):

internals/overlord/checkstate/checkers.go
```diff
@@ -128,9 +128,16 @@ func (c *execChecker) check(ctx context.Context) error {
...
-	cmd.Env = make([]string, 0, len(c.environment)) // avoid nil to ensure we don't inherit parent env
+	// Similar to services and exec, inherit the daemon's environment.
+	environment := osutil.Environ()
 	for k, v := range c.environment {
+		// Requested environment takes precedence.
+		environment[k] = v
+	}
+
+	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
+	cmd.Env = make([]string, 0, len(environment)) // avoid nil to ensure we don't inherit parent env
```
niemeyer, line 139: The comment and logic here make no sense now. We're explicitly inheriting the parent's environment above.

The line may just be deleted I believe.

niemeyer (changes requested):
Looks good. Just one more trivial:

internals/overlord/checkstate/checkers.go
benhoyt: Good catch. I think the line of code is still worthwhile to do a single allocation (instead of multiple allocations/copies as we append), but I've updated the comment to just `// avoid additional allocations`.

flotter (approved):

PR #236, opened 2023-06-14
Improve error messages with exec, especially if working dir missing

### LINKED ISSUE #235: Misleading error message in container.exec
When running the following code:
```python
self.container = self.unit.get_container("mycontainer")
self.container.exec(["realcommand"], working_dir="/path/that/doesnt/exist")
stdout, stderr = process.wait_output()
if stderr:
    logger.error(stderr)
logger.info(stdout)
```

It returns the following error:
```
fork/exec /usr/local/bin/realcommand: no such file or directory
```

This error is very misleading. `/usr/local/bin/realcommand` does exist. What doesn't exist is `/path/that/doesnt/exist` that was set as `working_dir` of exec call.

Note: this is the original conversation on charmhub mattermost https://chat.charmhub.io/charmhub/pl/b16f95fnt7rmugynwfg5sjbo7a

benhoyt:
This PR improves the error messages with `pebble exec` (at the API level), mainly so that if the working directory you specify doesn't exist, it's clear that the "no such file or directory" message is referring to the working dir and not the executable.

For example, the current error messages (shown below) looks like the `/usr/bin/echo` binary doesn't exist, when it's actually the `-w` working dir you've specified that doesn't exist:

```
$ pebble exec -w /noexist -- /usr/bin/echo foo
- exec command "/usr/bin/echo" (fork/exec /usr/bin/echo: no such file or directory)
```

This update isn't perfect, because it doesn't check the permissions of the working directory (I believe `x` is required), but definitely an improvement for the case at hand.

Fixes #235

### DIFF
--- a/internals/daemon/api_exec.go
+++ b/internals/daemon/api_exec.go
@@ -64,7 +64,20 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
 	// Check up-front that the executable exists.
 	_, err := exec.LookPath(payload.Command[0])
 	if err != nil {
-		return statusBadRequest("%v", err)
+		return statusBadRequest("cannot find executable %q", payload.Command[0])
+	}
+
+	// Also check that the working directory exists, to avoid a confusing
+	// error message later that implies the command doesn't exist:
+	//
+	//  fork/exec /usr/local/bin/realcommand: no such file or directory
+	//
+	// Note that this check still doesn't check that the permissions are
+	// correct to use it as a working directory, but this is a good start.
+	if payload.WorkingDir != "" {
+		if !osutil.IsDir(payload.WorkingDir) {
+			return statusBadRequest("cannot find working directory %q", payload.WorkingDir)
+		}
 	}
 
 	// Convert User/UserID and Group/GroupID combinations into raw uid/gid.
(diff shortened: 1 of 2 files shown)


internals/daemon/api_exec.go
```diff
@@ -64,7 +64,20 @@ func v1PostExec(c *Command, req *http.Request, _ *userState) Response {
 	// Check up-front that the executable exists.
 	_, err := exec.LookPath(payload.Command[0])
```
flotter, line 65: I thought at first that perhaps the executable lookup could happen relative to the working directory, and therefore I was wondering if we should check the working directory first. It seems that ```LookPath()``` will only consult $PATH or an absolute path for verification, so that does not seem like a valid use case.

flotter (approved):
Looks good to me.

benhoyt:
Interestingly, a similar fix was done in `os/exec` [here](https://github.com/golang/go/commit/7aba72baaae64792707076724307f6bdc7fec44f), but it was updated [here](https://github.com/golang/go/commit/d36c095da920cffef16da7f37ce725f9a07f6ae0) to only apply if `SysProcAttr` is nil (which it isn't in our case)

benhoyt:
Minor change to exec error messages -- merging without further review.

PR #238, opened 2023-06-14
feat(cli): implement personality for application name-agnostic help manuals

anpep:
* Implements the concept of personality, two variables that contain information about the application identity, such as the program binary name and the application display name.  This enables the implementation of Pebble-based software that does not necessarily have the same name as Pebble.
 * Help strings throughout the application have been slightly modified so self-references to `pebble`/Pebble are removed without changing the semantics of the help manual.


internals/cli/cmd_enter.go
```diff
@@ -160,7 +160,7 @@ func (cmd *cmdEnter) Execute(args []string) error {
 	case runStop = <-runReadyCh:
 	case runPanic := <-runResultCh:
 		if runPanic == nil {
-			panic("internal error: pebble daemon stopped early")
+			panic("internal error: daemon stopped early")
```
atesburak, line 163: In case of multiple services wouldn't the program name be useful in a panic?

internals/cli/cmd_exec.go
```diff
@@ -71,7 +71,7 @@ stderr are output locally.
 To avoid confusion, exec options may be separated from the command and its
 arguments using "--", for example:
 
-pebble exec --timeout 10s -- echo -n foo bar
+	exec --timeout 10s -- echo -n foo bar
```
benhoyt, line 74: I think removing the program here make these examples worse; it's a bit confusing without the program name prefix, because you'd never actually run it that way. Any way to include it, or do we have an ordering problem where `Personality` has to be set (by the main program) before init time?

If there's an ordering problem, can we do the default `addCommand` calls at parse / run time rather than init time?

internals/cli/cmd_help.go
```diff
@@ -101,11 +102,11 @@ type manfixer struct {
...
 	if !w.done {
 		w.done = true
-		if bytes.HasPrefix(buf, []byte(".TH pebble 1 ")) {
+		if bytes.HasPrefix(buf, []byte(fmt.Sprintf(".TH %s 1 ", cmd.Personality.ProgramName))) {
 			// io.Writer.Write must not modify the buffer, even temporarily
-			n, _ := w.Buffer.Write(buf[:9])
+			n, _ := w.Buffer.Write(buf[:5+len(cmd.Personality.ProgramName)])
 			w.Buffer.Write([]byte{'8'})
-			m, err := w.Buffer.Write(buf[10:])
+			m, err := w.Buffer.Write(buf[6+len(cmd.Personality.ProgramName):])
```
benhoyt, line 109: I'm not sure what's correct here, but it looks like these values will be different to what they were before. They were `buf[:9]` and `buf[10:]`, now they'll be `buf[:11]` and `buf[12:]`.

internals/cli/cmd_help.go
```diff
@@ -191,37 +192,35 @@ var helpCategories = []helpCategory{{
...
 
 func printHelpHeader() {
-	fmt.Fprintln(Stdout, longPebbleDescription)
+	fmt.Fprintln(Stdout, longPebbleDescription())
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleUsage)
+	fmt.Fprintf(Stdout, "Usage: %s <command> [<options>...]\n", cmd.Personality.ProgramName)
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleHelpCategoriesIntro)
+	fmt.Fprintln(Stdout, `Commands can be classified as follows:`)
```
benhoyt, line 207: Nit: no need for the backticks here.

internals/cli/cmd_help.go
```diff
@@ -191,37 +192,35 @@ var helpCategories = []helpCategory{{
...
 }
 
 func printHelpAllFooter() {
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleHelpAllFooter)
+	fmt.Fprintf(Stdout, strings.TrimSpace(`
+Set the PEBBLE environment variable to override the configuration directory
+(which defaults to %s). Set PEBBLE_SOCKET to override
+the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).
```
benhoyt, line 215: Just to clarify, the environment variables will always be `PEBBLE` and `PEBBLE_SOCKET`, even when the program name is `foobar`?

internals/cli/cmd_exec.go
anpep: This might seem like a trivial issue to fix, and I've proposed many options in the past. However, a brief discussion sufficed to conclude that the simplest approach was to try and let go of self-references in help strings.

It's not ideal, but it's the solution with less impact in overall complexity of the code base.

internals/cli/cmd_help.go
anpep: Yes. It's the way to go for now (not to mention, `TERMUS` environment variable would cause a collision here).

internals/cli/cmd_exec.go
benhoyt: Fair enough. I guess we could also do something like `<program> exec --timeout ...`, but I'm not sure that's any clearer. Oh well, as is is okay.

internals/cli/cmd_help.go
benhoyt: I take your word for it that the test passes! But I am curious how the arithmetic works out? Can you explain why 10 before is equivalent to 6+len("pebble")==12 now?

paul-rodriguez (approved):

benhoyt (approved):

flotter (commented):
Looks close! Few ideas to simplify mostly.

internals/cli/cmd_help.go
flotter, line 109: The current Pebble on master, according to my manpage, looks like this. So I think this may have been broken?
```
pebbl8(1)                                            General Commands Manual                                           pebbl8(1)

NAME
       pebble - Tool to interact with pebble

SYNOPSIS
       pebble [OPTIONS]
```       
       
I think this intention is to write all the bytes up to before the 1, then replace it with 8, and write the remaining bytes. The first one should have been [:11] and the second [12:]

internals/cli/cli.go
```diff
@@ -179,8 +180,8 @@ func Parser(cli *client.Client) *flags.Parser {
 	}
 	flagopts := flags.Options(flags.PassDoubleDash)
 	parser := flags.NewParser(&optionsData, flagopts)
-	parser.ShortDescription = "Tool to interact with pebble"
-	parser.LongDescription = longPebbleDescription
+	parser.ShortDescription = fmt.Sprintf("Tool to interact with %s", cmd.Personality.DisplayName)
```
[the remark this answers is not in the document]
atesburak: Makes sense to limit Sprintf to non-string operations for the sake of time and simple readibility.

internals/cli/cmd_exec.go
atesburak: I would still prefer a placeholder that can be used in general as the options without the program itself doesn't feel very clean.

internals/cli/cli.go
```diff
@@ -30,6 +30,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/canonical/pebble/client"
+	"github.com/canonical/pebble/cmd"
```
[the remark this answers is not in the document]
benhoyt: Yeah, perhaps. I don't love `version`, because then it's not at all clear it's the `cmd` package. If anything, I'd suggest `cmdpkg`. However, I'd suggest going the other way and change the receiver name to just `c`, for example `func (c *cmdEnter)`. Single-letter receiver names are the norm in Go, and there's a precedent for that in `cmd_changes.go` already.

internals/cli/cli.go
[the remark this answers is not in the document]
benhoyt: I agree with @flotter that we should do this *where it's clearer and more readable*. But I think it should be a code readability choice, and not a performance concern (50 nanoseconds is not an issue in this context -- [benchmark code](https://gist.github.com/benhoyt/256f27786184be80e621e42c7f3ee271)):

```
$ go test -run=^$ -bench=.
goos: linux
goarch: amd64
pkg: github.com/canonical/pebble/internals/cli
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
BenchmarkSprintf-16    	15079508	        72.37 ns/op
BenchmarkConcat-16     	46966413	        22.65 ns/op
PASS
```

internals/cli/cmd_help.go
```diff
@@ -101,11 +102,11 @@ type manfixer struct {
 func (w *manfixer) Write(buf []byte) (int, error) {
 	if !w.done {
 		w.done = true
-		if bytes.HasPrefix(buf, []byte(".TH pebble 1 ")) {
+		if bytes.HasPrefix(buf, []byte(fmt.Sprintf(".TH %s 1 ", cmd.Personality.ProgramName))) {
```
benhoyt, line 105: It's somewhat subjective, but here's a case where I personally think the `Sprintf` is more readable than string concat, because the big long `cmd.Personality.ProgramName` variable name gets in the way of things. It seems more clear with the printf template string where the fiddly spaces are.

```go
fmt.Sprintf(".TH %s 1 ", cmd.Personality.ProgramName)
// vs
".TH " + cmd.Personality.ProgramName + " 1 "
```

internals/cli/cmd_run.go
```diff
@@ -31,16 +31,16 @@ import (
...
+The run command starts the daemon and runs the configured environment.
 
 Additional arguments may be provided to the service command with the --args option, which
-must be terminated with ";" unless there are no further Pebble options.  These arguments
+must be terminated with ";" unless there are no further program options.  These arguments
 are appended to the end of the service command, and replace any default arguments defined
 in the service plan. For example:
 
-    $ pebble run --args myservice --port 8080 \; --hold
+    run --args myservice --port 8080 \; --hold
```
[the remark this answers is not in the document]
benhoyt: Yeah, I agree that the examples currently don't feel clear. I like `<binary>` because it's short, but it might be a bit confusing with "binary numbers", especially for `<binary> add`. Possibly `<program>` or `<executable>` or even `<pebble>`?

internals/cli/cmd_help.go
```diff
@@ -191,37 +192,35 @@ var helpCategories = []helpCategory{{
...
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleUsage)
+	fmt.Fprintf(Stdout, "Usage: %s <command> [<options>...]\n", cmd.Personality.ProgramName)
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleHelpCategoriesIntro)
+	fmt.Fprintln(Stdout, "Commands can be classified as follows:")
 }
 
 func printHelpAllFooter() {
 	fmt.Fprintln(Stdout)
```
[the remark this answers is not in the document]
benhoyt: Yeah, it's because `cli.Stdout` is a variable changed for testing, so the output has to go there rather than `os.Stdout` (which is where `Println` output goes).

internals/cli/cmd_help.go
anpep: That's interesting because that's not the output I'm getting...
```
pebble(8)                   System Manager's Manual                  pebble(8)

NAME
       pebble - System and service manager
```

```
.TH pebble 8 "22 June 2023"
.SH NAME
pebble \- System and service manager
.SH SYNOPSIS
\fBpebble\fP [OPTIONS]

```

Will investigate, thanks!

internals/cli/cli.go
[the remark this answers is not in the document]
anpep: I agree that using `fmt` here is unnecessary. I just think it makes the code more readable in the sense that the format string reads a lot like the final, user-facing string, and is easier to preview the potential output, also making (again, in my opinion) strings easier to modify.

internals/cli/cli.go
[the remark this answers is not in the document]
anpep: Let's stick with `c` (as a side note, I think `rcmd` comes from *run command*, which is something I think I saw in `cmd_run.go` and replicated here wrongly 🤡)

internals/cli/cmd_run.go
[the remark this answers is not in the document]
anpep: I think `<pebble>` would be the fittest. However, in Termus for instance, there are absolutely no references to `pebble` in any of the help strings (that's the point of this PR!), so for some people outside Canonical the `pebble` reference could be really confusing.

I'd love to hear @niemeyer's opinion on this matter 👍

internals/cli/cmd_help.go
anpep: As a note to myself, using personality here is a bug because the `pebble(1)` string actually comes from the binary name (`go-flags` finds this for us)

internals/cli/cmd_run.go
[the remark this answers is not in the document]
atesburak: IMO <pebble> could be misleading but I think we could be vocal about Pebble's personality here with something like <pebble_personality> or <pebble_child> etc... A single word is always better looking but not sure if it's enough to avoid ambiguity.

internals/cli/cmd_help.go
```diff
@@ -191,37 +192,35 @@ var helpCategories = []helpCategory{{
...
-	pebbleUsage               = "Usage: pebble <command> [<options>...]"
-	pebbleHelpCategoriesIntro = "Commands can be classified as follows:"
-	pebbleHelpAllFooter       = "Set the PEBBLE environment variable to override the configuration directory \n" +
-		"(which defaults to " + defaultPebbleDir + "). Set PEBBLE_SOCKET to override \n" +
-		"the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).\n" +
-		"\n" +
-		"For more information about a command, run 'pebble help <command>'."
-	pebbleHelpFooter = "For a short summary of all commands, run 'pebble help --all'."
-)
+func longPebbleDescription() string {
```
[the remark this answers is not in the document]
anpep: I'd rather use format strings in places where it makes the actual user-facing string (in this case the command description) more readable. Very interested in Gustavo's opinion here.

paul-rodriguez (approved):

atesburak (approved):
Looks on par up to last week's discussion

anpep:
Actually, this is missing the logic for replacing <pebble> with something
else from personality; however I think it would be a great starting point
to merge this along with the CLI API and then implement the replacement
logic, because it has to be done in the parser code which has been
rewritten in the CLI API PR.

anpep:
This PR is ready for review now, but changes will be required when #232 is merged.

paul-rodriguez (approved):

paul-rodriguez (approved):

internals/cli/cmd_run.go
```diff
@@ -31,16 +31,16 @@ import (
 	"github.com/canonical/pebble/internals/systemd"
 )
 
-const cmdRunSummary = "Run the pebble environment"
+const cmdRunSummary = "Run the <display anem> environment"
```
flotter, line 34: Typo on second word.

flotter (approved):
Looks good to me. (Just the one typo to fix)

internals/cli/cmd_run.go
anpep: Nice catch!

benhoyt: Even better if it were "amen"! :-)

paul-rodriguez (approved):
Seems ok

internals/cli/cmd_help.go
```diff
@@ -196,37 +199,35 @@ var HelpCategories = []HelpCategory{{
...
 
 func printHelpHeader() {
-	fmt.Fprintln(Stdout, longPebbleDescription)
+	fmt.Fprintln(Stdout, longPebbleDescription())
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleUsage)
+	fmt.Fprintf(Stdout, "Usage: %s <command> [<options>...]\n", cmd.Personality.ProgramName)
 	fmt.Fprintln(Stdout)
-	fmt.Fprintln(Stdout, pebbleHelpCategoriesIntro)
+	fmt.Fprintln(Stdout, "Commands can be classified as follows:")
```
niemeyer, line 214: Same.

internals/cli/cmd_help.go
niemeyer: Same.. let's please not refactor logic around in unnecessary ways when all this PR should really care about is using the variables that are being established. Makes the PR unnecessarily noisy and harder to review for no benefit.

internals/cli/cmd_help.go
```diff
@@ -196,37 +199,35 @@ var HelpCategories = []HelpCategory{{
...
+the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).
+
+For more information about a command, run '%s help <command>'.
+	`)+"\n", defaultPebbleDir, cmd.Personality.ProgramName)
 }
 
 func printHelpFooter() {
 	printHelpAllFooter()
-	fmt.Fprintln(Stdout, pebbleHelpFooter)
+	fmt.Fprintf(Stdout, "For a short summary of all commands, run '%s help --all'.\n", cmd.Personality.ProgramName)
```
niemeyer, line 230: Same.

internals/cli/cmd_run.go
```diff
@@ -31,16 +31,16 @@ import (
...
 The run command starts pebble and runs the configured environment.
 
 Additional arguments may be provided to the service command with the --args option, which
-must be terminated with ";" unless there are no further Pebble options.  These arguments
+must be terminated with ";" unless there are no further program options.  These arguments
 are appended to the end of the service command, and replace any default arguments defined
 in the service plan. For example:
 
-    $ pebble run --args myservice --port 8080 \; --hold
+    <program name> run --args myservice --port 8080 \; --hold
```
niemeyer, line 43: Lost a `$` here. That's okay if it's fixing the style that is already in use everywhere else. If it's mixed, let's please leave that alone and do it in a separate PR addressing the style alone, since there will likely be other strings not related to the program name that will need changing too.

internals/cli/cmd_warnings.go
```diff
@@ -28,14 +28,15 @@ import (
...
+	"github.com/canonical/pebble/cmd"
 	"github.com/canonical/pebble/internals/osutil"
 )
 
 const cmdWarningsSummary = "List warnings"
 const cmdWarningsDescription = `
 The warnings command lists the warnings that have been reported to the system.
 
-Once warnings have been listed with 'pebble warnings', 'pebble okay' may be used to
+Once warnings have been listed with 'warnings', 'okay' may be used to
```
niemeyer, line 39: This is changing the help text in a way that IMO makes it less clear: it's not "warnings", it's "pebble warnings", a command. Once more, let's please not change the help text in this PR. Let's focus on the task at hand so that we can merge this. If there's a desire to improve the text itself anyhow, this shoud be done and discussed separately instead of inside a bulk change touching a large number of files.

cmd/version.go
```diff
@@ -24,3 +24,17 @@ func MockVersion(version string) (restore func()) {
 	Version = version
 	return func() { Version = old }
 }
+
+// PersonalityInfo contains information about the application identity.
+// These values will be used across the CLI.
+type PersonalityInfo struct {
```
niemeyer, line 30: After reviewing all of the PR, there's apparently no benefit in using this abstraction. How about we simply have cmd.ProgramName and cmd.DisplayName?

niemeyer (changes requested):
Thanks! Quite a few comments, but pretty simple stuff mostly:

internals/cli/cmd_help.go
```diff
@@ -124,40 +127,40 @@ func (w *manfixer) flush() {
 	io.Copy(Stdout, strings.NewReader(str))
 }
 
-func (cmd cmdHelp) Execute(args []string) error {
+func (rcmd cmdHelp) Execute(args []string) error {
```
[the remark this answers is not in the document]
anpep: Looking more into it, seems like `rcmd` was used exclusively in `cmd_run.go` (with the *r*, I assume, from *r*un). I picked the convention of importing the `cmd` package as `cmdpkg`, like it's used in other places (e.g. import `path` as `pathpkg`). Feel free to unresolve this conversation if you're not happy with this decision (:

internals/cli/cmd_help.go
```diff
@@ -196,37 +199,35 @@ var HelpCategories = []HelpCategory{{
...
+func longPebbleDescription() string {
+	return fmt.Sprintf(strings.TrimSpace(`
+%s lets you control services and perform management actions on the
+system that is running them
+	`), cmd.Personality.DisplayName)
+}
 
 func printHelpHeader() {
-	fmt.Fprintln(Stdout, longPebbleDescription)
+	fmt.Fprintln(Stdout, longPebbleDescription())
```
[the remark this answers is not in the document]
anpep: If you're referring to calling this `printLongPebbleDescription` and have the result printed instead of returned, we do need the return value of this function in `cli.go`.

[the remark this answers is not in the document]
niemeyer: The suggestion was to do less: not change the variable name but tune the text, and use the same function we use elsewhere for performing the variable replacement. Am I missing something?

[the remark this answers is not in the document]
anpep: I did miss the point of your feedback (: I will instead keep the same variables and call `cli.applyPersonality()` throughout the code, is that what you mean? Thanks!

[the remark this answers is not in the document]
niemeyer: Yeah, thanks :+1:

internals/cli/cmd_help.go
```diff
@@ -106,12 +108,13 @@ type manfixer struct {
 func (w *manfixer) Write(buf []byte) (int, error) {
 	if !w.done {
 		w.done = true
-		if bytes.HasPrefix(buf, []byte(".TH pebble 1 ")) {
-			// io.Writer.Write must not modify the buffer, even temporarily
```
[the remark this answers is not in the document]
anpep: I just reimplemented this bit without paying much attention to this seemingly important comment. However, I'm now failing to understand why the buffer should not be manipulated, what am I missing?

paul-rodriguez (approved):

anpep:
@benhoyt What's the state of this? Do you think it's merge ready?

cmd/version.go
```diff
@@ -24,3 +24,9 @@ func MockVersion(version string) (restore func()) {
 	Version = version
 	return func() { Version = old }
 }
+
+// ProgramName represents the name of the application binary (i.e. pebble)
```
benhoyt, line 28: Let's drop the i.e. and parenthesis. That's already clear from the default in the code. Same below.

```suggestion
// ProgramName represents the name of the application binary.
```

internals/cli/cli.go
```diff
@@ -237,6 +238,11 @@ func Parser(cli *client.Client) *flags.Parser {
 	return parser
 }
 
+func applyPersonality(s string) string {
+	r := strings.NewReplacer("{{.ProgramName}}", cmd.ProgramName, "{{.DisplayName}}", cmd.DisplayName)
```
benhoyt, line 242: Not that it matters much in this context, but the replacer would normally be created once, as it can do (relatively) expense setup once:

```go
var personalityReplacer  = strings.NewReplacer(...)

func applyPersonality(s string) string {
    return personalityReplacer.Replace(s)
}
```

internals/cli/cli.go
```diff
@@ -338,7 +344,7 @@ func errorToMessage(e error) (normalMessage string, err error) {
 		}
 	case client.ErrorKindSystemRestart:
 		isError = false
-		msg = "pebble is about to reboot the system"
+		msg = fmt.Sprintf("%s is about to reboot the system", cmd.ProgramName)
```
benhoyt, line 347: I would probably use `DisplayName` here, as it's really talking about the product, not the CLI command.

internals/cli/cmd_enter.go
```diff
@@ -178,7 +179,7 @@ func (cmd *cmdEnter) Execute(args []string) error {
 	case runStop = <-runReadyCh:
 	case runPanic := <-runResultCh:
 		if runPanic == nil {
-			panic("internal error: pebble daemon stopped early")
+			panic(fmt.Sprintf("internal error: %s daemon stopped early", cmdpkg.ProgramName))
```
benhoyt, line 182: IMO we should just drop this one and just say "daemon" or "service manager" (depending on what we do elsewhere).

internals/cli/cmd_help.go
```diff
@@ -103,17 +105,19 @@ func (cmd *cmdHelp) setParser(parser *flags.Parser) {
...
-	done bool
+	done        bool
+	programName string
 }
 
 func (w *manfixer) Write(buf []byte) (int, error) {
 	if !w.done {
 		w.done = true
-		if bytes.HasPrefix(buf, []byte(".TH pebble 1 ")) {
+		prefix := ".TH " + w.programName + " "
```
benhoyt, line 115: We probably had this discussion already ages ago, but can you please explain again why this changes from `".TH pebble 1 "` to `".TH pebble "`, and why the slicing lengths change below?

internals/cli/cmd_help.go
```diff
@@ -175,7 +179,7 @@ type HelpCategory struct {
 // HelpCategories helps us by grouping commands
 var HelpCategories = []HelpCategory{{
 	Label:       "Run",
-	Description: "run pebble",
+	Description: "run {{.DisplayName}}",
```
benhoyt, line 182: What about "run the service manager"?

internals/cli/cmd_run.go
```diff
@@ -31,16 +31,16 @@ import (
 	"github.com/canonical/pebble/internals/systemd"
 )
 
-const cmdRunSummary = "Run the pebble environment"
+const cmdRunSummary = "Run the {{.DisplayName}} environment"
 const cmdRunDescription = `
 The run command starts pebble and runs the configured environment.
```
benhoyt, line 36: Need to override this instance too? Probably with `{{.DisplayName}}`.

internals/cli/cmd_run.go
```diff
@@ -98,7 +98,7 @@ func (rcmd *cmdRun) run(ready chan<- func()) {
 			// This exit code must be in system'd SuccessExitStatus.
 			panic(&exitStatus{42})
 		}
-		fmt.Fprintf(os.Stderr, "cannot run pebble: %v\n", err)
+		fmt.Fprintf(os.Stderr, "cannot run %s: %v\n", cmd.ProgramName, err)
```
benhoyt, line 101: What about "cannot run daemon" or "cannot run service manager"?

internals/cli/cmd_signal.go
```diff
@@ -28,7 +28,7 @@ const cmdSignalDescription = `
 The signal command sends a signal to one or more running services. The signal
 name must be uppercase, for example:
 
-pebble signal HUP mysql nginx
+	{{.ProgramName}} signal HUP mysql nginx
```
benhoyt, line 31: Any reason this should be indented?

benhoyt (changes requested):
Some changes requested. Note also that the usage help still says "Usage: pebble <command> ...". Can we change that? For example, I've changed `ProgramName` to `elbbep` locally, and get this:

```
$ go run ./cmd/pebble run -h
Usage:
  pebble run [run-OPTIONS]
...
```

internals/cli/cli.go
paul-rodriguez: Wouldn't this cause the function to not change behavior when `cmd.ProgramName` is modified dynamically?

benhoyt: Ooh, yes, you're right -- my mistake.

internals/cli/cmd_signal.go
anpep: It's indented everywhere else (: Also, the pattern is to even include the shell caret (`$`) which I'm not sure if I should include here as well for consistency.

internals/cli/cmd_help.go
anpep: The fact that I don't recall the details about this issue validates your point even further (: I'll look at it, fail again to find a better way to do this, and really document it this time.

(Also: see https://github.com/canonical/go-flags/issues/5)

anpep: So, apparently this is not changing from `.TH pebble 1 ` to `.TH pebble `. What it's doing is identify the `.TH <program name> ` prefix, find it, and replace the man section number afterwards by an `8` (instead of the default `1`).

The slicing lengths change from fixed magic numbers to the actual lengths of the prefix so that the code appears to be less magic than it was before.

The `w.programName` bit comes from the actual go-flags understanding of what is the program name (`cmd.parser.Name`).

anpep:
> Note also that the usage help still says "Usage: pebble ...". Can we change that?

I'll have to look if go-flags allows for such a change. Otherwise I'm afraid we might be stuck with it taking the binary name from argv0.

internals/cli/cmd_signal.go
benhoyt: The other example (in `cmd_exec.go`) doesn't use indentation or the `$ ` prefix. How about we drop them here and in `cmd_run.go` for consistency?

```
$ rg '^\s*pebble ' -C2
cmd_signal.go
29-name must be uppercase, for example:
30-
31:pebble signal HUP mysql nginx
32-`
33-

cmd_exec.go
40-arguments using "--", for example:
41-
42:pebble exec --timeout 10s -- echo -n foo bar
43-`
44-
```

run-checks
```diff
@@ -0,0 +1,183 @@
+#!/bin/bash
```
benhoyt, line 1: I don't mind considering adding a script to help running all checks, however, I'd prefer to do it in a separate PR with its own discussion. So let's remove this from this PR.

If we do add a script for this, let's keep it ultra-simple with no fancy Bash logic. I agree with @mwhudson's lightning talk about "stop writing Bash" :-), as it's error prone and hard to maintain. I'd probably suggest a `Makefile` with a `make check` target.

I personally don't mind typing `go test ./...`, but I'd be interested to see what other Gophers do/recommend.

internals/cli/cmd_help.go
benhoyt: Thanks. If you could add an updated comment about what's going on here, that would be great.

internals/cli/cmd_run.go
benhoyt: Looks like this comment was resolved but not done? The `cmdRunDescription` still says "The run command starts pebble..."

internals/cli/cmd_enter.go
benhoyt: That's fine -- I'd be happy to reinstate it as `{{.DisplayName}}`.

benhoyt:
> > Note also that the usage help still says "Usage: pebble ...". Can we change that?
> 
> I'll have to look if go-flags allows for such a change. Otherwise I'm afraid we might be stuck with it taking the binary name from argv0.

Looks like we could change it with this tweak:

```diff
diff --git a/internals/cli/cli.go b/internals/cli/cli.go
index 4de25ac..86c46e8 100644
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -166,6 +166,7 @@ func Parser(cli *client.Client) *flags.Parser {
 
        flagOpts := flags.Options(flags.PassDoubleDash)
        parser := flags.NewParser(&defaultOpts, flagOpts)
+       parser.Command.Name = cmd.ProgramName
        parser.ShortDescription = "System and service manager"
        parser.LongDescription = applyPersonality(longPebbleDescription)
 ```

**Review by benhoyt — CHANGES_REQUESTED** [review]
Few more minor comments. See note about modifying the "Usage:" thing above.

**Comment by benhoyt** [comment]
One more thing. Now that the `pebble notices` command is merged, we need to update that to use `{{.ProgramName}}` too. I've changed `ProgramName` and `DisplayName` locally to something other than "pebble" / "Pebble", and get this (the `run` one I've already noted above):

```
$ go run ./cmd/pebble help --man >manpage 
$ grep -i pebble manpage
may then be acknowledged by running 'pebble okay'. When a notice repeats, it
The \fIrun\fP command starts pebble and runs the configured environment.
```

File: run-checks (hunk 30 above)
Comment (anpep) [reply run-checks:1]: Oops! Accidentally committed this file (shouldn't be here)

File: internals/cli/cmd_warnings.go [hunk 31]
```diff
@@ -209,7 +197,7 @@ func lastWarningTimestamp() (time.Time, error) {
 	f, err := os.Open(warnFilename(user.HomeDir))
 	if err != nil {
 		if os.IsNotExist(err) {
-			return time.Time{}, fmt.Errorf("you must have looked at the warnings before acknowledging them. Try '%s warnings'.", cmd.ProgramName)
+			return time.Time{}, fmt.Errorf("you must have looked at the warnings before acknowledging them. Try 'pebble warnings'.")
```
paul-rodriguez: Is this intentional? I don't get why this should refer to pebble explicitly.

paul-rodriguez (approved):
run-checks should be removed. I was confused by one of the changes putting back an explicit "pebble".

internals/cli/cmd_warnings.go
anpep: It's a silly little mistake (: Good catch!

benhoyt (approved):
Looks good to me now, thanks!

PR #239, opened 2023-06-19
Allow non-root user to specify current user/group when running commands

### LINKED ISSUE #166: [bug] when started as a nonroot user, Pebble doesn't run services with the user's own UID
**Let's assume the nonroot user:group "ubuntu:ubuntu" with UID:GID "1000:1000"**.

### SOTA

At the moment, it is possible to run Pebble as a nonroot user. So, given the following layer:

```yaml
summary: test layer

services:
  svc1:
    override: replace
    command: /app.py
    startup: enabled
```

`pebble run` would work just fine:

```bash
# ps -u ubuntu
    PID TTY          TIME CMD
     75 pts/1    00:00:00 bash
    102 pts/1    00:00:00 pebble
    149 pts/1    00:00:00 python3
(...)
```

### Problem

However, if a user*/group* attribute is specified in the layer, then we would expect:
 - if `user` **is not** "me" (i.e. the user running Pebble), then fail
   - this is the case :heavy_check_mark: 
 - if `user` **is** "me", then continue
   - this is NOT the case, as Pebble fails to start the service, with permission denied :x: 

E.g.:

```yaml
summary: test layer

services:
  svc1:
    override: replace
    command: /app.py
    startup: enabled
    user-id: 1000
    group-id: 1000
```
  
and the `pebble run`, will not start the `svc1`. And we I try to force start it, I get:

```bash
# pebble start svc1
- Start service "svc1" (cannot start service: fork/exec /app.py: operation not permitted)
```

### Conclusion

I believe this is being caused by https://github.com/canonical/pebble/blob/master/internal/overlord/servstate/handlers.go#L342, where Pebble is trying to set the Credentials attribute for the `cmd`, whenever `uid` and `gid` are defined (no matter what they are). I'd expect instead, that if the give UID and GID match the Pebble user, then the SysProcAttr->Credential can be ignored.

benhoyt:
When user and group (or user ID and group ID) options are given when running services and they're equal to the current user and group, avoid setting os/exec's `syscall.Credential` `Uid` and `Gid`, because they require root. In the case when they're explicitly asking to run the command as the current user, we can just run it directly, without requiring root.

This goes for services, exec (one-shot commands), and health checks of type "exec".

This PR also fixes a long-standing bug in the user/group options for "exec" health checks not working at all. This was due to not setting `cmd.SysProcAttr`, so the `cmd.SysProcAttr.Credential = ...` line would panic with a nil pointer error.

Also a drive-by fix to the README to update to the new terminology. See also https://github.com/canonical/operator/pull/950

Fixes #166

### DIFF
--- a/README.md
+++ b/README.md
@@ -667,7 +667,7 @@ The Go client is used primarily by the CLI, but is importable and can be used by
 
 We try to never change the underlying HTTP API in a backwards-incompatible way, however, in rare cases we may change the Go client in a backwards-incompatible way.
 
-In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the Python Operator Framework used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
+In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the [`ops` library](https://github.com/canonical/operator) used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
 
 ## Roadmap / TODO
 
--- a/internals/osutil/user.go
+++ b/internals/osutil/user.go
@@ -124,3 +124,21 @@ func NormalizeUidGid(uid, gid *int, username, group string) (*int, *int, error) 
 	}
 	return uid, gid, nil
 }
+
+// IsCurrent reports whether the given user ID and group ID are those of the
+// current user.
+func IsCurrent(uid, gid int) (bool, error) {
+	current, err := userCurrent()
+	if err != nil {
+		return false, err
+	}
+	currentUid, err := strconv.Atoi(current.Uid)
+	if err != nil {
+		return false, err
+	}
+	currentGid, err := strconv.Atoi(current.Gid)
+	if err != nil {
+		return false, err
+	}
+	return uid == currentUid && gid == currentGid, nil
+}
--- a/internals/overlord/checkstate/checkers.go
+++ b/internals/overlord/checkstate/checkers.go
@@ -148,9 +148,16 @@ func (c *execChecker) check(ctx context.Context) error {
 		return err
 	}
 	if uid != nil && gid != nil {
-		cmd.SysProcAttr.Credential = &syscall.Credential{
-			Uid: uint32(*uid),
-			Gid: uint32(*gid),
+		isCurrent, err := osutil.IsCurrent(*uid, *gid)
+		if err != nil {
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -667,7 +667,7 @@ The Go client is used primarily by the CLI, but is importable and can be used by
 
 We try to never change the underlying HTTP API in a backwards-incompatible way, however, in rare cases we may change the Go client in a backwards-incompatible way.
 
-In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the Python Operator Framework used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
+In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the [`ops` library](https://github.com/canonical/operator) used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
```
benhoyt, line 670: Unrelated change per updated Juju Terminology doc. See also https://github.com/canonical/operator/pull/950

internals/osutil/user.go
```diff
@@ -124,3 +124,21 @@ func NormalizeUidGid(uid, gid *int, username, group string) (*int, *int, error)
...
+
+// IsCurrent reports whether the given user ID and group ID are those of the
+// current user.
+func IsCurrent(uid, gid int) (bool, error) {
+	current, err := userCurrent()
+	if err != nil {
+		return false, err
+	}
+	currentUid, err := strconv.Atoi(current.Uid)
+	if err != nil {
```
niemeyer, line 136: We might shortcut it with `err != nil || uid != currentUid`, in which case we might use the same pattern below and return a bare `true, nil` at the end.

internals/overlord/checkstate/checkers.go
```diff
@@ -148,9 +148,16 @@ func (c *execChecker) check(ctx context.Context) error {
 		return err
 	}
 	if uid != nil && gid != nil {
-		cmd.SysProcAttr.Credential = &syscall.Credential{
-			Uid: uint32(*uid),
-			Gid: uint32(*gid),
+		isCurrent, err := osutil.IsCurrent(*uid, *gid)
+		if err != nil {
+			logger.Debugf("Cannot determine if uid %d gid %d is current user", *uid, *gid)
```
niemeyer, line 153: A `...: %v` error suffix might come handy if this is ever an actual issue.

niemeyer (approved):
Thanks for the fixes. Two non-important comments below, but I'm going ahead and merging it anyway since our timezones are off and it will take a while to merge it otherwise.

PR #243, opened 2023-06-21
[RFC] shutdown/notify: allow for custom implementations

flotter:
The current shutdown and notify logic in pebble is designed with the
assumption that it will run under systemd, or at least that some
systemd components, such as ```shutdown```, is installed.

Allow custom flavours of pebble to change the default assumptions.

- The daemon now defines a notification and shutdown interface, and
  configures the default implementation to use systemd.

- The shutdown logic (of which we currently only support reboot) is
  moved to the systemd package, as its based on systemd.

- Two globals daemon.Notify and daemon.Shutdown now provides a way
  to override the defaults.

This is an alternative approach to what we have in https://github.com/canonical/pebble/pull/197

An hypothetical derivative project based on Pebble can now supply its own shutdown and notification implementations before starting the daemon:

```
daemon.Notify = custom.Notifier 
daemon.Shutdown = custom.Shutdown
:
d, err := daemon.New(&dopts)
:
d.Start()
```

### DIFF
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -28,7 +28,6 @@ import (
 	"github.com/canonical/pebble/cmd"
 	"github.com/canonical/pebble/internals/daemon"
 	"github.com/canonical/pebble/internals/logger"
-	"github.com/canonical/pebble/internals/systemd"
 )
 
 var shortRunHelp = "Run the pebble environment"
@@ -114,7 +113,7 @@ func runWatchdog(d *daemon.Daemon) (*time.Ticker, error) {
 			case <-wt.C:
 				// TODO: poke the API here and only report WATCHDOG=1 if it
 				//       replies with valid data.
-				systemd.SdNotify("WATCHDOG=1")
+				daemon.Notify.Send("WATCHDOG=1")
 			case <-d.Dying():
 				return
 			}
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -23,7 +23,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"os/exec"
 	"os/signal"
 	"runtime"
 	"strings"
@@ -36,7 +35,6 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/canonical/pebble/internals/logger"
-	"github.com/canonical/pebble/internals/osutil"
 	"github.com/canonical/pebble/internals/osutil/sys"
 	"github.com/canonical/pebble/internals/overlord"
 	"github.com/canonical/pebble/internals/overlord/checkstate"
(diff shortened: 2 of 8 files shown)


internals/daemon/daemon.go
```diff
@@ -717,22 +742,6 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
 	return fmt.Errorf("expected reboot did not happen")
 }
 
-var shutdownMsg = "reboot scheduled to update the system"
```
flotter, line 720: This has been moved to ```internals/systemd/shutdown```. Note that I am reusing the ```shutdown``` command group name here, where the command offers halt, power-off and reboot functionality.

internals/daemon/daemon_test.go
```diff
@@ -724,32 +741,6 @@ func (s *daemonSuite) TestRestartSystemWiring(c *check.C) {
 	c.Check(rebootAt.After(approxAt) || rebootAt.Equal(approxAt), check.Equals, true)
 }
 
-func (s *daemonSuite) TestRebootHelper(c *check.C) {
```
flotter, line 727: This was moved to ```internals/systemd/shutdown_test.go```.

internals/daemon/daemon_test.go
```diff
@@ -944,9 +931,6 @@ func (s *daemonSuite) TestRestartIntoSocketModeNoNewChanges(c *check.C) {
 		time.Sleep(5 * time.Millisecond)
 	}
 
-	c.Assert(d.standbyOpinions.CanStandby(), check.Equals, false)
-	f, _ := os.Create(notifySocket)
-	f.Close()
```
flotter, line 949: See comment above.

anpep (approved):
Really clean approach!
Since I've seen pebble have platform-specific code, maybe we could consider moving `internals/systemd/{notify,shutdown}.go -> internals/systemd/{notify,shutdown}_linux.go`, then have `_darwin.go` return `ErrDarwin`.

internals/cli/cmd_run.go
```diff
@@ -114,7 +113,7 @@ func runWatchdog(d *daemon.Daemon) (*time.Ticker, error) {
 			case <-wt.C:
 				// TODO: poke the API here and only report WATCHDOG=1 if it
 				//       replies with valid data.
-				systemd.SdNotify("WATCHDOG=1")
+				daemon.Notify.Send("WATCHDOG=1")
```
benhoyt, line 116: I'm not really sure what `WATCHDOG=1` does, but it seems like it's pretty specific to systemd? In that case, could it not be a string in the `Send` interface, but just included in the systemd implementation? Or something like `Send(daemon.SendWatchdog)` and `Send(daemon.SendReady)` (not the best API, but spitballing here).

internals/daemon/daemon.go
```diff
@@ -47,13 +45,40 @@ import (
...
+// notifySupervisor interface defines the required API for
+// sending notifications to a supervisor daemon.
+type notifySupervisor interface {
+	Available() bool
+	Send(state string) error
+}
+
+// shutdown defines the required API for reboot, poweroff
+// and halt control of the underlying system
+type shutdown interface {
```
benhoyt, line 57: Because external packages need to implement these interfaces (and might want to pass around the type), shouldn't they be exported/capitalised?

internals/daemon/daemon.go
```diff
@@ -47,13 +45,40 @@ import (
...
+	// not have systemd, and will use this system manager directly.
+
+	// Notify provides a way to send a notification to a supervisor
+	// daemon, if one exists. The default implementation assumes
+	// systemd and should be replaced if not available.
+	Notify notifySupervisor = systemd.Notifier
+
+	// Shutdown provides the system specific functionality to
+	// halt, poweroff or reboot.
+	Shutdown shutdown = systemd.Shutdown
```
benhoyt, line 77: It seems really unfortunate that these need to be globals when the `Daemon` is created via the more typical `daemon.New` pattern (that is, not a global). Can these somehow be part of the `Options` struct and not be globals of the daemon package? But I guess the tricky part is then having the alternative CLI customize it.

One approach would be to add Notify and Shutdown to `daemon.Options`, but then have a `ConfigureDaemon` that's part of the CLI's "personality". The signature would be `ConfigureDaemon func(opts *daemon.Options)`, and the alternative CLI would do something like:

```go
cmd.Personality.ConfigureDaemon = func(opts *daemon.Options) {
    opts.Notify = mySpecialNotify
    opts.Shutdown = mySpecialShutdown
}
```

I dunno, maybe it's not any better, but at least it would confine the messy globals to one place, the "personality" struct. Or we could put the globals in the `internals/cli` package -- putting them in the CLI still seems better to me than in the daemon package.

atesburak: I like the idea of confinement but is this really a part of the Personality, I can see this behaviour growing with other options, so a new struct to embody them?

anpep: I don't think this is part of Personality, because (i) it does not concern other components apart from the daemon, and (ii) Termus *will* have its own initialization of the daemon during bootstrap.

I'm on board @benhoyt's idea of having those be inside `daemon.Options` 😃

PR #244, opened 2023-06-22
testutil: support empty string args in fakecommand

flotter:
testutil: support empty string args in fakecommand

FakeCommand allows unit tests to check which command-line arguments
were passed, by recording a log of calls with their arguments.

The following examples cannot be tested with FakeCommand:

1. shutdown -r "+0" ""

2. echo ""

3. configure --program-prefix ""

The empty string argument confuses the command parser in testutil.

The issue is that both the argument (\000) and command delimiter
(\000\000) uses null characters, which makes the command split
operation split on the first occurrence of two null chars. This
is an invalid assumption for an empty string argument, which will
be presented as two argument delimiters, without anything in
between, appearing like a command delimiter.

To add support for this:

- Change the command delimiter from \000\000 => \000\f\n\r (magic sequence)

- Update the Calls() function to split the lines correctly.

Testutil can now be used for testing the following command, including the
case where msg is an empty string for the wall message:

shutdown [OPTIONS...] [TIME] [WALL...]

cmd := exec.Command("shutdown", "-r", fmt.Sprintf("+%d", mins), msg)

### DIFF
--- a/internals/testutil/exec.go
+++ b/internals/testutil/exec.go
@@ -92,7 +92,7 @@ type FakeCmd struct {
 // faking commands that need "\n" in their args (like zenity)
 // we use the following convention:
 // - generate \0 to separate args
-// - generate \0\0 to separate commands
+// - generate \0\f\n\r magic sequence to separate commands
 var scriptTpl = `#!/bin/bash
 printf "%%s" "$(basename "$0")" >> %[1]q
 printf '\0' >> %[1]q
@@ -102,7 +102,7 @@ for arg in "$@"; do
      printf '\0'  >> %[1]q
 done
 
-printf '\0' >> %[1]q
+printf '\f\n\r' >> %[1]q
 %s
 `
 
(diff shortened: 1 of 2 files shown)


flotter:
This was an enhancement I did for https://github.com/canonical/pebble/pull/243 which is now closed.

atesburak (approved):

benhoyt (approved):
Fine with me.

anpep (approved):

jnsgruk:
@flotter you mentioned that this was for #243 which is now closed -- is this PR still relevant? If so I think we're good to merge it.

flotter:
@jnsgruk I did the work to enhance the testutil library to test the changes in the now closed PR. When I closed the PR, I decided that this may be a useful enhancement to keep, so I broke this out into its own PR.

jnsgruk (approved):

Issue #159, opened 2022-10-18
Add "context" option for pebble exec and exec health checks


benhoyt:
Per discussion with Gustavo at our spec meeting last night, we'd like to add an API parameter context to pebble exec (command line arg --context) and to exec-type health checks, which would be the name of the service whose context to run the command in. It would use all the relevant context from the service definition, specifically [environment variables and user and group](https://github.com/canonical/pebble/blob/26c8f010363c93dfd8ccdd5730c553a073a17370/internal/plan/plan.go#L74-L79).

Once we [add `working-dir`](https://github.com/canonical/pebble/issues/158) we'll include that in the context too.

Note that we'd also like a way to say "inherit the context / env vars from pebble itself". We could do that with `--context=-` or `--context=pebble` or even `--context=` (empty string). We didn't come to a final decision on that syntax though.

[Spec OP034.](https://docs.google.com/document/d/1hdzrbvrsQomK04FCUTW-pueAG9F3dvrxyISGL5rzM0g/edit)

benhoyt:
Done in #246, though we decided to spell it "service-context".

PR #245, opened 2023-06-26
Add support for service's to configure working directory

### LINKED ISSUE #158: Add working-dir option to service definition
Both `pebble exec` (one-shot commands) and exec-type health checks have an optional `working-dir` argument, which allows you to specify the working directory the command runs in. Services, oddly, don't have this. We should add `working-dir` to the [plan.Service struct here](https://github.com/canonical/pebble/blob/26c8f010363c93dfd8ccdd5730c553a073a17370/internal/plan/plan.go#L74-L79).

Then once `--context` is implemented (for one-shot commands and exec health checks) that would also be included in the context. See https://github.com/canonical/pebble/issues/125.

benhoyt:
Fixes #158 (in preparation for #159)

### DIFF
--- a/internals/overlord/servstate/manager_test.go
+++ b/internals/overlord/servstate/manager_test.go
@@ -1763,3 +1763,62 @@ type fakeLogManager struct{}
 func (f fakeLogManager) ServiceStarted(serviceName string, logs *servicelog.RingBuffer) {
 	// no-op
 }
+
+func (s *S) TestNoWorkingDir(c *C) {
+	// Service command should run in current directory (package directory)
+	// if "working-dir" config option not set.
+	dir := c.MkDir()
+	err := os.Mkdir(filepath.Join(dir, "layers"), 0755)
+	c.Assert(err, IsNil)
+	manager, err := servstate.NewManager(s.st, s.runner, dir, nil, nil, fakeLogManager{})
+	c.Assert(err, IsNil)
+	defer manager.Stop()
+
+	outputPath := filepath.Join(dir, "output")
+	layer := parseLayer(c, 0, "layer", fmt.Sprintf(`
+services:
+    nowrkdir:
+        override: replace
+        command: /bin/sh -c "pwd >%s; sleep %g"
+`, outputPath, shortOkayDelay.Seconds()+0.01))
+	err = manager.AppendLayer(layer)
+	c.Assert(err, IsNil)
+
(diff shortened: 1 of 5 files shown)


internals/overlord/servstate/manager_test.go
```diff
@@ -1763,3 +1763,62 @@ type fakeLogManager struct{}
...
+	manager, err := servstate.NewManager(s.st, s.runner, dir, nil, nil, fakeLogManager{})
+	c.Assert(err, IsNil)
+	defer manager.Stop()
+
+	outputPath := filepath.Join(dir, "output")
+	layer := parseLayer(c, 0, "layer", fmt.Sprintf(`
+services:
+    nowrkdir:
+        override: replace
+        command: /bin/sh -c "pwd >%s; sleep %g"
```
barrettj12, line 1782: So this doesn't depend on which directory you run `go test` from?

internals/plan/plan.go
```diff
@@ -154,6 +155,9 @@ func (s *Service) Merge(other *Service) {
 	if other.Group != "" {
 		s.Group = other.Group
 	}
+	if other.WorkingDir != "" {
+		s.WorkingDir = other.WorkingDir
+	}
```
barrettj12, line 160: Presumably you need to add something to the `Service.Copy` method too.

barrettj12 (changes requested):

internals/plan/plan.go
barrettj12: and maybe add a plan test to check this parses correctly?

internals/overlord/servstate/manager_test.go
benhoyt: No, `go test` is explicitly documented to run the binary for the test package in the package directory.

internals/plan/plan.go
benhoyt: You don't actually need it for `Service.Copy` because the value fields are copied by [this line](https://github.com/canonical/pebble/blob/8902cbe618e4b1bcfbc16a2a6c597c6c49aa88c7/internals/plan/plan.go#L98).

benhoyt: > maybe add a plan test to check this parses correctly?

Good *plan*. I'll do that tomorrow.

niemeyer (approved):
Looks good, thanks!

Not merging waiting for the point you mention in @barrettj12's review.

PR #246, opened 2023-06-26
Add "service context" support for exec and health checks

benhoyt:
This adds the ability to run exec commands (one-shot commands) and exec-based health checks using the context from an existing service, specifically, reusing its environment variables, user/group, and working directory. Any options in the exec command or health check, if present, override the service context (and the environment variables are merged).

Note that #245 -- support for `working-dir` in the service configuration -- should be merged first.

This is based on spec [OP034](https://docs.google.com/document/d/1hdzrbvrsQomK04FCUTW-pueAG9F3dvrxyISGL5rzM0g/edit).

### DIFF
--- a/client/exec.go
+++ b/client/exec.go
@@ -30,6 +30,12 @@ type ExecOptions struct {
 	// Required: command and arguments (first element is the executable).
 	Command []string
 
+	// Optional service name to run the command in the context of this service,
+	// that is, inherit its environment variables, user/group settings,
+	// and working directory. The other options in this struct will override
+	// the service context; Environment will be merged on top of the service's.
+	Context string
+
 	// Optional environment variables.
 	Environment map[string]string
 
@@ -75,6 +81,7 @@ type ExecOptions struct {
 
 type execPayload struct {
 	Command     []string          `json:"command"`
+	Context     string            `json:"context,omitempty"`
 	Environment map[string]string `json:"environment,omitempty"`
 	WorkingDir  string            `json:"working-dir,omitempty"`
 	Timeout     string            `json:"timeout,omitempty"`
@@ -123,6 +130,7 @@ func (client *Client) Exec(opts *ExecOptions) (*ExecProcess, error) {
 	}
 	payload := execPayload{
 		Command:     opts.Command,
+		Context:     opts.Context,
 		Environment: opts.Environment,
 		WorkingDir:  opts.WorkingDir,
 		Timeout:     timeoutStr,
--- a/internals/cli/cmd_exec.go
+++ b/internals/cli/cmd_exec.go
@@ -39,6 +39,7 @@ type cmdExec struct {
 	GroupID        *int          `long:"gid"`
 	Group          string        `long:"group"`
 	Timeout        time.Duration `long:"timeout"`
+	Context        string        `long:"context"`
 	Terminal       bool          `short:"t"`
 	NoTerminal     bool          `short:"T"`
 	Interactive    bool          `short:"i"`
@@ -56,6 +57,7 @@ var execDescs = map[string]string{
 	"gid":     "Group ID to run command as",
 	"group":   "Group name to run command as (group's GID must match gid if both present)",
 	"timeout": "Timeout after which to terminate command",
+	"context": "Run the command in the context of this service (overridden by -w, --env, --user, --uid, --group, --gid)",
 	"t":       "Allocate remote pseudo-terminal and connect stdout to it (default if stdout is a TTY)",
 	"T":       "Disable remote pseudo-terminal allocation",
 	"i":       "Interactive mode: connect stdin to the pseudo-terminal (default if stdin and stdout are TTYs)",
@@ -144,6 +146,7 @@ func (cmd *cmdExec) Execute(args []string) error {
 
 	opts := &client.ExecOptions{
 		Command:     command,
+		Context:     cmd.Context,
 		Environment: env,
 		WorkingDir:  cmd.WorkingDir,
 		Timeout:     cmd.Timeout,
--- a/internals/overlord/checkstate/manager.go
+++ b/internals/overlord/checkstate/manager.go
@@ -100,55 +100,28 @@ func newChecker(config *plan.Check, p *plan.Plan) checker {
 		}
 
 	case config.Exec != nil:
-		var (
-			environment     = make(map[string]string)
-			userID, groupID *int
-			user, group     string
-			workingDir      string
-		)
-		if config.Exec.Context != "" {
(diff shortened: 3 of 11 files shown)


client/exec.go
```diff
@@ -30,6 +30,12 @@ type ExecOptions struct {
 	// Required: command and arguments (first element is the executable).
 	Command []string
 
+	// Optional service name to run the command in the context of this service,
```
barrettj12, line 33: Grammar doesn't quite parse here. Maybe
```
Optional: name of a service in whose context this command will be run,
```

internals/cli/cmd_exec.go
```diff
@@ -56,6 +57,7 @@ var execDescs = map[string]string{
 	"gid":     "Group ID to run command as",
 	"group":   "Group name to run command as (group's GID must match gid if both present)",
 	"timeout": "Timeout after which to terminate command",
+	"context": "Run the command in the context of this service (overridden by -w, --env, --user, --uid, --group, --gid)",
```
barrettj12, line 60: Maybe "Inherit the context (env, user/group settings, working dir) of the named service"

Also with the list: if we change anything, we'll have to either update this list, or (more likely) forget to and it will go out of sync. Better not to list things here IMO, just say "overridden by exec options" as you have in other doc comments.

internals/plan/plan.go
```diff
@@ -765,6 +771,12 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 					Message: fmt.Sprintf("plan check %q command invalid: %v", name, err),
 				}
 			}
+			if check.Exec.Context != "" && !serviceNames[check.Exec.Context] {
```
barrettj12, line 774: Why can't you just check `combined.Services` here?
```go
if _, ok := combined.Services[check.Exec.Context]; !ok { ... }
```

internals/plan/plan.go
```diff
@@ -765,6 +771,12 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+				return nil, &FormatError{
+					Message: fmt.Sprintf("plan check %q context %q is not a service name",
+						name, check.Exec.Context),
```
barrettj12, line 777: Not a huge fan of the error message here. IMO it's not `%q is not a service name` but more like `service %q not found`. Maybe
`plan check %q context: service %q not found`
or
`plan check %q context specifies non-existent service %q`

internals/plan/plan.go
```diff
@@ -1105,3 +1117,74 @@ func ReadDir(dir string) (*Plan, error) {
 	}
 	return plan, err
 }
+
+// MergeServiceContext merges the overrides on top of the service context
+// specified by serviceName, returning a new ContextOptions value.
+func MergeServiceContext(p *Plan, serviceName string, overrides ContextOptions) (ContextOptions, error) {
+	if serviceName == "" {
+		return overrides, nil
```
barrettj12, line 1125: What's the rationale behind this block?

barrettj12 (approved):

internals/overlord/checkstate/manager.go
```diff
@@ -65,7 +65,7 @@ func (m *CheckManager) PlanChanged(p *plan.Plan) {
 		ctx, cancel := context.WithCancel(context.Background())
 		check := &checkData{
 			config:  config,
-			checker: newChecker(config),
+			checker: newChecker(config, p),
 			ctx:     ctx,
 			cancel:  cancel,
```
niemeyer, line 70: This is not about this PR, but looking at this logic makes me think there's a bug here that needs to be addressed separately.  The cancel here is a request. Right above we're calling cancel() as if this would be a synchronous operation, but it's not. The goroutine may take its time to run. This may cause very awkward outcomes, such as an old check running after a new check. We need to use tomb or something similar to make the stopping of checkers become synchronous.

internals/overlord/checkstate/manager.go
```diff
@@ -100,15 +100,28 @@ func newChecker(config *plan.Check) checker {
...
 	case config.Exec != nil:
+		overrides := plan.ContextOptions{
+			Environment: config.Exec.Environment,
+			UserID:      config.Exec.UserID,
+			User:        config.Exec.User,
+			GroupID:     config.Exec.GroupID,
+			Group:       config.Exec.Group,
+			WorkingDir:  config.Exec.WorkingDir,
+		}
+		merged, err := plan.MergeServiceContext(p, config.Exec.Context, overrides)
```
niemeyer, line 111: That's a very clean implementation. Thank you.

internals/plan/plan.go
```diff
@@ -1105,3 +1116,75 @@ func ReadDir(dir string) (*Plan, error) {
...
+		merged.GroupID = &copied
+	}
+	merged.Group = service.Group
+	merged.WorkingDir = service.WorkingDir
+
+	// Merge in fields from the overrides, if set.
+	for k, v := range overrides.Environment {
+		merged.Environment[k] = v
+	}
+	if overrides.UserID != nil {
```
niemeyer, line 1161: A couple of functions would turn all of these into oneliners and make these read like a nice sequence:
```
func copyIntPtr(v *int) *int
func copyStrPtr(v *string) *string
```

niemeyer (approved):
Super nice work, folks.

What follows is a couple of trivial suggestions, and one bug that is unrelated to this PR.

I would also like to brainstorm/bikeshed quickly about the "context" name next Tuesday, but a preemptive +1 on the PR already.

internals/overlord/checkstate/manager.go
benhoyt: Indeed! @flotter Actually fixed this exact bug a couple of weeks ago. It was causing intermittent issues in CI, but of course as you point out it could be a problem in production too.

@flotter Note that in future (perhaps for your servstate fix :-) we should probably use tombs for this rather than wait groups, to be consistent with other parts of Pebble. Pebble already uses tombs extensively elsewhere, so it won't be a new dependency. I should have mentioned that in the code review.

internals/plan/plan.go
benhoyt: Good call for `copyIntPtr` -- done. We don't need `copyStrPtr` as the strings aren't actually pointers (empty string means not set). Also note that we still need to have the `if overrides.UserID != nil` tests, because you only want to set the merged version at all if the override is non-nil. However, the helper function still saves an awkward line of code in each place, so I've added it.

benhoyt:
Code review changes address. This is done now, but I'll wait till after Tuesday to merge in case we decide on a different name than "context" during the bikeshed.

benhoyt:
Updated "context" to "service-context" per discussion.

PR #247, opened 2023-06-26
daemon: FakeCommand usage requires reaper.Start()

flotter:
The usage of testutil.FakeCommand requires the test environment to start and stop the reaper in the case where the reaper option is enabled.

For example, in internals/daemon/daemon_test.go:

:
cmd := testutil.FakeCommand(c, "shutdown", "", true) :

However, in daemon_test.go, the code is currently relying on the service manager, provided by the overlord, to start the reaper. This is not a safe solution as not all test implementations may actually run the real overlord code, and even if they do, we have a potential race condition.

daemon.Init() -> overlord.New() -> servstate.NewManager() -> reaper.Start()

- Add reaper.Start() and reaper.Stop() to the daemon test setup and teardown.

- Add a reaper based test for testutil.FakeCommand().

### DIFF
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -37,6 +37,7 @@ import (
 	"github.com/canonical/pebble/internals/overlord/restart"
 	"github.com/canonical/pebble/internals/overlord/standby"
 	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/reaper"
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Looks fine to me, thanks

anpep (approved):

jnsgruk (approved):
Nice, thanks for the analysis on this Fred. Going ahead with merge as this is just an improvement to the test suite.

PR #248, opened 2023-06-26
cmd: add detection for containers and pid1

flotter:
The system manager currently makes some assumptions about the environment it is running in.

For example, it assumes that a shutdown program is available in userspace and accessible with PATH configured appropriately.

internals/daemon/daemon.go:
:
cmd := exec.Command("shutdown", "-r", ...
:

This patch adds two detection mechanisms that will allow code to make environment specific decisions in the future (not part of this patch):

- cmd.IsConfined() returns true if running inside a container runtime

- cmd.IsInit() returns true if the system manager was started as PID 1

In addition, the overlord code currently disables reboot failure detection if the system manager is running as PID 1. However, this change is only required for container runtimes, and not generically.

- Update the boot id workaround code to only apply for container runtimes.


cmd/version.go
```diff
@@ -16,11 +16,64 @@ package cmd
...
 
-func MockVersion(version string) (restore func()) {
-	old := Version
-	Version = version
-	return func() { Version = old }
+var (
+	// Version will be overwritten at build-time via mkversion.sh
+	Version = "unknown"
+
+	pid2ProcPath = "/proc/2/status"
```
hpidcock, line 31: This file is called "version.go". It should remain unchanged, all this should go into its own file.

cmd/version.go
```diff
@@ -16,11 +16,64 @@ package cmd
...
+// Containerised returns true if we are running inside a container runtime
+// such as lxd or Docker. The detection is only intended for Linux host
+// environments as it looks for the Linux kernel kthreadd process, started
+// by the Linux kernel after init. Kernel initialised processes do not have
+// a parent as its not a child of init (PID1) and as a result has its ppid
+// set to zero. We use this property to detect the absence of a container
+// runtime. If /proc is not mounted, the detection will assume its a
+// container, so make sure early mounts are completed before running this
+// check on an unconfined host.
+func Containerised() bool {
```
hpidcock, line 47: Let's refactor this function to breakout the detection logic from the synchronisation logic so that it can be tested properly.

```
func Containerised() bool {
	containerOnce.Do(func() {
		containerRuntime, _ = detectContainerised(pid2ProcPath)
	})
	return containerRuntime
}

func detectContainerised(procPath string) (bool, error) {
	s, err := ioutil.ReadFile(procPath);
	if err != nil {
		return true, err
	}
	lines := strings.Split(string(s), "\n")
	for _, l := range lines {
		kv := strings.Split(l, "\t")
		if zeroPPid(kv) {
			return false, nil
		}
	}
	return true, nil
}
```

cmd/version.go
```diff
@@ -16,11 +16,64 @@ package cmd
...
+		}
+	})
+
+	return containerRuntime
+}
+
+// zeroPPid returns true if a PPid key with a zero value was found,
+// otherwise false.
+func zeroPPid(kv []string) (found bool) {
+	if kv[0] == "PPid:" {
```
hpidcock, line 67: need a len check here

cmd/version.go
```diff
@@ -16,11 +16,64 @@ package cmd
...
+		if ppid, err := strconv.Atoi(kv[1]); err == nil && ppid == 0 {
```
hpidcock, line 68: should return errors for tests

cmd/export_test.go
```diff
@@ -0,0 +1,50 @@
...
+// has run.
+func MockVersion(version string) (restore func()) {
+	old := Version
+	Version = version
+	return func() { Version = old }
+}
+
+// ResetContainerInit forces the container runtime check
+// to retry with globals reset
+func ResetContainerInit() {
```
hpidcock, line 47: We don't need this if we seperate out the logic from the variable/sync

cmd/export_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+package cmd
+
+import (
+	"sync"
+)
+
+// MockPid2ProcPath assigns a temporary path to where the PID2
+// status can be found.
+func MockPid2ProcPath(path string) (restore func()) {
```
hpidcock, line 23: Same here, if we allow the implementation to be tested directly, we don't need this.

hpidcock (changes requested):
A few comments but otherwise looks ok.

cmd/version.go
```diff
@@ -16,11 +16,64 @@ package cmd
...
+// environments as it looks for the Linux kernel kthreadd process, started
+// by the Linux kernel after init. Kernel initialised processes do not have
+// a parent as its not a child of init (PID1) and as a result has its ppid
+// set to zero. We use this property to detect the absence of a container
+// runtime. If /proc is not mounted, the detection will assume its a
+// container, so make sure early mounts are completed before running this
+// check on an unconfined host.
+func Containerised() bool {
+	containerOnce.Do(func() {
+		if s, err := ioutil.ReadFile(pid2ProcPath); err == nil {
```
cjdcordeiro, line 49: I am not 100% sure this would always work. Here's an example:

```bash
$ docker run --rm --pid host ubuntu sh -c "[ -d /proc/2 ] && ps -AfwT" | head -3
UID          PID    SPID    PPID  C STIME TTY          TIME CMD
root           1       1       0  0 Jun23 ?        00:09:13 /sbin/init
root           2       2       0  0 Jun23 ?        00:00:00 [kthreadd]
```

Here are some other reasons that make me think this might not be a universal rule:
 - the fact that containers have a PID 1 and then other processes with PID >2 is due to the container runtime's way of managing the container. Afaiu, this implementation may differ from runtime to runtime, so, can we be absolutely sure that we won't ever have a PID 2 inside a container (today or in the future)?
 - some runtimes allow the deployment-time injection of an init process into the containers. I've tried this with Docker (`docker run --init --rm -it ubuntu`), and PID 2 was still not present...but could it be though? Could the introduction of init systems into the container, by the container runtime, void this assessment?

flotter: @cjdcordeiro Thanks for this! Let me look into this. What did you do to be able to see the real PID2 inside the container ?

woky (commented):
Nifty! It seems there's no more straightforward way to from a child PID namespace that a process is in a child PID namespace. Besides comments by Harry, LGTM.

cmd/version.go
woky: We look for `/.dockerenv` [1] and `/run/.containerenv` [2] instead of or in addition to checking for kernel threads. The latter is apparently not stable interface and may disappear but it's still being created and scripts depend on it, e.g. remediation scripts generated by OpenSCAP.

[1] https://superuser.com/questions/1021834/what-are-dockerenv-and-dockerinit
[2] https://docs.podman.io/en/latest/markdown/podman-run.1.html

flotter: I have a feeling the proper answer is no single solution is bulletproof. 

http://git.annexia.org/?p=virt-what.git;a=blob;f=virt-what.in;h=896bfd8b3cb40e8c265597c33eec5494e6fcfd49;hb=HEAD

@cjdcordeiro @woky I wonder if we should start by covering a list of cases from here, and then keep it up to date as technologies change. Anything we use is bound to stop working at some point, so I am thinking we should have a list of checks to detect containers, otherwise we assume its hardware/vm?

flotter: (above is the red-hat virt-what source)

I will respin this PR with the latest feedback, breaking out this container detection into its own file, and structure it so that checks can easily be added.

woky: > Anything we use is bound to stop working at some point, so I am thinking we should have a list of checks to detect containers, otherwise we assume its hardware/vm?

Sounds good. The isolation is too good to be detected reliably. :-)

cmd/version.go
flotter: Resolved. Placed in ```detect.go```.

cmd/version.go
flotter: Split it out, much better, thanks.

cmd/version.go
flotter: Updated, everything can not return an error where it makes sense.

cmd/version.go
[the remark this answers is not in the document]
flotter: Changed, thanks.

cmd/version.go
flotter: New proposal to follow soon. Resolving this for now.

cmd/version.go
flotter: Fixed, although code moved around a bit.

flotter:
Sorry people for talking your time for reviews on this, but it was agreed the scope at which I was trying to solve my own little problem is too inflated here, so I am going to have to close this for now. Perhaps this will be needed again.

The next person to find this closed PR with a hope to detect container runtimes / VMs - have a look here:

[virt-what from Redhat](https://git.annexia.org/?p=virt-what.git;a=blob;f=virt-what.in;hb=HEAD)

[systemd-detect-virt from Systemd](https://github.com/systemd/systemd/blob/main/src/basic/virt.c)

Issue #224, opened 2023-05-19
Intermittent failure in servstate tests


barrettj12:
```
panic: internal error: reaper must be started

goroutine 275 [running]:
github.com/canonical/pebble/internal/reaper.WaitCommand(0xc000292b00, 0xc000044708, 0x7f2824, 0xb8c1e8)
	/home/runner/work/pebble/pebble/internal/reaper/reaper.go:192 +0x794
github.com/canonical/pebble/internal/overlord/servstate.(*serviceData).startInternal.func2(0xc000292b00, 0xc000353a10, 0x5, 0xc0000830e0, 0xc0000ddb80)
	/home/runner/work/pebble/pebble/internal/overlord/servstate/handlers.go:414 +0x51
created by github.com/canonical/pebble/internal/overlord/servstate.(*serviceData).startInternal
	/home/runner/work/pebble/pebble/internal/overlord/servstate/handlers.go:413 +0x111e
FAIL	github.com/canonical/pebble/internal/overlord/servstate	2.850s
```

See the full test logs [here](https://github.com/canonical/pebble/actions/runs/5020872866/jobs/9002731241?pr=221).

I can't quite tell from the logs which test is failing.

barrettj12:
Still observing this [here](https://github.com/canonical/pebble/actions/runs/5459284932/jobs/9935129193?pr=218).

flotter:
@barrettj12 Hi Jordan, yes, I will tackle servstate next ...

flotter:
https://github.com/canonical/pebble/pull/266 Should now fix this.

PR #249, opened 2023-06-27
checkstate: make tests more robust

flotter:
**checkstate: make check runners exit cleanly**

The checks manager uses contexts to cancel check runners. This happens
during checkstate.PlanChanged() when the previous checkers are all
terminated, and the new checker definition is applied.

However, context cancellation is asynchronous (as highlighted in the
context documentation), so the check runners will continue to complete
even after the context cancellations returned.

Due to this race condition, the runner can still launch a check even after
the context cancellation is complete.

This problem is reproduced here:
https://gist.github.com/flotter/f6bf32d7cc33cb1d53a3ddc39fc3b64c

The result of this is easily seen the following unit test failure:

panic: internal error: reaper must be started

goroutine 75 [running]:
github.com/canonical/pebble/internals/reaper.StartCommand
	/home/flotter/pebble/internals/reaper/reaper.go:164
github.com/canonical/pebble/internals/overlord/checkstate.(*execChecker).check()
	/home/flotter/pebble/internals/overlord/checkstate/checkers.go:170
github.com/canonical/pebble/internals/overlord/checkstate.(*checkData).runCheck()
	/home/flotter/pebble/internals/overlord/checkstate/manager.go:199
github.com/canonical/pebble/internals/overlord/checkstate.(*checkData).loop()
	/home/flotter/pebble/internals/overlord/checkstate/manager.go:182
created by github.com/canonical/pebble/internals/overlord/checkstate.(*CheckManager).PlanChanged
	/home/flotter/pebble/internals/overlord/checkstate/manager.go:74

In this particular case, the problem is that test teardown stops the reaper
process which is responsible for reaping check commands, while the final
(edge case) run of the check is still completing.

Following the same logic, this could also cause check overlap between terminating
previous checks and starting new checks in checkstate.PlanChanged().

The following changes fix the problem:

- Use a WaitGroup to represent the group of active checks in the check manager

- Wait() on the group to complete before creating the next group of checks
  when PlanChanged is called.

- Each check runner will use a deferred exit to decrement the group counter.

**checkstate: make tests more robust**

The following changes are made:

- Remove a check which depends on check output in TestTimeout as
  there is no way to remove the race between the 'echo FOO'
  completing before the timeout is triggered, without significant rework.

  This failure can easily be reproduced by running the test under
  heavy cpu load (i.e. stress --cpu 10 --io 4)

  manager_test.go:140:
      c.Assert(check.ErrorDetails, Equals, "FOO")
  ... obtained string = ""
  ... expected string = "FOO"

- Make waitCheck() more conservative when waiting for a check
  runner iteration to complete. The worse case timeout can
  safely be much longer as this does not have a general impact
  on test duration.

  This failure can easily be reproduced by running the test under
  heavy cpu load (i.e. stress --cpu 10 --io 4)

  manager_test.go:134:
      check := waitCheck(c, mgr, "chk1", func(check *CheckInfo) bool {
          return check.Status != CheckStatusUp
      })
  manager_test.go:283:
      c.Fatalf("timed out waiting for check %q", name)
  ... Error: timed out waiting for check "chk1"

### DIFF
--- a/internals/overlord/checkstate/manager.go
+++ b/internals/overlord/checkstate/manager.go
@@ -27,6 +27,7 @@ import (
 // CheckManager starts and manages the health checks.
 type CheckManager struct {
 	mutex           sync.Mutex
+	group           sync.WaitGroup
 	checks          map[string]*checkData
 	failureHandlers []FailureFunc
 }
@@ -58,6 +59,12 @@ func (m *CheckManager) PlanChanged(p *plan.Plan) {
 	for _, check := range m.checks {
 		check.cancel()
 	}
+	// Wait for all context cancellations to propagate and allow
+	// each goroutine to cleanly exit.
+	m.group.Wait()
+
+	// Set the size of the next wait group
+	m.group.Add(len(p.Checks))
 
 	// Then configure and start new checks.
 	checks := make(map[string]*checkData, len(p.Checks))
@@ -65,6 +72,7 @@ func (m *CheckManager) PlanChanged(p *plan.Plan) {
 		ctx, cancel := context.WithCancel(context.Background())
 		check := &checkData{
 			config:  config,
+			group:   &m.group,
 			checker: newChecker(config),
 			ctx:     ctx,
 			cancel:  cancel,
@@ -155,6 +163,7 @@ const (
 // checkData holds state for an active health check.
 type checkData struct {
 	config  *plan.Check
+	group   *sync.WaitGroup
 	checker checker
 	ctx     context.Context
 	cancel  context.CancelFunc
@@ -171,6 +180,10 @@ type checker interface {
 }
 
 func (c *checkData) loop() {
+	// Schedule a notification on exit to indicate another
+	// checker in the group is complete.
+	defer c.group.Done()
+
 	logger.Debugf("Check %q starting with period %v", c.config.Name, c.config.Period.Value)
 
 	ticker := time.NewTicker(c.config.Period.Value)
--- a/internals/overlord/checkstate/manager.go
+++ b/internals/overlord/checkstate/manager.go
@@ -27,7 +27,7 @@ import (
 // CheckManager starts and manages the health checks.
 type CheckManager struct {
 	mutex           sync.Mutex
-	group           sync.WaitGroup
+	wg              sync.WaitGroup
 	checks          map[string]*checkData
 	failureHandlers []FailureFunc
 }
@@ -61,10 +61,10 @@ func (m *CheckManager) PlanChanged(p *plan.Plan) {
 	}
 	// Wait for all context cancellations to propagate and allow
 	// each goroutine to cleanly exit.
-	m.group.Wait()
+	m.wg.Wait()
 
 	// Set the size of the next wait group
-	m.group.Add(len(p.Checks))
+	m.wg.Add(len(p.Checks))
 
 	// Then configure and start new checks.
 	checks := make(map[string]*checkData, len(p.Checks))
@@ -72,14 +72,16 @@ func (m *CheckManager) PlanChanged(p *plan.Plan) {
 		ctx, cancel := context.WithCancel(context.Background())
 		check := &checkData{
 			config:  config,
-			group:   &m.group,
 			checker: newChecker(config),
 			ctx:     ctx,
 			cancel:  cancel,
 			action:  m.callFailureHandlers,
 		}
 		checks[name] = check
-		go check.loop()
+		go func() {
+			defer m.wg.Done()
+			check.loop()
+		}()
 	}
 	m.checks = checks
 }
@@ -163,7 +165,6 @@ const (
 // checkData holds state for an active health check.
 type checkData struct {
 	config  *plan.Check
-	group   *sync.WaitGroup
 	checker checker
 	ctx     context.Context
 	cancel  context.CancelFunc
@@ -180,10 +181,6 @@ type checker interface {
 }
 
 func (c *checkData) loop() {
-	// Schedule a notification on exit to indicate another
-	// checker in the group is complete.
-	defer c.group.Done()
-
 	logger.Debugf("Check %q starting with period %v", c.config.Name, c.config.Period.Value)
 
 	ticker := time.NewTicker(c.config.Period.Value)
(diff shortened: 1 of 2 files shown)


internals/overlord/checkstate/manager_test.go
```diff
@@ -45,12 +45,14 @@ func (s *ManagerSuite) SetUpSuite(c *C) {
 	setLoggerOnce.Do(func() {
 		logger.SetLogger(logger.New(os.Stderr, "[test] "))
 	})
+}
```
flotter, line 48: Moved reaper.Start() and reaper.Stop() to test level. The command cleanup should happen before the next test starts, and it also provides a more robust test environment for race conditions, as now each test must cleanly exit before the reaper is stopped.

Note that the reaper is bound to the package level test binary process ID, and therefore moving it to the test setup and teardown make it incompatible for parallel testing. However, parallel testing is strictly opt-in, so this is still a valid requirement for this particular package tests.

internals/overlord/checkstate/manager_test.go
```diff
@@ -161,17 +162,15 @@ func (s *ManagerSuite) TestCheckCanceled(c *C) {
 		},
 	})
 
-	// Wait for command to start (output file grows in size)
-	prevSize := 0
+	// Wait for command to start (output file is not zero in size)
```
flotter, line 165: Previous logic did not make 100% sense, so I simplified it a bit.

c.MkDir() is test managed and removed after the test. The tempfile is therefore always zero to start with, and so we can simply check for a non-zero value. In the original code, the ```prevSize = len(b)``` appeared redundant as the previous check would break out before it could happen.

internals/overlord/checkstate/manager.go
```diff
@@ -27,6 +27,7 @@ import (
 // CheckManager starts and manages the health checks.
 type CheckManager struct {
 	mutex           sync.Mutex
+	group           sync.WaitGroup
```
benhoyt, line 30: Naming nit: this name begs the question, "group of what?" These are often called `wg` by Gophers, or just `waitGroup` if you prefer to spell it out.

internals/overlord/checkstate/manager.go
```diff
@@ -58,13 +59,20 @@ func (m *CheckManager) PlanChanged(p *plan.Plan) {
...
+	// Set the size of the next wait group
+	m.group.Add(len(p.Checks))
 
 	// Then configure and start new checks.
 	checks := make(map[string]*checkData, len(p.Checks))
 	for name, config := range p.Checks {
 		ctx, cancel := context.WithCancel(context.Background())
 		check := &checkData{
 			config:  config,
+			group:   &m.group,
```
benhoyt, line 75: I think it's a bit simpler to keep the wait group out of the `checkData` struct, so the check itself doesn't have to care about it and all the `WaitGroup` stuff is managed here. You'd do this by saying:

```go
for name, config := range p.Checks {
    ctx, cancel := context.WithCancel(context.Background())
    check := &checkData{...} // no "group" in here anymore
    checks[name] = check
    go func() {
        defer m.group.Done()
        check.loop()
    }()
}
```

internals/overlord/checkstate/manager_test.go
```diff
@@ -269,8 +267,20 @@ func (s *ManagerSuite) TestFailures(c *C) {
 	c.Assert(failureName, Equals, "")
 }
 
+// waitCheck is a time based approach to wait for a checker run to complete.
+// The timeout value does not impact the general time it takes for tests to
+// complete, but determines a worse case waiting period before giving up.
```
benhoyt, line 272: Nit: "worse case" -> "worst case"

internals/overlord/checkstate/manager_test.go
```diff
@@ -269,8 +267,20 @@ func (s *ManagerSuite) TestFailures(c *C) {
...
+// The timeout value must take into account single core or very busy machines
+// so it makes sense to pick a conservative number here as failing a test
+// due to a busy test resource is more extensive than waiting a few more
+// seconds.
 func waitCheck(c *C, mgr *CheckManager, name string, f func(check *CheckInfo) bool) *CheckInfo {
-	for i := 0; i < 100; i++ {
+	// Worse case waiting time for checker run(s) to complete. This
```
benhoyt, line 278: Nit: same as above.

benhoyt (approved):
Thank you! This is really good work and detailed analysis. Much appreciated!

In terms of review: approved, but with a couple of very small suggestions.

anpep (approved):
At last! Thanks for the great work

internals/overlord/checkstate/manager.go
flotter: Changed to ```wg```, thanks.

internals/overlord/checkstate/manager.go
flotter: Wow, crazy cool. Makes a lot of sense.

internals/overlord/checkstate/manager_test.go
flotter: Fixed.

internals/overlord/checkstate/manager_test.go
```diff
@@ -137,7 +139,6 @@ func (s *ManagerSuite) TestTimeout(c *C) {
 	c.Assert(check.Failures, Equals, 1)
 	c.Assert(check.Threshold, Equals, 1)
 	c.Assert(check.LastError, Equals, "exec check timed out")
-	c.Assert(check.ErrorDetails, Equals, "FOO")
```
flotter, line 140: As explained in the commit message, this check is too racey. There is no way there is any guarantee this would be done before the timeout cuts off the command execution. Given that this test focuses on the timeout mechanism, and not command output logging, I feel it is justified to remove this to improve test robustness.

internals/overlord/checkstate/manager_test.go
```diff
@@ -185,7 +184,6 @@ func (s *ManagerSuite) TestCheckCanceled(c *C) {
 	stopChecks(c, mgr)
 
 	// Ensure command was terminated (output file didn't grow in size)
-	time.Sleep(50 * time.Millisecond)
```
flotter, line 188: stopChecks() calls PlanChanged(), which terminates the previous checks synchronously now, so this is not needed.

benhoyt (approved):

benhoyt:
Small stability improvements, no API changes -- going to merge this without further review. We can always tweak in subsequent PRs if needed.

barrettj12:
@flotter I'm seeing a similar intermittent failure in the `servstate` tests - see #224. Maybe a similar fix could help here too?

PR #250, opened 2023-06-29
daemon: add ability to use syscall based reboot

flotter:
Reboot in the daemon is currently relying on userspace having a shutdown command. This may not always be the case, as Pebble is used in a wide variety of environments.

Add a SetRebootMode() function to the daemon to allow the daemon to be initialised with a syscall based reboot implementation. The default reboot mode using systemd remains unchanged.


internals/daemon/daemon.go
```diff
@@ -717,21 +717,55 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
 	return fmt.Errorf("expected reboot did not happen")
 }
 
-var shutdownMsg = "reboot scheduled to update the system"
+var (
+	rebootMsg     = "reboot scheduled to update the system"
```
benhoyt, line 721: Should this be a `const`, or maybe even just a literal string in the one place it's used?

internals/daemon/daemon.go
```diff
@@ -717,21 +717,55 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
 	return nil
 }
 
-var reboot = rebootImpl
+var shutdownSyscall = func() {
+	// As per the requirements of the reboot syscall, we have to
+	// first call sync.
+	unix.Sync()
+	// This syscall can fail (EINVAL) if invalid arguments are
+	// supplied, which should not be the case, but let's panic if
```
benhoyt, line 749: Wouldn't it error if running as a non-root user, too?

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,35 @@ services:
 	c.Assert(tasks, HasLen, 1)
 	c.Check(tasks[0].Kind(), Equals, "stop")
 }
+
+func (s *daemonSuite) TestSyscallRebootDelay(c *C) {
+	waitState := 0
```
benhoyt, line 1153: These are booleans, right? I think you'd normally use `true` or `false` for this.

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,35 @@ services:
...
+	old := shutdownSyscall
+	shutdownSyscall = func() {
+		waitState = 1
+	}
+	defer func() {
+		shutdownSyscall = old
+	}()
+	syscallReboot(time.Millisecond * 25)
+	c.Assert(waitState, Equals, 0)
+	time.Sleep(time.Millisecond * 50)
```
benhoyt, line 1163: There are data races that `go test -race` is showing caused by writing the global vars with `time.Sleep` as the synchronization mechanism. Per your excellent work in the other PR, I think we're going to have to use proper synchronization here. That might mean we have to expose a bit more (a channel?) in the non-test code.

benhoyt: Or maybe you can send on a channel inside the modified `shutdownSyscall`?

internals/daemon/daemon.go
flotter: Changed the the proposed ```const``` for now for minimum change. I think this could be a separate point of discussion, because I am not sure this single global specific message will be solution going forward. It feels better to supply it with the reboot request so that context specific messages can be supplied. For now, I would prefer to keep the scope of this PR strictly to allow ```shutdown```-less reboots, if that is OK.

internals/daemon/daemon.go
flotter: yes, thanks!

Since the reboot happens non-blocking, I decided to do a pre-check for the required capability (CAP_SYS_BOOT) at the start of the request, before we kick of the timed request. This means EPERM is ruled out, which would make the panic associated with the reboot only possible for an argument error, which means a coding bug, which will be easy to rule out.

internals/daemon/daemon_test.go
flotter: Cleaned up. Sorry about the state of the tests - I rushed out the PR for overnight feedback, but the tests needed more attention :(

internals/daemon/daemon_test.go
flotter: Should look much nicer now. Let me know what you think. Sorry for waiting your time here earlier ...

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,98 @@ services:
...
+		return nil
+	})()
+	defer mockShutdownSyscall(func() {
+		wait <- 1
+	})()
+
+	period := time.Millisecond * 25
+	syscallReboot(period)
+	start := time.Now()
+	<-wait
```
benhoyt, line 1180: These kind of waits should probably always be in a `select` with an alternative timeout branch, so the test doesn't hang forever in case of a bug or edge case we've forgotten. Similar below. Something like this:

```go
select {
case <-wait:
case <-time.After(10 * time.Second):
	c.Fail("timed out waiting for X")
}
```

internals/daemon/daemon_test.go
benhoyt: No problemo. Looks good now, except for the note about having a timeout on the waits.

benhoyt (approved):

internals/daemon/daemon_test.go
flotter: Fixed. Thanks for the feedback.

benhoyt (approved):

internals/daemon/daemon.go
```diff
@@ -717,21 +717,79 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
 		return osutil.OutputErr(out, err)
 	}
 	return nil
 }
 
-var reboot = rebootImpl
+var (
+	// checkCapSysBoot returns nil if the system has the correct
+	// permissions to issue a reboot request to the Linux kernel
+	checkCapSysBoot = func() error {
```
niemeyer, line 746: The way we typically do this is to define a real function, and then define a variable that can be overriden, so code inside the export_test.go file can replace it, using a function prefixed with `Fake` and that returns a restoring function named `restore`.

With that said, all of this logic feels like it's adding significant complexity to the code with no benefit, because the test is testing the test code. In other words, all of this allows the test to override the actual complexity of the implementation with mock versions, in a way that simply call two mock functions. If we have actual bugs in this logic, almost certainly it will be in the logic of these two functions, which remains untested, and not in the trivial lines inside the function that calls them. We're pretty much testing that AfterFunc works. Can we please make this whole logic normal?

internals/daemon/daemon.go
```diff
@@ -717,21 +717,79 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
+	if err != nil {
+		return err
+	}
+
+	if rebootDelay < 0 {
+		rebootDelay = 0
+	}
+	// This has to be non-blocking, and scheduled for a future
+	// point in time to mimic shutdown.
+	time.AfterFunc(rebootDelay, shutdownSyscall)
```
niemeyer, line 790: What happens here when the syscall is dispatched?  The shutdown logic will actually ask Pebble to be shutdown, which will terminate the services correctly, I think, while this will just have a sudden death moment ahead?  Might be better to shutdown pebble and call the syscall immediately instead of waiting for death at an arbitrary and harsh point in the future.

internals/daemon/daemon.go
```diff
@@ -28,9 +28,9 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
+	"golang.org/x/sys/unix"
```
niemeyer, line 33: Why the change from syscall into an external package?

internals/daemon/daemon.go
```diff
@@ -717,21 +717,79 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
+	// checkCapSysBoot returns nil if the system has the correct
+	// permissions to issue a reboot request to the Linux kernel
+	checkCapSysBoot = func() error {
+		var caps unix.CapUserData
+		// We deliberately use v1 caps here due to:
+		// https://github.com/golang/go/issues/44312
+		hdr := unix.CapUserHeader{Version: unix.LINUX_CAPABILITY_VERSION_1}
+		err := unix.Capget(&hdr, &caps)
+		if err == nil {
+			if (int32(caps.Effective) & (1 << unix.CAP_SYS_BOOT)) == 0 {
```
niemeyer, line 753: Do we need this logic for the use cases at hand?  Can't we simply try and report the failures?

internals/daemon/daemon.go
```diff
@@ -717,21 +717,79 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
+		// first call sync.
+		unix.Sync()
+		// This syscall can fail (EINVAL/EPERM) if invalid arguments are
+		// supplied or CAP_SYS_BOOT capability is missing. We cover the
+		// latter case in a separate capability check, so this will not
+		// happen here, but let's panic if we see something to make sure
+		// we catch anything unexpected.
+		err := unix.Reboot(unix.LINUX_REBOOT_CMD_RESTART)
+		if err != nil {
+			panic("internal error: reboot syscall failed")
```
niemeyer, line 771: An internal error is something that we did which could never fail unless we, the developers, made a mistake. A syscall can fail outside of our control. So this is not an internal error, and this should probably not be a panic either for similar reasons.

niemeyer (changes requested):
Thanks, Fred. Some notes and questions below.

internals/daemon/daemon.go
anpep: The `syscall` package is deprecated. See https://groups.google.com/g/golang-dev/c/cEASnHIXmLI/m/WUThor2mBwAJ

anpep (changes requested):

internals/daemon/daemon.go
```diff
@@ -717,21 +717,79 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
 	return fmt.Errorf("expected reboot did not happen")
 }
 
-var shutdownMsg = "reboot scheduled to update the system"
+const rebootMsg = "reboot scheduled to update the system"
```
atesburak, line 720: Since we are using this once, can we just use the string literal instead and add the new variable if/when we need it in the future?

internals/daemon/daemon.go
flotter: I've decided to revert to syscall, as the entire Pebble is using this package. I think a clean migration as a separate exercise would be better, once we are ready to migrate.

internals/daemon/daemon.go
flotter: As discussed, this check is removed, and reboot syscall error logged (```logger.Notice```).

internals/daemon/daemon.go
flotter: As agreed, this no longer panics, and gets logged to ```logger.Notice```.

internals/daemon/daemon.go
```diff
@@ -717,21 +717,79 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
+			panic("internal error: reboot syscall failed")
+		}
+	}
+)
+
+// syscallReboot performs a reboot using direct Linux kernel syscalls.
+//
+// Note: Reboot message not currently supported.
+func syscallReboot(rebootDelay time.Duration) error {
+	err := checkCapSysBoot()
```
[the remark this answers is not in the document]
flotter: Code removed.

internals/daemon/daemon.go
flotter: I am going to limit the scope of the changes to an absolute minimum. Your proposal makes sense, but this message should be properly considered in the bigger context, which I think is more suited for a separate exercise.

benhoyt (commented):
Looks fine overall to me, just a few really nitty comments.

internals/daemon/daemon.go
```diff
@@ -717,21 +717,55 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
+	}
+	// This has to be non-blocking, and scheduled for a future
+	// point in time to mimic shutdown.
+	time.AfterFunc(rebootDelay, func() {
+		// As per the requirements of the reboot syscall, we
+		// have to first call sync.
+		syncSyscall()
+		err := rebootSyscall(syscall.LINUX_REBOOT_CMD_RESTART)
+		if err != nil {
+			logger.Noticef("reboot syscall failed : %v", err)
```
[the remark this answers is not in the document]
flotter: I changed this to, starting with a verb and removing the space.

"Failed on reboot syscall: %v"

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,139 @@ services:
...
+	wait := make(chan int)
+	defer mockSyncSyscall(func() {})()
+	defer mockRebootSyscall(func(cmd int) error {
+		if cmd == syscall.LINUX_REBOOT_CMD_RESTART {
+			wait <- 1
+		}
+		return nil
+	})()
+
+	period := time.Millisecond * 25
```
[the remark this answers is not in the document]
flotter: Fixed everywhere in my PR. thanks.

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,139 @@ services:
...
+	timeout := time.Second * 10
```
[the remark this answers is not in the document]
flotter: Removed, and code simplified with a better failure description.

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,139 @@ services:
...
+	syscallReboot(period)
+	start := time.Now()
+	select {
+	case <-wait:
+	case <-time.After(timeout): // exit test if we fail and get stuck
+		c.Fail()
+	}
+	elapse := time.Now().Sub(start)
```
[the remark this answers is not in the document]
flotter: Fixed, thanks.

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,139 @@ services:
...
+	SetSyscallReboot()
+	defer func() {
+		rebootHandler = commandReboot
+	}()
+	complete := make(chan int)
+	l := fakeLogger{c: complete}
+	old := logger.SetLogger(&l)
+	defer func() {
+		logger.SetLogger(old)
+	}()
```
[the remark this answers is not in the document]
flotter: Done. thanks.

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,139 @@ services:
...
+	select {
+	case <-wait:
+	case <-time.After(timeout): // exit test if we fail and get stuck
+		c.Fail()
+	}
+}
+
+type fakeLogger struct {
+	s string
+	c chan int
```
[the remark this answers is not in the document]
flotter: Improved.

internals/daemon/daemon_test.go
```diff
@@ -1148,3 +1148,139 @@ services:
...
+
+func mockRebootSyscall(f func(cmd int) error) (restore func()) {
+	old := rebootSyscall
+	rebootSyscall = f
+	return func() {
+		rebootSyscall = old
+	}
+}
+
+func (s *daemonSuite) TestSyscallPosRebootDelay(c *C) {
```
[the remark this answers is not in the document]
flotter: Done.

internals/daemon/daemon.go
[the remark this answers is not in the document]
benhoyt: That's better, though I actually meant we specifically decided on the word "Cannot" (it's not my favourite, but at least we settled on something for consistency).

anpep (approved):

anpep (approved):

internals/daemon/daemon_test.go
```diff
@@ -1176,3 +1177,136 @@ services:
 	c.Assert(tasks, HasLen, 1)
 	c.Check(tasks[0].Kind(), Equals, "stop")
 }
+
+type rebootSuite struct{}
+
+var _ = Suite(&rebootSuite{})
+
+func mockSyncSyscall(f func()) (restore func()) {
```
flotter, line 1185: mock -> fake

internals/daemon/daemon.go
```diff
@@ -710,21 +710,55 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
 
-var shutdownMsg = "reboot scheduled to update the system"
+const rebootMsg = "reboot scheduled to update the system"
 
-func rebootImpl(rebootDelay time.Duration) error {
+var rebootHandler = commandReboot
+
+// SetSyscallReboot replaces the default command-based reboot
+// with a direct Linux kernel syscall based implementation.
+func SetSyscallReboot() {
```
niemeyer, line 719: How about something like this:
```mode 
type RebootMode int

var (
    SystemdReboot RebootMode = iota + 1
    SyscallReboot
)

func SetRebootMode(mode RebootMode) { ... }
```

internals/daemon/daemon.go
```diff
@@ -710,21 +710,55 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
 	if out, err := cmd.CombinedOutput(); err != nil {
 		return osutil.OutputErr(out, err)
 	}
 	return nil
 }
 
-var reboot = rebootImpl
+var (
+	syncSyscall   = syscall.Sync
+	rebootSyscall = syscall.Reboot
```
niemeyer, line 738: IIRC, typically we use the package name as the prefix in such cases, as it becomes slightly more natural:
```
syscallSync
syscallReboot
```

internals/daemon/daemon_test.go
niemeyer: Yeah, and should be made "public" in the export_test.go package (FakeSyncSyscall...)

flotter: Fixed.

internals/daemon/daemon.go
flotter: Done, with minor tweaks.

internals/daemon/daemon.go
flotter: Done. It required me to reshuffle a few things because the name was used.

internals/daemon/daemon.go
```diff
@@ -710,21 +710,55 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
...
+)
+
+// syscallReboot performs a delayed async reboot using direct Linux
+// kernel syscalls.
+//
+// Note: Reboot message not currently supported.
+func syscallReboot(rebootDelay time.Duration) error {
```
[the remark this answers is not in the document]
flotter: Changed the two reboot functions to:

syscallModeReboot and systemdModeReboot.

PR #252, opened 2023-06-30
[JUJU-3777] Multiplex services to log targets using only a `log-target.services` field

barrettj12:
#223 was proving too complex, so we've come up with an alternative way to specify which services each log target should collect logs from.

The log target spec now has a `services` field, which accepts a list of services to collect log targets from.
```yaml
log-targets:
  tgt1:
    services: [svc1, svc2]
```

This replaces the old `log-target.selection` field, which has now been removed. The `services` field is also much more expressive than `selection`, so it also removes the need for the `service.log-targets` field. **Effectively, `log-target.services` is now the only place to specify which services' logs go to which log targets.**

Some bonus features:
- Use the `all` keyword to match all services (including future services):
  ```yaml
  tgt1:
    services: [all]
  ```

- Use `-svc` to remove a service from the list. This is especially useful when merging or after using `all`.
  ```yaml
  tgt1:
    services: [all, -svc1, -svc3]
  ```
  matches all services except `svc1` and `svc3`.

- Use `-all` to remove all service from the list. Again, useful when merging or after using `all`.
  ```yaml
  tgt1:
    services: [all, -svc3, -all]
  ```
  matches no services.

The `services` field is simply appended when merging. In combination with the above special syntaxes, this allows "replace" behaviour in a merge:
- To remove service(s) specified in a lower layer, you can just merge on top
  ```yaml
  tgt1:
    services: [-svc2, -svc4]
    override: merge
  ```

- To remove all services for a log target (i.e. opting out of log forwarding entirely), just merge on top
  ```yaml
  tgt1:
    services: [-all]
    override: merge
  ```

### Context / links

- Original log forwarding PR (closed w/o merging): #165
- Original log forwarding spec (not yet updated): [JU052](https://docs.google.com/document/d/12tJFhIbVIlOcblG6l9_u-K5bCib78lojRJ-hdGBxSsw/edit)
- Merged PR containing plan changes for log forwarding: #209
  - This PR essentially is a slight modification / follow-up to that PR.
- Rejected PRs trying to solve similar problems:
  - #217
  - #221
  - #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,7 +439,7 @@ my-target:
     services: [-svc1]
     override: merge
 ```
-then in the merged layer, `my-target` will collect logs from only `svc2`. You can also use `-all` to remove all services from the list.
+then in the merged layer, the `services` list will be merged to `[svc1, svc2, -svc1]`, which evaluates left to right as simply `[svc2]`. So `my-target` will collect logs from only `svc2`. You can also use `-all` to remove all services from the list.
 -->
 
 ## Container usage
--- a/README.md
+++ b/README.md
@@ -403,6 +403,45 @@ $ pebble run --verbose
 ...
 ```
 
+<!--
+TODO: uncomment this section once log forwarding is fully implemented
+TODO: add log targets to the Pebble layer spec below
+
+#### Log forwarding
+
+Pebble supports forwarding its services' logs to a remote Loki server or syslog receiver (via UDP/TCP). In the `log-targets` section of the plan, you can specify destinations for log forwarding, for example:
+```yaml
+log-targets:
+    loki-example:
+        override: merge
+        type: loki
+        location: http://10.1.77.205:3100/loki/api/v1/push
+        services: [all]
+    syslog-example:
+        override: merge
+        type: syslog
+        location: tcp://192.168.10.241:1514
+        services: [svc1, svc2]
+```
+
+For each log target, use the `services` key to specify a list of services to collect logs from. In the above example, the `syslog-example` target will collect logs from `svc1` and `svc2`.
+
+Use the special keyword `all` to match all services. In the above example, `loki-example` will collect logs from all services.
+
+To remove a service from a log target when merging, prefix the service name with a minus `-`. For example, if we have a base layer with
+```yaml
+my-target:
+    services: [svc1, svc2]
+```
+and override layer with
+```yaml
+my-target:
+    services: [-svc1]
+    override: merge
+```
+then in the merged layer, `my-target` will collect logs from only `svc2`. You can also use `-all` to remove all services from the list.
+-->
+
 ## Container usage
 
 Pebble works well as a local service manager, but if running Pebble in a separate container, you can use the exec and file management APIs to coordinate with the remote system over the shared unix socket.
--- a/README.md
+++ b/README.md
@@ -426,7 +426,7 @@ log-targets:
 
 For each log target, use the `services` key to specify a list of services to collect logs from. In the above example, the `syslog-example` target will collect logs from `svc1` and `svc2`.
 
-Use the special keyword `all` to match all services. In the above example, `loki-example` will collect logs from all services.
+Use the special keyword `all` to match all services, including services that might be added in future layers. In the above example, `loki-example` will collect logs from all services.
 
 To remove a service from a log target when merging, prefix the service name with a minus `-`. For example, if we have a base layer with
 ```yaml
@@ -439,7 +439,22 @@ my-target:
     services: [-svc1]
     override: merge
 ```
-then in the merged layer, the `services` list will be merged to `[svc1, svc2, -svc1]`, which evaluates left to right as simply `[svc2]`. So `my-target` will collect logs from only `svc2`. You can also use `-all` to remove all services from the list.
+then in the merged layer, the `services` list will be merged to `[svc1, svc2, -svc1]`, which evaluates left to right as simply `[svc2]`. So `my-target` will collect logs from only `svc2`.
+
+You can also use `-all` to remove all services from the list. For example, adding an override layer with
+```yaml
+my-target:
+    services: [-all]
+    override: merge
+```
+would remove all services from `my-target`, effectively disabling `my-target`. Meanwhile, adding an override layer with
+```yaml
+my-target:
+    services: [-all, svc1]
+    override: merge
+```
+would remove all services and then add `svc1`, so `my-target` would receive logs from only `svc1`.
+
 -->
 
 ## Container usage
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -780,7 +780,7 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 			}
 		}
 
-		if target.Location == "" && len(target.Services) > 0 {
+		if target.Location == "" {
 			return nil, &FormatError{
 				Message: fmt.Sprintf(`plan must define "location" for log target %q`, name),
 			}
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -523,8 +523,6 @@ func (t *LogTarget) Merge(other *LogTarget) {
 	if other.Location != "" {
 		t.Location = other.Location
 	}
-	// TODO: reduce log targets?
-	// e.g. [..., all, -svc1] -> [all, -svc1]
 	t.Services = append(t.Services, other.Services...)
 }
 
@@ -774,6 +772,7 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 			if _, ok := combined.Services[serviceName]; ok {
 				continue
 			}
+			// Check if this is `-svc`
 			serviceName = strings.TrimPrefix(serviceName, "-")
 			if _, ok := combined.Services[serviceName]; ok {
 				continue
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -255,7 +255,21 @@ func CommandString(base, extra []string) string {
 
 // LogsTo returns true if the logs from s should be forwarded to target t.
 func (s *Service) LogsTo(t *LogTarget) bool {
-	// TODO: implement
+	// Iterate backwards through t.Services until we find something matching
+	// s.Name.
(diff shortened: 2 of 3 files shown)


internals/plan/plan.go
```diff
@@ -553,9 +523,7 @@ func (t *LogTarget) Merge(other *LogTarget) {
 	if other.Location != "" {
 		t.Location = other.Location
 	}
-	if other.Selection != "" {
-		t.Selection = other.Selection
-	}
+	t.Services = append(t.Services, other.Services...)
```
barrettj12, line 526: I'm not sure if we should "reduce" the services list when merging. For example,
```
[svc1, svc2]  <--  [-svc1]       =  [svc1, svc2, -svc1]       ==>  [svc2]
[svc1, svc2]  <--  [all, -svc3]  =  [svc1, svc2, all, -svc3]  ==>  [all, -svc3]
[svc1, svc2]  <--  [-all, svc3]  =  [svc1, svc2, -all, svc3]  ==>  [svc3]
```

The way I've implemented `LogsTo`, it's not *necessary* to do this, but it might help to keep things more manageable (e.g. if we have to remarshal due to the user running `pebble plan`). It will complicate the logic though.

README.md
```diff
@@ -403,6 +403,45 @@ $ pebble run --verbose
...
+my-target:
+    services: [svc1, svc2]
+```
+and override layer with
+```yaml
+my-target:
+    services: [-svc1]
+    override: merge
+```
+then in the merged layer, `my-target` will collect logs from only `svc2`. You can also use `-all` to remove all services from the list.
```
benhoyt, line 442: Perhaps spell this out slightly:

```suggestion
then in the merged layer, the `services` list will be merged to `[svc1, svc2, -svc1]`, which evaluates left to right as simply `[svc2]`. So `my-target` will collect logs from only `svc2`. You can also use `-all` to remove all services from the list.
```

internals/plan/plan.go
```diff
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/canonical/pebble/internals/logger"
```
benhoyt, line 29: Nit: import ordering - this should go in the `canonical/pebble` section below.

internals/plan/plan.go
```diff
@@ -275,23 +255,22 @@ func CommandString(base, extra []string) string {
...
-//   - t.Selection is "opt-out" or empty, and s.LogTargets is empty; or
-//   - t.Selection is not "disabled", and s.LogTargets contains t.
 func (s *Service) LogsTo(t *LogTarget) bool {
-	if t.Selection == DisabledSelection {
-		return false
-	}
-	if len(s.LogTargets) == 0 {
-		if t.Selection == UnsetSelection || t.Selection == OptOutSelection {
+	// Iterate backwards through t.Services until we find something matching
+	// s.Name.
```
benhoyt, line 260: This is a really clever way to do it. I *think* I've convinced myself it has the same effect as keeping a set and evaluating left to right...

internals/plan/plan.go
benhoyt: I think what you've done here is the right approach. The field itself behaves according to the standard merge logic, so `pebble plan` should show the full list of merged items (rather than the evaluated result).

Separately we could consider adding a `pebble log-targets` or similar command which would show the status of the log targets: what services they've been evaluated as actually going to, how many logs sent, number of errors, that sort of thing. Future work...

internals/plan/plan.go
```diff
@@ -796,35 +764,32 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+		for _, serviceName := range target.Services {
+			if serviceName == "all" || serviceName == "-all" {
+				continue
+			}
+			if _, ok := combined.Services[serviceName]; ok {
+				continue
+			}
+			// Check if this is `-svc`
+			serviceName = strings.TrimPrefix(serviceName, "-")
+			if _, ok := combined.Services[serviceName]; ok {
```
benhoyt, line 777: It seems to me that only this second `combined.Services` check is needed. First trim any `-` prefix, then check if it's a service name. No need for the first one above.

internals/plan/plan.go
```diff
@@ -796,35 +764,32 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
 			return nil, &FormatError{
-				Message: fmt.Sprintf(`log target %q has invalid selection %q, must be %q, %q or %q`,
-					name, target.Selection, OptOutSelection, OptInSelection, DisabledSelection),
+				Message: fmt.Sprintf(`log target %q specifies unknown service %q`,
+					target.Name, serviceName),
 			}
 		}
 
-		if target.Location == "" && target.Selection != DisabledSelection {
+		if target.Location == "" && len(target.Services) > 0 {
```
benhoyt, line 786: Why do we only validation `location` if services is set? Shouldn't we always do that regardless?

internals/plan/plan.go
barrettj12: `TestLogsTo` should remove any doubt

internals/plan/plan.go
barrettj12: good catch

barrettj12: We also need to disallow service names starting with `-`

internals/plan/plan.go
barrettj12: I think so you can disable a target with
```
tgt1:
  services: []
  override: replace
```

We don't really need `location` if the target is not receiving any logs. I'm open to changing this though.

internals/plan/plan.go
barrettj12: Is it okay to add logging to `ParseLayer`? Is there any reason the plan package didn't include any logging previously?

internals/plan/plan.go
barrettj12: > The field itself behaves according to the standard merge logic, so `pebble plan` should show the full list of merged items

Yes, but that's a **choice** we've made here. I'm wondering if it's the right choice.

It would be equally valid to say "when merging, remove duplicates / reduce the list to a minimal equivalent set". (We previously did this with the `services.log-targets` field). It would behave the same either way, so I guess it's a matter of taste.

internals/plan/plan.go
benhoyt: Yeah, I realize we don't need it, but it just seems weird for the validation of `location` to depend on whether it's going somewhere right at the moment or not.

barrettj12: I guess now you can disable a target by merging this on top
```
tgt1:
  services: [-all]
  override: merge
```
so we can remove this exception to the rule.

internals/plan/plan.go
flotter: I think leaving the merge without evaluating it also help with debugging. If someone does not understand why a particular service disappeared, the complete history is there. If you evaluate it, its gone.

internals/plan/plan.go
```diff
@@ -275,23 +255,22 @@ func CommandString(base, extra []string) string {
...
-//   - t.Selection is not "disabled", and s.LogTargets contains t.
 func (s *Service) LogsTo(t *LogTarget) bool {
-	if t.Selection == DisabledSelection {
-		return false
-	}
-	if len(s.LogTargets) == 0 {
-		if t.Selection == UnsetSelection || t.Selection == OptOutSelection {
+	// Iterate backwards through t.Services until we find something matching
+	// s.Name.
+	for i := len(t.Services) - 1; i >= 0; i-- {
```
hpidcock, line 261: Seems like a reasonable way to do this, considering we won't be calling LogsTo for every log event, only during setup/configuration of the service/log-target.

internals/plan/plan.go
hpidcock: I am happy keeping this simple, we don't need to prematurely optimise here.

hpidcock (approved):
Awesome, glad we were able to come to a simple solution that doesn't really lose much of the functionality we were seeking.

benhoyt (approved):
Excellent, thanks for the updates -- looks good to me.

internals/plan/plan.go
```diff
@@ -796,35 +764,29 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
-		switch target.Selection {
-		case OptOutSelection, OptInSelection, DisabledSelection, UnsetSelection:
-			// valid, continue
-		default:
+		// Validate service names specified in log target
+		for _, serviceName := range target.Services {
+			if serviceName == "all" || serviceName == "-all" {
+				continue
+			}
+			// This could be `-svc` - try to trim a preceding
```
niemeyer, line 772: The sentence looks unfinished, and probably unnecessary as well. TrimPrefix reads just as easily as "try to trim a preceding". The line might be moved above the "all" check as well, so that we don't have to test both.

internals/plan/plan.go
```diff
@@ -963,6 +925,15 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 				Message: fmt.Sprintf("cannot use reserved service name %q", name),
 			}
 		}
+		// Deprecated service names
+		if name == "all" || name == "default" || name == "none" {
+			logger.Noticef("WARNING: using %q as service name is deprecated", name)
```
niemeyer, line 930: As a rule of thumb, error messages are lowercased and use no ending punctuation, while log messages are proper sentences starting with capital case and ending with punctuation.

README.md
```diff
@@ -403,6 +403,45 @@ $ pebble run --verbose
...
+    syslog-example:
+        override: merge
+        type: syslog
+        location: tcp://192.168.10.241:1514
+        services: [svc1, svc2]
+```
+
+For each log target, use the `services` key to specify a list of services to collect logs from. In the above example, the `syslog-example` target will collect logs from `svc1` and `svc2`.
+
+Use the special keyword `all` to match all services. In the above example, `loki-example` will collect logs from all services.
```
niemeyer, line 429: Would be nice to mention here the pattern `[-all, svc1]` as well, and the fact that `[all]` is not the same as listing every service, since this will catch future services too, while listing services individually will not.

niemeyer (approved):
That's much nicer, thanks Jordan. Only trivial comments below, and LGTM.

PR #253, opened 2023-07-04
daemon: remove listener closing race

flotter:
We use http.Server with a number of custom net.Listener instances.

This patch makes some changes to the daemon.Stop(...) function.

In order for the server instance to Shutdown() gracefully, the listener termination must be performed by the shutdown process itself. The server shutdown process is thread-safe, and tracks the listeners in use. Thread safety is important because the http.Server has two places where listeners can be terminated:

1. The thread hosting the Serve() method, using a deferred mutex protected closing function for all registered listeners.

2. The Shutdown() method (with mutex protection), which will close listeners async resuling in an error from the listener Accept(), but this special case is handled in Serve() and returns the ErrServerClosed error which means shutdown was done cleanly.

However, in daemon.go we also pre-close all the listeners before the call to Shutdown(), which causes two kinds of errors:

1. accept tcp [::]:<port>: use of closed network connection

This error is actually masked because of a tomb.Kill() call for the threads dealing with the http.Server. The errors are masked when the tomb state is dying however, in this case the tomb.Kill() is repeated later in the same function again.

2. close tcp [::]:<port>: use of closed network connection

This is the actual race condition that can sometimes be seen in CI logs. This race is caused by the two closing mechanism discussed above because we are bypassing the thread protection and closing the listeners manually first. The race happens when both the Serve() defer and the Shutdown attempts to close the same listener at the same time, with only one of them succeeding, and the other one showing the error above.

A standalone simulation of the problem can be seen here:

https://gist.github.com/flotter/483d233d38f0961a52f7ef7405887012

- Remove the early tomb.Kill(nil) in daemon.Stop(). This has no purpose apart from masking the errors during http.Server shutdown, as none of the tomb managed threads actually use the dying state to exit.

  The only purpose of Kill() in the daemon is to signal to top level level thread to call daemon.Stop(), and to terminate the watchdog thread.

- Remove the early manual listener closing code so its properly done by the Shutdown() call.


internals/daemon/daemon.go
```diff
@@ -561,24 +561,13 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 		logger.Noticef("Cannot stop running services: %v", err)
 	}
 
-	d.tomb.Kill(nil)
```
flotter, line 564: Kill() is used to signal tomb managed goroutines to exit. However, our http.Server goroutines does not use this directly. Instead, its used for suppressing server errors on exit, which masked part of the original issue in the first place. The daemon tomb is actually used at the top level to trigger a call of this function, so removing this early duplicate (its called again later in this function), makes Serve() errors getting properly reported.

The tomb effectively has two uses:

1. Daemon termination signal using the dying state
2. Waiting for http.Server goroutines to exit cleanly (and supressing errors - which can still happen on a reboot request, which uses the Kill() method - see daemon.Start())

The two are very decoupled right now, so in the future I would be tempted to use a separate channel for termination requests.

benhoyt (approved):
Logic sounds reasonable to me. Thanks again for the detailed analysis, @flotter!

benhoyt:
I also pinged @mvo5 on Mattermost, as from a quick glance at snapd's [daemon.go](https://github.com/snapcore/snapd/blob/master/daemon/daemon.go), this same issue applies there.

internals/daemon/daemon.go
flotter: I decided to undo this change, as it would be very difficult to prove this does not break something. Pebble can be terminated from a signal, in which case, Stop() would enter without the http.Server goroutines set do dying. Under this condition, if any connections fail to close, errors will be seen. Without Dying state, those errors will affect the shutdown logic.

niemeyer (changes requested):
> The thread hosting the Serve() method, using a deferred mutex protected closing function for all registered listeners.

Where are these close calls?  Both Serve goroutines are fired in the Start method, and there are no Close calls there?

benhoyt:
@niemeyer The close calls Fred's referring to are the `defer l.Close()` in the net/http `Server.Serve` method. Shutdown closes all the listeners and gracefully waits for things to finish, so we don't need to call Close manually at all.

benhoyt:
@niemeyer's comments from Mattermost for the record:

> I see, thanks for the details. It smells slightly strange as the initialization of these listeners is done separately in Init, and the goroutines are only started inside Start, so the disconnection feels brittle. It's also unclear to me how this logic I think is unchanged from snaps, and yet we don't have issues there. Wonder what happened in between.
>
> Easiest path is probably to just organize things a bit better. I don't see, for instance, what's the value of keeping these listeners around if Serve is going to kill them.

niemeyer (approved):
I think we can just merge this for now.  The interlock between the Init of the Daemon, and the Start/Stop pair, don't feel ideal. For example, if we Init the daemon, we already get open listeners, but to get rid of the open listeners, we need to Stop it.  Except, if we run Init without Stop, it will crash because d.serve is not there yet.  So it does need polishing.

With that said, given the above it also feels like dropping the explicit closing won't make it any worse, since Shutdown is called unconditionally on that sequence. So +1 on just merging this now and polishing it later.

PR #254, opened 2023-07-05
Enforce import order

barrettj12:
Apparently there is an implicit rule that Pebble imports should be arranged into three groups:
- standard library imports
- 3rd-party / non-Pebble imports
- Pebble imports (i.e. those beginning with `github.com/canonical/pebble`)

This rule was not being enforced anywhere except code review, and there were many Go files not following this convention.

I've fixed up all Go files so that they now follow this convention. I used the [`gci`](https://github.com/daixiang0/gci) tool to enforce the import order. Unfortunately `goimports` is not strict enough - it will only make sure that imports from the same group are not together. It will allow a group to be split up, so e.g. the following still passes:
```go
import (
  "github.com/canonical/pebble/internals/plan"

)
```

The easiest way to use `gci` is via `golangci-lint` (which `gci` is bundled inside). Run
```
golangci-lint run -c .github/.golangci.yml
```
to do the linting. If you add the `--fix` flag, `gci` will fix all the imports for you.

While I'm here, I resolved a stale TODO in `daemon_test.go` by changing the import of `gocheck` to use `.`, following the convention observed in (most) other test files.

Finally, I've added a `Lint` GitHub workflow to enforce this import order on future PRs.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          go-mod-file: 'go.mod'
+          cache: false
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          args: '-c .github/.golangci.yml'
+          skip-cache: true
```
benhoyt, line 20: Why are `skip-cache: true` and `cache: false` above required?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-mod-file: 'go.mod'
+          cache: false
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
```
benhoyt, line 16: Probably a silly question, but are we sure this fails CI when imports are incorrect?

benhoyt (approved):
Thanks! Definitely nice to enforce this via tooling rather than (clearly error-prone) human eyeballs. And nice to make `daemon_test.go` consistent and get rid of that TODO too.

HACKING.md
```diff
@@ -83,6 +83,39 @@ $ curl --unix-socket ~/pebble/.pebble.socket 'http://localhost/v1/services?names
 ```
 
 
+## Code style
```
flotter, line 86: ```snapd``` uses a CODING.md for describing topics related to style. Would it maybe a good idea, as I can see this section growing in the future.

flotter:
Just an idea for the future: I like the idea of having a single script (like ```snapd``` and others : ```run-checks```) that can perform the same checks the CI will in advance to save time iterating. Right now things are still simple, but I think it would soon be really useful to start adding ```spread``` like tests as well, and this means the turnaround time of a failed CI run can be painfully long.

flotter (approved):

.github/workflows/lint.yml
barrettj12: They're not required, but there's no real benefit to caching here considering the whole thing takes 9 seconds to run.

HACKING.md
barrettj12: Once it grows large enough to justify it, we can break it out into a separate file. No need for premature optimisation IMO.

.github/workflows/lint.yml
barrettj12: fair point, I'll do a sanity check tomorrow.

barrettj12: @benhoyt did a test [here](https://github.com/barrettj12/pebble/actions/runs/5469046259/jobs/9957477703). Linting indeed fails, but it's not very descriptive - it doesn't even say which files need fixing. I'll see if I can improve this output.

barrettj12: I've improved the action output on failure - [here](https://github.com/barrettj12/pebble/actions/runs/5469442058/jobs/9958385473) is a sample run.

benhoyt (approved):

jnsgruk (approved):
Looks good, thanks Jordan. Merging this as it's just an enforcement of agreed code style.

PR #256, opened 2023-07-06
feat(logfwd): log forwarding implementation II

barrettj12:
This PR contains the second part of #165, including the actual mechanics of log forwarding in Pebble. It builds upon #209 and #252. This includes an abstract `logClient` interface but doesn't include the specific Loki/syslog implementations - this will come in later PRs.

*This is a modification of #218, designed to fix fundamental issues identified with that PR.*

### Current design

For each log target, there is a single "gatherer", which receives logs from a bunch of services and writes them to a "client". The gatherer runs a control loop in a separate goroutine, which waits to receive logs on a channel, and writes these to the client.

For each service, the gatherer spawns a "log puller". Each puller runs in a separate goroutine, pulling logs from an iterator on the service's ringbuffer. Pulled logs are sent to the gatherer's control loop on the shared channel.

The `logClient` interface represents a client to a specific type of backend. In future PRs, we will add `lokiClient` and `syslogClient` types which implement `logClient`.

`logClient` includes two methods
```go
type logClient interface {
	Write(context.Context, servicelog.Entry) error
	Flush(context.Context) error
}
```

Client implementations have some freedom about the semantics of these methods.
- For a buffering client (e.g. HTTP), `Write` could add the log to the client's internal buffer, while `Flush` would prepare and send an HTTP request with the buffered logs.
- For a non-buffering client (TCP?), `Write` could serialise the log directly to the open connection, while `Flush` would be a no-op.

### Teardown

Gracefully stopping a log gatherer is a complex process with multiple steps.
- At the instant we attempt to stop the gatherer, it may be in the middle of flushing its client. So, wait a small amount of time for the client to finish flushing, but cancel the flush if this takes too long.
- The service may have emitted some final logs on shutdown. Give each puller some time to pull the final logs from its iterator - but again, force kill it if this is taking too long.
- Once the pullers are all finished, we must have received all the logs we're gonna get, so we can safely shut down the main loop.
- Do a final flush of the client to send off any remaining buffered logs.

All this logic is encapsulated in the `gatherer.stop()` method.

## QA

I've included some sample implementations of `logClient` [here](https://github.com/barrettj12/pebble/blob/logfwd-fake/internals/overlord/logstate/fake.go). They just print the logs to stdout. These can be used to verify that the log forwarding mechanics work properly.

Create a simple logging service, e.g.
```bash
#!/bin/bash
while true; do
  echo "Hello"
  sleep 1
done
```

and a simple plan using this service
```yaml
services:
  svc1: &logsvc
    command: /home/jb/git/canonical/pebble/logfwd-impl2/pebble/logsvc
    startup: enabled
    override: merge
  svc2: *logsvc

log-targets:
  tgt1:
    override: merge
    services: [all]
    type: loki
    location: unnecessary
```

Add the [`fake.go`](https://github.com/barrettj12/pebble/blob/logfwd-fake/internals/overlord/logstate/fake.go) file to the repo.

Comment out the following line
https://github.com/canonical/pebble/blob/3e904f9d22f297b68cba2dc33c9cf8e1bbbadd90/internals/overlord/logstate/gatherer.go#L356
and replace it with e.g.
```go
return &nonBufferingClient{}, nil          // unbuffered
return &bufferingClient{}, nil             // unlimited-size buffer, will flush on timeout only
return &bufferingClient{threshold: 3}, nil // buffer with max size: 3 logs
```
You might also want to change the gatherer's tick period:
https://github.com/canonical/pebble/blob/3e904f9d22f297b68cba2dc33c9cf8e1bbbadd90/internals/overlord/logstate/gatherer.go#L32

Run Pebble with
```
go run ./cmd/pebble run
```
and verify the logs are printing to stdout.

---

JUJU-3776


internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,190 @@
...
+		client:     client,
+		notifyChan: make(chan bool, 1),
+	}, nil
+}
+
+func (g *logGatherer) addIterator(serviceName string, iterator servicelog.Iterator, parser *servicelog.Parser) {
+	g.itLock.Lock()
+	defer g.itLock.Unlock()
+
+	iterator.Notify(g.notifyChan)
```
niemeyer, line 71: Given that the iteration is surviving across multiple plans while gatherers get removed, how do we clean up notifiers so it doesn't become a leak?

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,190 @@
...
+		if it.Next(nil) {
+			moreData = true
+		}
+	}
+
+	// If there is more data to read - put a token back in the notifyChan, so
+	// that we will continue to pull logs on the next loop iteration.
+	if moreData {
+		select {
+		case g.notifyChan <- true:
```
niemeyer, line 167: Again, this looks a bit brittle. If we process all the data currently in the iterator, we shoudn't have to feed our own watcher because we can only do something once we have more data, at which point the iterator itself will notify us that there's more data. If we hook up that pipeline between both sides in a nice way, this middle layer should get pretty "boring", in the good sense.

niemeyer (changes requested):
Thanks, Jordan! Even if there are some loose ends, this is clearly a good step in the right direction. Here is some early feedback to help shaping it up.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (p *logPuller) loop() error {
+	parser := servicelog.NewParser(p.iterator, parserSize)
+
+	for p.iterator.Next(p.tomb.Dying()) {
+		for parser.Next() {
+			if err := parser.Err(); err != nil {
+				return err
+			}
+			p.entryCh <- parser.Entry()
```
hpidcock, line 182: Should be a send in a select statement, with an case that handles cancellation via tomb dying channel.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,207 @@
...
+			continue
+		}
+
+		// Create new puller
+		newPullers[service.Name] = newLogPuller(g.entryCh, buffer)
+	}
+
+	// Old pullers for now-removed services need to be shut down.
+	for svc, puller := range g.pullers {
+		err := puller.tomb.Killf("gatherer closed")
```
hpidcock, line 100: gatherer didn't close, should be a different error.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,207 @@
...
+				g.targetName, svc, err)
+		}
+	}
+	g.pullers = newPullers
+}
+
+func (g *logGatherer) serviceStarted(service *plan.Service, buffer *servicelog.RingBuffer) {
+	g.pullersLock.Lock()
+	defer g.pullersLock.Unlock()
+	g.pullers[service.Name] = newLogPuller(g.entryCh, buffer)
```
hpidcock, line 112: need to check for old puller from a previous service right?

hpidcock: yeah so here we need to either wait for the previous puller to die (after it reaches the end of the old ring buffer), or have the gatherer track a list of dying pullers, so it can make sure to stop them.

I think we need to seperate the lifetime of a puller from g.pullers. All pullers spawned from a gatherer should be killed when the gatherer stops, but only after they have had a chance to flush.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,207 @@
...
+// stop closes the cancel channel, thereby terminating the main loop.
+func (g *logGatherer) stop() {
+	close(g.cancel)
+
+	g.pullersLock.Lock()
+	defer g.pullersLock.Unlock()
+	for _, puller := range g.pullers {
+		_ = puller.tomb.Killf("gatherer closed")
+	}
+}
```
hpidcock, line 154: we need to force the client to flush here, with a timeout.

internals/overlord/logstate/manager.go
```diff
@@ -15,26 +15,82 @@
...
+		if gatherer == nil {
+			// Create new gatherer
+			var err error
+			gatherer, err = newLogGatherer(target)
+			if err != nil {
+				logger.Noticef("Internal error: couldn't create gatherer: %v", err)
+				break
+			}
+
+			go gatherer.loop()
```
hpidcock, line 59: This should probably be handled by the newLogGatherer code.

internals/overlord/logstate/manager.go
```diff
@@ -15,26 +15,108 @@
...
+	newGatherers := make(map[string]*logGatherer, len(pl.LogTargets))
+
+	for _, target := range pl.LogTargets {
+		gatherer := m.gatherers[target.Name]
+		if gatherer == nil {
+			// Create new gatherer
+			var err error
+			gatherer, err = m.newGatherer(target)
+			if err != nil {
+				logger.Noticef("Internal error: couldn't create gatherer: %v", err)
```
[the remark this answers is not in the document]
hpidcock: I'd recommend, if we are not going to handle this failure here, to remove the error return from newLogGatherer, move the client creation to inside the gatherer.loop, so it can (re)create it as necessary. Not sure why the client itself could error in creation, possibly it doesn't error too?

internals/overlord/logstate/manager.go
```diff
@@ -15,26 +15,82 @@
...
+			delete(m.gatherers, target.Name)
+		}
+
+		// Update iterators for gatherer
+		gatherer.planChanged(pl, m.buffers)
+	}
+
+	// Old gatherers for now-removed targets need to be shut down.
+	for _, gatherer := range m.gatherers {
+		gatherer.stop()
```
hpidcock, line 73: I think if a gatherer is removed, it should be a fairly forceful remove, it shouldn't care about flushing clients, so maybe we pass a force argument into stop here.

internals/overlord/logstate/manager.go
```diff
@@ -44,5 +100,10 @@ func (m *LogManager) Ensure() error {
 
 // Stop implements overlord.StateStopper and stops all log forwarding.
 func (m *LogManager) Stop() {
-	// TODO: implement
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	for _, gatherer := range m.gatherers {
+		gatherer.stop()
```
hpidcock, line 107: Here we should wait for gatherers to finish and have their final chance to flush their buffers.

hpidcock (commented):
Great work so far, I think we need to be a bit more careful around defining a graceful termination path for pullers (go until the end of the ring buffer), gatherers (wait up to a timeout for the pullers to stop, then wait for up to a timeout for the client to flush).

The graceful shutdown path would be used when a service restarts or when pebble shuts down. All cases around reconfiguration can be much more forceful in ending all log forwarding for unwanted targets/services.

internals/overlord/logstate/manager.go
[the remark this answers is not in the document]
barrettj12: It's an "internal error" because the gatherer creation *should* never fail - all issues *should* be caught when validating the plan.

Client creation could error:
- if the `target.Type` is not valid (impossible to return a client here as we don't know what type of client)
- if the URL is not valid
- if some other configuration option is not valid

internals/overlord/logstate/gatherer.go
barrettj12: I'm not sure what you mean. We are killing the puller here because the gatherer has been asked to stop. Maybe I'm not properly understanding the semantics of the `Killf` argument?

internals/overlord/logstate/gatherer.go
barrettj12: `serviceStarted` will only be called for *new* services, so there is no need to check for old pullers.

Puller lifecycle is tied to the gatherer lifecycle. You're probably right that we should shut down old pullers in *planChanged* (rather than *serviceStarted*).

internals/overlord/logstate/gatherer.go
barrettj12: ah sorry, this is in `planChanged`. You're dead right

internals/overlord/logstate/gatherer.go
barrettj12: To have a timeout for the client flush here, we'd need to pass a channel into the `logClient.Flush` method and ask client implementations to respect that

internals/overlord/logstate/manager.go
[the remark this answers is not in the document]
barrettj12: "Handling" the error here is `continue`-ing on to the next log target. Again, this *should* never happen, so we don't really have a way to handle it other than telling the user what's wrong.

internals/overlord/logstate/gatherer.go
barrettj12: A service restart actually reuses the same ringbuffer. I've updated the code to (explicitly) handle service restarts.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+	pullersLock sync.Mutex
+
+	// Context to pass to client
+	// When we are asked to shutdown, we can cancel this context to ensure the
+	// client is not blocking us.
+	clientCtx    context.Context
+	clientCancel context.CancelFunc
+
+	// tomb for the main loop
+	tomb *tomb.Tomb
```
benhoyt, line 63: Any reason this needs to be a pointer? Similar with `logPuller.tomb`.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-g.tomb.Dying():
+			return nil
+
+		case <-ticker.C:
+			// Timeout - flush
+			err := g.client.Flush(g.clientCtx)
```
benhoyt, line 144: Could these calls use the tomb's context, as in `g.client.Flush(g.tomb.Context(nil))`? I wonder if we'd need `clientCtx` and `clientCancel` at all in that case -- when `stop` kills the tomb, I believe the context will be cancelled.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+	// Shut down pullers
+	g.pullersLock.Lock()
+	defer g.pullersLock.Unlock()
+	for _, puller := range g.pullers {
+		_ = puller.tomb.Killf("gatherer closed")
+	}
+
+	// Wait for the main loop to finish - then we can do a final flush of the client
+	_ = g.tomb.Wait()
+	err := g.client.Flush(g.clientCtx)
```
benhoyt, line 175: If we remove `clientCtx` / `clientCancel`, we can just use a new context `WithTimeout` here.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, fmt.Errorf("could not create log client: %w", err)
+	}
+
+	g := &logGatherer{
+		targetName: target.Name,
+		tickPeriod: tickPeriod,
+		client:     client,
+		entryCh:    make(chan servicelog.Entry),
+		pullers:    map[string]*logPuller{},
+		tomb:       &tomb.Tomb{},
```
benhoyt, line 78: If we change this to a non-pointer field, this line can be removed as the zero value will be fine.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+	// Start a timeout
+	time.AfterFunc(teardownTimeout, g.clientCancel)
+	// Kill the main loop
+	_ = g.tomb.Killf("gatherer closed")
+
+	// Shut down pullers
+	g.pullersLock.Lock()
+	defer g.pullersLock.Unlock()
+	for _, puller := range g.pullers {
+		_ = puller.tomb.Killf("gatherer closed")
```
benhoyt, line 170: Should the puller just have a reference to the gatherer's tomb, so that when the gatherer's tomb is killed, the pullers stop too, rather than having to do an explicit loop?

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+// stop closes the cancel channel, thereby terminating the main loop.
+// This method will block until gatherer teardown is complete.
+func (g *logGatherer) stop() {
+	// Start a timeout
+	time.AfterFunc(teardownTimeout, g.clientCancel)
+	// Kill the main loop
+	_ = g.tomb.Killf("gatherer closed")
```
benhoyt, line 164: Nit: "gatherer stopped"?

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+			select {
+			case p.entryCh <- parser.Entry():
+			case <-p.tomb.Dying():
+				break iteratorLoop
+			}
+		}
+	}
+
+	// We've been killed - close the iterator
+	return p.iterator.Close()
```
benhoyt, line 217: I guess we could `defer p.iterator.Close()` at the top of this function, and then the `break iteratorLoop` could become just `return` and we wouldn't need the label.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,255 @@
...
+			continue
+		}
+
+		// Create new puller
+		newPullers[service.Name] = newLogPuller(g.entryCh, buffer)
+	}
+
+	// Old pullers for now-removed services need to be shut down.
+	for _, puller := range g.pullers {
+		_ = puller.tomb.Killf("plan changed")
```
benhoyt, line 117: Should we debug-log this error?

benhoyt (commented):
Looking good! Left a few comments, all relatively minor.

internals/overlord/logstate/gatherer.go
barrettj12: I thought about this, but the pullers need a separate tomb because they sometimes need to stop independently of the gatherer. For example, if the plan changes, and the corresponding service is no longer forwarding logs to this target, we need to kill the puller's tomb even though the gatherer is not shutting down.

internals/overlord/logstate/gatherer.go
barrettj12: Killf semantics are confusing. The error returned is exactly the same obtained from the arguments you pass in - it doesn't indicate a problem running the method. I could change this to the following to avoid confusion?
```
puller.tomb.Kill(errors.New("plan changed"))
```

internals/overlord/logstate/gatherer.go
barrettj12: I guess not

internals/overlord/logstate/gatherer.go
barrettj12: Hmm, I think the issue is:
- when we call `gatherer.stop`, we want it to stop pulling logs right away
- But we want to allow a grace period (e.g. 5 seconds) to flush out the existing logs

So I think we need to distinguish between:
- a soft stop signal (i.e. "hey it's time to start shutting down") - this is the `tomb.Kill`
- a hard stop signal (i.e. "alright I'm pulling the plug, stop right now") - this is the `clientCancel`

As far as I know, `tomb` doesn't really give you a way to model this. We can discuss whether this is the correct behaviour though.

barrettj12: I've rewritten the `gatherer.stop` method to handle edge cases, and added comments throughout. Hopefully this makes it clear why (I think) we need separate contexts here.

benhoyt (approved):

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Kill the main loop once either:
+	// - all the pullers are done
+	// - timeoutMainLoop has passed
+	select {
+	case <-g.pullers.Done():
+	case <-time.After(timeoutMainLoop):
+	}
+
+	_ = g.tomb.Killf("gatherer stopped")
```
niemeyer, line 218: The convention with tomb is that when it's just dying for good reasons, we do a Kill(nil), which is the typical err == nil case. With the above message, tomb.Err() will actually become an error message, and it may eventually be interpreted as an actual problem by someone used to the typical convention.  It's also worth noting that Kill(nil) does not override an earlier error.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,256 @@
...
+	// Wait for final flush in the main loop
+	_ = g.tomb.Wait()
```
niemeyer, line 220: Why are we ignoring all errors?  Shouldn't this be reporting any errors found so that there's a way out for them?  Among other things, that's nice for example when you want to test that the logic is generally sane, as in tests for example, but it's often also a hint that the logic knows how to die properly. For instance: if all errors are currently ignored, how do we detect and recover from a situation where a network connection dies?  Do we have good recovery for this ATM?

niemeyer (changes requested):
Thanks a lot for this, Jordan. We're clearly approaching the end of the road.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// The main control loop for the logGatherer. loop receives logs from the
+// pullers on entryCh, and writes them to the client. It also flushes the
+// client periodically, and exits when the gatherer's tomb is killed.
+func (g *logGatherer) loop() error {
+	ticker := time.NewTicker(g.tickPeriod)
+	defer ticker.Stop()
+
+mainLoop:
```
flotter, line 161: Can the second Flush block be incorporated in the main loop? Having two contexts feel a bit confusing to me. If the exit was not blocked by the client context (e.g. the pullers exited cleanly, fast), the context would still be valid, and we can use it to Flush on exit. If the first client context is dead, that is most likely because the client is not responding, so creating a second context feels slightly too hopeful.

Can we not simply have:
```
    flush := func() {
        err := g.client.Flush(g.clientCtx)
	if err != nil {
		logger.Noticef("sending logs to target %q: %v", g.targetName, err)
	}
	
	:
		select {
		case <-g.tomb.Dying():
			flush()
			break mainLoop
		case <-ticker.C:
		        flush()
	:
	
```

flotter (commented):
Really impressive PR. I have learnt a lot from studying your code. I have really tried to give you a in depth review - I hope some things will be useful for you. If not, please discard.

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,153 @@
...
+	pg.mu.Lock()
+	defer pg.mu.Unlock()
+	if puller, ok := pg.pullers[serviceName]; ok {
+		// This should never happen, but just in case, shut down the old puller.
+		puller.kill()
+	}
+	pg.pullers[serviceName] = lp
+}
+
+func (pg *pullerGroup) RemoveOld(pl *plan.Plan) {
```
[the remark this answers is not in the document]
barrettj12: Fair point, I've moved all the plan logic into the gatherer.

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,153 @@
...
+		tgt := pl.LogTargets[pg.targetName]
+		if !svc.LogsTo(tgt) {
+			pg.remove(svcName)
+		}
+	}
+}
+
+// not thread safe, lock mu before calling
+func (pg *pullerGroup) remove(serviceName string) {
+	pg.pullers[serviceName].kill()
```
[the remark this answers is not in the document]
barrettj12: Ok, I've added an extra ctx.Done check in the logPuller.loop method, so it is now guaranteed to exit when the context is cancelled, and won't send any more logs. Hopefully this addresses your concerns?

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,256 @@
...
+// This method will block until gatherer teardown is complete.
+//
+// The teardown process has several steps:
+//   - If the main loop is in the middle of a flush when we call stop, this
+//     will block the pullers from sending logs to the gatherer. Hence, wait
+//     for the current flush to complete.
+//   - Wait for the pullers to pull the final logs from the iterator.
+//   - Kill the main loop.
+//   - Flush out any final logs buffered in the client.
+func (g *logGatherer) stop() {
```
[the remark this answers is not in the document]
barrettj12: Fair point, I've done this now.

internals/overlord/stateengine.go
```diff
@@ -136,10 +136,17 @@ func (se *StateEngine) Stop() {
...
+
+	var wg sync.WaitGroup
 	for _, m := range se.managers {
 		if stopper, ok := m.(StateStopper); ok {
-			stopper.Stop()
+			wg.Add(1)
+			go func() {
+				stopper.Stop()
+				wg.Done()
+			}()
```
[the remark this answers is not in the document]
barrettj12: Fair point, I've reverted the changes to the state engine.

We just need to ensure that the log manager is stopped after the service manager, so we can collect any final logs from the service. I've rejigged things in the overlord to ensure this happens.

internals/overlord/stateengine_test.go
```diff
@@ -36,21 +36,21 @@ func (ses *stateEngineSuite) TestNewAndState(c *C) {
 
 type fakeManager struct {
 	name                   string
-	calls                  *[]string
+	calls                  chan<- string
```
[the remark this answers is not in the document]
barrettj12: These were required by the previous changes to the state engine. They've been reverted now.

internals/overlord/logstate/gatherer.go
barrettj12: ah okay, wasn't aware of that convention. Makes a lot of sense.

internals/overlord/logstate/gatherer.go
barrettj12: This is during teardown, so if something goes wrong, we just have to move on. We can't really block Pebble shutdown because of a bad network connection. It's "best effort" here - try to send off the last logs, but if we can't, we just have to drop them. I think it's fair to log the error to the user though.

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (pg *pullerGroup) KillAll() {
+	pg.kill()
+}
+
+// Done returns a channel which can be waited on until all pullers have finished.
+func (pg *pullerGroup) Done() chan struct{} {
+	done := make(chan struct{})
+	go func() {
+		pg.wg.Wait()
```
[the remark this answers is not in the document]
barrettj12: This channel/method is only used during teardown, so not really a practical concern.

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,153 @@
...
+	pg.wg.Add(1) // this will be marked as done once loop finishes
+	go func() {
+		lp.loop()
+		pg.wg.Done()
+	}()
+
+	pg.mu.Lock()
+	defer pg.mu.Unlock()
+	if puller, ok := pg.pullers[serviceName]; ok {
+		// This should never happen, but just in case, shut down the old puller.
```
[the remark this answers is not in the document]
barrettj12: I'm opposed to a hard fail here - we should be able to continue in the face of unexpected circumstances, without bringing down the whole enterprise. Happy to log a warning though.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Wait up to timeoutPullers for the pullers to pull the final logs from the
+	// iterator and send to the main loop.
+	time.AfterFunc(timeoutPullers, g.pullers.KillAll)
+
+	// Kill the main loop once either:
+	// - all the pullers are done
+	// - timeoutMainLoop has passed
+	select {
+	case <-g.pullers.Done():
```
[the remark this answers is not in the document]
barrettj12: The piece of the puzzle you're missing is: a puller will exit of its own accord when its iterator is finished. The "ideal" shutdown sequence is:
- Service manager sends kill signal to service
- Service shuts down, possibly writing some final logs
- Service manager closes the service's ringbuffer, to indicate no more logs will be written (now added)
- The iterator continues to read up to the end of the buffer, at which point `iterator.Next` will return false - exiting the loop
- WaitGroup hits 0 and we can continue early.

The timeouts are just there in case something goes wrong (e.g. a service refuses to shut down, or writes too many logs for us to deal with).

I'm not sure I understand the rest of your comment, but maybe it is irrelevant in light of what I wrote above.

internals/overlord/logstate/gatherer.go
barrettj12: > Can the second Flush block be incorporated in the main loop?

It could be, but I like having it separate. Makes it more clear that "this is the final step".

> Having two contexts feel a bit confusing to me.

There's a very good reason why we need this, although it is a little complex.

At the moment `gatherer.Stop()` is called, we may be in the middle of flushing the client (e.g. waiting for a response from the server). This could potentially take a long time, and while we're inside the `Flush` call, the main loop is blocked, and can't process any logs from the pullers. So we need to set a deadline, after which we cancel the flush, so we can unblock the loop and process those final logs.

So we need to create a new context for the final flush, as the previous one may have been cancelled. Plus, we want to have separate lifetimes for the current flush (if any) and the final flush.

> If the first client context is dead, that is most likely because the client is not responding, so creating a second context feels slightly too hopeful.

It could be the server not responding, or maybe our request got lost in the ether, due to a bad connection or something. In any case, I think we should still make a best effort to ship the final logs.

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Currently active logPullers, indexed by service name
+	pullers map[string]*logPuller
+	// Mutex for pullers map
+	mu sync.RWMutex
+	// Common context for all pullers. Each puller uses a derived context so we
+	// can easily kill all pullers (if required) during teardown.
+	ctx context.Context
+	// Cancel func for ctx
+	kill context.CancelFunc
```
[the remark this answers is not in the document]
barrettj12: this one specifically is more of a "kill", as we only use it if the pullers are too slow to finish, and we have to force kill them. But the one on `logPuller` is more of a "cancel".

internals/overlord/logstate/puller.go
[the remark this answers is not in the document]
barrettj12: it's theoretically possible, but I don't think it's a major concern. The worst we'll get is duplicated logs, which can be deduped on the server side anyway.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,256 @@
...
+	if args.timeoutFinalFlush == 0 {
+		args.timeoutFinalFlush = timeoutFinalFlush
+	}
+	if args.newClient == nil {
+		args.newClient = newLogClient
+	}
+	return args
+}
+
+// planChanged is called by the LogManager when the plan is changed.
```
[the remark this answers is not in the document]
barrettj12: Sounds right! I added this to the doc comment.

internals/overlord/logstate/puller.go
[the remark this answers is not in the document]
flotter: True, its only called during stop(). Thank you.

internals/overlord/logstate/gatherer.go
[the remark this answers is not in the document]
flotter: Yes I completely missed that link. Services are sync-stopped in daemon.Stop() before the overlord.

internals/overlord/servstate/handlers.go
```diff
@@ -223,6 +223,9 @@ func (m *ServiceManager) doStop(task *state.Task, tomb *tomb.Tomb) error {
 	if service == nil {
 		return nil
 	}
+	// Close the ringbuffer to signal to consumers (e.g. log forwarding) that
+	// there are no more logs coming.
+	defer service.logs.Close()
```
flotter, line 228: I believe this is the critical piece of plumbing that is required to make the pullers exit as planned during Pebble shutdown? This hook is the stop handler, so it will only be used by external stop requests, including the tear-down function of the daemon. However, that function will only stop services in ```running``` and ```backoff``` state. So I am wondering what about all the other services that may have been started in the past (have valid ring buffers), but which are not running right now? (e.g. crashed service with restart action = ignore). Any puller associated with such as service will cause the exit to delay by the puller timeout delay. Not immediately sure the best way to solve this, could the termination happens in the ```overlord.exited()``` hook, but not sure if can be reused after closing it, because ```exited()``` if often only a temp transition?

internals/overlord/overlord.go
```diff
@@ -108,13 +108,15 @@ func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writ
...
 
 	o.logMgr = logstate.NewLogManager()
-	o.addManager(o.logMgr)
 
 	o.serviceMgr, err = servstate.NewManager(s, o.runner, o.pebbleDir, serviceOutput, restartHandler, o.logMgr)
 	if err != nil {
 		return nil, err
 	}
 	o.addManager(o.serviceMgr)
+	// The log manager should be stopped after the service manager, so we can
```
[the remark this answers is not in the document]
barrettj12: yeah you're right, it doesn't seem to make a difference

internals/overlord/servstate/handlers.go
barrettj12: Yeah, also if you stop a service which is in `starting` state (e.g. interrupting Pebble shortly after it starts), it doesn't work. Not sure where is the best place for this code.

internals/overlord/overlord.go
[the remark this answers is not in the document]
barrettj12: We've now implemented it so that the ring buffers are closed in `ServiceManager.Stop`, so it is important to have this ordering.

internals/overlord/servstate/handlers.go
barrettj12: Spoke to @benhoyt about this, we agreed it makes the most sense to have the ring buffer lifecycle tied to the service data (i.e. when we no longer need the service data, we no longer need the ring buffer either). So we've put the close in `ServiceManager.removeService`, and are ensuring we call `removeService` in the `ServiceManager.Stop` code.

internals/overlord/logstate/puller.go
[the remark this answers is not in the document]
barrettj12: We're now using (and waiting on) a tomb for the puller, which should resolve your issue about synchronising pullers.

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,174 @@
...
+
+	for _, puller := range pg.pullers {
+		puller.tomb.Kill(nil)
+	}
+}
+
+// Done returns a channel which can be waited on until all pullers have finished.
+func (pg *pullerGroup) Done() <-chan struct{} {
+	pg.mu.RLock()
+	pullers := pg.pullers
```
niemeyer, line 146: This is read-locking the access to the map field from pg, but then all actual accesses to the map are done concurrently in a goroutine below, which is a race condition that needs addressing.

There's also a second problem here: such Done channels are used cheaply to verify the conclusion of work. If you ook inside tomb (recommended), you'll notice that this literally a single channel throughout the entire lifetime of the object, and alive/dead are simply states of such channels. Here we instead see a goroutine being created everyt ime the method is called, and it blocks every time waiting for everything to get finished. This sort of pattern can easily become logical memory leaks in the system, with goroutines that never get collected. 

Can we please address both the race condition and that second issue at once?

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (pg *pullerGroup) Add(serviceName string, buffer *servicelog.RingBuffer, entryCh chan<- servicelog.Entry) {
+	lp := &logPuller{
+		iterator: buffer.TailIterator(),
+		entryCh:  entryCh,
+	}
+	lp.tomb.Go(func() error {
+		err := lp.loop()
+		// Once the puller exits, we should remove it from the group.
+		//pg.Remove(serviceName) - this seems to cause deadlock though
```
niemeyer, line 88: This issue needs to be understood and addressed. The possible cause is that issue in Done.

niemeyer (changes requested):
Thanks, Jordan. This is looking great.

There's still one important issue to address, unfortunately:

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,319 @@
...
+
+func newTimer() timer {
+	t := timer{
+		timer: time.NewTimer(1 * time.Hour),
+	}
+	t.Stop()
+	return t
+}
+
+func (t *timer) Finished() <-chan time.Time {
```
benhoyt, line 263: Naming nit: should this be called `Expired`, which is what the `time.Timer` docs call the end time?

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,319 @@
...
+	t.timer.Stop()
+	t.set = false
+	// Drain timer channel
+	select {
+	case <-t.timer.C:
+	default:
+	}
+}
+
+func (t *timer) EnsureSet(timeout time.Duration) {
```
benhoyt, line 277: Naming nit: would this just be better named `Reset`?

internals/overlord/servstate/manager.go
```diff
@@ -88,6 +88,9 @@ func (m *ServiceManager) Stop() {
 	if err != nil {
 		logger.Noticef("Cannot stop child process reaper: %v", err)
 	}
+	for name := range m.services {
```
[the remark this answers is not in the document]
barrettj12: I'm not quite sure what you mean by "servicelock" - do you mean `m.servicesLock`?

Since `m.services` can be accessed concurrently, we should lock `m.servicesLock` whenever we access it. This locking is done inside the `removeService` method.

Even if it's not possible for anything to concurrently access `m.services` here, it still doesn't hurt to lock it. It's safe, adds negligible overhead, and future-proofs against race conditions.

Does that address your concern?

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,164 @@
...
+	pg.pullers[serviceName] = lp
+}
+
+// List returns a list of all service names for which we have a currently
+// active puller.
+func (pg *pullerGroup) List() []string {
+	pg.mu.RLock()
+	defer pg.mu.RUnlock()
+
+	var svcs []string
```
benhoyt, line 110: Nit: the slightly more efficient idiom is `svcs := make([]string, 0, len(pg.pullers))`

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	lp.tomb.Go(lp.loop)
+	pg.tomb.Go(lp.tomb.Wait)
+
+	pg.pullers[serviceName] = lp
+}
+
+// List returns a list of all service names for which we have a currently
+// active puller.
+func (pg *pullerGroup) List() []string {
```
benhoyt, line 106: Would this be better called `ServiceNames`? (list of what?)

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,164 @@
...
+	if !pullerExists {
+		return
+	}
+
+	puller.tomb.Kill(nil)
+	delete(pg.pullers, serviceName)
+
+	err := puller.tomb.Wait()
+	if err != nil {
+		logger.Noticef("Error from log puller: %v", err)
```
benhoyt, line 134: Another log message nit: "Cannot"

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	pg.tomb.Kill(nil)
+}
+
+// Done returns a channel which can be waited on until all pullers have finished.
+func (pg *pullerGroup) Done() <-chan struct{} {
+	return pg.tomb.Dead()
+}
+
+func (pg *pullerGroup) Contains(serviceName string) bool {
```
benhoyt, line 153: Can the only-used-in-test methods be moved to an `export_test.go` file to make it clear they shouldn't be used in production code?

benhoyt (approved):
This looks good to me. Left a few very nitty comments (some of which are "take it or leave it" naming suggestions), but I'm happy with this.

internals/overlord/logstate/gatherer.go
barrettj12: No, because it's idempotent, which to me is not implied by the word "Reset"

internals/overlord/logstate/puller.go
barrettj12: I don't see how to rewrite this one with "Cannot", as there is no verb.

internals/overlord/logstate/puller.go
barrettj12: The testing here deals a lot with the internals, so it's "white-box" testing done inside the same package. i.e. there is no `logstate_test` package, you'll notice all the tests are inside the `logstate` package. So I don't think `export_test.go` makes sense here.

We could think about moving all the tests to a `logstate_test` package, but I feel like this would be a lot of extra work exporting otherwise-non-exported things, without a clear benefit in my mind.

Another point - why *shouldn't* `Contains` or `Len` be used in production code? Currently, it happens that they are not, but I don't really see an issue if we did need to use these in the future.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,319 @@
...
+	timer := newTimer()
+
+mainLoop:
+	for {
+		select {
+		case <-g.tomb.Dying():
+			break mainLoop
+
+		case <-timer.Finished():
+			// Mark timer as unset
```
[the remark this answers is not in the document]
barrettj12: It's pre-empting the question, "why do you need to call `timer.Stop` if the timer has already expired (so it should be stopped)?"

And the answer is, "we need to set the internal state of the timer so that it knows it is stopped, and we'll be able to set it again in the future".

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,319 @@
...
+			if err != nil {
+				logger.Noticef("Error sending logs to target %q: %v", g.targetName, err)
+			}
+
+		case entry := <-g.entryCh:
+			err := g.client.Write(g.clientCtx, entry)
+			if err != nil {
+				logger.Noticef("Error writing logs to target %q: %v", g.targetName, err)
+			}
+			// Set timer if not already set
```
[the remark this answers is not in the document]
barrettj12: this is fair

internals/overlord/logstate/gatherer.go
barrettj12: works for me

internals/overlord/logstate/puller.go
barrettj12: I'll call it `Services`

internals/overlord/logstate/puller.go
benhoyt: Ah yeah, fair enough, as it's a tomb.Wait

internals/overlord/logstate/puller.go
benhoyt: > why shouldn't Contains or Len be used in production code?

`Contains` and `Len` look a bit dubious, because they're inherently racey. The locking implies they're concurrency-safe, which they are at a low level (no data races), but not at a high level: as soon as you get the length, it could change. Kinda like how `len(channel)` in Go is normally incorrect: as soon as you have the number of buffered elements, it could be sent/received to on another goroutine and your length could be wrong. That's what it stood out to me.

But I don't have strong feelings here, and what you say regarding white-box testing makes sense.

benhoyt (approved):

internals/overlord/servstate/manager.go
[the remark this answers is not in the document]
hpidcock: reading m.services needs the servicesLock, since removeService needs the lock too there are probably two options:

Copy the names:
```
// Stop implements overlord.StateStopper and stops background functions.
func (m *ServiceManager) Stop() {
	err := reaper.Stop()
	if err != nil {
		logger.Noticef("Cannot stop child process reaper: %v", err)
	}
	m.servicesLock.Lock()
	services := make([]string, len(m.services))
	for name := range m.services {
		services = append(services, name)
	}
	m.servicesLock.Unlock()
	for _, name := range services {
		m.removeService(name)
	}
}
```

Add a removeAllServices method:
```
// Stop implements overlord.StateStopper and stops background functions.
func (m *ServiceManager) Stop() {
	err := reaper.Stop()
	if err != nil {
		logger.Noticef("Cannot stop child process reaper: %v", err)
	}
	m.removeAllServices()
}

func (m *ServiceManager) removeAllServices() {
	m.servicesLock.Lock()
	defer m.servicesLock.Unlock()

	for name, svc := range m.services {
		if svc.logs != nil {
			err := svc.logs.Close()
			if err != nil {
				logger.Noticef("Error closing service %q ring buffer: %v", name, err)
			}
		}
	}

	m.services = map[string]*serviceData{}
}
```

internals/overlord/logstate/fake.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/canonical/pebble/internals/servicelog"
+)
+
+// Fake sample implementations of logClient
+// TODO: remove this file before merging
```
hpidcock, line 12: Is this used by the tests?

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,164 @@
...
+	tomb tomb.Tomb
+}
+
+func newPullerGroup(targetName string) *pullerGroup {
+	pg := &pullerGroup{
+		targetName: targetName,
+		pullers:    map[string]*logPuller{},
+	}
+	pg.tomb.Go(func() error {
+		<-pg.tomb.Dying()
```
hpidcock, line 80: maybe just add a comment here

hpidcock (approved):
I'm happy where this ended up. Thank-you for all the hard work on it.

internals/overlord/logstate/fake.go
barrettj12: nope it's just for QA. Maybe I'll put in on a separate branch?

internals/overlord/servstate/manager.go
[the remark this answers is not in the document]
barrettj12: ah I see the issue now. Sorry @flotter I misunderstood your comment

internals/overlord/logstate/puller.go
barrettj12: ok, you've made a fair case as to why these shouldn't be used in production code. I'll lowercase both methods and add a note saying they are used for testing.

internals/overlord/logstate/gatherer.go
```diff
@@ -0,0 +1,319 @@
...
+		logger.Debugf("gatherer %q: pullers have finished", g.targetName)
+	case <-time.After(timeoutMainLoop):
+		logger.Debugf("gatherer %q: force killing main loop", g.targetName)
+	}
+
+	g.tomb.Kill(nil)
+	// Wait for final flush in the main loop
+	err := g.tomb.Wait()
+	if err != nil {
+		logger.Noticef("Internal error: cannot shut down gatherer: %v", err)
```
benhoyt, line 245:
```suggestion
		logger.Noticef("Cannot shut down gatherer: %v", err)
```

internals/overlord/logstate/puller.go
```diff
@@ -0,0 +1,177 @@
...
+
+func newPullerGroup(targetName string) *pullerGroup {
+	pg := &pullerGroup{
+		targetName: targetName,
+		pullers:    map[string]*logPuller{},
+	}
+	// This goroutine lives for the lifetime of the pullerGroup. This is so that,
+	// if needed, we can safely remove all pullers and then add more, without
+	// causing a panic (tombs can't be reused once all the tracked goroutines
+	// have finished).
```
benhoyt, line 82: Gustavo said this is a nice comment. :-)

niemeyer (approved):

benhoyt:
@barrettj12 We discussed this at our APAC spec review meeting live with Gustavo and agreed to merge. Merge party!

Issue #167, opened 2022-12-02
Pebble doesn't report errors starting startup-enabled services


barrettj12:
I have the following simple "service" at `root/log.sh`:
```bash
#!/bin/bash
while true
do
  echo "Hi I'm alive"
  sleep 1
done
```
and the following Pebble layer:
```yaml
services:
  log:
    command: /root/log.sh
    override: merge
    startup: enabled
```
If I don't have permission to run `/root/log.sh`:
```console
$ /root/log.sh
bash: /root/log.sh: Permission denied
```
then I would expect Pebble to report an error when I go `pebble run`, but it doesn't report anything:
```console
$ pebble run -v
2022-12-02T02:47:24.670Z [pebble] Started daemon.
2022-12-02T02:47:24.692Z [pebble] POST /v1/services 20.668906ms 202
2022-12-02T02:47:24.693Z [pebble] Service "log" starting: /root/log.sh
2022-12-02T02:47:24.693Z [pebble] Started default services with change 9.
```

benhoyt:
Yeah, I've noticed this too, and it's unexpected not to get any feedback (unless you specify `PEBBLE_DEBUG=1`). It's actually more general than just permission denied errors -- it's any error starting the startup-enabled services when the daemon starts.

Do you think it should log an error, or wait (up to 1 second per service, I think it is) and exit `pebble run` with an error message and non-zero exit code?

barrettj12:
It doesn't report `No such file or directory` errors either.

I think it should log an error right away - because errors like `permission denied` and `no such file or directory` are "unrecoverable" - trying again is not gonna help, some manual intervention is needed.

I'm not sure what Pebble's usual behaviour is, but I think exiting `pebble run` is probably not the right move - you still want the other services to run if possible, I presume?

hpidcock:
> I'm not sure what Pebble's usual behaviour is, but I think exiting `pebble run` is probably not the right move - you still want the other services to run if possible, I presume?

if the service is critical to the plan then it should have
```
on-failure: shutdown
```

barrettj12:
@aliamerj might be another good one for you to work on. Shouldn't be too difficult to surface this error, and it would be a vast improvement.

benhoyt:
[...]

PR #257, opened 2023-07-07
enter: start default services before executing subcommand 

rebornplusplus:
The `--run` option in the `enter` command should ensure that the default services have all started before executing the subcommand, as specified in the help info.

This, however, does not seem to be the case as the current implementation only sends a request to the API to start the default services and immediately afterwards, executes the subcommand. This can have undesired behaviour. For example, the following fails:

    $ pebble enter --run stop srv1
    2023-06-01T09:52:52.516Z [pebble] Started daemon.
    2023-06-01T09:52:52.546Z [pebble] POST /v1/services 29.059682ms 202
    2023-06-01T09:52:52.546Z [pebble] Started default services with change 386.
    2023-06-01T09:52:52.565Z [pebble] Service "srv1" starting: <cmd..>
    2023-06-01T09:52:52.584Z [pebble] POST /v1/services 37.51026ms 202
    - Stop service "srv1" (cannot stop service while starting)

It may also happen that the stop subcommand is executed, before the service hasn't even been initiated/started.

    $ pebble enter --run stop srv99
    2023-06-22T05:46:02.604Z [pebble] Started daemon.
    2023-06-22T05:46:02.619Z [pebble] Service "srv55" starting: <cmd..>
    2023-06-22T05:46:02.629Z [pebble] POST /v1/services 23.863368ms 202
    2023-06-22T05:46:02.630Z [pebble] Started default services with change 9.
    2023-06-22T05:46:02.637Z [pebble] Service "srv1" starting: <cmd..>
    2023-06-22T05:46:02.646Z [pebble] POST /v1/services 14.761043ms 202
    2023-06-22T11:46:02+06:00 INFO Service "srv99" has never been started.
    ...

This commit should ensure that the default services have all started, before executing the subcommand. It does so by ensuring the "change" made by the "autostart" request is ready, before running the subcommand. The status of the "change" can be "Error" once it's ready; the subcommand will still be executed after.


internals/cli/cmd_run.go
```diff
@@ -202,13 +211,21 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
...
-		close(ready)
+		if ready != nil {
+			// wait for the default services to start
+			autoStartReady = make(chan struct{}, 1)
+			go func() {
+				waitCmd := waitMixin{
+					hideProgress: true,
+				}
+				waitCmd.setClient(rcmd.client)
+				waitCmd.wait(changeID)
```
woky, line 223: The wait call can fail. The signature of `wait()` is
```
func (wmx waitMixin) wait(id string) (*client.Change, error) {
```
While the commit message makes it clear that we ignore `client.Change.Err != ""`, I think we should fail on non-nil error, because the cause may lead to further errors that will be hard to debug when we silently ignore the original error.

woky (changes requested):

internals/cli/cmd_run.go
woky: Thinking more about it: If we fail here, we have to stop the daemon gracefully (i.e. with services) as well. If that turns out to be too complicated, we should at least log the error from `wait()`. But there has to be more places where we fail on an error and I suspect the daemon is going to be stopped is in some `defer`-ed function.

rebornplusplus: Hiya, I took the easier route and just logged the error message for now. Let me know if that's okay.

woky (approved):

rebornplusplus:
@benhoyt @niemeyer -- will you please take a look at this PR? Thanks. :)

niemeyer:
> PS. The tests are failing randomly. I would appreciate it very much if any of you could re-trigger the workflow.

I'd appreciate even more if the tests could be fixed so they don't fail randomly.

niemeyer (changes requested):

benhoyt:
> > PS. The tests are failing randomly. I would appreciate it very much if any of you could re-trigger the workflow.
>
> I'd appreciate even more if the tests could be fixed so they don't fail randomly.

Agreed. However, note that Fred Lotter is [working on that](https://github.com/canonical/pebble/issues/224#issuecomment-1621067201) (these are the servstate failures).

internals/cli/cmd_run.go
```diff
@@ -202,13 +211,24 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
...
+			// wait for the default services to start
+			autoStartReady = make(chan struct{}, 1)
+			go func() {
+				waitCmd := waitMixin{
+					hideProgress: true,
+				}
+				waitCmd.setClient(rcmd.client)
+				_, err := waitCmd.wait(changeID)
+				if err != nil {
+					logger.Debugf("Error starting default services: %v", err)
```
benhoyt, line 225: Should this be `logger.Noticef` instead of debug? Or even a hard error and exit with a non-zero error code?

internals/cli/cmd_run.go
```diff
@@ -202,13 +211,24 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 		} else {
 			logger.Noticef("Started default services with change %s.", changeID)
 		}
-	}
 
-	var stop chan struct{}
-	if ready != nil {
-		stop = make(chan struct{}, 1)
-		ready <- func() { close(stop) }
-		close(ready)
```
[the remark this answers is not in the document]
benhoyt: Yeah, I'm having a bit of trouble following the logic too. Seems weird to be passing on a channel a function that closes another channel.

Somewhat related: https://github.com/canonical/pebble/issues/167 -- I wonder if `pebble run` should also wait for services to be started successfully before starting up (or if not waiting, exit if they don't start properly). It seems like it'd be better for that to be a hard error. But maybe that's a separate PR/discussion...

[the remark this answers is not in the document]
rebornplusplus: This was introduced in #188 while adding the `enter` command. The commits added an argument to the `rcmd.run` function in [`cmd_run.go`](https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_run.go#L82):
https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_run.go#L82
The `ready` channel is passed from [`cmd_enter.go`](https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_enter.go#L150-L157) and in other cases, it's typically `nil`. 
https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_enter.go#L150-L157
After running the `run()` function with the channel, execution [waited on that channel](https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_enter.go#L159-L166) and one other; the latter received values if there was any `panic()` while executing `run()`.
https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_enter.go#L159-L166
The `runReadyCh` in `cmd_enter.go` (`ready` in `cmd_run.go`) received a function to close the `stop` channel in `cmd_run.go`, after the daemon had started, the watchdog was run and the default services had begun in [`cmd_run.go`](https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_run.go#L207-L212).
https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_run.go#L207-L212
The `stop` channel, if it was closed or received any values would kill the daemon and stop the pebble server, as specified here:
https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_run.go#L229-L230
This behavior was utilized in [`cmd_enter.go`](https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_enter.go#L174-L176), to shut down the pebble daemon if there was no need to keep the "Service Manager" alive.
https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_enter.go#L174-L176

---

**New changes**

The desire is to send to the `ready` channel `cmd_run.go` after all the default services have "started". Previously, it only made an API call and signalled that the services have started, through the `ready` channel. This produced a few errors as I specified in https://github.com/canonical/pebble/pull/257#issue-1793475495. The new changes are supposed to avoid those scenarios and send to `ready` only after all the default services have started. This is being done by sending to the channel when the respective autostart "change" is "Ready".

In the new changes, after making the autostart POST call https://github.com/canonical/pebble/blob/8e96c74a21cc104ac722bf4160d7db6e11ddeb60/internals/cli/cmd_run.go#L199
a goroutine starts to wait on the changes to be "Ready".
```go
        if ready != nil {
            // wait for the default services to start
            autoStartReady = make(chan struct{}, 1)
            go func() {
                waitCmd := waitMixin{
                    hideProgress: true,
                }
                waitCmd.setClient(rcmd.client)
                _, err := waitCmd.wait(changeID)
                if err != nil {
                    logger.Debugf("Error starting default services: %v", err)
                }
                autoStartReady <- struct{}{}
            }()
        }
```
Only when the change has been "Ready", it sends to a new channel `autoStartReady`. Afterwards, the previous logic is used to send to the `ready` (`runReadyCh` in `cmd_enter.go`) channel, as before.
```go
        case <-autoStartReady:
            notifyReady()
        case <-stop:
            break out
```

internals/cli/cmd_run.go
rebornplusplus: I have pushed a commit which produces a hard error if there is any error from `waitCmd.wait`. @woky mentioned this previously as well. 

Although looking at the [`internals/cli/wait.go`](https://github.com/canonical/pebble/blob/master/internals/cli/wait.go#L45-L152) file, I wonder if we should raise a hard-error **only if** we have received a non-`nil` "change" from the `wait()` function and `change.Err` is not `nil`. Let me know what you think.

rebornplusplus: I agree. https://github.com/canonical/pebble/pull/257/commits/4098a71cd2ebe7125d54dd416e62554f3be8cfca should cover this case then.

internals/cli/cmd_run.go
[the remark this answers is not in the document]
benhoyt: Thanks. Still seems a little convoluted, but an obviously-simpler approach is not jumping out at me, so seems reasonable.

What do you think of my earlier question/comment:

> Somewhat related: #167 -- I wonder if pebble run should also wait for services to be started successfully before starting up (or if not waiting, exit if they don't start properly). It seems like it'd be better for that to be a hard error. But maybe that's a separate PR/discussion...

benhoyt (approved):
Added a comment, but looks reasonable to me. Is it possible to add a test for this?

internals/cli/cmd_run.go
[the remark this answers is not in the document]
rebornplusplus: > Somewhat related: https://github.com/canonical/pebble/issues/167 -- I wonder if pebble run should also wait for services to be started successfully before starting up (or if not waiting, exit if they don't start properly). It seems like it'd be better for that to be a hard error. But maybe that's a separate PR/discussion...

Concerning `pebble run`, I think it's better to exit and report if the services don't start properly. In `enter`, we need the wait because the subcommand which follows needs to be run after the services have started. That may not be the case for `run` though. 

Also, I am not sure and please correct me if I am wrong, but it seemed to be that if a startup-enabled service failed to start with an error, the services that followed in the queue were not started as well. So I think it's best to report any errors while starting up and exit afterwards.

woky (approved):
LGTM

internals/cli/cmd_enter_test.go
```diff
@@ -231,3 +232,40 @@ func (s *PebbleSuite) TestEnterHelpCommandHelpArg(c *C) {
...
+
+// TestEnterSubCommandWaits checks that the subcommand in enter
+// starts **after** the default services have started. It utilizes
+// the modification time of /proc/[PID] directory for checking.
+func (s *PebbleSuite) TestEnterSubCommandWaits(c *C) {
+	layerTemplate := dumbDedent(`
+		services:
+		  stat:
+		    override: replace
+		    command: /bin/sh -c 'stat -c "%y" /proc/$$ > $PEBBLE/stat; sleep 1;'
```
woky, line 244: Just curious; why the `sleep 1`?

rebornplusplus: To avoid getting the new error report this PR introduced:
```
cannot run pebble: error starting default services: cannot perform the following tasks:
- Start service "stat" (cannot start service: exited quickly with code 0)
```
This reminds me, I should add a test for that as well. Thank you.

rebornplusplus: Added in https://github.com/canonical/pebble/pull/257/commits/1281c967da6093bb3e2e6567a792f83dd860a079. :v:

internals/cli/cmd_run.go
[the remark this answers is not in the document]
rebornplusplus: I ran into a similar situation to #167 while adding the recent tests; I had to debug for some time before realizing "Oh wait, this must be that...! :man_facepalming:".
I echo my earlier comments. I don't think `pebble run` or `pebble enter --run` (without any subcommand) should wait for the services to start. Although I do somewhat understand the motivation in that. But I think, it's crucial that we report (and exit) if any default service fails to start.
In any case, it should be trivial to do so on top of this PR as a separate commit or even a new PR afterwards. Let me know what you think.

woky (approved):

cjdcordeiro (changes requested):
thanks for fixing this behavior @rebornplusplus. The overall behavioural changes to the `enter` startup sequence with `--run` LGTM!

My only comment is about this thread: 

> > Somewhat related: #167 -- I wonder if pebble run should also wait for services to be started successfully before starting up (or if not waiting, exit if they don't start properly). It seems like it'd be better for that to be a hard error. But maybe that's a separate PR/discussion...
> 
> Concerning `pebble run`, I think it's better to exit and report if the services don't start properly. In `enter`, we need the wait because the subcommand which follows needs to be run after the services have started. That may not be the case for `run` though.
> 
> Also, I am not sure and please correct me if I am wrong, but it seemed to be that if a startup-enabled service failed to start with an error, the services that followed in the queue were not started as well. So I think it's best to report any errors while starting up and exit afterwards.

Afaiu and according to our discussions, https://github.com/canonical/pebble/pull/257/commits/4098a71cd2ebe7125d54dd416e62554f3be8cfca# makes the Pebble daemon exit if 1 service fails to start correctly. IMO **this is NOT the right behaviour**, and even if it was, it deserves some discussion. The rationale is that Pebble services can be completely independent (unless one comes `after` or `before` another), and thus the failure of one service should not jeopardize the execution of another. **I suggest dropping the hard error** from that commit (having extra logging is still nice though).

Moreover, `--run` is supposed to mimic `pebble run`'s behaviour and thus this PR would introduce a discrepancy amongst the two (as alluded by @benhoyt in his comment above).

@benhoyt @rebornplusplus please let me know if you disagree. If so, we should probably have a chat about it. IMO the right approach is to actually address #167.

benhoyt:
@cjdcordeiro Yep, that makes sense and I agree. We should at least have more discussion if we're going to change that behaviour, and (separately) address #167.

rebornplusplus:
> > @benhoyt @rebornplusplus please let me know if you disagree. If so, we should probably have a chat about it. IMO the right approach is to actually address https://github.com/canonical/pebble/issues/167.
>
> @cjdcordeiro Yep, that makes sense and I agree. We should at least have more discussion if we're going to change that behaviour, and (separately) address https://github.com/canonical/pebble/issues/167.

Sounds good, let's discuss that separately. Meanwhile, I have reverted related changes in https://github.com/canonical/pebble/pull/257/commits/c01f729c560b6f1e7c829fde1badacbeeaff596a.

cjdcordeiro (approved):
Thanks for the changes ;)

rebornplusplus:
Hiya, this has been idle for some time. I have updated the PR to be in-line with the `master` branch. Please take another look at your convenience. Thanks!

internals/cli/cmd_run.go
```diff
@@ -225,13 +234,20 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
...
+		if ready != nil {
+			// wait for the default services to start
+			autoStartReady = make(chan error, 1)
+			go func() {
+				waitCmd := waitMixin{
```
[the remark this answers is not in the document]
benhoyt: Also, I think we need to modify the logging and error handling a bit now. We should only ever call `WaitChange` if the `AutoStart` succeeded (err is nil case). I think we should probably log "Cannot start default services" for the `AutoStart` error, and "Cannot wait for autostart change" for the `WaitChange` error. And, if `ready != nil`, only log "Started default services with change %s." if the `WaitChange` succeeds.

internals/cli/wait.go
```diff
@@ -31,7 +31,8 @@ var (
 )
 
 type waitMixin struct {
-	NoWait bool `long:"no-wait"`
+	NoWait       bool `long:"no-wait"`
```
benhoyt, line 34: Per above, I think we should revert the changes to this file and use `Client.WaitChange` instead.

internals/cli/cmd_enter_test.go
```diff
@@ -242,3 +243,40 @@ func (s *PebbleSuite) TestEnterHelpCommandHelpArg(c *C) {
...
```
benhoyt, line 255: I'm wondering if we can simplify, instead of using `stat` on `/proc/$$`, just output the date directly. `date --rfc-3339=ns > $PEBBLE/enter-wait`, for example. Similar with the exec command.

internals/cli/cmd_enter_test.go
```diff
@@ -242,3 +243,40 @@ func (s *PebbleSuite) TestEnterHelpCommandHelpArg(c *C) {
...
+		    startup: enabled
+	`)
+	layerPath := filepath.Join(s.pebbleDir, "layers", "001-stat.yaml")
+	writeTemplate(layerPath, layerTemplate, nil)
+
+	cmd := []string{"pebble", "enter", "--run", "exec", "/bin/sh", "-c", "stat -c '%y' /proc/$$"}
```
benhoyt, line 261: I don't think we need to go via `/bin/sh` here, just `exec <command>` directly should work, as we're not using any shell features.

benhoyt (changes requested):

internals/cli/cmd_run.go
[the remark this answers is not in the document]
rebornplusplus: Thanks! I have tried to reflect your suggestions in https://github.com/canonical/pebble/pull/257/commits/753eb86dd3a5bfea27584c6333cbe443218e7ad2. Please take a look.

internals/cli/cmd_run.go
```diff
@@ -225,13 +234,20 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
...
+					hideProgress: true,
+				}
+				_, err := waitCmd.wait(rcmd.client, changeID)
+				autoStartReady <- err
```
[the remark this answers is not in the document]
rebornplusplus: Done! Thanks.

internals/cli/cmd_enter_test.go
rebornplusplus: This is better actually. Can't remember why I didn't try this before, it has been so long..

internals/cli/cmd_enter_test.go
rebornplusplus: True

internals/cli/cmd_run.go
```diff
@@ -229,21 +229,39 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) (err error)
...
+	}
+
 	if !rcmd.Hold {
+		// Start the default services.
 		servopts := client.ServiceOptions{}
 		changeID, err := rcmd.client.AutoStart(&servopts)
 		if err != nil {
 			logger.Noticef("Cannot start default services: %v", err)
-		} else {
+		} else if ready == nil {
```
benhoyt, line 248: Actually, thinking about this now, I think we can slightly simplify this and fix (half the issue with) #167 so that `pebble run` at least logs errors starting default services. So what about starting the `WaitChange` goroutine in either case so we can log success or failure, but only notifying if `ready != nil`:

```go
// Start the default services (those configured with "startup: enabled").
servopts := client.ServiceOptions{}
changeID, err := rcmd.client.AutoStart(&servopts)
if err != nil {
	logger.Noticef("Cannot start default services: %v", err)
} else {
	// Wait for the default services to actually start and then notify
	// the ready channel (for the "enter" command).
	go func() {
		logger.Debugf("Waiting for default services to autostart with change %s.", changeID)
		_, err := rcmd.client.WaitChange(changeID, nil)
		if err != nil {
			logger.Noticef("Cannot wait for autostart change %s: %v", changeID, err)
		} else {
			logger.Noticef("Started default services with change %s.", changeID)
		}
		if ready != nil {
			notifyReady()
		}
	}()
}
```

Note that addition of a debug log before the `WaitChange` call, as that seems like it'd be useful for debugging.

internals/cli/cmd_run.go
```diff
@@ -229,21 +229,39 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) (err error)
...
 			logger.Noticef("Started default services with change %s.", changeID)
+		} else {
+			// If ready != nil (case for "enter" command), wait for the default
+			// services to start and then notify on ready channel.
+			go func() {
+				_, err := rcmd.client.WaitChange(changeID, nil)
+				if err != nil {
+					logger.Noticef("Cannot wait for autostart change %s", changeID)
```
benhoyt, line 256: This is reflected in the above suggestion, but we should include the `err` value here:
```suggestion
					logger.Noticef("Cannot wait for autostart change %s: %v", changeID, err)
```

internals/cli/cmd_enter_test.go
```diff
@@ -241,3 +242,38 @@ func (s *PebbleSuite) TestEnterHelpCommandHelpArg(c *C) {
 	c.Check(stdout, Matches, "(?s).*\\bThe help command displays information about commands\\..*")
 	c.Check(exitCode, Equals, 0)
 }
+
+// TestEnterSubCommandWaits checks that the subcommand in enter
+// starts **after** the default services have started.
+func (s *PebbleSuite) TestEnterSubCommandWaits(c *C) {
```
benhoyt, line 248: This is a nice test, thanks!

benhoyt (changes requested):
One more suggested simplification / improvement. Sorry for the back and forth -- let me know what you think.

internals/cli/cmd_run.go
rebornplusplus: I thought of this before pushing yesterday too. This is nice. I have added https://github.com/canonical/pebble/pull/257/commits/b55bbf8b9b72b167537c886ab2a8a1680ccf6903.

benhoyt (approved):
Looks good, thanks. I've tested locally with both `pebble run` and `pebble enter` as well.

PR #259, opened 2023-07-23
Merge snapcore/snapd/state changes

dmitry-lyfar:
Hello there! As discussed with @niemeyer, the latest changes in snapd's state management were merged into Pebble. Please find the summary below:

State:
- Add WaitStatus support that allows a task to wait until further action to continue its execution. The WaitStatus is treated mostly as DoneStatus, except it is not ready status.
- Add CopyState function.
- Add Change.AbortUnreadyLanes.
- Add Change.CheckTaskDependencies to check if tasks have circular dependencies.
- Add task and change callbacks invoked on a status change.
- Implement timings.GetSaver interface in state.State. Note, however, the snapcore/snapd/timings package that provides the interface is not included in this PR. Pebble has a similar "timing" package. It might be reasonable to introduce "timings" separately to replace the "timing" package to keep concerns (state and timings) separated.

State engine:
- Add StateStarterUp interface to perform possible expensive initialisation in a separate StartUp method.

Overlord:
- Implement the StateStarterUp interface.

Daemon:
- Call Overlord StartUp during the start process to complete its initialisation. Note, that there is no obvious requirement to introduce this separation for Pebble. It can still initialise Overlord in New(). The separation was mimicked to maintain better compatibility with possible future mergers of the snapcore/snapd/overlord/* changes and unit tests.

Misc:
- Add jsonutil package required by the new state's CopyState function.
- Rename osutil.CanStat to osutil.FileExist.
- Rename task.FakeTime to task.MockTime.
- Add testutil.ErrorIs and testutil.DeepUnsortedMatches checkers.

Unit-testing:

Pebble's unit tests fail occasionally if ran as:

`go test ./... -count 10`

This PR does not appear to introduce any regression to the current tests' behaviour.

### DIFF
--- a/internals/overlord/state/change.go
+++ b/internals/overlord/state/change.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 
 /*
- * Copyright (c) 2016 Canonical Ltd
+ * Copyright (C) 2016-2023 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as
@@ -23,8 +23,11 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
+
+	"github.com/canonical/pebble/internals/logger"
 )
 
 // Status is used for status values for changes and tasks.
@@ -37,7 +40,8 @@ const (
 	// to an aggregation of its tasks' statuses. See Change.Status for details.
 	DefaultStatus Status = 0
 
-	// HoldStatus means the task should not run, perhaps as a consequence of an error on another task.
+	// HoldStatus means the task should not run for the moment, perhaps as a
+	// consequence of an error on another task.
 	HoldStatus Status = 1
 
 	// DoStatus means the change or task is ready to start.
@@ -65,6 +69,12 @@ const (
 	// ErrorStatus means the change or task has errored out while running or being undone.
 	ErrorStatus Status = 9
 
+	// WaitStatus means the task was accomplished successfully but some
+	// external event needs to happen before work can progress further
+	// (e.g. on classic we require the user to reboot after a
+	// kernel snap update).
+	WaitStatus Status = 10
+
 	nStatuses = iota
 )
 
@@ -88,6 +98,8 @@ func (s Status) String() string {
 		return "Doing"
 	case DoneStatus:
 		return "Done"
+	case WaitStatus:
+		return "Wait"
 	case AbortStatus:
 		return "Abort"
 	case UndoStatus:
@@ -104,6 +116,18 @@ func (s Status) String() string {
 	panic(fmt.Sprintf("internal error: unknown task status code: %d", s))
 }
 
+// taskWaitComputeStatus is used while computing the wait status of a
+// change. It keeps track of whether a task is waiting or not waiting, or the
+// computation for it is still in-progress to detect cyclic dependencies.
+type taskWaitComputeStatus int
+
+const (
+	taskWaitStatusNotComputed taskWaitComputeStatus = iota
+	taskWaitStatusComputing
+	taskWaitStatusNotWaiting
+	taskWaitStatusWaiting
+)
+
 // Change represents a tracked modification to the system state.
 //
 // The Change provides both the justification for individual tasks
@@ -115,16 +139,16 @@ func (s Status) String() string {
 // while the individual Task values would track the running of
 // the hooks themselves.
 type Change struct {
-	state   *State
-	id      string
-	kind    string
-	summary string
-	status  Status
-	clean   bool
-	data    customData
-	taskIDs []string
-	lanes   int
-	ready   chan struct{}
+	state              *State
+	id                 string
+	kind               string
+	summary            string
+	status             Status
+	clean              bool
+	data               customData
+	taskIDs            []string
+	ready              chan struct{}
+	lastObservedStatus Status
 
 	spawnTime time.Time
 	readyTime time.Time
@@ -157,7 +181,6 @@ type marshalledChange struct {
 	Clean   bool                        `json:"clean,omitempty"`
 	Data    map[string]*json.RawMessage `json:"data,omitempty"`
 	TaskIDs []string                    `json:"task-ids,omitempty"`
-	Lanes   int                         `json:"lanes,omitempty"`
 
 	SpawnTime time.Time  `json:"spawn-time"`
 	ReadyTime *time.Time `json:"ready-time,omitempty"`
@@ -178,7 +201,6 @@ func (c *Change) MarshalJSON() ([]byte, error) {
 		Clean:   c.clean,
 		Data:    c.data,
 		TaskIDs: c.taskIDs,
-		Lanes:   c.lanes,
 
 		SpawnTime: c.spawnTime,
 		ReadyTime: readyTime,
@@ -206,7 +228,6 @@ func (c *Change) UnmarshalJSON(data []byte) error {
 	}
 	c.data = custData
 	c.taskIDs = unmarshalled.TaskIDs
-	c.lanes = unmarshalled.Lanes
 	c.ready = make(chan struct{})
 	c.spawnTime = unmarshalled.SpawnTime
 	if unmarshalled.ReadyTime != nil {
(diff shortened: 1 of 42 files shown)


internals/overlord/state/change.go
```diff
@@ -157,7 +181,6 @@ type marshalledChange struct {
 	Clean   bool                        `json:"clean,omitempty"`
 	Data    map[string]*json.RawMessage `json:"data,omitempty"`
 	TaskIDs []string                    `json:"task-ids,omitempty"`
-	Lanes   int                         `json:"lanes,omitempty"`
```
flotter, line 160: Is this safe for us to just remove? What will be the impact for current systems ?

flotter (commented):
Enormous amount of work, thank you.

I have a couple of first pass comments added with a few questions here:

1. I found that some changes pulled in here belong to commits with additional changes and comments providing the original context. It is extremely difficult for me to review some of the code without the original context explained. I wonder if some of the groups of changes would be better to cherry-pick from the original commit as a whole, e.g:
```
commit 4815875ea89eace1349d30fe863371e251e08ebd
overlord: only initialize loopTomb when needed
```
2. I would break this PR up into smaller logical commits. It makes reviewing easier as one can select only on commit, and use that as context to understand which files are impacted?

3. I wonder if some things like jsonutil and testutil could be moved to ```x-go``` to prevent this exercise from repeating itself on those common code areas?

dmitry-lyfar:
Hi @benhoyt , @flotter (cc @niemeyer).

After a discussion with @flotter  over the comments above, it is suggested to try the following approach for this PR instead:

* Separate changes into smaller logical blocks that can be reviewed and accepted independently. I’ll try to keep them close to ~500 lines.
* Do not rename FakeTime to MockTime as in the snapd’s code base.
* Do not introduce the StateStarterUp interface (i.e. separating the start up logic from a manager's creation in a StartUp method); all Pebble managers initialise swiftly enough at creation. Thus, we'll avoid numerous changes to the Daemon and its existing test suite. 
* Do not introduce [GetMaybeTimings](https://github.com/canonical/pebble/pull/259/files#diff-ebec3d25fa36b396e9ba4ec3a116de67f9938326751aed5f79b9b82df2c44958R487) method to State as it involves another dependency on the snapcore/snapd/timings.GetSaver interface and is not really useful in Pebble. It appears the approaches to measuring spans in Pebble and snapd diverted substantially (perhaps, Pebble [does not save/use](https://github.com/canonical/pebble/blob/e68699380cdc364a7c565162a632a91bb46205b4/internals/overlord/overlord.go#L231) timings now at all).

Thus, the changesets I imagine this PR to be divided into (roughly):
* Extend testutil with testutil.ErrorIs to prepare for the move of the updated state package state, change and task data entry functions. These functions return [NoStateError](https://github.com/snapcore/snapd/blob/cffb170b55841ae8f11343b4019b4f249c3d8736/overlord/state/state.go#L282) with a key name if a data key is not found contrary to the previously used ErrNoState. Thus, the direct comparison of a returned error with ErrNoState instead of errors.Is would fail.
* Extend testutil with testutil.DeepUnsortedMatches. This checker is used by 2 new large TaskRunner unit tests. It’s a useful checker to have for other Pebble unit tests too.
* (Optional) Update osutil’s CanStat -> FileExist and corresponding packages/tests using it. ([CanStat](https://github.com/canonical/pebble/blob/e68699380cdc364a7c565162a632a91bb46205b4/internals/osutil/stat.go#L30), [FileExists](https://github.com/snapcore/snapd/blob/cffb170b55841ae8f11343b4019b4f249c3d8736/osutil/stat.go#L30)). This is something I’m on the fence with as CanStat is used twice in the state’s package code: to load and copy the state. The majority of osutil.CanStat usage is elsewhere in the Pebble’s code base. Should we, instead, just keep the state package to use CanStat and avoid noisy changes?
* Add StartOfOperation time tracking to Overlord so that Overlord can use it to prune outdated changes (the new state package introduces the start of operation time to the state.Prune parameters so a change, when considered for pruning, will use the start of operation time as its spawn time if it was spawned before the latest overlord start). 
* Add new changes from the state package (excluding CopyState functionality). Also, update code and tests that would now fail on state entry functions error checking (patch, overlord, daemon).
* (Optional) Add CopyState function (copy.go from the original state package)  and the required safejson package. This  is not used by Pebble and, given it is a completely isolated piece of functionality, it can now be dropped and introduced when/if required.

Please give me your thoughts on the above. If it looks reasonable, I can go ahead, cancel this PR and send changes in stages.

benhoyt:
Yes, that looks very reasonable to me, thanks @dmitry-lyfar. Some specifics:

* You're right that Pebble doesn't use those span timings at all now, so I think not introducing that is the right call.
* I'm not sure why CanStat diverged from FileExists. There's only an "initial commit" in Pebble, so I suspect Gustavo went through and renamed / cleaned up a few things when he originally copied the Pebble code. I think `CanStat` says more what it's actually checking than `FileExists`, so `CanStat` is probably "better". So apart from the fact that it'd introduce a bunch of noise in the rest of Pebble, I'd leave that one out (and rename `FileExists` to `CanStat` in snapd if anything).
* CopyState -- yeah, it sounds like we should wait with that one for now.

niemeyer:
We talked about this today in our call and agreed to break it down. Some more detailed feedback:

> Separate changes into smaller logical blocks that can be reviewed and accepted independently. 

Sounds good, thank you.

> Do not rename FakeTime to MockTime as in the snapd’s code base.

Yes, we should go the opposite direction: snapd still needs to move towards Fake* terminology.  It was mixed, and when it was moved to Pebble it was standardized.

> Do not introduce the StateStarterUp interface (i.e. separating the start up logic from a manager's creation in a StartUp method); all Pebble managers initialise swiftly enough at creation. Thus, we'll avoid numerous changes to the Daemon and its existing test suite.

We do want to do this. The change in snapd was actually made on my request precisely after Pebble was split, because it makes little sense for machine start/restart logic to live inside a package that is responsible for general state management and work dispatch. So let's please integrate this change too, ideally on its own PR like is being proposed for everything else.

> Do not introduce [GetMaybeTimings](https://github.com/canonical/pebble/pull/259/files#diff-ebec3d25fa36b396e9ba4ec3a116de67f9938326751aed5f79b9b82df2c44958R487) method to State as it involves another 

Why would we not introduce it?  As I remember, we're not making extensive use of timers yet, but this is a nice feature that would be nice to have avaialble, and the fact we're not using this extensively makes the job easier supposedly.

> Thus, the changesets I imagine this PR to be divided into (roughly):

From those descriptions it's not clear what is being proposed that is a feature that does not exist in snapd itself and what is a feature that is just being migrated. So it's harder to agree/disagree. In general: what is being ported from snapd in support of general state logic is good. What is being created inside Pebble itself, or that is being ported from snapd but is unrelated to the state package itself, not so great but we can talk about it. Feel free to provide more details so we can consider specific changes, either in their own PR or in our calls.

For now, I suggest closing this PR.

dmitry-lyfar:
> We do want to do this. The change in snapd was actually made on my request precisely after Pebble was split, because it makes little sense for machine start/restart logic to live inside a package that is responsible for general state management and work dispatch. So let's please integrate this change too, ideally on its own PR like is being proposed for everything else.

Ack.

> > Do not introduce [GetMaybeTimings](https://github.com/canonical/pebble/pull/259/files#diff-ebec3d25fa36b396e9ba4ec3a116de67f9938326751aed5f79b9b82df2c44958R487) method to State as it involves another
> 
> Why would we not introduce it? As I remember, we're not making extensive use of timers yet, but this is a nice feature that would be nice to have avaialble, and the fact we're not using this extensively makes the job easier supposedly.

The interface implementation itself is tiny as it stores/retrieves raw data to the state. It is largerly used by the independent [timings](https://github.com/snapcore/snapd/tree/master/timings) package which I was thinking makes sense to merge with some changes that would use it as it is not directly related to the state management.

I will add the GetSaver implementation now so it can be used later when someone introduces timings and their usage in Pebble later on.

> > Thus, the changesets I imagine this PR to be divided into (roughly):
> 
> From those descriptions it's not clear what is being proposed that is a feature that does not exist in snapd itself and what is a feature that is just being migrated. So it's harder to agree/disagree. In general: what is being ported from snapd in support of general state logic is good. What is being created inside Pebble itself, or that is being ported from snapd but is unrelated to the state package itself, not so great but we can talk about it. Feel free to provide more details so we can consider specific changes, either in their own PR or in our calls.

Sounds good, closing this one now.

PR #261, opened 2023-07-27
client: public API

anpep:
This patch, similar in spirit to #238, introduces a new public API for
extending the Pebble client. The basic idea behind this patch is the client mixin
struct embedded in commands' structs no longer holds an instance of the
Pebble-specific `client.Client` struct, but instead take a `ClientGetter` interface
that implements a `Client()` method. `ClientGetter.Client()` always returns a
Pebble-specific `client.Client` struct.

For applications that intend to extend Pebble, this means that either the Pebble
client or a new, application-specific client can be used. In the latter case,
the application-specific client must implement the `ClientGetter` interface so
that a Pebble-specific `client.Client` struct can always be derived by the
facilities consuming the `ClientGetter` interface. The easiest way to implement
this is to embed the Pebble client:

```go
type PebbleClient = client.Client
type MyClient struct {
    *PebbleClient
}
```

Since the Pebble-specific `client.Client` is embedded, and the `ClientGetter`
interface requires `Client()` to be implemented with a pointer receiver, the
snippet above suffices to implement a client based off the Pebble-supplied one
without much hassle, and that provides lower-level facilities for communicating
with the daemon, such as `DoSync()`, `DoAsync()` and `DoAsyncFull()`.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -87,6 +87,17 @@ type doer interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
+// ClientGetter implementations must provide a way to convert themselves to
+// Pebble client instances.
+type ClientGetter interface {
+	// Client returns a Pebble client instance.
+	Client() *Client
+}
+
+type ClientSetter interface {
+	SetClient(ClientGetter)
+}
+
 // Config allows the user to customize client behavior.
 type Config struct {
 	// BaseURL contains the base URL where the Pebble daemon is expected to be.
@@ -116,6 +127,10 @@ type Client struct {
 	warningTimestamp time.Time
 
 	getWebsocket getWebsocketFunc
+}
+
+func (c *Client) Client() *Client {
+	return c
 }
 
 type getWebsocketFunc func(url string) (clientWebsocket, error)
@@ -326,13 +341,27 @@ func decodeInto(reader io.Reader, v interface{}) error {
 	return nil
 }
 
-// doSync performs a request to the given path using the specified HTTP method.
+// RequestInfo holds the information to perform a request to the daemon.
+type RequestInfo struct {
+	Method  string
+	Path    string
+	Query   url.Values
+	Headers map[string]string
+	Body    io.Reader
+}
+
+// ResultInfo is empty for now, but this is the mechanism that conveys
+// general information that makes sense to requests at a more general
+// level, and might be disconnected from the specific request at hand.
+type ResultInfo struct{}
+
+// DoSync performs a request to the given path using the specified HTTP method.
 // It expects a "sync" response from the API and on success decodes the JSON
 // response payload into the given value using the "UseNumber" json decoding
 // which produces json.Numbers instead of float64 types for numbers.
-func (client *Client) doSync(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) (*ResultInfo, error) {
+func (client *Client) DoSync(req *RequestInfo, v interface{}) (*ResultInfo, error) {
 	var rsp response
-	if err := client.do(method, path, query, headers, body, &rsp); err != nil {
+	if err := client.do(req.Method, req.Path, req.Query, req.Headers, req.Body, &rsp); err != nil {
 		return nil, err
 	}
 	if err := rsp.err(client); err != nil {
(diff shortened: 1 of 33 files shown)


benhoyt:
@anpep Would you be able to give more context and description about what problem this change is trying to solve, and how it does that? (And specifics like why `ContextGetter` is needed, that kind of thing.) That would be really helpful for reviewers.

internals/cli/cli.go
```diff
@@ -141,15 +141,15 @@ func fixupArg(optName string) string {
...
 
 type clientSetter interface {
-	setClient(*client.Client)
+	setClient(client.ClientGetter)
 }
 
-type clientMixin struct {
-	client *client.Client
+type ClientMixin struct {
+	client.ClientGetter
```
flotter, line 148: For other reviewers: This is essentially the heart of the proposal. We need to change the CLI to accept an Client Interface, not only the Pebble Client structure. This will allow another pebble derivative to embed the Pebble Client struct and extend it with new options, and then initialise the Parser with the extended version.

internals/cli/cli.go
```diff
@@ -141,15 +141,15 @@ func fixupArg(optName string) string {
...
 }
 
-func (ch *clientMixin) setClient(cli *client.Client) {
-	ch.client = cli
+func (ch *ClientMixin) setClient(cg client.ClientGetter) {
+	ch.ClientGetter = cg
```
flotter, line 152: Do we need a GetClient() to get the address of the derivative Client based structure on the Pebble derivative CLI command side?

client/client.go
```diff
@@ -326,11 +335,11 @@ func decodeInto(reader io.Reader, v interface{}) error {
 	return nil
 }
 
-// doSync performs a request to the given path using the specified HTTP method.
+// DoSync performs a request to the given path using the specified HTTP method.
 // It expects a "sync" response from the API and on success decodes the JSON
 // response payload into the given value using the "UseNumber" json decoding
 // which produces json.Numbers instead of float64 types for numbers.
-func (client *Client) doSync(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) (*ResultInfo, error) {
+func (client *Client) DoSync(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) (*ResultInfo, error) {
```
benhoyt, line 342: Why does `DoSync` need to be exported?

client/client.go
```diff
@@ -87,6 +87,11 @@ type doer interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
+type ClientGetter interface {
```
benhoyt, line 90: Normally in Go you'd define the interface type only where it's needed/consumed, for example in `internals/cli/cli.go`.

benhoyt (commented):
I don't understand how this helps. The `ClientMixin` still uses the concrete client type, it's just behind a method which returns a concrete `*client.Client`. Can you show a small example of how this would be used to explain how it helps?

client/client.go
anpep: Please re-review this PR since it makes much more sense now.
Answering to your concrete question, applications extending the Pebble client will definitely want these utility functions to be available in order to communicate with the Pebble daemon.

flotter (commented):
@anpep I did a detailed review as a patch on your proposal and sent it to you. I did it this way to test the proposal as its very difficult without the kernos side :)

benhoyt:
I think the changes to export `DoSync` (and similar methods) and the addition of `RequestInfo` are good.

However, I don't like the `ClientGetter` / `ClientSetter` stuff at all. It seems very indirect and implicit, and hard to follow how things are being wired up. This isn't really your fault -- you're extending the existing `clientMixin` / `clientSetter` pattern, which I've always found a bit messy and hard to follow anyway: you add this `clientMixin` to your command struct, and then because it implements `setClient` (determined by type assertion at runtime), the parser knows to call it, which magically fills in the `client` field for the command's `Execute` to use.

Side notes: it's also weird that a new client is created even for commands that don't need it (though it looks like `help` is the only one like that right now, so perhaps a bit of a theoretical concern). I also don't like how something called a "parser" is not just parsing, but *executing*, the commands -- but that's really a separate, go-flags issue.

I wonder if instead of getting further entrenched in this hard-to-understand implicit wiring, we can create and wire up the client more directly. I think there are several ways we could do this, but I think the most obvious is just to create the client in the command's `Execute` method. This is how `run` works with the daemon: the `Daemon` is created inside `Execute` (it's a little nested, but it is direct: Execute > run > runDaemon > daemon.New).

For example, consider `pebble changes`:

```go
func (c *cmdChanges) Execute(args []string) error {
	// These few lines are the new code:
	cl, err := client.New(clientConfigFromEnv())
	if err != nil {
		return err
	}
	defer maybePresentWarnings(cl)

	// This is existing code:
	changes, err := queryChanges(cl, &opts)
	if err != nil {
		return err
	}
	// ...
}
```

The `clientConfigFromEnv` function would be like `getEnvPaths`, but build a `*client.Config` from environment variables. I realise this would need to be overridden by another program like Termus that used Pebble. But there are various simple ways to do that, for example adding `EnvDir` and `EnvSocket` to the [personality struct](https://github.com/canonical/pebble/pull/238) to allow you to change the names of the environment variables, and default them to `PEBBLE` and `PEBBLE_SOCKET`.

The `maybePresentWarnings` is something that's currently in `Run` that most commands do at the end -- if there were warnings, they show `WARNING: There are %d new warnings. See 'pebble warnings'.`. So we need to keep that (and may need to make it exported for Termus et al).

This approach means a few lines of boilerplate in every `Execute` method, but it's clear and explicit, and it gets rid of the `clientMixin` / `setClient` messiness, as well as allowing you to create whatever type of client you want on Termus `Execute` methods.

What do you think? I have Tuesday evening meetings (EU morning time) today anyway, so I'm going to try to schedule a meeting today for us to chat further.

anpep:
Closing because this approach was ruled out (:

PR #265, opened 2023-07-28
daemon: allowing adding external http handlers

paul-rodriguez:
This PR turns the daemon API into an exported variable.

External tools depending on pebble may append to the variable to extend the daemon API, see snippet below.

```
type fakeHandler struct {
	cmd        *daemon.Command
	lastMethod string
}

func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	h.lastMethod = r.Method
	fmt.Println("Fake handler called")
}

func main() {

	var handler fakeHandler

	getCallback := func(c *daemon.Command, r *http.Request, s *daemon.UserState) daemon.Response {
		handler.cmd = c
		fmt.Println("getCallback called")
		return &handler
	}

	command := daemon.Command{
		Path:    "/v1/addedendpoint",
		GuestOK: true,
		GET:     getCallback,
	}
	
	// Here's the relevant line:
	daemon.API = append(daemon.API, &command)

	if err := cli.Run(); err != nil {
		fmt.Fprintf(cli.Stderr, "error: %v", err)
		os.Exit(1)
	}
	
	// Executable now handles requests for "/v1/addedendpoint"
}
```


internals/daemon/api.go
```diff
@@ -117,3 +117,7 @@ func v1SystemInfo(c *Command, r *http.Request, _ *userState) Response {
 	}
 	return SyncResponse(result)
 }
+
+func Register(c *Command) {
+	api = append(api, c)
+}
```
anpep, line 123: Please have a look at this snippet:
https://github.com/canonical/pebble/blob/00bcd1f7dda570bcdf42600545ddc1707d24365e/internals/daemon/api.go#L27-L97

I had a similar issue while coming up with an API for adding commands to the CLI. Initially, I came up with the `cli.AddCommand()` method, but my proposal was rejected. One of the major points that came up were that it would be easier to just export the `Command`s slice. This way, on applications extending Pebble, we could have a slice of `Command`s at the top side of the `main.go` file, exposing all the commands' information in a single place and acting as a reference that clearly exposes all available commands upfront to the developer.

I'd like to see a similar approach here. In applications extending Pebble, we should be able to just create a `Command` slice that we can append to the Pebble one that exposes all the custom API endpoints like in the snippet above, so that it is immediately obvious to developers what's in the box (:

internals/daemon/daemon.go
```diff
@@ -109,10 +109,10 @@ type Daemon struct {
 }
 
 // XXX Placeholder for now.
```
anpep, line 111: If we are exporting this, this shouldn't be a placeholder. I'd be reluctant to just strip it since it's used everywhere and seems valuable to store e.g. authenticated user information in the future. Curious to hear Ben's and Gustavo's opinion on this one, but I'd just include a brief comment actually explaining why this exists.

See this snippet from the client code:
https://github.com/canonical/pebble/blob/00bcd1f7dda570bcdf42600545ddc1707d24365e/client/client.go#L384-L387

anpep (changes requested):

internals/daemon/daemon.go
paul-rodriguez: You're talking about UserState right?  And yes at least in snapd it is used to persist authentication information across requests, so I will mention this.

anpep (approved):

internals/daemon/daemon_test.go
```diff
@@ -98,6 +98,29 @@ func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.lastMethod = r.Method
 }
 
+func (s *daemonSuite) TestRegisterCommand(c *C) {
```
flotter, line 101: As discussed, register -> add change in general as we are now exposing Api list.

flotter (approved):

benhoyt (approved):
Approved, but note we should change Api -> API

internals/daemon/daemon_test.go
```diff
@@ -98,6 +98,29 @@ func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	getCallback := func(c *Command, r *http.Request, s *UserState) Response {
+		handler.cmd = c
+		return &handler
+	}
+	command := Command{
+		Path:    endpoint,
+		GuestOK: true,
+		GET:     getCallback,
+	}
+	API = append(API, &command)
```
niemeyer, line 113: This needs to be removed in a defer so this is not polluting the environment for other tests.

niemeyer (approved):
Looks good. Just one detail needs to be fixed on the test.

flotter:
Gustavo reviewed  & approved, pending **comment fix** and **conflict resolved**

internals/daemon/daemon_test.go
```diff
@@ -110,7 +110,9 @@ func (s *daemonSuite) TestAddCommand(c *C) {
 		GuestOK: true,
 		GET:     getCallback,
 	}
+
 	API = append(API, &command)
+	defer func() { API = API[:len(API)-1] }()
```
flotter, line 116: Discussed with Paul. Would personally have favoured a more generic setup (backup) and tear-down (restore) approach, but given that modifying the API in tests at this point is unlikely going to common in the test suite, I am OK with this.

flotter (approved):

internals/daemon/daemon_test.go
niemeyer: Can we please do a slightly more careful job here and pop out &command, specifically?

paul-rodriguez: I've changed this to use Fred's solution as it doesn't require an additional search + delete function

internals/daemon/daemon_test.go
```diff
@@ -54,6 +54,7 @@ type daemonSuite struct {
 	err             error
 	notified        []string
 	restoreBackends func()
+	savedAPI        []*Command
```
flotter, line 57: How about:
```
type daemonSuite struct {
:
restoreAPI func()
```

```
func (s *daemonSuite) SetUpSuite(c *check.C) {
    backup := append([]*Command{}, API...)
    restoreAPI = func() {
        API = append([]*Command{}, backup...)
    }
}
```
and then finally in ```TearDownTest```:

```
func (s *daemonSuite) TearDownTest(c *C) {
    s.restoreAPI()
}
```

internals/daemon/daemon_test.go
```diff
@@ -67,9 +68,15 @@ func (s *daemonSuite) SetUpTest(c *C) {
...
 	}
+	if s.savedAPI == nil {
+		s.savedAPI = make([]*Command, len(API))
+		copy(s.savedAPI, API)
+	}
 }
 
 func (s *daemonSuite) TearDownTest(c *C) {
+	copy(API, s.savedAPI)
+	API = API[:len(s.savedAPI)]
```
niemeyer, line 79: This logic is incorrect if len(API) < len(s.savedAPI). Also, it seems a bit wasteful to be doing this when this is only touched by one trivial test.  Can we please go back to the prior approach, and simply drop the one offending entry that the test is adding?  We don't even need to iterate... we can just panic if that last element isn't the expected one, and then drop the last element.

niemeyer (changes requested):
Sorry, but still has issues. :(

internals/daemon/daemon_test.go
paul-rodriguez: Should be ok now.

niemeyer (approved):
Thanks!

PR #266, opened 2023-07-28
servstate: test stability improvements

flotter:
**servstate: backoff svcs should be stopped on exit**

The default behaviour today for a service that exits is that Pebble will
auto-restart the service. This process has a built in back-off strategy
which means that on exit a timer is set to restart the service after
the back-off period expires.

The way to prevent a service in back-off to restart is to explicitly stop
the service (this case is covered in a unit test).

When the Pebble daemon starts a graceful exit, it has to first stop all
running services by requesting this explicitly, and waiting for the change
request to complete.

However, services currently scheduled for restart (back-off) is not currently
stopped through this request. This is an issue because the restart mechanism
relies on the reaper package, which is stopped straight after running services
are stopped.

This means any service in back-off can find itself in one of the following
situations:

1. The most common case is that the service restart thread is killed by the
   garbage collector on process exit, so this appears clean.

2. The second scenario is the service restarts between the services stop
   request and the reaper kill request, in which case the service will
   continue to run after the Pebble process is dead (of course this is
   statistically very unlikely to happen, but possible).

3. The last scenario is that the service restart after the reaper is killed,
   but before the Pebble process exits. This will result in the "reaper not
   started" panic we often see in unit tests (this is also a statistically
   unlikely scenario, but possible).

This PR modifies the servstate function used for terminating all services
during shutdown (only) to also explicitly stop services in back-off.

**servstate: unit test teardown must stop svcs**

Currently running unit test services are overlapping each other as there
is no explicit services stop on test tear-down. This is not directly a
problem although it makes interpreting the unit test output impossible,
as the servstate messages you see during the test also includes output
from service events of previous tests.

There is however one real issue: services that restart after a
unit test exit (due to expiring back-off), requires the original
test reaper instance to start and wait on the service. The reaper is
correctly destroyed during test teardown, so as a result we constantly
get error messages like:

```panic: reaper not started```

This PR includes the following work:

- Reorganise the test setup and tear-down logic to allow consistent
  management of servstate manager instances (some tests use a default
  setup template, while others create a test specific instance).

- Enforce clean service shutdown (running and back-off) during test
  teardown.

- Arrange the unit tests to be first in the file, with helper
  functions at the end, to allow for easy reading.

**servstate: remove test races on svcs file output**

Unit Tests for servstate is extremely timing sensitive. If you run the unit
tests under heavy CPU load, changes are very good you will start to see
various failures. The easiest way to demonstrate this is to use the stress
utility to simulate heavy CPU load (stress --io 12 --cpu 12).

Unit Test output snippet:

```
START: manager_test.go:828: S.TestEnvironment

... obtained string = ""
... expected string = "" +
...     "PEBBLE_ENV_TEST_1=foo\n" +
...     "PEBBLE_ENV_TEST_2=bar bazz\n" +
...     "PEBBLE_ENV_TEST_PARENT=from-parent\n"

FAIL: manager_test.go:828: S.TestEnvironment
```

This specific race condition is caused because unit tests often make
assumptions on how long it takes from the moment the service is started
until the moment the service command completes.

This PR will only address this very specific kind of race, and hopefully
in the process provide a solution that will be used in future tests.

Please note that using the stress utility on tests from this package will
still trigger lots of remaining races caused by very timing sensitive tests.

**servstate: unify log assert and done check**

The startTestServices() test helper uses a special entry in the
service command under test to write the service standard output
also to a log file that can be inspected.

This mechanism suffers from a race condition (as highlighted in
https://github.com/canonical/pebble/issues/264) because when the
content of the log file is loaded, the service may not yet have
completed writing to the log.

Since standard output is also verified separately through a
different mechanism, the follservstate: unify log assert and done check

The startTestServices() test helper uses a special entry in the
service command under test to write the service standard output
also to a log file that can be inspected.

This mechanism suffers from a race condition (as highlighted in
https://github.com/canonical/pebble/issues/264) because when the
content of the log file is loaded, the service may not yet have
completed writing to the log.

Since standard output is also verified separately through a
different mechanism, the following changes are made:

- Enhance the global "done check" (previous called "done file") to
  check completion per service. This effectively adds the
  capability previously provided by the log assert mechanism.

- Use the existing "done check" mechanism to wait until the service
  command-line is complete up to the point of the check.

- Only now verify the stdout buffer content as checked previously.

- Remove the log mechanism all together.

### DIFF
--- a/internals/overlord/servstate/manager.go
+++ b/internals/overlord/servstate/manager.go
@@ -548,7 +548,14 @@ func (m *ServiceManager) SetServiceArgs(serviceArgs map[string][]string) error {
 	return m.appendLayer(newLayer)
 }
 
-// servicesToStop returns a slice of service names to stop, in dependency order.
+// servicesToStop is used during service manager shutdown to cleanly terminate
+// all running services. Running services include both services in the
+// stateRunning and stateBackoff, since a service in backoff state can start
+// running once the timeout expires, which creates a race on service manager
+// exit. If it starts just before, it would continue to run after the service
+// manager is terminated. If it starts just after (before the main process
+// exits), it would generate a runtime error as the reaper would already be dead.
+// This function returns a slice of service names to stop, in dependency order.
 func servicesToStop(m *ServiceManager) ([]string, error) {
 	releasePlan, err := m.acquirePlan()
 	if err != nil {
@@ -568,15 +575,15 @@ func servicesToStop(m *ServiceManager) ([]string, error) {
 		return nil, err
 	}
 
-	// Filter down to only those that are running.
+	// Filter down to only those that are running or in backoff
 	m.servicesLock.Lock()
 	defer m.servicesLock.Unlock()
-	var running []string
+	var notStopped []string
 	for _, name := range stop {
 		s := m.services[name]
-		if s != nil && s.state == stateRunning {
-			running = append(running, name)
+		if s != nil && (s.state == stateRunning || s.state == stateBackoff) {
+			notStopped = append(notStopped, name)
 		}
 	}
-	return running, nil
+	return notStopped, nil
 }
--- a/internals/overlord/servstate/manager_test.go
+++ b/internals/overlord/servstate/manager_test.go
@@ -77,6 +77,7 @@ type S struct {
 	log          string
 	logBuffer    bytes.Buffer
 	logBufferMut sync.Mutex
+	logOutput    writerFunc
 
 	st *state.State
 
@@ -148,20 +149,10 @@ func (s *S) SetUpTest(c *C) {
 	s.BaseTest.SetUpTest(c)
 	s.st = state.New(nil)
 
-	s.dir = c.MkDir()
-	os.Mkdir(filepath.Join(s.dir, "layers"), 0755)
-
-	s.log = filepath.Join(s.dir, "log.txt")
-	data := fmt.Sprintf(planLayer1, s.log, s.log)
-	err := ioutil.WriteFile(filepath.Join(s.dir, "layers", "001-base.yaml"), []byte(data), 0644)
-	c.Assert(err, IsNil)
-	err = ioutil.WriteFile(filepath.Join(s.dir, "layers", "002-two.yaml"), []byte(planLayer2), 0644)
-	c.Assert(err, IsNil)
-
 	s.logBufferMut.Lock()
 	s.logBuffer.Reset()
 	s.logBufferMut.Unlock()
-	logOutput := writerFunc(func(p []byte) (int, error) {
+	s.logOutput = writerFunc(func(p []byte) (int, error) {
 		s.logBufferMut.Lock()
 		defer s.logBufferMut.Unlock()
 		return s.logBuffer.Write(p)
@@ -169,10 +160,6 @@ func (s *S) SetUpTest(c *C) {
 
 	s.runner = state.NewTaskRunner(s.st)
 	s.stopDaemon = make(chan struct{})
-	manager, err := servstate.NewManager(s.st, s.runner, s.dir, logOutput, testRestarter{s.stopDaemon}, fakeLogManager{})
-	c.Assert(err, IsNil)
-	s.AddCleanup(manager.Stop)
-	s.manager = manager
 
 	restore := servstate.FakeOkayWait(shortOkayDelay)
 	s.AddCleanup(restore)
@@ -181,91 +168,23 @@ func (s *S) SetUpTest(c *C) {
 }
 
 func (s *S) TearDownTest(c *C) {
+	// Make sure all services are stopped
+	s.stopRunningServices(c)
+	// Kill the reaper
+	s.manager.Stop()
+	// General test cleanup
 	s.BaseTest.TearDownTest(c)
 }
 
-type testRestarter struct {
-	ch chan struct{}
-}
-
-func (r testRestarter) HandleRestart(t restart.RestartType) {
-	close(r.ch)
-}
-
-func (s *S) assertLog(c *C, expected string) {
-	s.logBufferMut.Lock()
-	defer s.logBufferMut.Unlock()
-	data, err := ioutil.ReadFile(s.log)
-	if os.IsNotExist(err) {
-		c.Fatal("Services have not run")
-	}
-	c.Assert(err, IsNil)
-	c.Assert(string(data), Matches, "(?s)"+expected)
-	c.Assert(s.logBuffer.String(), Matches, "(?s)"+expected)
-}
-
-func (s *S) logBufferString() string {
-	s.logBufferMut.Lock()
-	defer s.logBufferMut.Unlock()
-	str := s.logBuffer.String()
-	s.logBuffer.Reset()
-	return str
-}
-
 func (s *S) TestDefaultServiceNames(c *C) {
+	s.setupDefaultServiceManager(c)
 	services, err := s.manager.DefaultServiceNames()
 	c.Assert(err, IsNil)
 	c.Assert(services, DeepEquals, []string{"test1", "test2"})
 }
 
-func (s *S) ensure(c *C, n int) {
-	for i := 0; i < n; i++ {
-		s.runner.Ensure()
-		s.runner.Wait()
-	}
-}
-
-func (s *S) startServices(c *C, services []string, nEnsure int) *state.Change {
-	s.st.Lock()
-	ts, err := servstate.Start(s.st, services)
-	c.Check(err, IsNil)
-	chg := s.st.NewChange("test", "Start test")
-	chg.AddAll(ts)
-	s.st.Unlock()
-
-	// Num to ensure may be more than one due to the cross-task dependencies.
-	s.ensure(c, nEnsure)
-
-	return chg
-}
-
-func (s *S) stopServices(c *C, services []string, nEnsure int) *state.Change {
-	s.st.Lock()
-	ts, err := servstate.Stop(s.st, services)
-	c.Check(err, IsNil)
-	chg := s.st.NewChange("test", "Stop test")
-	chg.AddAll(ts)
-	s.st.Unlock()
-
-	// Num to ensure may be more than one due to the cross-task dependencies.
-	s.ensure(c, nEnsure)
-
-	return chg
-}
-
-func (s *S) startTestServices(c *C) {
-	chg := s.startServices(c, []string{"test1", "test2"}, 2)
-	s.st.Lock()
-	c.Check(chg.Status(), Equals, state.DoneStatus, Commentf("Error: %v", chg.Err()))
-	s.st.Unlock()
-
-	s.assertLog(c, ".*test1\n.*test2\n")
-
-	cmds := s.manager.RunningCmds()
-	c.Check(cmds, HasLen, 2)
-}
-
 func (s *S) TestStartStopServices(c *C) {
+	s.setupDefaultServiceManager(c)
 	s.startTestServices(c)
 
 	if c.Failed() {
@@ -276,6 +195,7 @@ func (s *S) TestStartStopServices(c *C) {
 }
 
 func (s *S) TestStartStopServicesIdempotency(c *C) {
+	s.setupDefaultServiceManager(c)
 	s.startTestServices(c)
 	if c.Failed() {
 		return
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: internals/overlord/servstate/manager.go [hunk 1]
```diff
@@ -548,7 +548,13 @@ func (m *ServiceManager) SetServiceArgs(serviceArgs map[string][]string) error {
 	return m.appendLayer(newLayer)
 }
 
-// servicesToStop returns a slice of service names to stop, in dependency order.
+// servicesToStop is used during service manager shutdown to cleanly terminate
```
benhoyt, line 551: This is a great comment, however, it now doesn't say what the return value means (a slice of service names to stop, in dependency order). Maybe prefix the new comment with the old one?

internals/overlord/servstate/manager_test.go
```diff
@@ -525,24 +395,28 @@ func (s *S) TestCurrentUserGroup(c *C) {
...
+	err = s.manager.AppendLayer(layer)
 	c.Assert(err, IsNil)
 
 	chg := s.startServices(c, []string{"usrtest"}, 1)
 	s.st.Lock()
 	c.Assert(chg.Err(), IsNil)
 	s.st.Unlock()
 
+	// Let's wait until we know the output file is available
+	s.waitForFileToAppear(c)
```
benhoyt, line 411: What about calling this `waitForDoneFile`, so its not ambiguous what file it's waiting for? And we probably don't need this comment every time the function is called -- a clear name should be enough.

internals/overlord/servstate/manager_test.go
```diff
@@ -1160,22 +1013,19 @@ checks:
...
-		}
-		b, _ := ioutil.ReadFile(tempFile)
-		if string(b) == "x\n" {
-			break
-		}
-		time.Sleep(5 * time.Millisecond)
-	}
+
+	// Let's wait until we know the output file is available
+	s.waitForFileToAppear(c)
```
benhoyt, line 1024: Yeah, this is much simpler than the for loop approach before.

internals/overlord/servstate/manager_test.go
```diff
@@ -506,14 +377,13 @@ func (s *S) TestStartBadCommand(c *C) {
 }
 
 func (s *S) TestCurrentUserGroup(c *C) {
-	// Don't re-use s.manager, because we're adding a layer and service, and
-	// that would conflict with other tests like TestServiceLogs.
+	// This test supplies its own service manager instance and
+	// therefore should not call the default setup.
```
benhoyt, line 381: This comment (also below) is probably unnecessary now. All tests supply their own service manager -- some via `defaultServiceManagerSetup` and some inline.

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+	// Let's wait until we know the output file is available
+	s.waitForFileToAppear(c)
+
 	output, err := ioutil.ReadFile(outputPath)
 	c.Assert(err, IsNil)
 	c.Check(string(output), Equals, dir+"\n")
 }
+
+// Default service manager setup
+func (s *S) defaultServiceManagerSetup(c *C) {
```
benhoyt, line 1637: Two nits: 1) the comment is unnecessary as it just repeats the name, 2) I think the name would be clearer as `setupDefaultServiceManager` (a verb).

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+
+	s.manager, err = servstate.NewManager(s.st, s.runner, s.dir, s.logOutput, testRestarter{s.stopDaemon}, fakeLogManager{})
+	c.Assert(err, IsNil)
+}
+
+// Return on timeout or when the file appears. This is used to determine
+// when the expected service output is actually available, not when the
+// service starts to run.
+func (s *S) waitForFileToAppear(c *C) {
+	timeout := time.After(time.Duration(time.Second * 10))
```
benhoyt, line 1655: No need for `time.Duration(...)` conversion (similar below). Also slightly more common to have the number before the unit (like you write quantities in physics), like `time.After(10 * time.Second)`.

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+found:
+	for {
+		select {
+		case <-timeout:
+			c.Fatal("timeout waiting for service output")
+		default:
```
benhoyt, line 1661: Probably a bit more natural to use a `time.Ticker` here, rather than a `default` clause with a `time.Sleep`:

```go
timeout := time.After(10 * time.Second)
ticker := time.NewTicker(time.Millisecond)
defer ticker.Stop()
for {
    select {
    case <-timeout:
        ...
    case <-ticker.C:
        ...
    }
}
```

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+			stat, err := os.Stat(s.donePath)
+			if err == nil && stat.Mode().IsRegular() {
+				// Delete it so we can reuse this feature
+				// in the same test.
+				os.Remove(s.donePath)
+				break found
```
benhoyt, line 1667: Could avoid the labelled break by just doing `return` here.

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+		}
+	}
+}
+
+// Make sure services are all stopped before the next test starts.
+func (s *S) stopRunningServices(c *C) {
+	taskSet, err := servstate.StopRunning(s.st, s.manager)
+	if err != nil {
+		c.Fatal("error stopping running test services")
+		return
```
benhoyt, line 1679: Shouldn't need `return` after a `c.Fatal` call. Could this whole `if` block be `c.Assert(err, IsNil)`?

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+	}
+	if taskSet == nil {
+		return
+	}
+
+	// One change to stop them all.
```
benhoyt, line 1685: One change to find them. One change to bring them all, and in the darkness bind them!

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+	chg.AddAll(taskSet)
+	s.st.EnsureBefore(0) // start operation right away
+	s.st.Unlock()
+
+	// Wait for a limited amount of time for them to stop.
+	timeout := time.After(time.Duration(time.Second * 10))
+stop:
+	for {
+		s.runner.Ensure()
+		s.runner.Wait()
```
benhoyt, line 1697: Why are these needed? And won't `s.runner.Wait()` mean that after this call it will always be ready? See for example https://github.com/canonical/pebble/blob/00bcd1f7dda570bcdf42600545ddc1707d24365e/internals/daemon/api_changes.go#L204-L212

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+		// Exit loop if change is complete
+		select {
+		case <-chg.Ready():
+			break stop
+		case <-timeout:
+			c.Fatal("timeout waiting for services to stop")
+		default:
+		}
+	}
+	return
```
benhoyt, line 1708: No need for `return` at the end of a function.

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+		s.runner.Ensure()
+		s.runner.Wait()
+
+		// Exit loop if change is complete
+		select {
+		case <-chg.Ready():
+			break stop
+		case <-timeout:
+			c.Fatal("timeout waiting for services to stop")
+		default:
```
benhoyt, line 1705: If what I said above is true, we shouldn't need the `default` case here.

internals/overlord/servstate/manager_test.go
```diff
@@ -1753,72 +1560,363 @@ func (s *S) TestStopRunning(c *C) {
...
+	timeout := time.After(time.Duration(time.Second * 10))
+stop:
+	for {
+		s.runner.Ensure()
+		s.runner.Wait()
+
+		// Exit loop if change is complete
+		select {
+		case <-chg.Ready():
+			break stop
```
benhoyt, line 1702: Could again avoid the labelled break with a `return` here.

benhoyt (commented):
This is great, thanks for the TLC, @flotter! I'd been thinking of this kind of cleanup to this test file for a while, but had never gotten to it.

I've added a bunch of comments, but they're all nits / minor.

One thing I wondered about was whether `S` should have the fields like `manager` at all. Maybe it'd be clearer if the setup method just returned the manager, so everything is scoped to just the test at hand. I'm guessing it's a bit awkward because then you have to pass around the log buffer and state and runner to the relevant funcs. But I'm interested in your take. I'm happy leaving that too (forever or for future work) -- it's fine to make incremental progress.

internals/overlord/servstate/manager_test.go
flotter: Like your proposal - applying everywhere.

internals/overlord/servstate/manager_test.go
flotter: Thank you.

internals/overlord/servstate/manager_test.go
flotter: Done. Thanks.

internals/overlord/servstate/manager_test.go
flotter: Fixed in two places. Thank you.

internals/overlord/servstate/manager_test.go
flotter: Improved. Thanks.

internals/overlord/servstate/manager_test.go
flotter: Yes, doh.

internals/overlord/servstate/manager_test.go
flotter: Yes, Thanks.

internals/overlord/servstate/manager_test.go
flotter: Doh! Sorry that was bad.

internals/overlord/servstate/manager_test.go
flotter: I will come back with a more concrete answer here after studying the code more. However, the hint is in the ensure() function used already in the tests:

```
func (s *S) ensure(c *C, n int) {
        for i := 0; i < n; i++ {
                s.runner.Ensure()
                s.runner.Wait()
        }
}
```

Basically, a change consists of a taskset, which can consist of a diverse set of tasks with cross task deps. In the normal Pebble world the overlord dedicated loop iterates over multiple ensures to complete a change. In the test world, no loop exist, so we either have to know how many ensure iterations we need, e.g.:

```
        // Num to ensure may be more than one due to the cross-task dependencies.
        s.ensure(c, nEnsure)
```

Or we have to loop until the change is done.
```

File: internals/overlord/servstate/manager_test.go (hunk 14 above)
Comment (flotter) [reply internals/overlord/servstate/manager_test.go:1702]: Done! Thanks.

File: internals/overlord/servstate/manager_test.go (hunk 10 above)
Comment (flotter) [reply internals/overlord/servstate/manager_test.go:1685]: :-)

File: internals/overlord/servstate/manager_test.go [hunk 15]
```diff
@@ -148,124 +151,46 @@ func (s *S) SetUpTest(c *C) {
...
+	// Sync feature to ensure service action is complete. See
+	// waitForFileToAppear()
+	s.donePath = filepath.Join(c.MkDir(), "done")
 }
 
 func (s *S) TearDownTest(c *C) {
+	// Make sure all services are stopped
+	s.stopRunningServices(c)
+	// Kill the reaper
+	s.manager.Stop()
```
paul-rodriguez, line 180: Wouldn't inverting those two steps also have fixed the issue?

Looking deeper, it looks like stopping the manager first is necessary, otherwise you can still get races inside `stopRunningServices`.

internals/overlord/servstate/manager.go
```diff
@@ -568,15 +574,15 @@ func servicesToStop(m *ServiceManager) ([]string, error) {
...
+	// Filter down to only those that are running or in backoff
 	m.servicesLock.Lock()
 	defer m.servicesLock.Unlock()
-	var running []string
+	var notStopped []string
 	for _, name := range stop {
 		s := m.services[name]
-		if s != nil && s.state == stateRunning {
-			running = append(running, name)
+		if s != nil && (s.state == stateRunning || s.state == stateBackoff) {
```
paul-rodriguez, line 583: With this test, it looks like it's possible to miss services which are in the process of being restarted by the reaper (`stateStarting`), or in the process of terminating (`stateTerminating` and `stateKilling`), which would then go into backoff.

paul-rodriguez (commented):
Reaper vs stopRunningServices race?

internals/overlord/servstate/manager_test.go
flotter: @benhoyt let me know if my explanation does not convince you, or if I missed your point.

internals/overlord/servstate/manager_test.go
flotter: If you agree with the comments above, I think the ```default``` is needed to allow iterating to completion as fast as possible.

internals/overlord/servstate/manager.go
flotter:
```
        m.servicesLock.Lock()
        defer m.servicesLock.Unlock()
        var notStopped []string
        for _, name := range stop {
                s := m.services[name]
                if s != nil && (s.state == stateRunning || s.state == stateBackoff) {
                        notStopped = append(notStopped, name)
                }
        }
        return notStopped, nil

```

Note the first two lines - its a servstate wide lock that ensures only one thread is changing the captured state of any service at any one time.

paul-rodriguez:
Thanks for the clarifications ! :green_circle:

internals/overlord/servstate/manager_test.go
benhoyt: Ah, great -- thanks for the explanation!

benhoyt (approved):
Thanks! This is a great maintainability and reliability improvement.

benhoyt:
@flotter Do you think this is ready to merge? If so, I can go ahead and do that.

PR #269, opened 2023-08-02
Run unit tests only on Go 1.20

barrettj12:
As established in #171, Pebble no longer needs to support Go versions before Go 1.20. Hence, it no longer makes sense to run the unit tests with Go versions before 1.20.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,16 +12,17 @@ jobs:
 
     strategy:
       fail-fast: false
+      matrix:
+        go: ['1.20']
 
-    name: Go unit tests
-
+    name: Go ${{ matrix.go }}
     steps:
     - uses: actions/checkout@v3
 
     - name: Set up Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
-        go-version-file: 'go.mod'
+        go-version: ${{ matrix.go }}
 
     - name: Test and build
       run: |
@@ -32,14 +33,13 @@ jobs:
     runs-on: ubuntu-latest
 
     name: Root Tests
-
     steps:
     - uses: actions/checkout@v3
 
     - name: Set up Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
-        go-version-file: 'go.mod'
+        go-version: ${{ matrix.go }}
 
     - name: Test
       run: |
@@ -55,12 +55,11 @@ jobs:
     runs-on: ubuntu-latest
 
     name: Format check
-
     steps:
     - uses: actions/checkout@v3
 
     - name: Set up Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
         go-version-file: 'go.mod'
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,10 +12,8 @@ jobs:
 
     strategy:
       fail-fast: false
-      matrix:
-        go: ['1.20', '1.19', '1.18', '1.17']
 
-    name: Go ${{ matrix.go }}
+    name: Go unit tests
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -12,18 +12,16 @@ jobs:
 
     strategy:
       fail-fast: false
-      matrix:
-        go: ['1.20', '1.19', '1.18', '1.17']
```
benhoyt, line 16: I think we should leave the matrix in here. We'll want to run on Go 1.21 when it comes out (in the next week or two?).

barrettj12: Left comments on Mattermost, but I'll copy them here for the record.

I don't see why it's necessary to run unit tests with multiple Go versions. If I'm interpreting [this](https://go.dev/ref/mod#go-mod-file-go) right, running the tests with 1.21 will just use the 1.20 semantics, since that is the version declared in go.mod. In which case there is no reason to run the Go tests with 1.21 as well.

benhoyt: Just to be clear, I meant leave the matrix in but with just `['1.20']` in it, so we can easily add 1.21 later.

I think it's still useful to run on both supported versions. As I understand it, Go 1.21 will attempt to run with 1.20 semantics for language features and APIs, but when 1.21 comes out, running the tests on Go 1.21 as well as 1.20 will ensure that it definitely does run and pass all the tests on both. Even if it's just a matter of error messages being different in the two versions or something where our tests are too specific, it would be good to catch.

That's my take, anyway, though I don't feel particularly strongly about it.

barrettj12: > Even if it's just a matter of error messages being different in the two versions or something where our tests are too specific, it would be good to catch.

I just can't really see that happening, due to the Go compatibility promise. The one exception would be the [new loop semantics](https://github.com/golang/go/discussions/56010), but that's easy enough to test manually.

IMO we should test using the same version(s) of Go we build Pebble with. If we only build Pebble binaries using a single Go version, then we only need to test with that version.

So I don't see the benefit to running tests with multiple Go versions, but I'm interested to see what others think. @hpidcock @jnsgruk @flotter @anpep @tlm (@ anyone else) ?

benhoyt: > IMO we should test using the same version(s) of Go we build Pebble with. If we only build Pebble binaries using a single Go version, then we only need to test with that version.

That's fair enough -- I'd be fine with that. [Looks like](https://github.com/canonical/pebble/blob/6991b3c387ed673436afe6e7afea73f64b69d4ed/.github/workflows/binaries.yml#L30) we build binaries with the latest stable version of Go (though arguably we should explicitly pin a version there). Interested to hear what others think too.

jnsgruk: I'd prefer to stick with the matrix, but trim to just `['1.20']`.

We will likely not bump the actual go version of pebble to 1.21 immediately, but we will probably want to test on 1.21.

@barrettj12 if you can just change this to a single version for now, that would be handy.

While you're in there, can you adjust the binary builder to use the version of go from `go.mod`?

jnsgruk: I believe you can do that like so:

```yaml
- uses: actions/setup-go@v4
    with:
      go-version-file: "go.mod"
```

benhoyt (approved):
Works for me, thanks.

anpep (approved):
My review was not requested here, but I wanted to chip in since I was mentioned in the matrix discussion.

I think we should test only the Go version in the `go.mod` file. Testing multiple versions only adds more running time to the CI and does not provide much value since, as previously discussed, the semantics of the version stated in `go.mod` are used regardless.

Also, these changes look good to me (: I'd be fine with removing the matrix but the current approach is functionally equivalent so I'd say let's stick to that until there's a more general consensus on the matter!

PR #271, opened 2023-08-02
Build Pebble snap with Go 1.20

cjdcordeiro:
Based on https://github.com/canonical/pebble/pull/171, this PR is simply making sure the Pebble snap is built with Go 1.20 instead of whatever the latest stable Go version is in the snap store.


jnsgruk:
Merging this as it's a trivial change to CI and build process.

PR #273, opened 2023-08-08
feat(overlord): allow adding external managers

paul-rodriguez:
This change enables an external application importing Pebble to extend the overlord functionality by adding additional managers.

Example overlord extension:
```go
package ovldext

type OverlordExtension struct {
	fooManager *foostate.Manager
	barManager *barstate.Manager
}

func New() *OverlordExtension {
    return &OverlordExtension{}
}

func (oe *OverlordExtension) ExtraManagers(o *Overlord) ([]overlord.StateManager, error) {
    oe.fooManager, err := foostate.NewManager(o.State(), o.TaskRunner())
    if err != nil {
        return nil, err
    }
    oe.barManager, err = barstate.NewManager(o.State(), o.TaskRunner())
    if err != nil {
        return nil, err
    }
    return []overlord.StateManager{oe.fooManager, oe.barManager}, nil
}

func (oe *OverlordExtension) FooManager() *foostate.Manager { return oe.fooManager }

func (oe *OverlordExtension) BarManager() *barstate.Manager { return oe.barManager }

func Extension(o *overlord.Overlord) *OverlordExtension {
	return o.Extension().(*OverlordExtension)
}
```

Example startup code:
```go
    dopts := daemon.Options{
        Dir:        pebbleDir,
        SocketPath: socketPath,

        // Supply overlord extension
        OverlordExtension: ovldext.New()            
    }
    
    d, err := daemon.New(&dopts)
    if err != nil {
        return err
    }
```

External HTTP request handler example:
```go  
    :  
    // We can now get access to the overlord
    o := d.Overlord()
    
    // Get access to the custom state manager to perform the HTTP request
    fooManager := ovldext.Extension(o).FooManager()
    :
```

### DIFF
--- a/.github/.golangci.yml
+++ b/.github/.golangci.yml
@@ -12,3 +12,7 @@ issues:
   # these values ensure that all issues will be surfaced
   max-issues-per-linter: 0
   max-same-issues: 0
+
+run:
+  timeout: 5m
+
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
         id: lint
         with:
           version: latest
-          args: '-c .github/.golangci.yml --out-format=colored-line-number'
+          args: '-c .github/.golangci.yml --out-format=colored-line-number --timeout 2m0s'
           skip-cache: true
 
       - name: Print error message
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
         id: lint
         with:
           version: latest
-          args: '-c .github/.golangci.yml --out-format=colored-line-number --timeout 2m0s'
+          args: '-c .github/.golangci.yml --out-format=colored-line-number'
           skip-cache: true
 
       - name: Print error message
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -217,6 +217,10 @@ func userFromRequest(state interface{}, r *http.Request) (*UserState, error) {
 	return nil, nil
 }
 
+func (d *Daemon) Overlord() *overlord.Overlord {
+	return d.overlord
+}
+
 func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	st := c.d.state
 	st.Lock()
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -791,7 +791,14 @@ func New(opts *Options) (*Daemon, error) {
 		httpAddress:         opts.HTTPAddress,
 	}
 
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, opts.OverlordExtension)
+	ovldOptions := overlord.Options{
+		PebbleDir:      opts.Dir,
+		RestartHandler: d,
+		ServiceOutput:  opts.ServiceOutput,
+		Extension:      opts.OverlordExtension,
+	}
+
+	ovld, err := overlord.New(ovldOptions)
 	if err == errExpectedReboot {
 		// we proceed without overlord until we reach Stop
 		// where we will schedule and wait again for a system restart.
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -72,10 +72,9 @@ type Options struct {
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
 
-	// NewManagerFuncs is an optional slice of NewManagerFunc which will
-	// be passed to the overlord.New to create custom managers at overlord
-	// creation.
-	NewManagerFuncs []overlord.NewManagerFunc
+	// OverlordExtension is an optional struct used to extend the capabilities
+	// of the Overlord.
+	OverlordExtension overlord.Extension
 }
 
 // A Daemon listens for requests and routes them to the right command
@@ -792,7 +791,7 @@ func New(opts *Options) (*Daemon, error) {
 		httpAddress:         opts.HTTPAddress,
 	}
 
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, opts.NewManagerFuncs)
+	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, opts.OverlordExtension)
 	if err == errExpectedReboot {
 		// we proceed without overlord until we reach Stop
 		// where we will schedule and wait again for a system restart.
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -72,7 +72,7 @@ type Options struct {
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
 
-	// OverlordExtension is an optional struct used to extend the capabilities
+	// OverlordExtension is an optional interface used to extend the capabilities
 	// of the Overlord.
 	OverlordExtension overlord.Extension
 }
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -72,10 +72,10 @@ type Options struct {
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
 
-	// ManagerGenerators is an optional slice of ManagerGenerator which will
+	// NewManagerFuncs is an optional slice of NewManagerFunc which will
 	// be passed to the overlord.New to create custom managers at overlord
 	// creation.
-	ManagerGenerators []overlord.ManagerGenerator
+	NewManagerFuncs []overlord.NewManagerFunc
 }
 
 // A Daemon listens for requests and routes them to the right command
@@ -792,7 +792,7 @@ func New(opts *Options) (*Daemon, error) {
 		httpAddress:         opts.HTTPAddress,
 	}
 
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, opts.ManagerGenerators)
+	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, opts.NewManagerFuncs)
 	if err == errExpectedReboot {
 		// we proceed without overlord until we reach Stop
 		// where we will schedule and wait again for a system restart.
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -798,7 +798,7 @@ func New(opts *Options) (*Daemon, error) {
 		Extension:      opts.OverlordExtension,
 	}
 
-	ovld, err := overlord.New(ovldOptions)
+	ovld, err := overlord.New(&ovldOptions)
 	if err == errExpectedReboot {
 		// we proceed without overlord until we reach Stop
 		// where we will schedule and wait again for a system restart.
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -71,6 +71,11 @@ type Options struct {
 	// ServiceOuput is an optional io.Writer for the service log output, if set, all services
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
+
+	// ManagerGenerators is an optional slice of ManagerGenerator which will
+	// be passed to the overlord.New to create custom managers at overlord
+	// creation.
+	ManagerGenerators []overlord.ManagerGenerator
 }
 
 // A Daemon listens for requests and routes them to the right command
@@ -787,7 +792,7 @@ func New(opts *Options) (*Daemon, error) {
 		httpAddress:         opts.HTTPAddress,
 	}
 
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput)
+	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, opts.ManagerGenerators)
 	if err == errExpectedReboot {
 		// we proceed without overlord until we reach Stop
 		// where we will schedule and wait again for a system restart.
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -120,6 +120,12 @@ func (f *fakeExtension) ExtraManagers(o *overlord.Overlord) ([]overlord.StateMan
 	return result, nil
 }
 
+type otherFakeExtension struct{}
+
+func (otherFakeExtension) ExtraManagers(o *overlord.Overlord) ([]overlord.StateManager, error) {
+	return nil, nil
+}
+
 func (s *daemonSuite) TestExternalManager(c *C) {
 	d, err := New(&Options{
 		Dir:               s.pebbleDir,
@@ -133,10 +139,34 @@ func (s *daemonSuite) TestExternalManager(c *C) {
 	c.Assert(err, IsNil)
 	extension, ok := d.overlord.Extension().(*fakeExtension)
 	c.Assert(ok, Equals, true)
-	c.Assert(extension, NotNil)
 	manager := extension.mgr
 	c.Assert(manager.id, Equals, "expected")
 	c.Assert(manager.ensureCalls, Equals, 1)
+}
+
+func (s *daemonSuite) TestNoExtension(c *C) {
+	d, err := New(&Options{
+		Dir:         s.pebbleDir,
+		SocketPath:  s.socketPath,
+		HTTPAddress: s.httpAddress,
+	})
+	c.Assert(err, IsNil)
+
+	extension := d.overlord.Extension()
+	c.Assert(extension, IsNil)
+}
+
+func (s *daemonSuite) TestWrongExtension(c *C) {
+	d, err := New(&Options{
+		Dir:               s.pebbleDir,
+		SocketPath:        s.socketPath,
+		HTTPAddress:       s.httpAddress,
+		OverlordExtension: &fakeExtension{},
+	})
+	c.Assert(err, IsNil)
+
+	_, ok := d.overlord.Extension().(*otherFakeExtension)
+	c.Assert(ok, Equals, false)
 }
 
 func (s *daemonSuite) TestAddCommand(c *C) {
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -112,8 +112,8 @@ func (m *fakeManager) Ensure() error {
 
 func (s *daemonSuite) TestExternalManager(c *C) {
 	generators := []overlord.ManagerGenerator{
-		func(o overlord.ManagerProvider) (any, overlord.StateManager) {
-			return fakeManager{}, &fakeManager{id: "expected"}
+		func(o overlord.ManagerProvider) (any, overlord.StateManager, error) {
+			return fakeManager{}, &fakeManager{id: "expected"}, nil
 		},
 	}
 
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -111,23 +111,23 @@ func (m *fakeManager) Ensure() error {
 }
 
 func (s *daemonSuite) TestExternalManager(c *C) {
-	generators := []overlord.ManagerGenerator{
-		func(o overlord.ManagerProvider) (any, overlord.StateManager, error) {
+	generators := []overlord.NewManagerFunc{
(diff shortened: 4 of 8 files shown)


internals/daemon/daemon.go
```diff
@@ -783,7 +794,14 @@ func New(opts *Options) (*Daemon, error) {
...
 
-	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput)
+	ManagerGenerators = append(
+		ManagerGenerators,
+		func(o *overlord.Overlord) (overlord.ManagerTag, overlord.StateManager) {
+			commandMgr := cmdstate.NewManager(o.TaskRunner())
+			return CommandManagerTag, commandMgr
+		})
+
+	ovld, err := overlord.New(opts.Dir, d, opts.ServiceOutput, ManagerGenerators)
```
flotter, line 804: Instead of using this scheme to now add all built-in overlord managers as well as external ones:

1. Why do we not leave the overlord code as is today, with your ```stateengine``` tweak to make the ```taskRunner``` special.

2. Add a daemon access method to get to the overlord instance? Then we can do this:

```
func (d *Daemon) Overlord() *overlord.Overlord {
    return d.overlord
}
```

Then in ```internals/cli/cmd_run.go``` equivalent in an importer project, we can do:

```
        d, err := daemon.New(&dopts)
        if err != nil {
                return err
        }
        
        // Add custom managers
        c := customManager.New(d.Overlord().State(), d.Overlord().TaskRunner(), ...)
        d.Overlord().StateEngine().AddManager(c)
        
        if err := d.Init(); err != nil {
                return err
        }

```

flotter (commented):
Some brainstorming ideas for discussion.

internals/daemon/daemon.go
paul-rodriguez: Yours is a valid way of doing it.
I wanted to avoid exposing the overlord. I don't have a strong preference either way.

internals/overlord/overlord.go
```diff
@@ -381,6 +402,10 @@ func (o *Overlord) SettleObserveBeforeCleanups(timeout time.Duration, beforeClea
 	return o.settle(timeout, beforeCleanups)
 }
 
+func (o *Overlord) GetManagerByTag(tag ManagerTag) StateManager {
+	return o.customManagers[tag]
+}
+
 // State returns the system state managed by the overlord.
 func (o *Overlord) State() *state.State {
 	return o.stateEng.State()
```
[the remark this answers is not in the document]
paul-rodriguez: We still need a way to refer to a manager from another manager (or at least, when building the dependent manager). This is what `GetManagerByTag`  is supposed to accomplish.

But it's true that if you build all your custom managers in the same place then you don't strictly need this.

anpep (changes requested):
I don't think the approach taken here is suitable for a public-facing API.

A simpler approach I suggest: having separate `func New(...) (*Overlord, error)` and `func (*Overlord) Init() error` functions, then have a `func (*Overlord) AddManager(StateManager) error` that we can simply call before `o.Init()`. 

What do you think?

paul-rodriguez:
> I don't think the approach taken here is suitable for a public-facing API.
> 
> A simpler approach I suggest: having separate `func New(...) (*Overlord, error)` and `func (*Overlord) Init() error` functions, then have a `func (*Overlord) AddManager(StateManager) error` that we can simply call before `o.Init()`.
> 
> What do you think?

I don't like letting the user of that code be bitten by forgetting Init or calling AddManager after Init. Right now Overlord is more or less RAII, let's keep it that way if possible.

Here the only purpose of Init would be to ensure the TaskRunner is the last manager. With the changes I made to the StateEngine, Init is not necessary. You can AddManager in cmd_run as Fred suggested.

internals/overlord/stateengine.go
```diff
@@ -56,8 +56,24 @@ type StateEngine struct {
 	state   *state.State
 	stopped bool
 	// managers in use
-	mgrLock  sync.Mutex
-	managers []StateManager
+	mgrLock    sync.Mutex
+	managers   []StateManager
+	taskRunner StateManager
```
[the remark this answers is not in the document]
flotter: We discussed this, and just to clarify my preference is to remove the ```orderedManagers()``` method rendering a new list which simply adds the taskRunner at the end, to simply call respective ```taskRunner``` method after iterating through the other managers. We decided to get some feedback from Gustavo and pick his preference.

flotter (commented):
There appears to be a linter failure in one of the tests.

internals/overlord/overlord.go
```diff
@@ -427,20 +424,14 @@ func Fake() *Overlord {
...
 	o.runner = state.NewTaskRunner(s)
 	return o
 }
 
-// AddManager adds a manager to a fake overlord. It cannot be used for
-// a normally initialized overlord those are already fully populated.
 func (o *Overlord) AddManager(mgr StateManager) {
-	if o.inited {
-		panic("internal error: cannot add managers to a fully initialized Overlord")
-	}
```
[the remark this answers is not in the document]
flotter: We agreed to use the ```o.StateEngine().AddManager(...)``` directly, instead of modifying the ```o.AddManager(...)``` test method (for use with FakeOverlord), for adding managers. This allows us to leave the ```inited``` flag and its usage unchanged for now.

anpep (approved):
Excellent work on this @paul-rodriguez!

.github/workflows/lint.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         id: lint
         with:
           version: latest
-          args: '-c .github/.golangci.yml --out-format=colored-line-number'
+          args: '-c .github/.golangci.yml --out-format=colored-line-number --timeout 2m0s'
```
anpep, line 20: Why is this required though?

benhoyt (approved):
It's a bit awkward, but I understand why the task runner is special, and this seems reasonable to me -- it's definitely nice and simple!

.github/workflows/lint.yml
flotter: The default appears too little for the linter. Lots of PRs for various projects increase this.

@paul-rodriguez It looks like this PR from @barrettj12 will set it to 5m using the [run configuration](https://golangci-lint.run/usage/configuration/#run-configuration) key ```timeout``` which is directly set in the config file ```.github/.golangci.yml```.

Maybe we should do the same, which means no conflicting attempts setting timeouts. 5m used by Jordan seems future safe?

flotter (approved):

internals/overlord/overlord.go
```diff
@@ -64,21 +64,42 @@ type Overlord struct {
...
+// handles to hook an external manager into the overlord's environment.
+type ManagerProvider interface {
+	State() *state.State
+	TaskRunner() *state.TaskRunner
+}
+
+// ManagerGenerator is passed to Overlord to create a manager
+// The return value is a key, value pair where the key has to be a unique
+// identifier for the manager being created.
+type ManagerGenerator func(ManagerProvider) (key any, manager StateManager)
```
flotter, line 88: I think we will have to include returning an error to comply with how each manager instantiation can fail, and how each of these errors are propagated to the ```overlord.New()``` error return.

flotter (commented):
Looks great Paul. I think this is perfect for a first round of reviews, including a proposal discussion with Gustavo. I've taken the liberty to update the example in the PR description for a quick highlight of this scheme for understanding the proposal.

benhoyt (commented):
It's a little indirect, but I don't see a significantly cleaner way to achieve what we're trying to achieve here. I've made a few naming suggestions and suggested a couple of ways to simplify the args passed to the new-manager func.

internals/daemon/daemon_test.go
```diff
@@ -99,6 +100,41 @@ func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		},
+	}
+
+	d, err := New(&Options{
+		Dir:               s.pebbleDir,
+		SocketPath:        s.socketPath,
+		HTTPAddress:       s.httpAddress,
+		ManagerGenerators: generators,
+	})
+	c.Assert(err, IsNil)
```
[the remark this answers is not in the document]
paul-rodriguez: I don't think this is the case: we merely create a daemon with New, but we don't Init or Start it.

internals/overlord/overlord.go
```diff
@@ -130,12 +151,28 @@ func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writ
...
 
 	return o, nil
 }
 
+func (o *Overlord) tagManager(tag any, mgr StateManager) {
+	o.externalManagers[tag] = mgr
+	o.addManager(mgr)
+}
+
+func (o *Overlord) GetExternalManager(tag any) StateManager {
```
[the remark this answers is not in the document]
paul-rodriguez: It's meant to be "managers external to pebble". But "external to the overlord" is also a valid interpretation.

internals/overlord/overlord.go
```diff
@@ -64,21 +64,42 @@ type Overlord struct {
...
+	logMgr           *logstate.LogManager
+	externalManagers map[any]StateManager
 }
 
-// New creates a new Overlord with all its state managers.
+// ManagerProvider is the interface that ManagerGenerator depends on
+//
+// Overlord implements ManagerProvider as it provides the necessary
+// handles to hook an external manager into the overlord's environment.
+type ManagerProvider interface {
```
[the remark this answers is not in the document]
paul-rodriguez: I went with 1. but then realized I was creating boilerplate code implementing those functions when I could just be using the ones of Overlord directly.

The reason we're not going with 3. is to avoid exposing the whole Overlord, something which Gustavo insisted on after seeing the previous version.

.github/.golangci.yml
```diff
@@ -11,4 +11,8 @@ linters-settings:
 issues:
   # these values ensure that all issues will be surfaced
   max-issues-per-linter: 0
-  max-same-issues: 0
\ No newline at end of file
+  max-same-issues: 0
+
+run:
+  timeout: 5m
```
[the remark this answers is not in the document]
paul-rodriguez: In my own testing it's more like two minutes, but yeah this is suspicious.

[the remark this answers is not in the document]
paul-rodriguez: Created issue https://github.com/canonical/pebble/issues/286

flotter:
Current status: Further discussions required. Current design unlikely to progress in its current form.

internals/daemon/daemon_test.go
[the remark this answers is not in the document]
benhoyt: Yeah, you're probably right. However, currently it calls `overlord.New`, which calls `servstate.NewManager`, which starts the (global) reaper with `reaper.Start()`. However, the right fix here is probably to not do that -- which it looks like @flotter is proposing anyway for different reasons in https://github.com/canonical/pebble/issues/284

So probably leave this as is for now.

internals/daemon/daemon.go
```diff
@@ -71,6 +71,10 @@ type Options struct {
 	// ServiceOuput is an optional io.Writer for the service log output, if set, all services
 	// log output will be written to the writer.
 	ServiceOutput io.Writer
+
+	// OverlordExtension is an optional struct used to extend the capabilities
```
flotter, line 75: Super nitty comment: I would refer to it as an interface, not struct. Also optional does not exactly explain how to not use it. Here are some thoughts for inspiration.
```
// OverlordExtension is an interface allowing the extension of the default Overlord. If no extension is needed, nil should be supplied.
```

internals/overlord/overlord.go
```diff
@@ -70,15 +76,38 @@ type Overlord struct {
 	commandMgr *cmdstate.CommandManager
 	checkMgr   *checkstate.CheckManager
 	logMgr     *logstate.LogManager
+
+	// Overlord extension
```
flotter, line 80: I think the comment can be dropped. It seems quite clear from the type.

internals/overlord/overlord.go
```diff
@@ -70,15 +76,38 @@ type Overlord struct {
...
+type ManagerEnvironment interface {
+	State() *state.State
+	TaskRunner() *state.TaskRunner
 }
 
-// New creates a new Overlord with all its state managers.
+// NewManagerFunc is passed to Overlord to create a manager
+// The return value is a key, value pair where the key has to be a unique
+// identifier for the manager being created.
+type NewManagerFunc func(ManagerEnvironment) (key any, manager StateManager, err error)
```
flotter: Leftover from previous proposal. Can be deleted I believe ?

internals/overlord/overlord.go
```diff
@@ -70,15 +76,38 @@ type Overlord struct {
...
+
+// New creates a  Overlord with all its state managers.
```
flotter, line 84: nit: typo: ```an Overlord```

internals/overlord/overlord.go
```diff
@@ -130,12 +159,28 @@ func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writ
...
 	o.checkMgr.NotifyCheckFailed(o.serviceMgr.CheckFailed)
 
-	// the shared task runner should be added last!
+	if extension != nil {
+		extraManagers, err := o.extension.ExtraManagers(o)
+		if err != nil {
+			return nil, err
+		}
+		for _, manager := range extraManagers {
+			o.addManager(manager)
```
flotter, line 154: nit: I would replace this with ```o.stateEng.AddManager(manager)```. The original purpose of ```addManager``` was to assign manager instances to the overlord struct (snapd), which has been removed. This wrapper adds no value. I think we can clean up (delete the method) the code as part of this?

flotter (commented):
Great! Added a few nitty and minor comments, as I think we may have time to do a detailed code review during our next meeting.

internals/overlord/overlord.go
paul-rodriguez: I wouldn't delete the `Overlord.addManager` method in this PR, but I will use `stateEng.AddManager` if you think it's better.

internals/daemon/daemon.go
paul-rodriguez: See the comments of the other fields in the Options struct: optional means "either something or nil" (pretty sure this is a widely used Golang convention). And the default behavior in case of "nil" is not to extend (I think this is obvious).

flotter: OK thanks. Makes sense!

internals/daemon/daemon_test.go
```diff
@@ -99,6 +100,45 @@ func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		HTTPAddress:       s.httpAddress,
+		OverlordExtension: &fakeExtension{},
+	})
+	c.Assert(err, IsNil)
+
+	err = d.overlord.StateEngine().Ensure()
+	c.Assert(err, IsNil)
+	extension, ok := d.overlord.Extension().(*fakeExtension)
+	c.Assert(ok, Equals, true)
+	c.Assert(extension, NotNil)
```
paul-rodriguez, line 136: remove this, add bad case test

internals/overlord/overlord.go
```diff
@@ -70,15 +76,23 @@ type Overlord struct {
...
 
-// New creates a new Overlord with all its state managers.
+// New creates an Overlord with all its state managers.
 // It can be provided with an optional restart.Handler.
-func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writer) (*Overlord, error) {
+func New(
+	pebbleDir string,
+	restartHandler restart.Handler,
+	serviceOutput io.Writer,
+	extension Extension) (*Overlord, error) {
```
paul-rodriguez, line 89: put all args into an options struct

flotter (approved):
Great work Paul. Looks super nice.

Regarding the PR description example, feel free to remote it or change it if you want. I find it essential assisting reviewers get the complete picture for things talking about some hypothetical external use-case (especially when discussing in person), but do not feel like we have to keep it there, or that version of it.

internals/overlord/overlord.go
```diff
@@ -49,6 +49,25 @@ var (
 	defaultCachedDownloads = 5
 )
 
+// Extension represents an extension of the Overlord.
+type Extension interface {
+	// ExtraManagers is called when building an Overlord.
```
benhoyt, line 54: I don't think this is quite enough of a comment. Perhaps:

```suggestion
	// ExtraManagers is called when building an Overlord, allowing additional
	// managers to be created and added to the overlord.
```

internals/overlord/overlord.go
```diff
@@ -49,6 +49,25 @@ var (
...
+// Options is the arguments passed to construct an Overlord.
+type Options struct {
+	// PebbleDir is the path to the pebble directory. It must be provided.
+	PebbleDir string
+	// RestartHandler is an optional structure to handle restart requests.
+	RestartHandler restart.Handler
+	// ServiceOutput is an optional output for the logging manager.
+	ServiceOutput io.Writer
+	// Extension is an optional struct defining additional components to add to
+	// the overlord.
```
benhoyt, line 67: Yeah, the wording here is tricky, but I agree with @flotter it's weird/confusing to say "struct" when this is clearly an interface. (Which could be fulfilled by any type, even a `func` type, or another interface.)

I think the updated comment on `Extension.ExtraManagers` will help, and maybe after that we can change this to:

```suggestion
	// Extension allows extending the overlord with additional managers.
```

benhoyt (approved):
This seems reasonable to me. Couple of comments about comments.

internals/overlord/overlord.go
```diff
@@ -130,12 +158,28 @@ func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writ
...
 	return o, nil
 }
 
+func (o *Overlord) Extension() Extension {
+	return o.extension
+}
+
 func (o *Overlord) addManager(mgr StateManager) {
 	o.stateEng.AddManager(mgr)
 }
```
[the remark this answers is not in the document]
paul-rodriguez: forgot about that, thanks!

internals/overlord/overlord.go
```diff
@@ -130,14 +158,26 @@ func New(pebbleDir string, restartHandler restart.Handler, serviceOutput io.Writ
...
+
+	// TaskRunner must be the last manager added to the StateEngine,
+	// because TaskRunner runs all the tasks required by the managers that ran
+	// before it.
 	o.stateEng.AddManager(o.runner)
 
 	return o, nil
 }
 
-func (o *Overlord) addManager(mgr StateManager) {
```
anpep, line 139: Yes!

anpep (approved):
Stellar work on this Paul, thank you!

internals/overlord/overlord.go
```diff
@@ -49,6 +49,25 @@ var (
 	defaultCachedDownloads = 5
 )
 
+// Extension represents an extension of the Overlord.
+type Extension interface {
+	// ExtraManagers is called when building an Overlord, it allows additional
+	// managers to be added to the Overlord.
```
niemeyer, line 55:
```
// ExtraManagers allows additional StateManagers to be used.
```

Whether this is called when building the overlord or elsewhere is up to the implementation of the overlord.

internals/overlord/overlord.go
```diff
@@ -49,6 +49,25 @@ var (
...
+
+// Options is the arguments passed to construct an Overlord.
+type Options struct {
+	// PebbleDir is the path to the pebble directory. It must be provided.
+	PebbleDir string
+	// RestartHandler is an optional structure to handle restart requests.
+	RestartHandler restart.Handler
+	// ServiceOutput is an optional output for the logging manager.
+	ServiceOutput io.Writer
+	// Extension, if set, defines additional components to add to the overlord.
```
niemeyer, line 67:
```
// Extension allows extending the overlord with externally defined features.
```

This is what the extension is. The documentation at the moment is describing ExtraManagers, but that's up to the Extension itself to document.

niemeyer (approved):
Thank you, that turned out well. Only documentation tuning suggested below.

flotter:
Last two corrections added. Approved by Gustavo. Merging.

PR #275, opened 2023-08-15
fix(servstate): use WaitDelay to avoid Command.Wait blocking on stdin/out/err

### LINKED ISSUE #149: Pebble service is not restarted on `on-check-failure` if the OS process was killed
# Issue

If the OS process related to a pebble service is killed (like using `pkill --signal SIGKILL -f process_name`) and that service has a health check that instructs pebble to restart it on a check failure (`on-check-failure` field from the pebble layer spec), pebble tries to restart the process but finishes with an `no such process` error. The number of check failures continues to grow and pebble doesn't try to, for example, just start the service, as there is no old OS process for that service to be stopped before the service start again.

There are more details in an [example repo](https://github.com/marceloneppel/patroni-restart-test) that I used to test that behaviour. I added some more description and some logs comparing what happens with pebble and also when using another tool, in that case supervisor, to start the service and later kill the OS process.

Is there a way to solve that issue (making the service restart correctly using the pebble checks) using other fields from the pebble layer? I found a workaround that I implemented in my charm code, but I would like to confirm if this is a bug or not.

benhoyt:
Use `os.exec`'s [`Cmd.WaitDelay`](https://pkg.go.dev/os/exec#Cmd.WaitDelay) to ensure `cmd.Wait()` returns in a reasonable timeframe if the goroutines that `cmd.Start()` uses to copy stdin/out/err are blocked when copying due to a sub-subprocess holding onto them. Read more details about the issue in https://github.com/golang/go/issues/23019 and the proposed solution (that was added in Go 1.20) in https://github.com/golang/go/issues/50436.

This solves [issue 149](https://github.com/canonical/pebble/issues/149), where Patroni wasn't restarting properly even after a `KILL` signal was sent to it. I had originally mis-diagnosed this problem as an issue with Pebble not tracking the process tree of processes that daemonise and change their process group (which is still an issue, but is not causing this problem). The Patroni process wasn't being marked as finished at all due to being blocked on the `cmd.Wait()`. Patroni starts sub-processes and "forwards" stdin/out/err, so the copy goroutines block. Thankfully Go 1.20 introduced `WaitDelay` to allow you to easily work around this exact problem.

The fix itself is [this one-liner](https://github.com/canonical/pebble/pull/275):

```go
s.cmd.WaitDelay = s.killDelay() // defaults to 5 seconds
```

This will really only be a problem for services, but we make the same change for exec and exec health checks as it won't hurt there either.

Also, as a drive-by, this PR also canonicalises some log messages: our style is to start with an uppercase letter (for logs, not errors) and to use "Cannot X" rather than "Error Xing".

Fixes #149.

### DIFF
--- a/internals/daemon/api_logs.go
+++ b/internals/daemon/api_logs.go
@@ -126,7 +126,7 @@ func (r logsResponse) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			err = encoder.Encode(newJSONLog(<-fifo))
 		}
 		if err != nil {
-			logger.Noticef("error writing logs: %v", err)
+			logger.Noticef("Cannot write logs: %v", err)
 			return false
 		}
 		flushWriter(w)
@@ -186,7 +186,7 @@ func (r logsResponse) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			// Otherwise encode and output log directly.
 			err := encoder.Encode(newJSONLog(log))
 			if err != nil {
-				logger.Noticef("error writing logs: %v", err)
+				logger.Noticef("Cannot write logs: %v", err)
 				return
 			}
 			if follow {
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -168,7 +168,7 @@ func (c *Command) canAccess(r *http.Request, user *UserState) accessResult {
 	if err == nil {
 		isUser = true
 	} else if err != errNoID {
-		logger.Noticef("Unexpected error when attempting to get UID: %s", err)
+		logger.Noticef("Cannot parse UID from remote address %q: %s", r.RemoteAddr, err)
 		return accessForbidden
 	}
 
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -168,7 +168,7 @@ func (c *Command) canAccess(r *http.Request, user *UserState) accessResult {
 	if err == nil {
 		isUser = true
 	} else if err != errNoID {
-		logger.Noticef("unexpected error when attempting to get UID: %s", err)
+		logger.Noticef("Unexpected error when attempting to get UID: %s", err)
 		return accessForbidden
 	}
 
@@ -530,7 +530,7 @@ func (d *Daemon) HandleRestart(t restart.RestartType) {
 		defer d.mu.Unlock()
 		d.restartSocket = true
 	default:
-		logger.Noticef("internal error: restart handler called with unknown restart type: %v", t)
+		logger.Noticef("Internal error: restart handler called with unknown restart type: %v", t)
 	}
 	d.tomb.Kill(nil)
 }
@@ -760,12 +760,12 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 		// might get rolled back!!
 		restart.ClearReboot(st)
 		clearReboot(st)
-		logger.Noticef("pebble was restarted while a system restart was expected, pebble retried to schedule and waited again for a system restart %d times and is giving up", rebootMaxTentatives)
+		logger.Noticef("Pebble was restarted while a system restart was expected, pebble retried to schedule and waited again for a system restart %d times and is giving up", rebootMaxTentatives)
 		return nil
 	}
 	st.Set("daemon-system-restart-tentative", nTentative)
 	d.state = st
-	logger.Noticef("pebble was restarted while a system restart was expected, pebble will try to schedule and wait for a system restart again (tenative %d/%d)", nTentative, rebootMaxTentatives)
+	logger.Noticef("Pebble was restarted while a system restart was expected, pebble will try to schedule and wait for a system restart again (tenative %d/%d)", nTentative, rebootMaxTentatives)
 	return errExpectedReboot
 }
 
--- a/internals/overlord/servstate/handlers.go
+++ b/internals/overlord/servstate/handlers.go
@@ -400,6 +400,18 @@ func (s *serviceData) startInternal() error {
 	s.cmd.Stdout = logWriter
 	s.cmd.Stderr = logWriter
 
+	// Add WaitDelay to ensure cmd.Wait() returns in a reasonable timeframe if
+	// the goroutines that cmd.Start() uses to copy Stdin/Stdout/Stderr are
+	// blocked when copying due to a sub-subprocess holding onto them. Read
+	// more details in these issues:
+	//
+	// - https://github.com/golang/go/issues/23019
+	// - https://github.com/golang/go/issues/50436
+	//
+	// This isn't the original intent of kill-delay, but it seems reasonable
+	// to reuse it in this context.
+	s.cmd.WaitDelay = s.killDelay()
+
 	// Start the process!
 	logger.Noticef("Service %q starting: %s", serviceName, s.config.Command)
 	err = reaper.StartCommand(s.cmd)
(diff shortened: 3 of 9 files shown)


internals/daemon/daemon.go
```diff
@@ -168,7 +168,7 @@ func (c *Command) canAccess(r *http.Request, user *UserState) accessResult {
 	if err == nil {
 		isUser = true
 	} else if err != errNoID {
-		logger.Noticef("unexpected error when attempting to get UID: %s", err)
+		logger.Noticef("Unexpected error when attempting to get UID: %s", err)
```
barrettj12, line 171: `Cannot get UID: %s` ?

internals/overlord/servstate/handlers.go
barrettj12, line 1: Can we add a regression test for these changes?

internals/daemon/daemon.go
benhoyt: Updated and improved, thanks.

hpidcock:
We might need special handling of [exec.ErrWaitDelay](https://pkg.go.dev/os/exec#pkg-variables)

benhoyt:
@hpidcock No, I don't think we need to handle `ErrWaitDelay` specially. Normally pebble services are terminated with a signal, so will have a "terminated" exit code and that will be the error reported. `ErrWaitDelay` is only returned if the `wait-delay` kicks in *and* the service terminates successfully (with a zero exit code) -- but we do want an error in that case as it means the children haven't cleaned up I/O properly. Additionally, in [`reaper.WaitCommand`](https://github.com/canonical/pebble/blob/836feaee563078928d295c01f98f4b228a22e17a/internals/reaper/reaper.go#L214-L226) we expect a non-success `err` from `cmd.Wait()` due to the reaper already waiting for PIDs, so we always get a syscall error from "wait" or "waitid".

internals/overlord/servstate/handlers.go
benhoyt: Okay, I've added a regression test now, `TestWaitDelay`. Easier said than done! When I comment out the `s.cmd.WaitDelay = s.killDelay()` line, the test fails with the following failure:

```
FAIL: manager_test.go:1656: S.TestWaitDelay

manager_test.go:1687:
    c.Assert(chg.Err(), IsNil)
... value *state.changeError = &state.changeError{errors:[]state.taskError{state.taskError{task:"Stop service \"waitdelay\"", error:"cannot stop service: process still running after SIGTERM and SIGKILL"}}} ("cannot perform the following tasks:\n- Stop service \"waitdelay\" (cannot stop service: process still running after SIGTERM and SIGKILL)")
```

internals/overlord/servstate/manager_test.go
```diff
@@ -1661,6 +1653,44 @@ services:
...
+	testExecutable, err := os.Executable()
+	c.Assert(err, IsNil)
+	layer := parseLayer(c, 0, "layer", fmt.Sprintf(`
+services:
+    waitdelay:
+        override: replace
+        command: %s
+        environment:
+            PEBBLE_TEST_WAITDELAY: 1
+        kill-delay: 200ms
```
barrettj12, line 1669: this delay could be a lot shorter to speed up this test.

barrettj12 (approved):

internals/overlord/servstate/manager_test.go
benhoyt: Yeah, I think you're right -- no reason this has to be long. If the `WaitDelay` logic is not working, it'll fail hard even with a short timeout. Reduced to 10ms.

internals/overlord/servstate/handlers.go
```diff
@@ -400,6 +400,18 @@ func (s *serviceData) startInternal() error {
 	s.cmd.Stdout = logWriter
 	s.cmd.Stderr = logWriter
 
+	// Add WaitDelay to ensure cmd.Wait() returns in a reasonable timeframe if
+	// the goroutines that cmd.Start() uses to copy Stdin/Stdout/Stderr are
+	// blocked when copying due to a sub-subprocess holding onto them. Read
+	// more details in these issues:
```
flotter, line 406: Since the SIGTERM and SIGKILL commands terminate the entire process group, the issue is only occurring when a child process promotes itself to become a process group leader (```setpgid```) or if it comes a new session leader (```setsid```, which implies the previous as well). This is when it holds on to resources, because the termination logic failed to kill it. Not sure it if may be useful to add enough to capture that this is not always an issue.

flotter (commented):
Looks great. I have one concern reusing ```killDelay``` - please check if my understand makes sense or not. 

There is one subtlety which this new feature may create in the future - since a service restart were never able to restart child process that promoted themselves outside the process group (or session), in the past that may have forced people to someone find ways to kill everything or restart. What we are now enabling is leaving a child behind running, and killing the file descriptors previously used to capture the output (is my understanding correct)? If this is true, a service restart will cause the new primary process to log output without having the stdout of the previous children included (This is assuming the service knows how to deal with reusing the running child, else it may even start another one)?

internals/overlord/servstate/handlers.go
```diff
@@ -400,6 +400,18 @@ func (s *serviceData) startInternal() error {
...
+	//
+	// - https://github.com/golang/go/issues/23019
+	// - https://github.com/golang/go/issues/50436
+	//
+	// This isn't the original intent of kill-delay, but it seems reasonable
+	// to reuse it in this context.
+	s.cmd.WaitDelay = s.killDelay()
```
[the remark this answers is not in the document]
benhoyt: Yeah, fair point. This race was actually happening (sometimes I saw `Cannot send SIGKILL to process: no such process` in the logs). As you point out, it's harmless but best avoided. I've made it a bit less than the kill-delay (90% of it) -- this avoids introducing another config value and avoids using a fixed constant.

internals/overlord/servstate/handlers.go
benhoyt: Added - thanks

internals/overlord/servstate/handlers.go
```diff
@@ -400,6 +400,21 @@ func (s *serviceData) startInternal() error {
...
+	// the process group. Read more details in these issues:
+	//
+	// - https://github.com/golang/go/issues/23019
+	// - https://github.com/golang/go/issues/50436
+	//
+	// This isn't the original intent of kill-delay, but it seems reasonable
+	// to reuse it in this context. Use a value slightly less than the kill
+	// delay (90% of it) to avoid racing with trying to send SIGKILL (to a
+	// process that has already exited).
+	s.cmd.WaitDelay = time.Duration(s.killDelay().Seconds() * 0.9 * float64(time.Second))
```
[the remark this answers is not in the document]
benhoyt: Yeah, I thought about the d*9/10 method, but then I thought "that might overflow". However, it'll only overflow if kill-delay is ~32 years, so I think we're good. :-) Updated.

Regarding the race: yes, strictly speaking you're right, but because 1) it's very difficult to avoid this race, and 2) it doesn't hurt anything if it happens, I think it's okay.

internals/daemon/api_logs.go
```diff
@@ -126,7 +126,7 @@ func (r logsResponse) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			err = encoder.Encode(newJSONLog(<-fifo))
 		}
 		if err != nil {
-			logger.Noticef("error writing logs: %v", err)
+			logger.Noticef("Cannot write logs: %v", err)
```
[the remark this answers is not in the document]
benhoyt: In this case it would have made the PR easier to review, yeah, and it'd be more of an "atomic" commit. However, I'll leave them in there now.

PR #276, opened 2023-08-15
Add section about log/error messages to HACKING.md

barrettj12:
In the spirit of documenting Pebble conventions, add a section about log/error message conventions to `HACKING.md`, so I don't forget.

Also, I must have forgotten to add the "Code style" section to the table of contents, so add this in.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -120,7 +120,7 @@ import (
 
 ### Log and error messages
 
-**Log messages** (i.e. those passed to `logger.Noticef`) should begin with a capital letter, and use "Cannot X" rather than "Error Xing":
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -115,6 +118,18 @@ import (
 )
 ```
 
+### Log and error messages
+
+**Log messages** (i.e. those passed to `logger.Noticef`) should begin with a capital letter, and use "Cannot X" rather than "Error Xing":
```
benhoyt, line 123:
```suggestion
**Log messages** (that is, those passed to `logger.Noticef` or `logger.Debugf`) should begin with a capital letter, and use "Cannot X" rather than "Error Xing":
```

benhoyt (approved):
LGTM, thanks!

PR #279, opened 2023-08-18
test(daemon): don't start/stop reaper is setup/teardown

### LINKED ISSUE #186: Intermittent test failure: daemonSuite.TestRebootHelper
```
FAIL: daemon_test.go:727: daemonSuite.TestRebootHelper

daemon_test.go:744:
    c.Assert(err, check.IsNil)
... value *os.SyscallError = &os.SyscallError{Syscall:"waitid", Err:0xa} ("waitid: no child processes")
```

Full logs here:
https://github.com/canonical/pebble/actions/runs/3929632318/jobs/6729056974

benhoyt:
Currently the TestRebootHelper test fails intermittently with the following failure ([example log](https://github.com/canonical/pebble/actions/runs/5888109857/job/15968661905?pr=267)):

```
daemon_test.go:773:
    c.Assert(err, IsNil)
... value *os.SyscallError = &os.SyscallError{Syscall:"wait", Err:0xa} ("wait: no child processes")
```

This is because we're running the reaper, but systemdModeReboot doesn't use or know about the reaper. So avoid starting/stopping the reaper in the setup and teardown; daemon.New and Daemon.Stop -- via the service manager) starts/stop it in any case. And set the FakeCommand withReaper flag to false here as the reaper isn't running there now.

Fixes #186

### DIFF
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -38,7 +38,6 @@ import (
 	"github.com/canonical/pebble/internals/overlord/restart"
 	"github.com/canonical/pebble/internals/overlord/standby"
 	"github.com/canonical/pebble/internals/overlord/state"
-	"github.com/canonical/pebble/internals/reaper"
 	"github.com/canonical/pebble/internals/systemd"
 	"github.com/canonical/pebble/internals/testutil"
 )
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):
Nice catch, thanks :)

anpep (approved):

PR #283, opened 2023-08-18
test(servstate): enhance checkstate and reaper tests

flotter:
The folllowing commits are included:

**test(servstate): enhance zombie reaper test**

TestReapZombies have some timing sensitive code which can result in CI failures under heavy CPU load.

This PR adds a file based trigger mechanism to ensure there is a controlled window of opportunity to:

- Read the grandchild PID from the child process stdout.

- Observe the grandchild process in zombie status before its reaped.

**test(servstate): remove checks related races**

The following tests had a check which relied on the checkstate failure counter:

- TestOnCheckFailureRestartWhileRunning
- TestOnCheckFailureIgnore
- TestOnCheckFailureShutdown
  
However, this counter increments every time after the check timeout expires, even after the service was notified (which happens once only). This counter is a detail that is covered by the checkstate unit tests. Expecting an exact value from the async testing thread makes the test timing sensitive and racey, without adding any value from the servstate package perspective.

Change the failure counter check to >= 1 to remove timing sensitivity from the following tests.


internals/overlord/servstate/manager_test.go
```diff
@@ -54,14 +54,31 @@ const (
...
-	if os.Getenv("PEBBLE_TEST_CREATE_ZOMBIE") == "1" {
+	if os.Getenv("PEBBLE_TEST_CREATE_CHILD") == "1" {
 		err := createZombie()
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "cannot create zombie: %v\n", err)
 			os.Exit(1)
 		}
+		donePath := os.Getenv("PEBBLE_TEST_EXIT_CHILD")
+		if donePath == "" {
+			fmt.Fprintf(os.Stderr, "failed to obtain valid path from environment variable\n")
```
benhoyt, line 65:
```suggestion
			fmt.Fprintf(os.Stderr, "PEBBLE_TEST_EXIT_CHILD must be set\n")
```

internals/overlord/servstate/manager_test.go
```diff
@@ -1492,65 +1509,89 @@ func (s *S) TestReapZombies(c *C) {
...
+		case <-timeout:
+			c.Fatalf("timed out waiting for grandchild to zombify")
+		}
 	}
 
-	// Wait till the main process terminates
+	// Wait till the child terminates (test2 exits)
+	fd, err := os.OpenFile(exitChildPath, os.O_RDONLY|os.O_CREATE, 0666)
+	c.Assert(err, IsNil)
+	fd.Close()
```
benhoyt, line 1577: I'm not sure what this OpenFile ... immediately Close sequence does -- can you clarity in the comment?

internals/overlord/servstate/manager_test.go
```diff
@@ -1758,14 +1799,31 @@ func (r testRestarter) HandleRestart(t restart.RestartType) {
 	close(r.ch)
 }
 
-func (s *S) logBufferString() string {
+// readClearLogBuffer reads and clears the current log buffer. If you need
+// to poll for a specific stdout/stderr message, this is not suitable.
+func (s *S) readClearLogBuffer() string {
```
benhoyt, line 1804: Naming: let's call this `readAndClearLogBuffer` for clarity -- now it looks like it's reading the "clear log buffer".

benhoyt (approved):
Excellent, thanks. Just a couple of nit comments.

internals/overlord/servstate/manager_test.go
flotter: I was trying to ```touch``` a file, so that it exists and allows the other process to proceed? Maybe there is a more compact way, but I could not find one :)

internals/overlord/servstate/manager_test.go
flotter: Done!

internals/overlord/servstate/manager_test.go
flotter: Done.

internals/overlord/servstate/manager_test.go
benhoyt: Perhaps `os.Create`? Though it's not quite the same flags -- but I'm guessing that doesn't matter in this case.

But my main concern is just adding something like "so that it exists and allows the other process to proceed" to the comment. It's hard to connect the current comment with creating the file.

benhoyt (approved):

PR #285, opened 2023-08-21
feat(cli): enable extended client access

anpep:
This patch gets rid of the `clientMixin` in command structs in favor of passing a `CmdOptions` struct to the `New` (formerly `Builder`) function in `CmdInfo`.

Example usage:

#### Command implementation
```go
type cmdFoo struct {
        /* ... */
	client *MyClient
}

func init() {
	AddCommand(&CmdInfo{
		/* ... */
		New: func(opts *CmdOptions) flags.Commander {
			return &cmdFoo{client: &MyClient{opts.Client}}
		},
	})
}

func (cmd *cmdFoo) Execute(args []string) error {
        return cmd.client.Foo()
}
```

#### Client implementation
```go
import "github.com/canonical/pebble/client"

type MyClient struct {
        *client.Client
}

func (c *MyClient) Foo() error {
        /* ... */
}
```

### DIFF
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -52,7 +52,7 @@ const defaultPebbleDir = "/var/lib/pebble/default"
 // ErrExtraArgs is returned  if extra arguments to a command are found
 var ErrExtraArgs = fmt.Errorf("too many arguments for command")
 
-// CmdOptions contains state that's required during each command execution.
+// CmdOptions exposes state made accessible during command execution.
 type CmdOptions struct {
 	// Client is the Pebble client instance
 	Client *client.Client
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -153,11 +153,11 @@ type defaultOptions struct {
 // Parser creates and populates a fresh parser.
 // Since commands have local state a fresh parser is required to isolate tests
 // from each other.
-func Parser(cli *client.Client) *flags.Parser {
+func Parser(c *client.Client) *flags.Parser {
 	// Implement --version by default on every command
 	defaultOpts := defaultOptions{
 		Version: func() {
-			printVersions(cli)
+			printVersions(c)
 			panic(&exitStatus{0})
 		},
 	}
@@ -182,55 +182,55 @@ func Parser(cli *client.Client) *flags.Parser {
 	}
 
 	// Add all commands
-	for _, c := range commands {
-		obj := c.New(&CmdOptions{Client: cli, Parser: parser})
+	for _, cmdInfo := range commands {
+		obj := cmdInfo.New(&CmdOptions{Client: c, Parser: parser})
 
 		var target *flags.Command
-		if c.Debug {
+		if cmdInfo.Debug {
 			target = debugCmd
 		} else {
 			target = parser.Command
 		}
-		cmd, err := target.AddCommand(c.Name, c.Summary, strings.TrimSpace(c.Description), obj)
+		cmd, err := target.AddCommand(cmdInfo.Name, cmdInfo.Summary, strings.TrimSpace(cmdInfo.Description), obj)
 		if err != nil {
-			logger.Panicf("internal error: cannot add command %q: %v", c.Name, err)
+			logger.Panicf("internal error: cannot add command %q: %v", cmdInfo.Name, err)
 		}
-		cmd.PassAfterNonOption = c.PassAfterNonOption
+		cmd.PassAfterNonOption = cmdInfo.PassAfterNonOption
 
 		// Extract help for flags and positional arguments from ArgsHelp
 		flagHelp := map[string]string{}
 		positionalHelp := map[string]string{}
-		for specifier, help := range c.ArgsHelp {
+		for specifier, help := range cmdInfo.ArgsHelp {
 			if flagRegexp.MatchString(specifier) {
 				flagHelp[specifier] = help
 			} else if positionalRegexp.MatchString(specifier) {
 				positionalHelp[specifier] = help
 			} else {
-				logger.Panicf("internal error: invalid help specifier from %q: %q", c.Name, specifier)
+				logger.Panicf("internal error: invalid help specifier from %q: %q", cmdInfo.Name, specifier)
 			}
 		}
 
 		// Make sure all argument descriptions are set
 		opts := cmd.Options()
 		if len(opts) != len(flagHelp) {
-			logger.Panicf("internal error: wrong number of flag descriptions for %q: expected %d, got %d", c.Name, len(opts), len(flagHelp))
+			logger.Panicf("internal error: wrong number of flag descriptions for %q: expected %d, got %d", cmdInfo.Name, len(opts), len(flagHelp))
 		}
 		for _, opt := range opts {
 			if description, ok := flagHelp["--"+opt.LongName]; ok {
-				lintDesc(c.Name, opt.LongName, description, opt.Description)
+				lintDesc(cmdInfo.Name, opt.LongName, description, opt.Description)
 				opt.Description = description
 			} else if description, ok := flagHelp["-"+string(opt.ShortName)]; ok {
-				lintDesc(c.Name, string(opt.ShortName), description, opt.Description)
+				lintDesc(cmdInfo.Name, string(opt.ShortName), description, opt.Description)
 				opt.Description = description
 			} else if !opt.Hidden {
-				logger.Panicf("internal error: %q missing description for %q", c.Name, opt)
+				logger.Panicf("internal error: %q missing description for %q", cmdInfo.Name, opt)
 			}
 		}
 
 		args := cmd.Args()
 		for _, arg := range args {
 			if description, ok := positionalHelp[arg.Name]; ok {
-				lintArg(c.Name, arg.Name, description, arg.Description)
+				lintArg(cmdInfo.Name, arg.Name, description, arg.Description)
 				arg.Name = fixupArg(arg.Name)
 				arg.Description = description
 			}
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -52,6 +52,14 @@ const defaultPebbleDir = "/var/lib/pebble/default"
 // ErrExtraArgs is returned  if extra arguments to a command are found
 var ErrExtraArgs = fmt.Errorf("too many arguments for command")
 
+// CmdOptions contains state that's required during each command execution.
+type CmdOptions struct {
+	// Client is the Pebble client instance
+	Client *client.Client
+	// Parser is the instance of the flags.Parser that processed the command.
+	Parser *flags.Parser
+}
+
 // CmdInfo holds information needed by the CLI to execute commands and
 // populate entries in the help manual.
 type CmdInfo struct {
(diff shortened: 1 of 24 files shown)


internals/cli/cli.go
```diff
@@ -139,16 +139,16 @@ func fixupArg(optName string) string {
 	return optName
 }
 
-type clientSetter interface {
-	setClient(*client.Client)
+type ClientInitializer interface {
+	Init(*client.Client)
```
benhoyt, line 143: Name bike-shedding, but:

Should this be called `InitClient`? I know it's sort of redundant as Go will only match the method if the types match as well, but `Init` seems too generic: is it initing the command, the client, or something else? Cf: before it was `setClient`, not `set`.

Then again, it seems this is really setting the client, not initializing it. So why not `ClientSetter` / `SetClient` more like it was?

benhoyt (commented):
Just a comment about the name.

I'd be fine with this. However, are you able to link to the previous discussion where I pushed back on the whole concept of the `clientMixin`? I can't find that right now, and I'm interesting in going back over my thoughts.

internals/cli/cli.go
flotter: We had a quick conversation about this. We will take this as is to Gustavo for now. The ```Init()``` choice does not immediately make sense in the context of Pebble, but the external project extending the client will actually have an ```Init()``` that does more, creating a new extended client instance. So its more that ```Init()``` now allows you to create a new extended client, but you do not have to use this facility if you are happy with the default Pebble client instance.

flotter (approved):

paul-rodriguez (approved):
I prefer the new naming.

internals/cli/cmd_add.go
```diff
@@ -32,12 +32,13 @@ label (or append if the label is not found).
...
 
 type cmdAdd struct {
-	clientMixin
 	Combine    bool `long:"combine"`
 	Positional struct {
 		Label     string `positional-arg-name:"<label>" required:"1"`
 		LayerPath string `positional-arg-name:"<layer-path>" required:"1"`
 	} `positional-args:"yes"`
+
+	client *client.Client
```
flotter, line 41: super nit: Any reason why you moved this all the way down and introduced a space?

internals/cli/cmd_run.go
```diff
@@ -60,8 +60,8 @@ var sharedRunEnterArgsHelp = map[string]string{
 }
 
 type cmdRun struct {
-	clientMixin
 	sharedRunEnterOpts
+	client *client.Client
```
flotter, line 64: Space precedes client elsewhere.

internals/cli/cmd_services.go
```diff
@@ -29,11 +29,11 @@ about all services if none are specified.
 `
 
 type cmdServices struct {
-	clientMixin
 	timeMixin
 	Positional struct {
 		Services []string `positional-arg-name:"<service>"`
 	} `positional-args:"yes"`
+	client *client.Client
```
flotter, line 36: Space precedes client elsewhere, but here and further down its removed.

paul-rodriguez (approved):
Great job !

internals/cli/cli.go
```diff
@@ -192,13 +183,7 @@ func Parser(cli *client.Client) *flags.Parser {
...
 	// Add all commands
 	for _, c := range commands {
-		obj := c.Builder()
-		if x, ok := obj.(clientSetter); ok {
-			x.setClient(cli)
-		}
-		if x, ok := obj.(parserSetter); ok {
-			x.setParser(parser)
-		}
+		obj := c.New(&CmdOptions{Client: cli, Parser: parser})
```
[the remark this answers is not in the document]
paul-rodriguez: I do agree to renaming `c` to `command` and `cli` to `client` in here.

internals/cli/cmd_add.go
anpep: In order to "cuddle" (group) the arguments and command members.

internals/cli/cli.go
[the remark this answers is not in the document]
anpep: We cannot use `client` (refers to the package name). I still like `New` better than `Builder`. Will look into this (:

internals/cli/cmd_run.go
anpep: But here we don't have more than one struct embedding so I don't think it makes sense to separate the members with a space.

internals/cli/cmd_services.go
anpep: Good catch (:

anpep: Here, we have: 1 embedded struct + 1 embedded inline struct + 1 member.
My rule is: only append a space after 2 or more lines of the same category. In this case we don't have it so I didn't add a space.

internals/cli/cli.go
```diff
@@ -67,9 +75,8 @@ type CmdInfo struct {
 	// command, and in the Pebble man page.
 	Description string
 
-	// Builder is a function that creates a new instance of the command
-	// struct containing an Execute(args []string) implementation.
-	Builder func() flags.Commander
+	// New is a function that creates a new instance of the command.
+	New func(*CmdOptions) flags.Commander
```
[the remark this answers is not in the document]
anpep: I still prefer `New` and it's consistent with the cheap initialization mantra.

internals/cli/cli.go
```diff
@@ -52,6 +52,14 @@ const defaultPebbleDir = "/var/lib/pebble/default"
 // ErrExtraArgs is returned  if extra arguments to a command are found
 var ErrExtraArgs = fmt.Errorf("too many arguments for command")
 
+// CmdOptions contains state that's required during each command execution.
```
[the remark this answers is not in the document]
anpep: Nice phrasing, thanks!

flotter (approved):

internals/cli/wait.go
```diff
@@ -31,9 +31,9 @@ var (
 )
 
 type waitMixin struct {
-	clientMixin
 	NoWait    bool `long:"no-wait"`
 	skipAbort bool
+	client    *client.Client
```
flotter, line 36: I think we have a slight inconsistency left here, which is not critical, but I will point it out in anycase.

1. We require the remaining ```MiXin``` helpers to include a client, which feels a bit odd. These ```MiXin``` helpers are used to include a common set of functionality often used.
2. All of a sudden, if you wish to have some common args, you need to leave out the command client definition, for some of the ```MiXins``` in the command struct, and even not all of them.
3. By specifying it inside the ```MiXin```, we make these helpers ugly to export in the future, because unlike the external command that may use a custom client, they have to then be setup with a Pebble client (they are private today - should they be Public for use by external commands - maybe later on demand if we really need them)?

So here are my thoughts:

- Leave them private until we need them
- Remove the client definition inside the ```MiXin``` struct, and rather supply it to the member.
- This also allows multiple helper ```MiXins``` to be cleaner to support in the same command.

e.g:

```go
func (wmx waitMixin) wait(client client.Client, id string) (*client.Change, error) {
```

flotter (approved):

benhoyt (approved):
Yeah, I like this -- it's a tiny bit more verbose, but it seems much more direct and Go-like than the "mixin" concept.

internals/cli/cli.go
```diff
@@ -52,6 +52,14 @@ const defaultPebbleDir = "/var/lib/pebble/default"
 // ErrExtraArgs is returned  if extra arguments to a command are found
 var ErrExtraArgs = fmt.Errorf("too many arguments for command")
 
+// CmdOptions exposes state made accessible during command execution.
+type CmdOptions struct {
+	// Client is the Pebble client instance
+	Client *client.Client
+	// Parser is the instance of the flags.Parser that processed the command.
```
niemeyer, line 59: Comments should add information which is not obvious, otherwise they're a negative asset: adds noise and work for readers, with no extra value. The two examples above are typical cases of this, as we're literally saying "foo is an instance of Foo". Can we please just drop both?

internals/cli/cli.go
```diff
@@ -139,34 +146,18 @@ func fixupArg(optName string) string {
...
-
 type defaultOptions struct {
 	Version func() `long:"version" hidden:"yes" description:"Print the version and exit"`
 }
 
 // Parser creates and populates a fresh parser.
 // Since commands have local state a fresh parser is required to isolate tests
 // from each other.
-func Parser(cli *client.Client) *flags.Parser {
+func Parser(c *client.Client) *flags.Parser {
```
niemeyer, line 156: That looks like an unnecessary change. It's removing clarity, and creating conflicts across merges because the meaning of `c` has been shifted in the code below, and in the conversations that we have around this code, as we've experienced live yesterday.

internals/cli/cli.go
```diff
@@ -191,61 +182,55 @@ func Parser(cli *client.Client) *flags.Parser {
...
 
 		var target *flags.Command
-		if c.Debug {
+		if cmdInfo.Debug {
 			target = debugCmd
 		} else {
 			target = parser.Command
 		}
-		cmd, err := target.AddCommand(c.Name, c.Summary, strings.TrimSpace(c.Description), obj)
+		cmd, err := target.AddCommand(cmdInfo.Name, cmdInfo.Summary, strings.TrimSpace(cmdInfo.Description), obj)
```
niemeyer, line 194: Same: why would we change short variable names unnecessarily and reuse them for different purposes inside a patch that is already a massive reorganization?  That creates deltas like the above, which are unnecessary noise in an otherwise good PR.

internals/cli/cmd_add.go
niemeyer: I'm with Fred here, can we please preserve the position of the client, and indeed have it as the first field of every struct that requires it?  This is the common ground for all commands that take it, so it shoud be early in the definition.

internals/cli/cmd_autostart.go
```diff
@@ -51,7 +54,7 @@ func (cmd cmdAutoStart) Execute(args []string) error {
 		return err
 	}
 
-	if _, err := cmd.wait(changeID); err != nil {
+	if _, err := cmd.wait(cmd.client, changeID); err != nil {
```
niemeyer, line 57: If we're handing the client in, is there a benefit in making this a mixin?

internals/cli/cmd_run.go
niemeyer: For the record, Fred is on a good line of thinking in his reviews: he's looking for and suggesting organization and consistency. Might be worth paying a bit more attention to his nitpicks.

internals/cli/cmd_services.go
niemeyer: For the record, Fred is on a good line of thinking in his reviews: he's looking for and suggesting organization and consistency. Might be worth paying a bit more attention to his nitpicks.

> My rule is: only append a space after 2 or more lines of the same category. In this case we don't have it so I didn't add a space.

It's nice that you're also looking for consistency, but I don't even understand how to apply this rule given the above. To me it also looks off considering what you had elsewhere. Probably worth rethinking it a bit so people can have more empathy for it.

niemeyer (approved):
This turned out very well, thank you. Only minor comments below and LGTM with these considered.

internals/cli/cmd_autostart.go
anpep: It's functionality shared by `autostart`, `replan`, `restart`, `start` and `stop` commands. I think a refactor here is possible (and desirable) but I'd rather leave it out of the scope of this PR.

paul-rodriguez (approved):

internals/cli/cmd_services.go
anpep: I've updating the spacing in all cases to stay more consistent and simpler: let's group members such as `client` or `parser` on the top, and whatever has to do with CLI options after a blank line in between.

flotter:
Discussed final state with Gustavo and got go-ahead to merge.

PR #289, opened 2023-08-25
feat(client): Export `DoSync`/`DoAsync` and `RequestInfo`.

anpep:
This patch exports `client.Do{Sync,Async}` methods and moves the arguments to a `RequestInfo` struct. The idea is that applications extending Pebble can call these helper methods in order to perform HTTP requests to the daemon through an extended client (see #285).

### DIFF
--- a/client/changes.go
+++ b/client/changes.go
@@ -89,8 +89,10 @@ type changeAndData struct {
 // Change fetches information about a Change given its ID.
 func (client *Client) Change(id string) (*Change, error) {
 	var chgd changeAndData
-	_, err := client.doSync("GET", "/v1/changes/"+id, nil, nil, nil, &chgd)
-	if err != nil {
+	if _, err := client.DoSync(&RequestInfo{
+		Method: "GET",
+		Path:   "/v1/changes/" + id,
+	}, &chgd); err != nil {
 		return nil, err
 	}
 
@@ -111,7 +113,11 @@ func (client *Client) Abort(id string) (*Change, error) {
 	}
 
 	var chg Change
-	if _, err := client.doSync("POST", "/v1/changes/"+id, nil, nil, &body, &chg); err != nil {
+	if _, err := client.DoSync(&RequestInfo{
+		Method: "POST",
+		Path:   "/v1/changes/" + id,
+		Body:   &body,
+	}, &chg); err != nil {
 		return nil, err
 	}
 
@@ -158,7 +164,11 @@ func (client *Client) Changes(opts *ChangesOptions) ([]*Change, error) {
 	}
 
 	var chgds []changeAndData
-	_, err := client.doSync("GET", "/v1/changes", query, nil, nil, &chgds)
+	_, err := client.DoSync(&RequestInfo{
+		Method: "GET",
+		Path:   "/v1/changes",
+		Query:  query,
+	}, &chgds)
 	if err != nil {
 		return nil, err
 	}
@@ -190,7 +200,11 @@ func (client *Client) WaitChange(id string, opts *WaitChangeOptions) (*Change, e
 		query.Set("timeout", opts.Timeout.String())
 	}
 
-	_, err := client.doSync("GET", "/v1/changes/"+id+"/wait", query, nil, nil, &chgd)
+	_, err := client.DoSync(&RequestInfo{
+		Method: "GET",
+		Path:   "/v1/changes/" + id + "/wait",
+		Query:  query,
+	}, &chgd)
 	if err != nil {
 		return nil, err
 	}
--- a/client/changes.go
+++ b/client/changes.go
@@ -89,10 +89,11 @@ type changeAndData struct {
 // Change fetches information about a Change given its ID.
 func (client *Client) Change(id string) (*Change, error) {
 	var chgd changeAndData
-	if _, err := client.DoSync(&RequestInfo{
+	_, err := client.DoSync(&RequestInfo{
 		Method: "GET",
 		Path:   "/v1/changes/" + id,
-	}, &chgd); err != nil {
+	}, &chgd)
+	if err != nil {
 		return nil, err
 	}
 
@@ -113,11 +114,12 @@ func (client *Client) Abort(id string) (*Change, error) {
 	}
 
 	var chg Change
-	if _, err := client.DoSync(&RequestInfo{
+	_, err := client.DoSync(&RequestInfo{
 		Method: "POST",
 		Path:   "/v1/changes/" + id,
 		Body:   &body,
-	}, &chg); err != nil {
+	}, &chg)
+	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


flotter:
This PR is required to truly make the Client API changes in https://github.com/canonical/pebble/pull/285 usable from a client extension implementation. This is currently in draft form, mainly to capture some proposals on how to best do this.

client/client.go
```diff
@@ -326,13 +326,27 @@ func decodeInto(reader io.Reader, v interface{}) error {
 	return nil
 }
 
-// doSync performs a request to the given path using the specified HTTP method.
+// RequestInfo holds the information to perform a request to the daemon.
+type RequestInfo struct {
```
paul-rodriguez, line 330: Is there a reason not to call this `Request` ?

client/client.go
```diff
@@ -326,13 +326,27 @@ func decodeInto(reader io.Reader, v interface{}) error {
...
+
+// DoSync performs a request to the given path using the specified HTTP method.
 // It expects a "sync" response from the API and on success decodes the JSON
 // response payload into the given value using the "UseNumber" json decoding
 // which produces json.Numbers instead of float64 types for numbers.
-func (client *Client) doSync(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) (*ResultInfo, error) {
+func (client *Client) DoSync(req *RequestInfo, v interface{}) (*ResultInfo, error) {
 	var rsp response
-	if err := client.do(method, path, query, headers, body, &rsp); err != nil {
+	if err := client.do(req.Method, req.Path, req.Query, req.Headers, req.Body, &rsp); err != nil {
```
paul-rodriguez, line 349: You're passing the entire struct to `client.do`, so you might as well just pass `req`?

client/client.go
```diff
@@ -354,22 +368,28 @@ func (client *Client) doSync(method, path string, query url.Values, headers map[
...
+// DoAsync performs a request to the given path using the specified HTTP method.
+// It expects an "async" response from the API and on success returns the
+// change ID.
+func (client *Client) DoAsync(req *RequestInfo) (changeID string, err error) {
+	_, changeID, err = client.DoAsyncFull(req)
 	return
 }
 
-func (client *Client) doAsyncFull(method, path string, query url.Values, headers map[string]string, body io.Reader) (result json.RawMessage, changeID string, err error) {
+// DoAsync performs a request to the given path using the specified HTTP method.
```
paul-rodriguez, line 379: Typo/copypaste error here

client/client.go
```diff
@@ -354,22 +368,28 @@ func (client *Client) doSync(method, path string, query url.Values, headers map[
...
+// It expects an "async" response from the API and on success returns the raw
+// JSON response from the daemon alongside the change ID.
+func (client *Client) DoAsyncFull(req *RequestInfo) (result json.RawMessage, changeID string, err error) {
```
paul-rodriguez, line 382: Would it be better to call this `DoAsyncRaw` ?

paul-rodriguez (commented):
Naming suggestions

client/client.go
benhoyt: I agree `Request` is nicer, except for the existing `ResultInfo` type has the `Info` suffix, so it's probably better to match that for consistency.

client/client.go
```diff
@@ -326,13 +326,27 @@ func decodeInto(reader io.Reader, v interface{}) error {
...
+// general information that makes sense to requests at a more general
+// level, and might be disconnected from the specific request at hand.
+type ResultInfo struct{}
+
+// DoSync performs a request to the given path using the specified HTTP method.
 // It expects a "sync" response from the API and on success decodes the JSON
 // response payload into the given value using the "UseNumber" json decoding
 // which produces json.Numbers instead of float64 types for numbers.
-func (client *Client) doSync(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) (*ResultInfo, error) {
+func (client *Client) DoSync(req *RequestInfo, v interface{}) (*ResultInfo, error) {
```
benhoyt, line 347: We `v` be better named `output` or similar?

client/client.go
benhoyt: Alternatively we just expose the "full" function but call it `DoAsync`, and callers can ignore the first return value if they don't care about it?

client/changes.go
```diff
@@ -89,8 +89,10 @@ type changeAndData struct {
 // Change fetches information about a Change given its ID.
 func (client *Client) Change(id string) (*Change, error) {
 	var chgd changeAndData
-	_, err := client.doSync("GET", "/v1/changes/"+id, nil, nil, nil, &chgd)
-	if err != nil {
+	if _, err := client.DoSync(&RequestInfo{
+		Method: "GET",
+		Path:   "/v1/changes/" + id,
+	}, &chgd); err != nil {
```
benhoyt, line 95: I think this construct would be clearer as it was, with the `if err != nil` as a separate statement. The `}, &chgd); err != nil {` on one line is a bit awkward. Similar throughout.

benhoyt (approved):
Approved, but some minor comments.

client/client.go
anpep: As @benhoyt said, it's `RequestInfo` for consistency so I'm resolving this. Feel free to unresolve if you'd like further discussion (:

benhoyt (approved):

client/client.go
hpidcock: ResultInfo doesn't have the Result. It is a placeholder. I would prefer Request.

client/client.go
```diff
@@ -326,13 +326,27 @@ func decodeInto(reader io.Reader, v interface{}) error {
...
+// general information that makes sense to requests at a more general
+// level, and might be disconnected from the specific request at hand.
+type ResultInfo struct{}
+
+// DoSync performs a request to the given path using the specified HTTP method.
 // It expects a "sync" response from the API and on success decodes the JSON
 // response payload into the given value using the "UseNumber" json decoding
 // which produces json.Numbers instead of float64 types for numbers.
-func (client *Client) doSync(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) (*ResultInfo, error) {
+func (client *Client) DoSync(req *RequestInfo, output interface{}) (*ResultInfo, error) {
```
hpidcock, line 347: I would prefer to rename `output` here to `result` to be consistent with the internal name inside the `response` struct and the DoAsync retval name.

hpidcock (approved):

benhoyt:
> ResultInfo doesn't have the Result. It is a placeholder. I would prefer Request.

@hpidcock That's fair. I'd be fine with `Request` for this -- it does make more sense than "request info".

anpep:
Closing this because it's basically exposing an implementation detail from the Pebble client that might not be suitable for other clients. Will iterate on a new proposal soon.

PR #290, opened 2023-08-28
docs: fix "service dependencies" section in README

benhoyt:
It currently says that "requires" does ordering ("Before ... first"), but that's not true -- "requires" only specifies dependencies. It "before" and "after" that does ordering.

Also note that limited usefulness of before/after as Pebble doesn't really check that a service is "up", just waits 1s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ For example, if service `nginx` requires `logger`, `pebble start nginx` will sta
 
 When multiple services need to be started together, they're started in order according to the `before` and `after` configuration, waiting 1 second for each to ensure the command doesn't exit too quickly. The `before` option is a list of services that must be started before this one (it may or may not `require` them). Or if it's easier to specify this ordering the other way around, `after` is a list of services that must be started after this one.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -244,11 +244,13 @@ If you want to force a service to restart even if its service configuration hasn
...
 
-If multiple dependencies need to be started at once, they're started in order according to the `before` and `after` configuration: `before` is a list of services that must be started before this one (but it doesn't `require` them). Or if it's easier to specify the other way around, `after` is a list of services that must be started after this one.
+When multiple services need to be started together, they're started in order according to the `before` and `after` configuration, waiting 1 second for each to ensure the command doesn't exit too quickly. The `before` option is a list of services that must be started before this one (it may or may not `require` them). Or if it's easier to specify this ordering the other way around, `after` is a list of services that must be started after this one.
+
+Note that currently, `before` and `after` are of limited usefulness, because Pebble only waits 1 second before moving on to start the next service, with no additional checks that the previous service is operating correctly. In future, health checks may be used for this, but right now the logic is simplistic.
```
hpidcock, line 253: I'd drop the comment about what it might do differently in the future.

hpidcock (approved):
Thanks for the clarification.

README.md
benhoyt: Fair call -- done.

jnsgruk (approved):
Thanks for the clarification :)

PR #291, opened 2023-08-29
docs: fix polarity of before/after in "service dependencies" section

benhoyt:
Thanks @troyanov for pointing out the mistake.

See also recent changes in #290 (though that didn't introduce this problem).


hpidcock (approved):
Much better. Thanks.

PR #294, opened 2023-09-02
Return multiple validation errors at once

### LINKED ISSUE #251: Return all validation errors at once when validating in `plan.CombineLayers`
There are multiple issues with the following plan:
```
services:
  svc1:
    startup: enabled
checks:
  chk1:
    override: replace
```

However, when this is run through `plan.CombineLayers` (which does a series of plan validations), it only returns a single problem:
```
layer "layer-0" must define "override" for service "svc1"
```

It would be more useful *not* to exit early, run all the validation checks, and then return an error message stating **everything** that's wrong with the plan, e.g.
```
invalid plan:
- layer "layer-0" must define "override" for service "svc1"
- plan must define "command" for service "svc1"
- plan must specify one of "http", "tcp", or "exec" for check "chk1"
```

aliamerj:
This PR aims to improve error reporting by returning all validation errors at once when validating in `plan.CombineLayers.`

## Changes
- Removed early return statements in validation logic to allow for the collection of multiple errors.
- Stored multiple error messages in a slice, which is then used to form a unified FormatError.

## Rationale
Prior to this change, the function would exit early upon encountering the first error during validation. This was less user-friendly as it required multiple iterations to identify all the errors. With this PR, all errors are collated into a single message, thereby improving the debugging experience.

## Testing
Manually tested to confirm that all error conditions are caught and reported in one go.

Fixes #251

### DIFF
--- a/internals/daemon/api_plan_test.go
+++ b/internals/daemon/api_plan_test.go
@@ -16,8 +16,11 @@ package daemon
 
 import (
 	"bytes"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"sort"
+	"strings"
 
 	. "gopkg.in/check.v1"
 	"gopkg.in/yaml.v3"
@@ -181,21 +184,192 @@ services:
 	s.planLayersHasLen(c, 1)
 }
 
-func (s *apiSuite) TestLayersCombineFormatError(c *C) {
+// ! missing override in one service
+func (s *apiSuite) TestLayersCombineOneError(c *C) {
+	testLayersCombineFormatError(s, c,
+		"services:\n dynamic:\n  command: echo dynamic\n",
+		`layer "base" must define "override" for service "dynamic"`)
+
+}
+
+func (s *apiSuite) TestLayersCombineServicesMultiErrors(c *C) {
+	input := `
+services:
+	svc1:
+	  override: replace
+	  command: ""
+	  on-success: invalid
+	  on-failure: invalid
+	  on-check-failure:
+		 svc1: foo
+	  backoff-factor: 0
+	svc2:
+	  override: replace
+	  command: invalidCommand
+	  on-success: invalid
+	  on-failure: invalid
+	  on-check-failure:
+		 svc1: foo
+	  backoff-factor: 0
+	svc3:
+	  override: foo
+	  command: ""
+	  on-success: invalid
+	  on-failure: invalid
+	  on-check-failure:
+		 svc3: foo
+	  backoff-factor: 0
+	svc4:
+	  command: cmd
+	  on-success: invalid
+	  on-failure: invalid
+	  on-check-failure:
+		 svc4: foo
+	  backoff-factor: 0
+	  `
+
+	expectedError := `multiple errors validating plan:
+- layer "base" has invalid "override" value for service "svc3"
+- layer "base" must define "override" for service "svc4"
+- plan service "svc2" on-success action "invalid" invalid
+- plan service "svc2" on-failure action "invalid" invalid
+- plan service "svc2" on-check-failure action "foo" invalid
+- plan service "svc2" backoff-factor must be 1.0 or greater, not 0
+- plan must define "command" for service "svc1"
+- plan service "svc1" on-success action "invalid" invalid
+- plan service "svc1" on-failure action "invalid" invalid
+- plan service "svc1" on-check-failure action "foo" invalid
+- plan service "svc1" backoff-factor must be 1.0 or greater, not 0
+`
+
+	testLayersCombineFormatError(s, c, input, expectedError)
+
+}
+
+func (s *apiSuite) TestLayersCombineChecksMultiErrors(c *C) {
+	input := `
+checks:
+  svc1:
+    override: replace
+    level: foo
+    period: 0
+    timeout: 0
+    http:
+      url: ""
+    tcp:
+      port: 0
+    exec:
+      command: ""
+      service-context: foo
+      user-id: 0
+  svc2:
+    override: replace
+    level: alive
+    period: 10s
+    timeout: 10s
+    exec:
+      command: invalid
+  svc3:
+    override: foo
+    level: alive
+    exec:
+      command: ""
+  svc4:
+    level: alive
+    exec:
+      command: cmd
+	  `
+
+	expectedError := `multiple errors validating plan:
+- layer "base" has invalid "override" value for check "svc3"
+- layer "base" must define "override" for check "svc4"
+- plan check "svc2" timeout must be less than period
+- plan check "svc1" level must be "alive" or "ready"
+- plan check "svc1" period must not be zero
+- plan check "svc1" timeout must not be zero
+- plan must set "url" for http check "svc1"
+- plan must set "port" for tcp check "svc1"
+- plan must set "command" for exec check "svc1"
+- plan check "svc1" service context specifies non-existent service "foo"
+- plan check "svc1" has invalid user/group: must specify group, not just UID
+- plan must specify one of "http", "tcp", or "exec" for check "svc1"
+`
+
+	testLayersCombineFormatError(s, c, input, expectedError)
+
+}
+
+func (s *apiSuite) TestLayersCombineLogTargetMultiErrors(c *C) {
+	input := `
+log-targets:
+  svc1:
+    override: replace
+    type: foo
+  svc2:
+    override: replace
+    type: ""
+    services: ["foo"]
+  svc3:
+    override: foo
+    type: ""
+  svc4:
+    type: loki
+    services: ["foo"]
+	  `
+
+	expectedError := `multiple errors validating plan:
+- layer "base" has invalid "override" value for log target "svc3"
+- layer "base" must define "override" for log target "svc4"
+- log target "svc1" has unsupported type "foo", must be "loki" or "syslog"
+- plan must define "location" for log target "svc1"
+- plan must define "type" ("loki" or "syslog") for log target "svc2"
+- log target "svc2" specifies unknown service "foo"
+- plan must define "location" for log target "svc2"
+
+`
+
+	testLayersCombineFormatError(s, c, input, expectedError)
+
+}
+
+func testLayersCombineFormatError(s *apiSuite, c *C, content string, response string) {
+	// Manually escape the string
+	content = strings.ReplaceAll(content, "\t", "    ")
+	escapedContent := strings.ReplaceAll(content, "\n", "\\n")
+	escapedContent = strings.ReplaceAll(escapedContent, "\"", "\\\"")
+	payload := fmt.Sprintf(`{"action": "add", "combine": true, "label": "base", "format": "yaml", "layer": "%s"}`, escapedContent)
 	writeTestLayer(s.pebbleDir, planLayer)
 	_ = s.daemon(c)
 	layersCmd := apiCmd("/v1/layers")
-
-	payload := `{"action": "add", "combine": true, "label": "base", "format": "yaml", "layer": "services:\n dynamic:\n  command: echo dynamic\n"}`
 	req, err := http.NewRequest("POST", "/v1/layers", bytes.NewBufferString(payload))
 	c.Assert(err, IsNil)
 	rsp := v1PostLayers(layersCmd, req, nil).(*resp)
+
 	rec := httptest.NewRecorder()
 	rsp.ServeHTTP(rec, req)
 	c.Assert(rec.Code, Equals, http.StatusBadRequest)
 	c.Assert(rsp.Status, Equals, http.StatusBadRequest)
 	c.Assert(rsp.Type, Equals, ResponseTypeError)
 	result := rsp.Result.(*errorResult)
-	c.Assert(result.Message, Matches, `layer "base" must define "override" for service "dynamic"`)
+	sortedExpected := sortErrorLines(response)
+	sortedActual := sortErrorLines(result.Message)
+	c.Assert(sortedActual, Equals, sortedExpected)
+}
 
+func sortErrorLines(errorStr string) string {
+	lines := strings.Split(errorStr, "\n")
+
+	// Filter out empty lines
+	var nonEmptyLines []string
+	for _, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if trimmed != "" {
+			nonEmptyLines = append(nonEmptyLines, trimmed)
+		}
+	}
+
+	// Sort the non-empty lines
+	sort.Strings(nonEmptyLines)
+
+	return strings.Join(nonEmptyLines, "\n")
 }
--- a/internals/daemon/api_plan_test.go
+++ b/internals/daemon/api_plan_test.go
@@ -16,11 +16,8 @@ package daemon
 
 import (
 	"bytes"
-	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"sort"
-	"strings"
 
 	. "gopkg.in/check.v1"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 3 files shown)


aliamerj:
This commit resolves Issue #251, which required the `plan.CombineLayers` function to perform all validation checks and return aggregated errors, rather than exiting early on the first error.

aliamerj (commented):
by mistake

aliamerj:
Adjusted the TestLayersCombineFormatError test to align with the updated error message style in the plan

benhoyt:
Note that this PR doesn't actually fix the issue mentioned in #251, so I've updated the title and description to match what it does do.

internals/plan/plan.go
```diff
@@ -535,6 +535,7 @@ func (e *FormatError) Error() string {
 // CombineLayers combines the given layers into a single layer, with the later
 // layers overriding earlier ones.
 func CombineLayers(layers ...*Layer) (*Layer, error) {
+	var errorMessages []string // To collect multiple errors
```
benhoyt, line 538: I wonder if we can avoid the `errorMessages = append(errorMessages, ...)` boilerplate by doing the following:

```go
var errorMessages []string
addError := func(message string) {
    errorMessages = append(errorMessages, message)
}
```

Or even:

```go
var errorMessages []string
addErrorf := func(format string, args ...any) {
    errorMessages = append(errorMessages, fmt.Sprintf(format, args...))
}
```

internals/plan/plan.go
```diff
@@ -776,17 +731,24 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
-				Message: fmt.Sprintf(`plan must define "location" for log target %q`, name),
-			}
+			errorMessages = append(errorMessages, fmt.Sprintf(`plan must define "location" for log target %q`, name))
+
+		}
+	}
+
+	// Ensure combined layers don't have errors
+	if len(errorMessages) > 0 {
+		fullErrorMessage := "\ninvalid plan:\n- " + strings.Join(errorMessages, "\n- ")
```
benhoyt, line 747: How about we keep these error messages the same when `len(errorMessages) == 1`, and note the multiple errors when >1. And let's drop the leading `\n`. Maybe:

```go
if len(errorMessages) == 1 {
    return nil, &FormatError{Message: errorMessages[0]}
}
if len(errorMessages) > 0 {
    message := "multiple errors validating plan:\n- " + strings.Join(errorMessages, "\n- ")
    return nil, &FormatError{Message: message}
}
```

benhoyt (changes requested):
Idea looks good overall. Couple of suggestions to tighten this up. And it looks like the tests are failing (you should be able to repro locally with `go test ./...`). Hopefully fewer tests will fail if you make the second change I've mentioned.

benhoyt:
Oops, sorry, this does fix #251, just not the `override` issue you mention further down -- updating again.

benhoyt:
@aliamerj Out of interest, how did you stumble across Pebble, and how are you using it? No need to answer if you don't want to, but I'm always curious about other people's use cases.

barrettj12 (commented):
Thanks @aliamerj, this is a great start. The main complication here is that some of the validation checks are dependent on other checks. For example, often we have two different checks:

A. checking that a required field `foo` is present
B. checking that `foo` has the correct format

If check A has already failed, it doesn't make sense to do check B. So we have to be careful about when we decide to do a check vs when we skip it because pre-checks have failed.

As a general rule, we should probably return at most one error per field. It's not that useful to see:
```
- plan must define "command" for service "foo"
- plan service "foo" command invalid: ...
```

aliamerj:
@benhoyt @barrettj12
Thank you for replying. I have read your comments and found them very helpful. I will try to fix this issue today. However, I was trying to match the use case provided by @barrettj12, so I set up the plan layer like this:

```yml
# ~/pebble/layers/123-some-label.yaml
services:
  svc1:
    startup: enabled
checks:
  chk1:
    override: replace
```
I also updated `plan.CombineLayers` to produce output like this:
```
invalid plan:
- layer "layer-0" must define "override" for service "svc1"
- plan must define "command" for service "svc1"
- plan must specify one of "http", "tcp", or "exec" for check "chk1"
```

# To avoid misunderstanding, please correct me if I'm wrong
In the error message handler, there should be two cases:

## 1 - If the `override` value is missing or invalid
       Then, just return one error message.
         Example:
            layer "layer-0" must define "override" for service "svc1" 

## 2- if the `override` set crorrectly
    Then collect multiple errors and return them.
       Example:
     
      invalid plan:
      - plan must define "command" for service "svc1"
      - plan must specify one of "http", "tcp", or "exec" for check "chk1"

aliamerj:
@benhoyt 
For your question on how I stumbled across Pebble, 
I'm actually new to this. I've been learning about Docker and was curious about how it works. In the process, so studyed namespaces and cgroups, and I've even made a [small project ](https://github.com/aliamerj/simple-container-manager) while learning these concepts.
since I really enjoy building CLI tools  , and i was hoping to work on real one .
I read about Pebble and am excited to contribute in it .

I have actually this crazy idea. I'm planning to contribute to Pebble daily, trying to address open issues. My goal is to work on a project that genuinely interests me while also gaining valuable work experience, since i am still an IT student in university and  trying to build good CV
I hope I can be a useful addition to your team.

aliamerj:
If I understand correctly, there's no need to set a flag for indicating "don't check the services." I will keep the return statements as they were for errors like missing or invalid "override." If the error message is not related to "override," I will collect them in a slice and return them at the end of the function.
what do you think ??

internals/plan/plan.go
```diff
@@ -560,15 +561,12 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 			case ReplaceOverride:
 				combined.Services[name] = service.Copy()
 			case UnknownOverride:
-				return nil, &FormatError{
-					Message: fmt.Sprintf(`layer %q must define "override" for service %q`,
-						layer.Label, service.Name),
-				}
+				errorMessages = append(errorMessages, fmt.Sprintf(`layer %q must define "override" for service %q`,
+					layer.Label, service.Name))
```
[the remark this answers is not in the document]
benhoyt: I think this part works fine as is, because if override is missing/invalid, we won't add to the `combined.Services` map, so it won't be checked below.

benhoyt:
Yep, your summary in [this comment](https://github.com/canonical/pebble/pull/294#issuecomment-1704813552) is what I was suggesting.

Nice `containers` project! I too find that it's really helpful to learn how something works by re-implementing parts of it.

> I have actually this crazy idea. I'm planning to contribute to Pebble daily, trying to address open issues.

That's a nice idea! However, just bear in mind that Pebble is a fairly slow-moving project, and has quite thorough code review requirements, because it's a core piece of several projects here at Canonical. Help is appreciated, but I just want to be up-front about the (relatively slow) pace this project moves at, and the limited time for code reviews on this end. So be patient. :-)

> I will keep the return statements as they were for errors like missing or invalid "override." If the error message is not related to "override," I will collect them in a slice and return them at the end of the function.

I don't think that's quite right. If there are multiple services, you still want to check each one of those and (potentially) return multiple errors. For example, this would return 3 "must define override for service" errors:

```
services:
  svc1:
    command: foo
  svc2:
    command: bar
  svc3:
    command: bazz
```

Regarding the code snippet you sent, I would suggest a more straight-forward translation of what we have today, like what @barrettj12 suggested. For each field, use an if/else so there's only one error per field, but then still append to the `errorMessages` slice. For example, this tidies up the `combined.Services` checking quite a bit:

```go
	for name, service := range combined.Services {
		if service.Command == "" {
			addErrorf(`plan must define "command" for service %q`, name)
		} else if _, _, err := service.ParseCommand(); err != nil {
			addErrorf("plan service %q command invalid: %v", name, err)
		}
		if !validServiceAction(service.OnSuccess) {
			addErrorf("plan service %q on-success action %q invalid", name, service.OnSuccess)
		}
		if !validServiceAction(service.OnFailure) {
			addErrorf("plan service %q on-failure action %q invalid", name, service.OnFailure)
		}
		for _, action := range service.OnCheckFailure {
			if !validServiceAction(action) {
				addErrorf("plan service %q on-check-failure action %q invalid", name, action)
			}
		}
		if !service.BackoffDelay.IsSet {
			service.BackoffDelay.Value = defaultBackoffDelay
		}
		if !service.BackoffFactor.IsSet {
			service.BackoffFactor.Value = defaultBackoffFactor
		} else if service.BackoffFactor.Value < 1 {
			addErrorf("plan service %q backoff-factor must be 1.0 or greater, not %g", name, service.BackoffFactor.Value)
		}
		if !service.BackoffLimit.IsSet {
			service.BackoffLimit.Value = defaultBackoffLimit
		}
	}
```

aliamerj:
Ah, I can run multiple services, this is what I was missing.
 In this case, I can use 'continue' to skip to the next service in the loop.

 Okay, I will push the update now to my fork and  If you have any suggestions, you can write here, and I will update it and push it tomorrow

barrettj12:
@aliamerj I don't think you want to use `continue` either. If `service.Command == ""`, then we just want to skip the `service.ParseCommand()` step - but we still want to run all the other service validation checks (actions, backoffs, etc)

aliamerj:
@barrettj12 
Thank you for the suggestion. I will work on them as soon as I finish this issue.

aliamerj:
> I don't think you want to use `continue` either. If `service.Command == ""`, then we just want to skip the `service.ParseCommand()` step - but we still want to run all the other service validation checks (actions, backoffs, etc)

okay but continue will do like this
if service.Command == "" then we all add the error message  and  check the next service 
and with this we all check all the services that been defined 
what i am missing ) ?

barrettj12:
@benhoyt already gave a perfect example of what the code should look like [here](https://github.com/canonical/pebble/pull/294#issuecomment-1705751435).

Maybe you can see how this would produce a different result to your current code. If it helps, you could run through the following example and work out what the error would be.
```yaml
services:
  svc1:
    command: ""
    on-success: invalid
    on-failure: invalid
```

aliamerj:
> @benhoyt already gave a perfect example of what the code should look like [here](https://github.com/canonical/pebble/pull/294#issuecomment-1705751435).
> 
> Maybe you can see how this would produce a different result to your current code. If it helps, you could run through the following example and work out what the error would be.
> 
> ```yaml
> services:
>   svc1:
>     command: ""
>     on-success: invalid
>     on-failure: invalid
> ```

i amg gettin this  when run ` go run ./cmd/pebble run`
```
2023-09-05T10:09:56.183Z [pebble] Started daemon.
2023-09-05T10:09:56.184Z [pebble] POST /v1/services 254.412µs 500
2023-09-05T10:09:56.184Z [pebble] Cannot start default services: layer "some-label" must define "override" for service "svc1
```

aliamerj:
and with 
``` yml
services:
  svc1:
    command: foo
  svc2:
    command: bar
  svc3:
    command: bazz

```
```
2023-09-05T10:14:00.337Z [pebble] Started daemon.
2023-09-05T10:14:00.337Z [pebble] POST /v1/services 268.782µs 500
2023-09-05T10:14:00.337Z [pebble] Cannot start default services: multiple errors validating plan:
- layer "some-label" must define "override" for service "svc2"
- layer "some-label" must define "override" for service "svc3"
- layer "some-label" must define "override" for service "svc1"
```

aliamerj:
Yeb, I should learn more about the process in Pebble and how it handles services. However, it may be that my implementation is correct; I just need to remove some `continue` statements from some loops for it to work fine

barrettj12:
@aliamerj I'm sorry, I think maybe I confused you with some of my comments. What you had initially was much closer to the correct code.

> i'm gettin this when run ` go run ./cmd/pebble run`
> 
> ```
> 2023-09-05T10:09:56.183Z [pebble] Started daemon.
> 2023-09-05T10:09:56.184Z [pebble] POST /v1/services 254.412µs 500
> 2023-09-05T10:09:56.184Z [pebble] Cannot start default services: layer "some-label" must define "override" for service "svc1
> ```

Sorry, bad example. Here's a better example:
```yaml
services:
  svc1:
    override: merge
    command: ""
    on-success: invalid
    on-failure: invalid
```

The point is, there are multiple errors in this plan, but your code will only return one error.

barrettj12:
> okay but continue will do like this
if service.Command == "" then we all add the error message and check the next service

Yes, but we **don't** want to move on to the next service yet. There are still other fields to check (e.g. the actions and backoffs). My point was, if we've already determined that `service.Command` is invalid because it's empty, then we shouldn't attempt to parse it either (service.ParseCommand).

aliamerj:
@barrettj12 
No, I apologize for wasting your time. I am still new to the project and am still  trying to understand its structure. Specifically, I'm trying to differentiate between main errors and sub-errors for each one.
so that what i understanded 
the function `plan.CombineLayers.` do two things validation and combining 
and with errors
there are 3 main branchs => 
# 1- errors tree for services
```
  Override --> command 
      |
      | ----> OnSuccess
      | ----> OnFailure
      |-----> BackoffFactor
```  
# 2- errors tree Checks
      Override --> level 
      |
      | ----> Period
      | ----> Timeout
      |-----> HTTP->  url
      |-----> TCP-> port
      |----> Exec -->  Command
               |----> ServiceContext
               |-----> user and group
  *plan must specify one of "http", "tcp", or "exec" for check 

# 3- LogTargets
 Override -->
     |->
     | ->

barrettj12:
> @barrettj12 No, I apologize for wasting your time.

It's okay, we're all trying to help each other here :)

I've reverted this PR to the commit aa1f5222185d67e1dd5ad53b12ed9ebb0a299826, as that was much closer to what we are looking for.

From here, the changes we want to see are:
- [ ] `addErrorf` utility function
- [ ] In the case of `service.Command == ""`, don't do the `service.ParseCommand()` check. **We are just skipping a single check here. Please don't `break` or `continue`, because we still want to check all the other fields for this service.**
- [ ] Similar to the previous point, if `check.Exec.Command == ""`, then don't attempt to parse the command with `shlex.Split`.
- [ ] [`Ben's comment`](https://github.com/canonical/pebble/pull/294/files#r1314388589) about just outputting the single error if there is only one.
- [ ] Update tests to match code changes.

Please ignore my earlier comment about the override - I was asking a question to Ben, rather than providing feedback, and he already said that no changes are needed here.

aliamerj:
@barrettj12 , thank very much 
I really appreciate that. 

What about the  override  
Should it be required in each part in the function and without them we don't know how to combine? 
So if they are missing or invalid should not check further in this specific Service , Checks or LogTarget

barrettj12:
Please just forget about the override, you don't need to make any changes here

internals/plan/plan.go
```diff
@@ -536,6 +536,7 @@ func (e *FormatError) Error() string {
 // layers overriding earlier ones.
 func CombineLayers(layers ...*Layer) (*Layer, error) {
 	var errorMessages []string // To collect multiple errors
+	addError := func(message string) { errorMessages = append(errorMessages, message) }
```
barrettj12, line 539: since addError is always called with fmt.Sprintf, let's pull the fmt.Sprintf call inside addError, and make it addErrorf

aliamerj: @barrettj12 
Is there anything else ??

internals/plan/plan.go
```diff
@@ -776,16 +727,14 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 			if _, ok := combined.Services[serviceName]; ok {
 				continue
 			}
-			return nil, &FormatError{
-				Message: fmt.Sprintf(`log target %q specifies unknown service %q`,
-					target.Name, serviceName),
-			}
+			addErrorf(fmt.Sprintf(`log target %q specifies unknown service %q`,
+				target.Name, serviceName))
+
```
barrettj12, line 732: remove extra newline

barrettj12 (changes requested):

barrettj12:
@aliamerj I'm surprised that all the plan tests are still passing. In this case, we want to have test coverage for the multi-error, so please add some new tests covering the case where a plan has multiple validation errors.

aliamerj:
> @aliamerj I'm surprised that all the plan tests are still passing. In this case, we want to have test coverage for the multi-error, so please add some new tests covering the case where a plan has multiple validation errors.

Thank you for the suggestion .
I will add new tests

barrettj12:
> I will add new tests

Awesome, thanks. Feel free to request another review from me once you have done this.

Any more questions/concerns, reach out and I will be happy to answer :)

aliamerj:
Initially, I was uncertain about whether to write unit tests or integration tests ( API tests). 
I went with API tests to comprehensively cover all possible error messages.

I've implemented four distinct testing functions:
I've implemented four distinct testing functions:

`TestLayersCombineOneError`: Single Error Message — This test case is designed to handle scenarios where only one error message is expected.

`TestLayersCombineServicesMultiErrors`: Multiple Service Errors — This test specifically targets multiple error messages within the 'services' section. It does so by providing completely invalid values in the services to trigger all possible error messages.

`TestLayersCombineChecksMultiErrors`: Checks Validation — This function operates similarly to `TestLayersCombineServicesMultiErrors`, but it focuses on the 'Checks' section.

`TestLayersCombineLogTargetMultiErrors`: Log Target Validation — This function also operates similarly but focuses on the 'Log Target' section.

internals/daemon/api_plan_test.go
```diff
@@ -181,20 +184,192 @@ services:
...
+- plan check "svc2" timeout must be less than period
+- plan check "svc1" level must be "alive" or "ready"
+- plan check "svc1" period must not be zero
+- plan check "svc1" timeout must not be zero
+- plan must set "url" for http check "svc1"
+- plan must set "port" for tcp check "svc1"
+- plan must set "command" for exec check "svc1"
+- plan check "svc1" service context specifies non-existent service "foo"
+- plan check "svc1" has invalid user/group: must specify group, not just UID
+- plan must specify one of "http", "tcp", or "exec" for check "svc1"
```
barrettj12, line 295: nit: this message is a bit unclear, it should probably say that the check can **only** specify one of `http`, `tcp`, `exec`

internals/daemon/api_plan_test.go
```diff
@@ -181,20 +184,192 @@ services:
...
+
 	rec := httptest.NewRecorder()
 	rsp.ServeHTTP(rec, req)
 	c.Assert(rec.Code, Equals, http.StatusBadRequest)
 	c.Assert(rsp.Status, Equals, http.StatusBadRequest)
 	c.Assert(rsp.Type, Equals, ResponseTypeError)
 	result := rsp.Result.(*errorResult)
-	c.Assert(result.Message, Matches, `layer "base" must define "override" for service "dynamic"`)
+	sortedExpected := sortErrorLines(response)
+	sortedActual := sortErrorLines(result.Message)
```
barrettj12, line 355: why is it necessary to sort the lines?

barrettj12 (changes requested):
Thanks @aliamerj - these unit tests look great, but unfortunately, they are in the wrong place. As is conventional with Go packages, the tests for `internals/plan/plan.go` can be found in [`internals/plan/plan_test.go`](https://github.com/canonical/pebble/blob/32e354b1db9de03b57422508fb50fcc6c2f8b4bd/internals/plan/plan_test.go). In fact, we already have testing infrastructure set up there, so you can just add your test cases to the end of the [`planTests` slice](https://github.com/canonical/pebble/blob/32e354b1db9de03b57422508fb50fcc6c2f8b4bd/internals/plan/plan_test.go#L84). You will find that a lot of the boilerplate you implemented in `testLayersCombineFormatError` is already written for you :)

barrettj12:
Just took another look at the actual implementation, and it looks perfect to me now. Let's just fix up those tests and we should be good to go :)

internals/daemon/api_plan_test.go
barrettj12: one more thing - if we get this message because more than one of `http`, `tcp`, `exec` is specified, then there's no need to validate those sections. So this error message should just have
```
plan must specify one of "http", "tcp", or "exec" for check "svc1"
```
and skip the validation for those sections, e.g. don't print
```
- plan must set "command" for exec check "svc1"
- plan check "svc1" service context specifies non-existent service "foo"
- plan check "svc1" has invalid user/group: must specify group, not just UID
```

aliamerj:
"I have converted the API tests for handling multiple error messages into unit tests. Specifically, I've created three unit tests to comprehensively check for errors related to different aspects  as before 
- `TestLayersCombineServicesMultiErrors` is designed to handle all invalid values and display errors related to services in a single test.
- `TestLayersCombineChecksMultiErrors` functions in a similar manner, but focuses on checks.
- `TestLayersCombineLogTargetMultiErrors` is dedicated to log targets.
In addition to these, I created an extra test for handling empty layers. This was done to achieve 100% test coverage for the layersCombine function."

Additionally, I found the ordering of error messages to be quite inconvenient while testing. To make it easier to match the expected and obtained lists of errors, I implemented a small function to sort the error messages. This ensures that the test is flexible enough to validate the error output regardless of the order in which the errors appear.

internals/plan/plan_test.go
```diff
@@ -81,9 +82,10 @@ type planTest struct {
 	stop    map[string][]string
 }
 
-var planTests = []planTest{{
-	summary: "Relatively simple layer with override on top",
-	input: []string{`
+var planTests = []planTest{
+	{
+		summary: "Relatively simple layer with override on top",
+		input: []string{`
```
barrettj12, line 88: Nothing should have changed here. Please undo these formatting changes and revert to the way things were originally.

internals/plan/plan_test.go
```diff
@@ -378,168 +380,168 @@ var planTests = []planTest{{
...
 				command: cmd -v [ --foo ] bar
 	`},
-}, {
-	summary: `Invalid service command: cannot have ] outside of [ ... ] group`,
-	error:   `plan service "svc1" command invalid: cannot parse service "svc1" command: cannot have \] outside of \[ ... \] group`,
-	input: []string{`
+	}, {
+		summary: `Invalid service command: cannot have ] outside of [ ... ] group`,
+		error:   `plan service "svc1" command invalid: cannot parse service "svc1" command: cannot have \] outside of \[ ... \] group`,
+		input: []string{`
```
barrettj12, line 527: again, these changes shouldn't have happened.

internals/plan/plan_test.go
```diff
@@ -1631,3 +1633,165 @@ func (s *S) TestMergeServiceContextOverrides(c *C) {
 		WorkingDir:  "/working/dir",
 	})
 }
+
+func (s *S) TestLayersCombineEmptyLayers(c *C) {
```
barrettj12, line 1637: delete this test

internals/plan/plan_test.go
```diff
@@ -1631,3 +1633,165 @@ func (s *S) TestMergeServiceContextOverrides(c *C) {
...
+- plan service "svc2" backoff-factor must be 1.0 or greater, not 0`
+
+	c.Assert(err, IsNil)                        // If you expect no errors from ParseLayer
+	combined, err := plan.CombineLayers(layer1) // Declare combined and err here
+	c.Assert(combined, IsNil)
+	c.Check(sortErrors(err.Error()), DeepEquals, sortErrors(expectedErrors))
+
+}
+
+func (s *S) TestLayersCombineChecksMultiErrors(c *C) {
```
barrettj12, line 1703: convert into a new test case at the end of the `planTests` slice

internals/plan/plan_test.go
```diff
@@ -1631,3 +1633,165 @@ func (s *S) TestMergeServiceContextOverrides(c *C) {
...
+- plan must specify one of "http", "tcp", or "exec" for check "svc5"`
+
+	c.Assert(err, IsNil)                        // If you expect no errors from ParseLayer
+	combined, err := plan.CombineLayers(layer1) // Declare combined and err here
+	c.Assert(combined, IsNil)
+	c.Check(sortErrors(err.Error()), DeepEquals, sortErrors(expectedErrors))
+
+}
+
+func (s *S) TestLayersCombineLogTargetMultiErrors(c *C) {
```
barrettj12, line 1760: convert into a new test case at the end of the `planTests` slice

internals/plan/plan_test.go
```diff
@@ -1631,3 +1633,165 @@ func (s *S) TestMergeServiceContextOverrides(c *C) {
...
+- plan must define "location" for log target "svc1"`
+
+	c.Assert(err, IsNil)                        // If you expect no errors from ParseLayer
+	combined, err := plan.CombineLayers(layer1) // Declare combined and err here
+	c.Assert(combined, IsNil)
+	c.Check(sortErrors(err.Error()), DeepEquals, sortErrors(expectedErrors))
+
+}
+
+func sortErrors(errMsg string) []string {
```
barrettj12, line 1793: remove this, we shouldn't be modifying the output of `CombineLayers`

internals/plan/plan_test.go
```diff
@@ -1631,3 +1633,165 @@ func (s *S) TestMergeServiceContextOverrides(c *C) {
...
+	c.Assert(err, IsNil) // Assert that there is no error
+	// Additional assertions to check that the returned *Layer is not nil and its fields are empty
+	c.Assert(combined, Not(IsNil))
+	c.Assert(len(combined.Services), Equals, 0)
+	c.Assert(len(combined.Checks), Equals, 0)
+	c.Assert(len(combined.LogTargets), Equals, 0)
+
+}
+
+func (s *S) TestLayersCombineServicesMultiErrors(c *C) {
```
barrettj12, line 1648: convert into a new test case at the end of the `planTests` slice

barrettj12 (changes requested):
Hi @aliamerj, I appreciate the enthusiasm here, but you need to read my review comments more carefully and actually follow them.

In my last review, I asked you to add new tests **at the end of the `planTests` slice.** Instead, you added separate top-level tests, and totally ignored the table-driven test infrastructure we have already developed here.

For your personal projects, you can do things however you like - but when you come to work on an existing open-source project, you have to follow the conventions of that project. Importantly, you also have to **listen to the maintainers** and take their feedback into account, otherwise your PRs won't get approved.

I have already spent a lot of time reviewing your PR - it probably would have taken me less time to implement it myself. Again, I appreciate that you are keen to contribute, but you really need to listen to my feedback, otherwise we're both wasting each other's time.

I will give you one more review. This time, please read my comments very carefully and follow them to the letter. Also go back and read my previous comments and make sure you have implemented those too.

As a final thing, once you have finished making changes, push up your changes, click on the "Files changed" tab **and give yourself a self-review.** Check that you haven't introduced any unnecessary changes such as reformatting things. (This is something you should do on every PR before requesting a review from someone else).

> "I have converted the API tests for handling multiple error messages into unit tests. Specifically, I've created three unit tests to comprehensively check for errors related to different aspects as before
> 
> * `TestLayersCombineServicesMultiErrors` is designed to handle all invalid values and display errors related to services in a single test.
> * `TestLayersCombineChecksMultiErrors` functions in a similar manner, but focuses on checks.
> * `TestLayersCombineLogTargetMultiErrors` is dedicated to log targets.

Please convert these into new test cases at the end of the planTests slice.

>   In addition to these, I created an extra test for handling empty layers. This was done to achieve 100% test coverage for the layersCombine function."

Please remove this. If this was necessary, there would already be test coverage here. Empty layers is not a realistic use case and not one that we care about, so we don't need a test here.

> Additionally, I found the ordering of error messages to be quite inconvenient while testing. To make it easier to match the expected and obtained lists of errors, I implemented a small function to sort the error messages. This ensures that the test is flexible enough to validate the error output regardless of the order in which the errors appear.

In general, this is a really bad idea. Unit tests should ideally compare a function's raw output to an expected value. As soon as you start processing the output, then you have to worry about bugs in your processing, and then you need tests for your tests, etc... it becomes a mess pretty quickly.

The ordering is also an essential part of the error message. If the order changes, it could indicate a breakage or unintended consequences of a change, and we would want to know about this. We shouldn't throw away this information. Unit tests are not supposed to be "flexible" - they are supposed to be precise.

The only case where I would consider doing this is if the output was randomised. However, this is not the case here - the output should be the same every time. Hence, we can just compare the output against an expected value (which is a constant).

aliamerj:
@barrettj12 I am really sorry for wasting your time
 i was trying to make sure all align with previous tests when you test
like in this test 
 ```go
func (s *S) TestMissingCommand(c *C) {
	// Combine fails if no command in combined plan
	layer1, err := plan.ParseLayer(1, "label1", []byte("{}"))
	c.Assert(err, IsNil)
	layer2, err := plan.ParseLayer(2, "label2", []byte(`
services:
    srv1:
        override: merge
`))
	c.Assert(err, IsNil)
	_, err = plan.CombineLayers(layer1, layer2)
	c.Check(err, ErrorMatches, `plan must define "command" for service "srv1"`)
	_, ok := err.(*plan.FormatError)
	c.Check(ok, Equals, true, Commentf("error must be *plan.FormatError, not %T", err))

	// Combine succeeds if there is a command in combined plan
	layer1, err = plan.ParseLayer(1, "label1", []byte(`
services:
    srv1:
        override: merge
        command: foo --bar
`))
	c.Assert(err, IsNil)
	layer2, err = plan.ParseLayer(2, "label2", []byte(`
services:
    srv1:
        override: merge
`))
	c.Assert(err, IsNil)
	combined, err := plan.CombineLayers(layer1, layer2)
	c.Assert(err, IsNil)
	c.Assert(combined.Services["srv1"].Command, Equals, "foo --bar")
}
```
again i am realy sorry for wasting your time

aliamerj:
the last request i made by  mistake

barrettj12:
@aliamerj are you still working on this? It's 80% there, would be great to get it across the line. I've laid out the path for you in my last review. If you need any further clarification or assistance, please don't hesitate to leave a comment here.

aliamerj:
@barrettj12 
Actually I am not but soon I will try to fix the issues with my previous commit . 

I really appreciate you reviewing my code and I will be more careful next time

benhoyt:
Going to close this due to lack of updates. We can always revisit the multiple validation errors concept later if desired.

PR #296, opened 2023-09-05
feat(daemon): add notices API

benhoyt:
This adds the API endpoints required for Pebble Notices (see spec [JU048](https://docs.google.com/document/d/16PJ85fefalQd7JbWSxkRWn0Ye-Hs8S1yE99eW7pk8fA/edit)).

The three endpoints are:

* `GET /v1/notices` with optional `?type=<type>&key=<key>&after=<timestamp>&timeout=<duration>`: get or wait for a list of notices, filtering by type/key/repeat-after timestamp
* `GET /v1/notices/{id}`: get a single notice by notice ID
* `POST /v1/notices`: record the occurrence of a (client) notice


internals/daemon/api_warnings_test.go
```diff
@@ -53,7 +53,7 @@ func (s *apiSuite) testWarnings(c *check.C, all bool, body io.Reader) (calls str
 	if all {
 		q.Set("select", "all")
 	}
-	req, err := http.NewRequest(method, "/v2/warnings?"+q.Encode(), body)
+	req, err := http.NewRequest(method, "/v1/warnings?"+q.Encode(), body)
```
barrettj12, line 56: what's happening here?

barrettj12 (approved):
Looks good to me, just a few minor comments.

internals/daemon/api.go
```diff
@@ -94,6 +94,15 @@ var API = []*Command{{
...
 	GET:    v1GetChecks,
+}, {
+	Path:   "/v1/notices",
+	UserOK: true,
+	GET:    v1GetNotices,
+	POST:   v1PostNotices,
+}, {
+	Path:   "/v1/notices/{id}",
+	UserOK: true,
+	GET:    v1GetNotice,
```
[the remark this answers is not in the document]
benhoyt: I don't disagree they're similar. However, I'm going to follow the existing convention here, for example `v1GetServices` and `v1GetService`.

internals/daemon/api_warnings_test.go
benhoyt: I believe this was a carry-over from snapd, which uses /v2/ for most API prefixes, and this was just forgotten when moving to Pebble. It didn't actually make the test fail, because of how we use `apiCmd()` above.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,154 @@
...
+	if err := decoder.Decode(&payload); err != nil {
+		return statusBadRequest("cannot decode request body: %v", err)
+	}
+
+	if payload.Action != "add" {
+		return statusBadRequest("invalid action %q", payload.Action)
+	}
+	if payload.Type != "client" {
+		return statusBadRequest(`invalid type %q (can only add "client" notices)`, payload.Type)
+	}
```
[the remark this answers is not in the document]
benhoyt: Yeah, however, in case we have other actions later (or allow other types later), it's best to be explicit and require it now. This is similar to how we do it in other APIs like [`v1PostChange`](https://github.com/canonical/pebble/blob/3819000f4b2835484c1e981034987426efeeed85/internals/daemon/api_changes.go#L246-L248).

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,154 @@
...
+
+// A very loose regex to ensure client keys are in the form "domain.com/key"
+var clientKeyRegexp = regexp.MustCompile(`([a-z0-9-_]+\.)+[a-z0-9-_]+/.+`)
+
+func v1GetNotices(c *Command, r *http.Request, _ *UserState) Response {
+	query := r.URL.Query()
+
+	typeStr := query.Get("type")
+	noticeType := state.NoticeTypeFromString(typeStr)
+	if typeStr != "" && noticeType == "" {
```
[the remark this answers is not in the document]
benhoyt: For better or worse, this is pretty typical for Go APIs: if the zero value is not meaningful, it's common to use is it as "not found" value. For example, [http.StatusText](https://pkg.go.dev/net/http#StatusText) and [unix.SignalNum](https://pkg.go.dev/golang.org/x/sys/unix#SignalNum).

internals/daemon/api_notices.go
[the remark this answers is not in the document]
barrettj12: my point is, `action` is already encoded in the HTTP verb. If the action was "change/modify", you should be using a PUT request, and for "remove", you should be using DELETE. What other actions would we need?

Adding fields to the API at a later stage should also be fine, since it is backward compatible and won't break the v1 semantic versioning. In that case, no need to put these in before we actually need them IMO.

internals/daemon/api_notices.go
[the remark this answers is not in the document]
barrettj12: well, I don't think we should follow bad practice just because everyone else is doing it. :)

It would also make this check a lot cleaner and more readable.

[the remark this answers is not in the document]
benhoyt: I don't consider it bad practice, though. Functions that return a single value are somewhat easier to deal with than multi-valued ones. (I do understand the disadvantage: they're less safe, because with `err`-returning ones you have to explicitly check the error.)

I also don't see how it would be a "lot cleaner and more readable", or why the current code is not clean or readable? What do you find confusing about it?

I grepped through the Go stdlib and found two functions in this class that. `http.StatusText` that I've already mentioned, and [`mime.TypeByExtension`](https://pkg.go.dev/mime#TypeByExtension), which also returns `""` in the not-found case. I think following stdlib precedent is a pretty compelling argument.

I didn't find any similar lookup functions that had an error return in the stdlib, though I may have missed them. All the functions with error returns seem to perform some OS or I/O operation. But those that are just simple lookups like this seem to not return error.

internals/daemon/api_notices.go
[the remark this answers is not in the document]
benhoyt: Once again, the precedent is strong with this one. I had this discussion with Gustavo a while ago, and Pebble's (and snap's) API are "vaguely REST-like if you squint" rather than REST-ful (using specific HTTP verbs for different actions and so on). Reads are done using GET and writes using POST, and we use meaningful URLs, but that's about it.

We use this "action" pattern for several other APIs in Pebble:

```
METH /path                actions
---------------------------------
POST /v1/changes/{id}     abort
GET  /v1/files            list read
POST /v1/files            make-dirs remove write
POST /v1/layers           add
POST /v1/services         autostart replan restart start stop
POST /v1/warnings         okay
```

So again, I'd much rather be consistent with that than break the pattern for the notices API.

As far as requiring the "action" field: I think you're right that it would be backwards-compatible to add it later if we needed it, and have the default action be "add" at that point. However, requiring it up-front is consistent with how we do it in the existing endpoints, and in light of the above API design, I think it's forward-looking to require it from the beginning. It certainly doesn't hurt to make things explicit.

internals/daemon/api_notices.go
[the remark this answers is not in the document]
barrettj12: Ok, well my concern is that with
```
noticeType := state.NoticeTypeFromString(typeStr)
if typeStr != "" && noticeType == "" {
```
it takes a while to understand what's happening here, it's not immediately clear, and you have to understand the internal details of the NoticeTypeFromString function to understand what's happening here (namely, that it returns "" when the input is invalid).

In contrast, with
```
noticeType, err := state.NoticeTypeFromString(typeStr)
if err != nil {
```
it's immediately clear that there was an error converting the string to a notice type.

Ultimately I suppose this is a matter of personal taste though.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,154 @@
...
+			return statusBadRequest("invalid timeout %q: %v", timeoutStr, err)
+		}
+
+		ctx, cancel := context.WithTimeout(r.Context(), timeout)
+		defer cancel()
+
+		notices, err = st.WaitNotices(ctx, filters)
+		if errors.Is(err, context.DeadlineExceeded) {
+			return statusGatewayTimeout("timed out after %s", timeout)
+		}
```
flotter, line 77: This is understandable, but it seems inconsistent with the timeout=0 case. If we do not get a filter hit within the given time, where time can be none, should we consider it an error?

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,154 @@
...
+		if errors.Is(err, context.Canceled) {
+			return statusBadRequest("request canceled")
+		}
```
flotter, line 80: Just to be sure, this cannot happen, right ?

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,154 @@
...
+	"errors"
+	"net/http"
+	"regexp"
+	"time"
+
+	"github.com/canonical/pebble/internals/overlord/state"
+)
+
+// A very loose regex to ensure client keys are in the form "domain.com/key"
+var clientKeyRegexp = regexp.MustCompile(`([a-z0-9-_]+\.)+[a-z0-9-_]+/.+`)
```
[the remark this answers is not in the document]
benhoyt: Fair enough. I've removed underscore from allowed domain chars (it's extremely rare and often disallowed anyway), and I tightened up the allowed chars in the key part -- though not quite as tight as you suggested. I've made it `[A-Za-z0-9./-]+` so that things like `/Foo-Bar/index.html` are allowed.

Oh, and I noticed I was missing the `^` and `$` (start and end) anchors! So I added those.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,154 @@
...
+		Key:   key,
+		After: after,
+	}
+	var notices []*state.Notice
+
+	st := c.d.overlord.State()
+	st.Lock()
+	defer st.Unlock()
+
+	timeoutStr := query.Get("timeout")
```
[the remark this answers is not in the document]
benhoyt: I can see what you're saying, and in the `/v1/changes/{id}/wait` endpoint we use "no timeout" to mean wait indefinitely (or till the request is cancelled). However, there the "wait" is explicit in the endpoint name. Besides, timeout working this way is what we agreed on in the spec, so I think I'll keep it as is, but leave this comment thread open for Gustavo's input.

internals/daemon/api_notices.go
benhoyt: Yeah, on reflection, I agree. I don't think the timeout expiring should be an error. The intention is that you use this as a long-polling endpoint, so it's a normal condition when the timeout elapses. I've just made it return an empty list of notices in that case.

internals/daemon/api_notices.go
benhoyt: No, this can happen if the client cancels the HTTP request (though whether the client would actually get the HTTP 400 I'm not sure...). In any case, there's a test for this.

internals/overlord/state/notices.go
```diff
@@ -405,7 +405,7 @@ func (s *State) WaitNotices(ctx context.Context, filter *NoticeFilter) ([]*Notic
 
 // Remove this and just use context.AfterFunc once we're on Go 1.21.
 func contextAfterFunc(ctx context.Context, f func()) func() {
-	stopCh := make(chan struct{}, 1)
+	stopCh := make(chan struct{})
```
benhoyt, line 408: These changes per code review comment [here](https://github.com/canonical/pebble/pull/295#discussion_r1348576840).

internals/daemon/api_notices.go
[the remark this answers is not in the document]
niemeyer: > I expect to see a timeout option from a command for which the default behaviour is to wait until the job is done (timeout=infinite).

A good implementation of this logic needs to deal with disconnects regardless, so it doesn't feel like the change here would help that logic significantly. Indeed, it seems better to implement the client logic pinging regularly than hanging on a socket that remains silent for a few years. :)

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+	if timeoutStr != "" {
+		// Wait up to timeout for notices matching given filter to occur
+		timeout, err := time.ParseDuration(timeoutStr)
+		if err != nil {
+			return statusBadRequest("invalid timeout %q: %v", timeoutStr, err)
```
niemeyer, line 79: The errors coming out of ParseDuration are usually nice, so we can do just:
```
("invalid timeout: %v", err)
```
Otherwise the string will show multiple times.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+		// an empty list of notices, not an error.
+		if err != nil && !errors.Is(err, context.DeadlineExceeded) {
+			return statusInternalError("cannot wait for notices: %s", err)
+		}
+	} else {
+		// No timeout given, fetch currently-available notices
+		notices = st.Notices(filter)
+	}
+
+	if len(notices) == 0 {
```
niemeyer, line 99: This is one of the rare cases where `== nil` is really what we want.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+		notices = []*state.Notice{} // avoid null result
+	}
+	return SyncResponse(notices)
+}
+
+func v1PostNotices(c *Command, r *http.Request, _ *UserState) Response {
```
niemeyer, line 105: We should also talk about taking user into account here too.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+	if payload.RepeatAfter != "" {
+		var err error
+		repeatAfter, err = time.ParseDuration(payload.RepeatAfter)
+		if err != nil {
+			return statusBadRequest("invalid repeat-after duration %q: %v", payload.RepeatAfter, err)
+		}
+	}
+
+	dataSize := 0
+	for k, v := range payload.Data {
```
niemeyer, line 141: If we unmarshal Data separately into a json.RawMessage, we can look into it before decoding.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+	noticeId, err := st.AddNotice(state.CustomNotice, payload.Key, &state.AddNoticeOptions{
+		Data:        payload.Data,
+		RepeatAfter: repeatAfter,
+	})
+	if err != nil {
+		return statusInternalError("%v", err)
+	}
+
+	result := struct {
+		ID string `json:"id"`
```
niemeyer, line 161: This is an API contract, so even if small, it probably deserves a proper type so future generations respect it a bit more. Is there a reason for this to not return the real `&Notice{ID: id}`?

niemeyer (changes requested):
Thanks, Ben. This looks very nice. A few minor comments, and the only issue we probably need to talk about is the user constraints.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+	"regexp"
+	"time"
+
+	"github.com/canonical/x-go/strutil"
+
+	"github.com/canonical/pebble/internals/overlord/state"
+)
+
+// A very loose regex to ensure custom keys are in the form "domain.com/key"
+var customKeyRegexp = regexp.MustCompile(`^([a-z0-9-]+\.)+[a-z0-9-]+/[A-Za-z0-9./-]+$`)
```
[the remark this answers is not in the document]
benhoyt: Fair call -- done.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+
+func v1GetNotices(c *Command, r *http.Request, _ *UserState) Response {
+	query := r.URL.Query()
+
+	typeStrs := strutil.MultiCommaSeparatedList(query["types"])
+	types := make([]state.NoticeType, 0, len(typeStrs))
+	for _, typeStr := range typeStrs {
+		noticeType := state.NoticeType(typeStr)
+		if !noticeType.Valid() {
+			return statusBadRequest("invalid notice type %q", typeStr)
```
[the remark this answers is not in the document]
benhoyt: Fair call -- done (with comment as to why).

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,178 @@
...
+		types = append(types, noticeType)
+	}
+
+	keys := strutil.MultiCommaSeparatedList(query["keys"])
+
+	afterStr := query.Get("after")
+	var after time.Time
+	if afterStr != "" {
+		var err error
+		after, err = time.Parse(time.RFC3339, afterStr)
```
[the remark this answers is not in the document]
benhoyt: I think it's borderline, as we're basically saving an if statement, but there were a couple of other similar fields, so I've added `parseOptionalTime` and `parseOptionalDuration` and used them where applicable.

internals/daemon/api_notices.go
benhoyt: Good call -- updated.

internals/daemon/api_notices.go
benhoyt: I don't really understand why that's better: I get that it means we could compare the data size directly (though that would include JSON punctuation and whitespace too), but then we have to unmarshal again after that to get the map.

internals/daemon/api_notices.go
benhoyt: I considered that, but the only field that's really interesting here is the ID. If we use `&Notice{ID: id}` directly, our `MarshalJSON` would include zero values for the `type`, `key`, `occurrences` and other fields -- which are normally required, so it seems problematic to make them "omitempty".

We could update `State.AddNotice` to return a `*Notice` instead of a `noticeID string`, and then this API would return a full `*Notice`. I guess that would have the nice property that the caller would get the exact times for `first-occurred` and so on, which might be useful?

internals/daemon/api_notices.go
benhoyt: Per voice discussion, we'll use a RawMessage per suggestion above and compare the size of that. It's only going to get smaller if anything (whitespace removed etc).

internals/daemon/api_notices.go
benhoyt: Per voice discussion, we'll add new daemon-local type `addedNotice` or similar with just the ID field.

internals/daemon/api.go
```diff
@@ -94,6 +94,15 @@ var API = []*Command{{
 	Path:   "/v1/checks",
 	UserOK: true,
 	GET:    v1GetChecks,
+}, {
+	Path:   "/v1/notices",
+	UserOK: true,
```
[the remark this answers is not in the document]
benhoyt: Per voice discussion, use use `UserOK` on all existing endpoints, so we'll leave as is for now. We can later merge in what snapd agrees on if appropriate.

internals/daemon/api_notices.go
benhoyt: See above.

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,169 @@
...
+	"regexp"
+
+	"github.com/canonical/x-go/strutil"
+
+	"github.com/canonical/pebble/internals/overlord/state"
+)
+
+// Ensure custom keys are in the form "domain.com/key" (but somewhat more restrictive).
+var customKeyRegexp = regexp.MustCompile(
+	`^[a-z0-9]+(-[a-z0-9]+)*(\.[a-z0-9]+(-[a-z0-9]+)*)+(/[a-z0-9]+(-[a-z0-9]+)*)+$`)
```
flotter, line 31: Just for my own knowledge, what is the strategy you take on line length (when to break it) ?

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,169 @@
...
+	query := r.URL.Query()
+
+	typeStrs := strutil.MultiCommaSeparatedList(query["types"])
+	types := make([]state.NoticeType, 0, len(typeStrs))
+	for _, typeStr := range typeStrs {
+		noticeType := state.NoticeType(typeStr)
+		if !noticeType.Valid() {
+			// Ignore invalid notice types (so requests from newer clients
+			// with unknown types succeed).
+			continue
```
flotter, line 52: I assume this has been considered, but just checking: from my understanding ignoring future supported notice types will also ignore invalid notice strings from the CLI. I did a quick check in the client proposal to confirm we do not perform a validation step on user input.

It feels to me that if I type something that is wrong, it would be better if the system tells me so, rather than not receiving anything for that invalid notice type.

Having said this, I have seen the same behaviour with logs. If I type the wrong service string, I just do not get any logs, no error.

Any thoughts ?

internals/daemon/api_notices.go
```diff
@@ -0,0 +1,169 @@
...
+		notices = st.Notices(filter)
+	}
+
+	if notices == nil {
+		notices = []*state.Notice{} // avoid null result
+	}
+	return SyncResponse(notices)
+}
+
+func v1PostNotices(c *Command, r *http.Request, _ *UserState) Response {
```
flotter, line 104: It looks to me that we can only ever produce one notice with each post. ```v1PostNotices``` seems to suggest otherwise?

flotter (approved):
Really nice. Always so much to learn from reviewing your code :)

internals/daemon/api_notices.go
benhoyt: I usually wrap comment text at 80 chars, but for code it's fairly informal! "Wrap at around 100" is probably my strategy, which is about what fits on my screen (with two code windows side by side on a widescreen monitor). In this case it was about 120 chars unwrapped, and I couldn't see the end, so I wrapped it. :-)

internals/daemon/api_notices.go
benhoyt: I had this originally rejected unknown types, but Gustavo said this in a [comment above](https://github.com/canonical/pebble/pull/296#discussion_r1350471123):

> This will make interacting with APIs in a backwards compatible manner harder, because someone coming from the future cannot do a selection on types that this system doesn't yet know about. It seems better to ignore such invalid types, after all we're respecting the contract and returning only the types matching the given filter. If a type is invalid, we'll have none of these.

I tried to capture that in my comment (probably not quite clear enough though).

And I think *not* rejecting them in the CLI is probably right, because then you can use an older client/CLI to talk to a newer server that supports additional notice types without it breaking.

internals/daemon/api_notices.go
benhoyt: I don't actually love these names either, but I've gone with convention. It looks like we normally use the plural form on the `/v1/things` endpoints (eg: `v1PostServices`), and the singular on endpoints like `/v1/changes/{id}` which deal with a single change (eg: `v1PostChange`).

flotter: OK. Good to know. I am all for conformance, thanks.

internals/daemon/api_notices.go
flotter: Yes, the argument does make sense. I guess its just a trade-off, and in general what I take from this is that enabling people that knows what they are doing in the most flexible is better than building in limitations to help people that is using the tool wrong. The latter case can be documented, while not having the former case has no workaround in the field. I think I like this, and I've learned an important consideration here.

internals/daemon/api.go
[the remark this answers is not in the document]
flotter: Just a side comment on this: I think the user story also comes into play if we allow the client to connect over TCP sockets (remote connections)? Project Crystal client would really benefit from remote client connections (I think we will schedule this in the next cycle), and if I recall my previous hack, I had to comment out some user permission checks. Just a heads-up that if this topic is touched, I would be happy to participate as a stakeholder.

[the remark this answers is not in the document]
niemeyer: In yesterday's call with snapd, we agree that the implementation with user support will be done by the snapd team, but first on Pebble, so that both of you can also easily participate in the process.

internals/daemon/api_changes.go
```diff
@@ -193,14 +193,13 @@ func v1GetChangeWait(c *Command, r *http.Request, _ *UserState) Response {
 		return statusNotFound("cannot find change with id %q", changeID)
 	}
 
-	timeoutStr := r.URL.Query().Get("timeout")
-	if timeoutStr != "" {
+	timeout, err := parseOptionalDuration(r.URL.Query().Get("timeout"))
```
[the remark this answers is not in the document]
niemeyer: Thanks :+1:

PR #297, opened 2023-09-05
feat(client): add Go client for notices

benhoyt:
This adds the Go client for Pebble Notices (see spec [JU048](https://docs.google.com/document/d/16PJ85fefalQd7JbWSxkRWn0Ye-Hs8S1yE99eW7pk8fA/edit)). It builds on top of the notices API in #296 (until that PR is merged, see [just the client diff here](https://github.com/benhoyt/pebble/compare/notices-api...benhoyt:pebble:notices-cli), or look at just the `client/*.go` changes).

Here is a summary of the new `Client` methods:

```go
Notify(opts *NotifyOptions) (string, error)
Notice(id string) (*Notice, error)
Notices(opts *NoticesOptions) ([]*Notice, error)
WaitNotices(ctx context.Context, opts *NoticesOptions, timeout time.Duration) ([]*Notice, error)
```


client/notices.go
```diff
@@ -0,0 +1,189 @@
...
+	Type NoticeType
+	// Key, if set, includes only notices with this key.
+	Key string
+	// After, if set, includes only notices that were last repeated after this time.
+	After time.Time
+}
+
+// Notice is a notification whose identity is the combination of Type and Key
+// that has occurred Occurrences number of times.
+type Notice struct {
```
barrettj12, line 81: is this not exactly the same struct as in the state code? If so, why duplicate it here? Can't we just reuse the same struct? (Maybe put it in a separate package if we don't want `client` to depend on `state`).

Ditto for `NoticeType` and `jsonNotice`.

client/notices.go
```diff
@@ -0,0 +1,189 @@
...
+type jsonNotice struct {
+	Notice
+	RepeatAfter string `json:"repeat-after,omitempty"`
+	ExpireAfter string `json:"expire-after,omitempty"`
+}
+
+// Notice fetches a single notice by ID.
+func (client *Client) Notice(id string) (*Notice, error) {
+	var jn *jsonNotice
+	_, err := client.doSync("GET", "/v1/notices/"+id, nil, nil, nil, &jn)
```
barrettj12, line 120: probably out of scope for this PR, but I think this is a great example where an args struct would be appropriate. That way, we can name the arguments (what are all these nils?), and not have to explicitly pass in nil values

client/notices.go
```diff
@@ -0,0 +1,189 @@
...
+	query.Set("timeout", timeout.String())
+
+	res, err := client.raw(ctx, "GET", "/v1/notices", query, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusGatewayTimeout {
+		return nil, nil
```
barrettj12, line 149: shouldn't we surface this error?

client/notices.go
benhoyt: We definitely don't want `client` to depend on `state`. We could put it in a common package as you say, but I think it's cleaner to just duplicate it -- it keeps the separation crystal clear, and it allows the state marshalling and API to evolve at a different rate than the client. For example, we might add internal notice types that we don't want to expose to the client, or new `Notice` fields that are only internally-relevant.

In this specific case, they're not quite the same ... if we did have a common package, it would have to expose the `JsonNotice` type for unmarshalling, but the real `Notice` type is actually different between the two (the `state` version intentionally doesn't export any fields, the `client` version intentionally exports all fields). We could probably still share some stuff, but it quickly gets messy and you have to find common APIs that work.

And once again, the precedent here is to not share code between server and client. (If the client were in a separate repo, we almost certainly wouldn't share it.)

I've often found the Go proverb to be true: "A little copying is better than a little dependency".

client/notices.go
benhoyt: Yeah, agreed: both that this is a good idea, and that it's out of scope for this PR. It's not really causing pain, but I'd happily approve a PR that made this change.

Incidentally, this was proposed as part of [the PR to expose client.DoSync etc](https://github.com/canonical/pebble/pull/289/files), but we decided not to go ahead with that one as it was considered an implementation detail of the client. But I still think it'd be fine to make the internal `doX` funcs take args structs.

client/notices.go
benhoyt: This is intentional, though I've now clarified that in the doc comment. I think it makes sense for this to not be an error, because this is intended as a "long-polling" style API, and the timeout you asked for just elapsed (not an error condition). It makes it easy for callers who are long polling to just check `len(notices) == 0` and try again.

client/notices.go
```diff
@@ -0,0 +1,192 @@
...
+		ID string `json:"id"`
+	}{}
+	_, err := client.doSync("POST", "/v1/notices", nil, nil, &body, &result)
+	if err != nil {
+		return "", err
+	}
+	return result.ID, err
+}
+
+type NoticesOptions struct {
```
flotter, line 70: This was called ```NoticeFilters``` in the state implementation. I think the reasoning is that in the client context, these are options to the client request itself. But for me as an outside person, having filter in the name is more descriptive, and having it the same name as the state implementation feels consistent.

client/notices.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Warnings are a subset of notices where the key is a human-readable
+	// warning message.
+	NoticeWarning NoticeType = "warning"
+)
+
+type jsonNotice struct {
+	Notice
+	RepeatAfter string `json:"repeat-after,omitempty"`
+	ExpireAfter string `json:"expire-after,omitempty"`
```
flotter, line 114: This is quite an interesting trick for me. So I take it that the unmarshalling will type match as well as matching the name, and therefore it will select your extended struct member?

client/notices.go
flotter: Ok interesting I see. 

So this is the reason why you had to split the ```doSync```? 

I made a comment on this topic here in the [API PR](https://github.com/canonical/pebble/pull/296/files#r1324706774). I completely second the fact that this should not error. My question would just be what is the point of sending back an error in the first place, as apposed to an empty notices list (like for this case where ```--timeout``` is not supplied).

flotter (commented):
Looks great. A couple of questions for the purpose of learning from this :)

client/notices.go
benhoyt: No, the reason I needed to split `doSync` was to use `raw()` in `WaitNotices` so I could pass the context through. I've added a comment to that effect.

client/client.go
```diff
@@ -338,6 +338,10 @@ func (client *Client) doSync(method, path string, query url.Values, headers map[
 	if err := rsp.err(client); err != nil {
 		return nil, err
 	}
+	return client.finishSync(rsp, v)
+}
+
+func (client *Client) finishSync(rsp response, v interface{}) (*ResultInfo, error) {
```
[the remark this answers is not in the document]
benhoyt: No, not readability -- the reason I needed to split `doSync` was to use `raw()` in `WaitNotices` so I could pass the context through. I've added a comment to that effect. Also, it doesn't apply to `doAsync` as the "finishing" code for that is quite different.

client/notices.go
```diff
@@ -0,0 +1,192 @@
...
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+type NotifyOptions struct {
```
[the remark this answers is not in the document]
benhoyt: Perhaps in the abstract, but *all* of the other `Client` methods take `FooOptions` struct, so I don't think this one should diverge from that.

client/notices.go
benhoyt: Per above comment, all the other `Client` methods take a `FooOptions`, so I don't think these should diverge. I agree it'd be nice to match the name in state, but users of client shouldn't need to know about state so I think that's a non-issue. I think we should go for consistency with the parameter name of the other `Client` methods here.

client/notices.go
benhoyt: Yeah, this approach was new to me as well (I stole it from `client/warnings.go`). I don't think it's about the types, it's because the fields in the struct itself (`jsonNotice`) take precedence over / override the same-named fields in the embedded struct.

client/notices.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Keys, if not empty, includes only notices whose key is one of these.
+	Keys []string
+
+	// After, if set, includes only notices that were last repeated after this time.
+	After time.Time
+}
+
+// Notice is a notification whose identity is the combination of Type and Key
+// that has occurred Occurrences number of times.
```
niemeyer, line 82: Perhaps something along the lines of:
```
// Notice holds details of an event that was observed and reported either inside
// the server itself or externally via the API. Besides the ID field itself, the Type
// and Key fields together also uniquely identify a particular notice, and when a new
// notification is made with matching Type and Key, the previous notice is updated
// appropriately instead of a new one being created.
```
This intro should hopefully make some of these fields more obvious.

client/notices.go
```diff
@@ -0,0 +1,200 @@
...
+	ChangeUpdateNotice NoticeType = "change-update"
+
+	// A custom notice reported via the Pebble client API or "pebble notify".
+	// The key and data fields are provided by the user. The key must be in
+	// the format "mydomain.io/mykey" to ensure well-namespaced notice keys.
+	CustomNotice NoticeType = "custom"
+
+	// Warnings are a subset of notices where the key is a human-readable
+	// warning message.
+	WarningNotice NoticeType = "warning"
```
niemeyer, line 110: Seems a bit premature considering we have no supporting code for this, and have no timeframing for supporting it either. Same for ChangeUpdateNotice above: I think we should add the types together with the supporting code.

client/notices.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// If the timeout elapses before any matching notices arrive, it's not
+// considered an error: WaitNotices returns a nil slice and a nil error.
+func (client *Client) WaitNotices(ctx context.Context, serverTimeout time.Duration, opts *NoticesOptions) ([]*Notice, error) {
+	query := makeNoticesQuery(opts)
+	query.Set("timeout", serverTimeout.String())
+
+	// We need to use client.raw() here to pass the context through (and we
+	// don't want retries).
+	res, err := client.raw(ctx, "GET", "/v1/notices", query, nil, nil)
```
niemeyer, line 150: Related to earlier comment, wE should be able to use the new API here instead of splitting doSync in two.

client/notices.go
```diff
@@ -0,0 +1,200 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/url"
+	"time"
+)
+
+type NotifyOptions struct {
+	// Key is the custom notice's key. Must be in "domain.com/key" format.
+	Key string
```
niemeyer, line 27: This is hiding the fact we have a Type here, and making it an internal option, but then when the notice comes back we suddenly have the Type detail available.  I think we should make these details known and used here, even if for now only a CustomNotice type can be emitted.

client/notices.go
niemeyer: We already have that struct now, after the Requester PR has landed.

client/notices.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+type jsonNotice struct {
+	Notice
+	RepeatAfter string `json:"repeat-after,omitempty"`
+	ExpireAfter string `json:"expire-after,omitempty"`
+}
+
+// Notice fetches a single notice by ID.
+func (client *Client) Notice(id string) (*Notice, error) {
```
niemeyer, line 120: Commenting here to avoid mixing with the unrelated thread below.

Taking a raw string from the outside world and blindly appending it to the GET path with zero validation is anxiety inducing. This is the kind of thing that can be exploited in an application we have no visibility over and become an actual hack.

niemeyer (changes requested):
Thanks for the PR. Looks good, with just high-level tuning.

client/notices.go
benhoyt: Yeah, fair enough -- I've added `Type` (the server ensures it's set and equal to "custom").

client/notices.go
benhoyt: I can see what you're saying, but I'm not so sure. Go will safely and properly percent-escape any disallowed characters, and the server will unescape (and will not find the notice).

However, I've added a simple client-side validation -- that's slightly more permissive than the integers we produce now for future-proofing -- see what you think.

If you think this is a good approach, I can create a PR to do something similar in [`client/changes.go`](https://github.com/canonical/pebble/blob/90805472b01d879e8c8e17c306e7f51849357cf4/client/changes.go#L92), where we do the same thing.

client/client.go
```diff
@@ -338,6 +338,10 @@ func (client *Client) doSync(method, path string, query url.Values, headers map[
 	if err := rsp.err(client); err != nil {
 		return nil, err
 	}
+	return client.finishSync(rsp, v)
```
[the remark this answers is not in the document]
benhoyt: Yep, this came before that was merged. Merged master and updated this. It's nice!

client/notices.go
```diff
@@ -0,0 +1,200 @@
...
+
+type NotifyOptions struct {
+	// Key is the custom notice's key. Must be in "domain.com/key" format.
+	Key string
+
+	// Data are optional key=value pairs for this occurrence of the notice.
+	Data map[string]string
+
+	// RepeatAfter allows the notice to repeat after this duration. If this is
+	// zero (the default), the notice will always repeat.
```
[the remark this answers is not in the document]
benhoyt: Done (with a slight tweak to add "if not zero").

client/notices.go
benhoyt: Done.

client/notices.go
benhoyt: Yeah, fair enough -- removed.

client/notices.go
benhoyt: I've opened https://github.com/canonical/pebble/issues/318 to track that.

PR #302, opened 2023-09-11
refactor(cli): pass client config to cli.Run()

anpep:
This patch modifies the signature of the `cli.Run()` function so that the Pebble client config is passed to it in order to build a Pebble client for the CLI commands.  This adds the flexibility needed for external applications to override the configuration of the Pebble client.

### DIFF
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -243,9 +243,6 @@ var (
 	osExit      = os.Exit
 )
 
-// ClientConfig is the configuration of the Client used by all commands.
-var clientConfig client.Config
-
 // exitStatus can be used in panic(&exitStatus{code}) to cause Pebble's main
 // function to exit with a given exit code, for the rare cases when you want
 // to return an exit code other than 0 or 1, or when an error return is not
@@ -258,7 +255,7 @@ func (e *exitStatus) Error() string {
 	return fmt.Sprintf("internal error: exitStatus{%d} being handled as normal error", e.code)
 }
 
-func Run() error {
+func Run(config *client.Config) error {
 	defer func() {
 		if v := recover(); v != nil {
 			if e, ok := v.(*exitStatus); ok {
@@ -270,9 +267,7 @@ func Run() error {
 
 	logger.SetLogger(logger.New(os.Stderr, "[pebble] "))
 
-	_, clientConfig.Socket = getEnvPaths()
-
-	cli, err := client.New(&clientConfig)
+	cli, err := client.New(config)
 	if err != nil {
 		return fmt.Errorf("cannot create client: %v", err)
 	}
@@ -353,7 +348,7 @@ func errorToMessage(e error) (normalMessage string, err error) {
 	return msg, nil
 }
 
-func getEnvPaths() (pebbleDir string, socketPath string) {
+func GetEnvPaths() (pebbleDir string, socketPath string) {
 	pebbleDir = os.Getenv("PEBBLE")
 	if pebbleDir == "" {
 		pebbleDir = defaultPebbleDir
--- a/internals/cli/cli_test.go
+++ b/internals/cli/cli_test.go
@@ -136,7 +136,7 @@ func (s *PebbleSuite) TestErrorResult(c *C) {
 	restore := fakeArgs("pebble", "warnings")
 	defer restore()
 
-	err := cli.RunMain()
+	err := cli.Run(cli.ClientConfig)
 	c.Assert(err, ErrorMatches, `cannot do something`)
 }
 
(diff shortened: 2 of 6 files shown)


paul-rodriguez (approved):
Makes sense to me.

internals/cli/cli.go
```diff
@@ -243,9 +243,6 @@ var (
 	osExit      = os.Exit
 )
 
-// ClientConfig is the configuration of the Client used by all commands.
-var clientConfig client.Config
```
benhoyt, line 247: It's kinda weird that this was a global anyway, rather than local to `Run()`.

internals/cli/cli.go
```diff
@@ -258,7 +255,7 @@ func (e *exitStatus) Error() string {
 	return fmt.Sprintf("internal error: exitStatus{%d} being handled as normal error", e.code)
 }
 
-func Run() error {
+func Run(config *client.Config) error {
```
benhoyt, line 258: Maybe name this `clientConfig` to be clear (eg: that it's not daemon or other config).

internals/cli/cli.go
```diff
@@ -353,7 +348,7 @@ func errorToMessage(e error) (normalMessage string, err error) {
 	return msg, nil
 }
 
-func getEnvPaths() (pebbleDir string, socketPath string) {
+func GetEnvPaths() (pebbleDir string, socketPath string) {
```
benhoyt, line 351: If we do add `client.NewConfigFromEnv`, this probably doesn't need to be exported anymore.

internals/cli/cli_test.go
```diff
@@ -136,7 +136,7 @@ func (s *PebbleSuite) TestErrorResult(c *C) {
 	restore := fakeArgs("pebble", "warnings")
 	defer restore()
 
-	err := cli.RunMain()
+	err := cli.Run(cli.ClientConfig)
```
benhoyt, line 139: Why not just `cli.Run(&client.Config{})` for this (and below)?

internals/cli/export_test.go
```diff
@@ -20,9 +20,7 @@ import (
 	"github.com/canonical/pebble/client"
 )
 
-var RunMain = Run
-
-var ClientConfig = &clientConfig
+var ClientConfig *client.Config = &client.Config{}
```
benhoyt, line 23: Per above, I'm not sure why we'd need to export this, instead of using `&client.Config{}` directly.

internals/cli/cli.go
anpep: We would still need this to be exported so we can `GetEnvPaths()` from the `client` package though, am I wrong?

benhoyt: Ah, true, I wasn't paying close enough attention that `GetEnvPaths` is in `cli` but `NewConfigFromEnv` is in `client`.

Actually, maybe `GetEnvPaths` should move to the `client` package? It's (sort of) a client concern, or half of it is. :-) Or maybe we should have `client.SocketPath()` and then just have an internal `cli.pebbleDir()` for the `run` command. Though I realize that will mean a tiny bit of duplication of the `PEBBLE` logic -- but it may be conceptually cleanest.

Oh, either way we should probably drop the `Get` prefix.

internals/cli/cli_test.go
anpep: That's fair (and actually the way I implemented it at first--only to realize there's no trivial way to get rid of the `cli.ClientConfig` global, so I used it anyway).

internals/cli/export_test.go
anpep: Because at one point some test expect this to be initialized after a call to `cli.Run()` from within the tests, and those fail without a valid client config initialized with an open Pebble socket.

internals/cli/cli.go
niemeyer: This is clearly an internal helper function with API not intended to be exposed. Let's please not publish it.

Also, as a more general note: I realize that most of the work on going in your team is about exposing functionality, but it feels like too often the reviews are heavy on "Please don't expose things that were built privately without considering how to make them public.". Maybe some of these could be anticipated by internal reviews inside the team itself.

niemeyer (changes requested):
Maybe this makes sense given some external context, but without it the underlying motivation for this change is not clear,. It's changing the Run method so it takes a parameter from the configuration, except that parameter is currently obtained from an environment variable, which is already a very public API and won't change any time soon. So if the goal is to be able to change that variable in tests, can't we simply use the environment variable itself? But maybe that's not the goal?

flotter:
@anpep I believe this PR will be superseded by https://github.com/canonical/pebble/pull/310 so I will close this for now to reduce the PR backlog on the Pebble repo. We can always reopen this if my assumptions were not correct.

PR #303, opened 2023-09-11
feat(daemon): add ErrorResponse to allow external error creation

flotter:
https://github.com/canonical/pebble/pull/265 provides the ability for the daemon REST API to be extended by appending new handlers to the API list.

For example, the following new reboot command can be added now:

```go
// requestReboot asks the daemon to gracefully shut down the system
// and issue a reboot.
func requestReboot(d *daemon.Daemon) daemon.Response {
        d.HandleRestart(restart.RestartSystem)
        result := deviceResult{}
        return daemon.SyncResponse(result)
}
```

Note that ```SyncResponse``` was already public, which allows the code above to work.

Following this patch, we can now reply with a suitable error response:

```go
:
    switch(...) {
    case reboot:
        requestReboot(...)
    default:    
        return daemon.ErrorResponse(http.StatusBadRequest,"invalid media type %q", mediaType)
    {
:
```


benhoyt (approved):
This seems reasonable to me, especially given that `SyncResponse` and `AsyncResponse` are already exported.

paul-rodriguez (approved):
This is very straightforward

internals/daemon/response.go
```diff
@@ -114,29 +114,29 @@ func (r *resp) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
-type errorValue interface{}
+type ErrorValue interface{}
 
-type errorResult struct {
+type ErrorResult struct {
 	Message string     `json:"message"` // note no omitempty
-	Kind    errorKind  `json:"kind,omitempty"`
-	Value   errorValue `json:"value,omitempty"`
+	Kind    ErrorKind  `json:"kind,omitempty"`
+	Value   ErrorValue `json:"value,omitempty"`
```
niemeyer, line 134: Similar here. This is already public in the client package already, and might be imported from there both here in and in KernOS, depending on whether the client package is already easily importable here.

niemeyer (changes requested):
Some remarks below. Please let me know how it sounds.

internals/daemon/response.go
```diff
@@ -114,29 +114,29 @@ func (r *resp) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
-	errorKindNotFound          = errorKind("not-found")
-	errorKindPermissionDenied  = errorKind("permission-denied")
-	errorKindGenericFileError  = errorKind("generic-file-error")
+	ErrorKindLoginRequired     = ErrorKind("login-required")
+	ErrorKindDaemonRestart     = ErrorKind("daemon-restart")
+	ErrorKindSystemRestart     = ErrorKind("system-restart")
+	ErrorKindNoDefaultServices = ErrorKind("no-default-services")
+	ErrorKindNotFound          = ErrorKind("not-found")
+	ErrorKindPermissionDenied  = ErrorKind("permission-denied")
+	ErrorKindGenericFileError  = ErrorKind("generic-file-error")
```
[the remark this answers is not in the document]
flotter: I've studied the snapd code and that is exactly what was done there. Taking inspiration from the snapd client:

I created a an ```errors.go``` in the client and moved the ```Error``` definition there (from ```client.go```), including placing all the Kinds in there as well. All in once place.

internals/daemon/response.go
```diff
@@ -114,29 +114,29 @@ func (r *resp) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
 )
 
-type errorValue interface{}
+type ErrorValue interface{}
```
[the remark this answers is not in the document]
flotter: Fixed.

internals/daemon/response.go
flotter: Tried this in the next proposal. Closing for further new comments to follow.

internals/daemon/response.go
```diff
@@ -194,11 +194,11 @@ type errorResponder func(string, ...interface{}) Response
...
-	statusMethodNotAllowed = makeErrorResponder(405)
-	statusInternalError    = makeErrorResponder(500)
-	statusGatewayTimeout   = makeErrorResponder(504)
+	StatusBadRequest       = makeErrorResponder(400)
+	StatusUnauthorized     = makeErrorResponder(401)
+	StatusForbidden        = makeErrorResponder(403)
+	StatusNotFound         = makeErrorResponder(404)
+	StatusMethodNotAllowed = makeErrorResponder(405)
+	StatusInternalError    = makeErrorResponder(500)
+	StatusGatewayTimeout   = makeErrorResponder(504)
```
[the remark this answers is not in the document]
flotter: done.

client/client.go
```diff
@@ -403,26 +403,6 @@ type response struct {
 	Maintenance *Error `json:"maintenance"`
 }
 
-// Error is the real value of response.Result when an error occurs.
-type Error struct {
-	Kind    string      `json:"kind"`
-	Value   interface{} `json:"value"`
-	Message string      `json:"message"`
-
-	StatusCode int
```
flotter, line 412: I have looked at snapd and pebble, and I cannot actually find this being used. I've built snapd + unit tests, and pebble + unit tests without this line, and all works. I propose we drop this until a real need arises. The Status code and test version is available in the ```response struct```, of which this ```Error``` is unmarshalled from ```response.Result```.

internals/daemon/response.go
```diff
@@ -169,22 +150,26 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, string(f))
 }
 
+func Reponsef(status int, format string, v ...interface{}) Response {
```
flotter, line 153: As per the proposal, we can now use this from a derivative project using ```http.<error>``` values.

client/errors.go
```diff
@@ -0,0 +1,43 @@
...
+func (e *Error) Error() string {
+	return e.Message
+}
+
+type ErrorKind string
+
+// Error kinds for use as a response result.
+const (
+	ErrorKindLoginRequired     = ErrorKind("login-required")
+	ErrorKindNoDefaultServices = ErrorKind("no-default-services")
```
paul-rodriguez, line 33: This name doesn't really tell us what's wrong. Was there an expectation of default services existing?

client/errors.go
```diff
@@ -0,0 +1,43 @@
...
+	ErrorKindNotFound          = ErrorKind("not-found")
+	ErrorKindPermissionDenied  = ErrorKind("permission-denied")
+	ErrorKindGenericFileError  = ErrorKind("generic-file-error")
```
paul-rodriguez, line 36: This might as well be `ErrorKindFileError`

and also

8020mi

internals/daemon/response.go
```diff
@@ -169,22 +150,26 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	res := &client.Error{}
+	if len(v) == 0 {
+		res.Message = format
+	} else {
+		res.Message = fmt.Sprintf(format, v...)
+	}
+	if status == 401 {
```
paul-rodriguez, line 160:
```suggestion
	if status == http.StatusUnauthorized {
```

internals/daemon/response.go
```diff
@@ -169,22 +150,26 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, string(f))
 }
 
+func Reponsef(status int, format string, v ...interface{}) Response {
+	res := &client.Error{}
+	if len(v) == 0 {
+		res.Message = format
+	} else {
+		res.Message = fmt.Sprintf(format, v...)
+	}
```
paul-rodriguez, line 159:
```suggestion
	res.Message = fmt.Sprintf(format, v...)
```

client/client.go
niemeyer: Yeah, seems okay to remove, but I'd argue that it's for a different reason: the key reason why supposedly it's helpful to have this code here is because it would be nice to know that the server is rejecting our request based on external factors (e.g. it's a proxy, and it doesn't want to talk to us at all).  Ironically, our code currently only sets the status code when it detects that the response unmarshals an error type properly, and by then we're already past the point of caring about the status code.  That's why it seems irrelevant to have the status code here, at least as the logic stands, and it also explains why no code is using it. So +1 on dropping it for now.

client/errors.go
niemeyer: Here "default services" is being used in the context of an operation. When we ask for an autostart, this will start whatever was supposed to start "by default", unless there's nothing to start by default. That's what the error is saying.

It's unfortunately too late to bikeshed about the error kind, because those are public constants exposed in the API, and real code depends on them.

client/errors.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (e *Error) Error() string {
+	return e.Message
+}
+
+type ErrorKind string
+
+// Error kinds for use as a response result.
+const (
+	ErrorKindLoginRequired     = ErrorKind("login-required")
```
niemeyer, line 32: Can we please drop the `Kind` on the variables?  It seems overly verbose for the public API (the type should still be named `ErrorKind`).

internals/daemon/response.go
niemeyer: > res.Message = fmt.Sprintf(format, v...)

No, there's an actual difference here. Try this:  `Responsef(404, "-%- funny external error -%-")`.

Please feel free to document the difference, so the next person doesn't feel tempted.

internals/daemon/response.go
niemeyer: +1

niemeyer (approved):
Nice refactoring, thanks. Only a few superficial details:

client/errors.go
anpep: While `not-found` and `permission-denied` are typically used in contexts involving files, I think those are generic enough that can be used to indicate e.g. the non-existence of a service, so these shouldn't really be in a "file management errors" bag.

Also, there's a lot of stuff that can go wrong when dealing with files (e.g. `EBUSY`, `EXDEV`) for which we don't have suitable error kinds.

internals/daemon/response.go
niemeyer: This is existing logic in existing code, right?

anpep (approved):
Thanks for your work on this, I think those changes are solid!

client/client.go
niemeyer: Actually, we may need to roll this back. It looks like the status code is being intentionally used already in some endpoints, and is part of tests too:

https://github.com/canonical/pebble/pull/299#discussion_r1327321448

niemeyer (changes requested):
One more detail about the StatusCode conversation above:

client/errors.go
flotter: As per Gustavo's clarification, we cannot change the error content at this stage.

internals/daemon/response.go
flotter: As per Gustavo's comment, I will document the detail here: We do have an -'f' type mechanism, but with a special exist clause that if no arguments are supplied, formatting is disabled. In this sense, I think its OK for now to proceed with the ```Responsef``` naming. I will document the details.

internals/daemon/response.go
flotter: Thanks @paul-rodriguez , resolved.

internals/daemon/response.go
paul-rodriguez: Just to note, my point was indeed that an `f` function should behave consistently and always format even if there's no format argument.

client/client.go
flotter: Hi Gustavo. I am not immediately sure how to best improve this:

[rsp.err](https://github.com/canonical/pebble/blob/cb90b548699363124ad7b009cecb2c3bfc301457/client/client.go#L426), as you know, currently discards the status code of any reponse type apart from type ```error```. This means the ```health``` [test](https://github.com/canonical/pebble/pull/299/files#r1328310832) code that sets the status code on reply for a ```sync``` type, will not result in an error. The status code currently is discarded.

flotter: In the client responder interface, I am offering the Status Code now as part of the common  return (as you have seen). Could this provide the plumbing for code to get access to the status code consistently, instead of through the error ?

```go
// DecoderFunc allows the client access to the HTTP response. See the SetDecoder
// description for more details.
type DecoderFunc func(ctx context.Context, res *http.Response, opts *RequestOptions, result interface{}) (*RequestResponse, error)

type Requester interface {
	// Do must support the following cases:
	//
	// 1. Sync response
	// 2. Async response
	// 3. Error response
	// 4. Websocket creation request
	// 5. Raw HTTP body request
	//
	// See the default implementation for information on what is expected in
	// each use case.
	Do(ctx context.Context, opts *RequestOptions, result interface{}) (*RequestResponse, error)

	// SetDecoder allows for client specific processing to be hooked into
	// the sync and async response decoding process. The decoder is also
	// responsible for unmarshalling the result, and populating the
	// RequestReponse.
	SetDecoder(decoder DecoderFunc)
}

// RequestOptions allows setting up a specific request.
type RequestOptions struct {
	Method  string
	Path    string
	Query   url.Values
	Headers map[string]string
	Body    io.Reader
	Async   bool
}

// RequestResponse defines a common response associated with requests.
type RequestResponse struct {
	StatusCode int
	Type       string
	Change     string
}
```

client/errors.go
flotter: Fixed.

client/client.go
flotter: I have a MM agreement with Gustavo that I will remove StatusCode from the Error type, but we will consistently expose it through the new interface planned in the next PR, as hinted above.

internals/daemon/response.go
flotter: Documented. Resolving.

client/client.go
benhoyt: Hmm, I don't like this removal of `StatusCode` at all -- this is an exported field of a public package, so I think we're playing fast and loose with backwards-compatibility. It *might* be safe to remove in this instance (we don't know), but as Pebble and the `pebble/client` package become more widely used, it's likely that someone is using that field.

I'm also not sure that @niemeyer's statement is true: "That's why it seems irrelevant to have the status code here, at least as the logic stands", because if there's an error you might want to drill down on the response code. For example, a package user could be calling `client.Change(id)`, and on error, type-asserting to `*client.Error` and checking whether the status code is 404 (change ID not found).

In this case the daemon returns an `Error.Kind` of `""`, so you can't test that. (This should probably be considered a bug -- I think the server should set `Kind` to `ErrorKindNotFound`, but it currently does not.)

Example (I've tested this locally and it correctly distinguishes between not-found and other server error codes correctly):

```go
	chg, err := cli.Change(id)
	if err != nil {
		clientErr, ok := err.(*client.Error)
		if ok && clientErr.StatusCode == 404 {
			// Handle not-found specially
			fmt.Printf("change ID %q not found!\n", chid)
		}
		return err
	}
```

By removing `Error.StatusCode`, this code would break. I agree it's somewhat unlikely this is being used (and clearly not in snapd or Pebble). But I don't think we can assume we don't have other `pebble/client` package users.

internals/daemon/response.go
```diff
@@ -169,22 +150,31 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, string(f))
 }
 
+// Reponsef behaves similar to the other printf style functions as long as additional
+// arguments beyond the format string are provided.
```
benhoyt, line 154: The note below already states the formatting quirk. So I suggest we change this comment to what the function actually does. Something like:

```suggestion
// Responsef builds an error Response that returns the status and formatted message.
```

internals/daemon/response.go
```diff
@@ -169,22 +150,31 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+//
+// Note: If no arguments are provided, formatting is disabled, and the format string
+// is used as is and not interpreted in any way.
+func Reponsef(status int, format string, v ...interface{}) Response {
```
benhoyt, line 158: Note there's a typo: `Reponsef` -> `Responsef`.

However, this doesn't just return a formatted response, it's an *error* response. I think we need to convey that. Maybe `Errorf` or `ErrorResponsef` (the shorter version is nice, but might be too confusing with `fmt.Errorf`).

Alternatively, we could keep it `Responsef` but make `status==http.StatusOK` do something sensible. Not sure if that's a good idea.

benhoyt (changes requested):
Sorry for the negativity here, but I really don't think breaking backwards-compatibility (for these minor adjustments) is worth it, and I also don't think we should import `client` from `daemon`. Detailed comments below.

client/client.go
flotter: @benhoyt @niemeyer I think I successfully mislead Gustavo in the claim that a Sync or ASync reponse can only be of type Sync or Async ... I see your point. The ```resp.err(...)``` happens before the check for type, and therefore allows such a call to return an error type! Glad you caught this.

client/errors.go
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2023 Canonical Ltd
```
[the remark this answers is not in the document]
flotter: OK no problem.

client/errors.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package client
+
+// Error is the real value of response.Result when an error type response occurs.
+type Error struct {
+	Message string      `json:"message"`
+	Kind    ErrorKind   `json:"kind,omitempty"`
```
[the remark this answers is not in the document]
flotter: I tried to combine the two (client and daemon), but I now see the light. My plan is to keep exported definition intact.

client/errors.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (e *Error) Error() string {
+	return e.Message
+}
+
+type ErrorKind string
+
+// Error kinds for use as a response result.
+const (
+	ErrorLoginRequired     = ErrorKind("login-required")
```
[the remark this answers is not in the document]
flotter: Yep mixed up the fact that the daemon errors were private, and therefore justified a rethink when made public, while the client errors were already public! doh!!

internals/cli/cli.go
```diff
@@ -329,17 +329,17 @@ func errorToMessage(e error) (normalMessage string, err error) {
 
 	var msg string
 	switch cerr.Kind {
-	case client.ErrorKindLoginRequired:
+	case client.ErrorLoginRequired:
```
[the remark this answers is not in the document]
flotter: Fixed.

internals/daemon/api_changes_test.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"gopkg.in/check.v1"
 
+	"github.com/canonical/pebble/client"
```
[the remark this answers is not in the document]
flotter: Understood, thanks. This PR was a great lesson.

internals/daemon/api_files.go
```diff
@@ -31,6 +31,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/canonical/pebble/client"
```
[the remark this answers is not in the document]
flotter: Done.

internals/daemon/response.go
```diff
@@ -114,26 +115,6 @@ func (r *resp) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
 	w.Write(bs)
 }
 
-type errorKind string
```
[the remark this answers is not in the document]
flotter: Done.

internals/daemon/response.go
```diff
@@ -114,26 +115,6 @@ func (r *resp) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
-	errorKindLoginRequired     = errorKind("login-required")
-	errorKindDaemonRestart     = errorKind("daemon-restart")
-	errorKindSystemRestart     = errorKind("system-restart")
-	errorKindNoDefaultServices = errorKind("no-default-services")
-	errorKindNotFound          = errorKind("not-found")
-	errorKindPermissionDenied  = errorKind("permission-denied")
-	errorKindGenericFileError  = errorKind("generic-file-error")
-)
-
-type errorValue interface{}
```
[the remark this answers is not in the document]
flotter: Yes. Thanks.

internals/daemon/response.go
flotter: Done, thanks.

internals/daemon/response.go
flotter: I am going to leave this comment open for @niemeyer as he proposed the name. Happy to tweak it :)

benhoyt (approved):
Thanks for the changes -- I'm good with this change now!

internals/daemon/response.go
```diff
@@ -116,22 +116,25 @@ func (r *resp) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
 	errorKindNoDefaultServices = errorKind("no-default-services")
 	errorKindNotFound          = errorKind("not-found")
 	errorKindPermissionDenied  = errorKind("permission-denied")
 	errorKindGenericFileError  = errorKind("generic-file-error")
 )
 
-type errorValue interface{}
+// Maintenance error kinds for use only inside the maintenance field of responses.
+const (
+	errorKindSystemRestart = errorKind("system-restart")
```
flotter, line 130: not split.

client/files_test.go
```diff
@@ -360,7 +360,7 @@ func (cs *clientSuite) TestMakeDirFailsOnDirectory(c *C) {
 	clientErr, ok := err.(*client.Error)
 	c.Assert(ok, Equals, true)
 	c.Assert(clientErr.Message, Equals, "could not bar")
-	c.Assert(clientErr.Kind, Equals, "permission-denied")
+	c.Assert(clientErr.Kind, Equals, client.ErrorKindPermissionDenied)
```
flotter, line 363: undo.

internals/daemon/response.go
```diff
@@ -169,22 +172,30 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, string(f))
 }
 
+// Responsef builds an error Response that returns the status and formatted message.
+//
+// Note: If no arguments are provided, formatting is disabled, and the format string
+// is used as is and not interpreted in any way.
+func Responsef(status int, format string, v ...interface{}) Response {
```
flotter, line 179: ```ErrorResponse()```

internals/daemon/response.go
```diff
@@ -169,22 +172,30 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, string(f))
 }
 
+// Responsef builds an error Response that returns the status and formatted message.
+//
+// Note: If no arguments are provided, formatting is disabled, and the format string
```
flotter, line 177: Remove "Note :"

internals/daemon/response.go
flotter: Fixed.

internals/daemon/response.go
flotter: Fixed.

internals/daemon/response.go
flotter: Fixed.

client/files_test.go
flotter: Done

internals/daemon/response.go
```diff
@@ -169,22 +168,30 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, string(f))
 }
 
+// Responsef builds an error Response that returns the status and formatted message.
```
benhoyt, line 171: Nit: This name needs to match the function name.

internals/daemon/response.go
```diff
@@ -169,22 +168,30 @@ func (f fileResponse) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+//
+// If no arguments are provided, formatting is disabled, and the format string
+// is used as is and not interpreted in any way.
+func ErrorResponse(status int, format string, v ...interface{}) Response {
```
benhoyt, line 175: Are we sure we don't want this to have the `f` prefix? That's fairly typical in Go on these formatting functions. I don't have strong feelings either way, just want to make sure dropping it wasn't accidental.

benhoyt (approved):
This looks much better to me now, and a very lightweight PR.

internals/daemon/response.go
flotter: Need to fix.

internals/daemon/response.go
flotter: Thanks Ben. The final call was in favour of sticking with it this way. Thank you for checking!

internals/daemon/response.go
flotter: Changed to ```ErrorResponse```, resolving.

internals/daemon/response.go
flotter: Fixed.

PR #308, opened 2023-09-15
feat(client): add `Requester` interface

anpep:
The `Requester` interface aims to replace the set of `raw`, `do`, `doSync` and `doAsync` internal methods from the `Client` and abstract them out to an interface that can be plugged into the `Client` and used by internal client API implementations and clients on external applications extending Pebble.

> [!IMPORTANT]
> This draft PR **only focuses on the interface and a reference implementation**. For this reason, no existing code from the client was removed or refactored, so right now there are a lot of redundancies and a huge room for refactoring the `requester` implementation.

I'd like to gather feedback on the `Requester` interface itself.

Also, I'd appreciate some feedback on the handling of `maintenance` and `warningCount`/`warningTimestamp`.

#### Example usage for a synchronous API
```go
// Change fetches information about a Change given its ID.
func (client *Client) Change(id string) (*Change, error) {
	var chgd changeAndData
        ctx := context.Background()
	_, err := client.Requester.RequestSync(ctx, "GET", "/v1/changes/"+id, &RequestOptions{
		Output: &chgd,
	})
	if err != nil {
		return nil, err
	}
	chgd.Change.data = chgd.Data
	return &chgd.Change, nil
}
```

#### Example usage for an asynchronous API
```go
// Exec starts a command with the given options, returning a value
// representing the process.
func (client *Client) Exec(opts *ExecOptions) (*ExecProcess, error) {
	/* ... */

	var body bytes.Buffer
	err := json.NewEncoder(&body).Encode(&payload)
	if err != nil {
		return nil, fmt.Errorf("cannot encode JSON payload: %w", err)
	}
	info, err := client.Requester.RequestAsync(ctx, "POST", "/v1/exec", &RequestOptions{
		Headers: map[string]string{"Content-Type": "application/json"},
		Body: &body,
	})
	if err != nil {
		return nil, err
	}

	var result execResult
	if err = json.Unmarshal(info.Result, &result); err != nil {
		return nil, fmt.Errorf("cannot unmarshal JSON response: %w", err)
	}

	changeID := info.ChangeID
	taskID := result.TaskID

	/* ... */
}
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -120,6 +120,23 @@ type Client struct {
 	getWebsocket getWebsocketFunc
 }
 
+type WarningHandler interface {
+	SetWarnings(count int, timestamp time.Time)
+}
+
+type MaintenanceHandler interface {
+	SetMaintenance(error)
+}
+
+func (c *Client) SetWarnings(count int, timestamp time.Time) {
+	c.warningCount = count
+	c.warningTimestamp = timestamp
+}
+
+func (c *Client) SetMaintenance(err error) {
+	c.maintenance = err
+}
+
 type getWebsocketFunc func(url string) (clientWebsocket, error)
 
 type clientWebsocket interface {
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -159,10 +161,129 @@ func New(config *Config) (*Client, error) {
...
+	if r.userAgent != "" {
+		req.Header.Set("User-Agent", r.userAgent)
+	}
+	for h, v := range opts.Headers {
+		req.Header.Set(h, v)
+	}
+
+	retry := time.NewTicker(doRetry)
+	defer retry.Stop()
+	timeout := time.After(doTimeout)
```
benhoyt, line 223: Should these intervals be configurable in the `RequestOptions`?

client/client.go
```diff
@@ -159,10 +161,129 @@ func New(config *Config) (*Client, error) {
...
+		}
+	}
+	if r.maintenanceHandler != nil {
+		r.maintenanceHandler(rsp.Maintenance)
+	}
+	if err := rsp.err(nil); err != nil {
+		return nil, err
+	}
+
+	if rsp.Type == "sync" {
```
benhoyt, line 260: It seems to me the caller should be able to say "I'm expecting a `sync` (or `async`) request". If you're expecting an `async` but get a `sync`, I think this will succeed, but it shouldn't.

Whether that means they pass in an expected request type, or we have two different functions again, I'm not sure.

benhoyt (commented):
Overall I like this consolidation. Couple of things, however:

1) Per my comment below, I think we do still need a way for the caller to say "I'm expecting an `async` response, and it's an error if I don't get one". Whether that means separate `RequestSync` and `RequestAsync` functions, or something you pass in, I'm not sure. I'd lean towards two functions.
2) Why does this need to be an interface? Will there ever be more than one implementation? Or is the idea that the interface will live in `client`, and the implementation will live in `internals/clientutil` or similar?

client/client.go
```diff
@@ -159,10 +161,129 @@ func New(config *Config) (*Client, error) {
...
+	baseURL            *url.URL
+	userAgent          string
+	client             *http.Client
+	maintenanceHandler func(error)
+	warningHandler     func(count int, timestamp time.Time)
+}
+
+func (r *requester) Request(ctx context.Context, verb, path string, opts *RequestOptions) (*ResponseInfo, error) {
+	url := *r.baseURL
+	url.Path = pathpkg.Join(url.Path, path)
```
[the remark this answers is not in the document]
anpep: Yes! I keep forgetting we are not targeting old golang versions anymore (:

client/client.go
anpep: I've not seen this value changed anywhere, but after seeing your notices PR I think we at least should provide a way to disable the retry/timeout. Will look into this

client/client.go
anpep: I will look into this. Thank you!

niemeyer:
Dropped the `Polish` tag as _polishing_ is generally making things behave the same way but better. This is a major API change.

flotter:
@anpep I have taken of this PR as per our agreement, and following conversations with Gustavo I have made some changes and pushed it into a branch I own. I will close this PR for now to reduce the traffic on the Pebble repo.

PR #309, opened 2023-09-19
fix(cmdstate): better error message if $HOME dir doesn't exist

### LINKED ISSUE #189: Protect against nonexistent user home directories when running pebble exec as a user
# Issue

When one runs `useradd <user>` to create a user, the home directory for the user is not guaranteed to be created. The following is a statement from the [man page of useradd](https://linux.die.net/man/8/useradd):
> d, --home HOME_DIR
    The new user will be created using HOME_DIR as the value for the user's login directory. The default is to append the LOGIN name to BASE_DIR and use that as the login directory name. The directory HOME_DIR does not have to exist but will not be created if it is missing.

Despite this missing home directory, one can sign in as the user and execute commands (albeit with a warning displayed):
```
root@mysql-k8s-0:/# su - mysql 
su: warning: cannot change directory to /home/mysql: No such file or directory
$ pwd 
/
```

However, (and may be due to the warning), running `pebble exec` with such a user fails:
```
root@mysql-k8s-1:/# /charm/bin/pebble exec --user mysql --group mysql pwd
- exec command "pwd" (fork/exec /usr/bin/pwd: no such file or directory)
```

benhoyt:
Currently if you create a user without a home directory (or a home directory that doesn't exist), `pebble exec` fails with a hard-to-understand error message which implies the executable you're trying to run doesn't exist. (As #189 points out, you can easily create a user without creating the home directory.)

For the [first version](https://github.com/canonical/pebble/pull/309/commits/ccf677e23f1ddea506abe8e043160fe7e4d72984) of this PR, we made that not fail (and use "/" as the working dir). However, as @hpidcock pointed out, that's problematic, and it probably should fail. Otherwise the caller might call `rm -rf bin` with `HOME=/home/ubuntu`, which if it doesn't exist will fall back to deleting `/bin` -- likely not what you want.

So it's better to return an error in this case. The charm or whoever's calling Pebble will need to fix their charm, by:

1) Explicitly specifying a working dir which does exist, or
2) Creating the HOME directory.

To test this, build and run Pebble with `sudo` in one terminal:

```
$ go build ./cmd/pebble
$ sudo ./pebble run
2023-09-19T05:08:20.457Z [pebble] Started daemon.
...
```

Then create a user with a non-existent home directory (I've created `tstuser` with non-existent HOME dir `/home/tstuser` for this example). Then use `sudo ./pebble exec`:

```
# BEFORE THIS PR
$ sudo ./pebble exec --user tstuser --group tstuser -- pwd
- exec command "pwd" (fork/exec /usr/bin/pwd: no such file or directory)

# AFTER THIS PR
$ sudo ./pebble exec --user tstuser --group tstuser -- pwd
error: cannot call exec: home directory "/home/tstuser" does not exist
```

Fixes #189

### DIFF
--- a/internals/daemon/api_exec.go
+++ b/internals/daemon/api_exec.go
@@ -69,19 +69,6 @@ func v1PostExec(c *Command, req *http.Request, _ *UserState) Response {
 		return statusBadRequest("cannot find executable %q", payload.Command[0])
 	}
 
-	// Also check that the working directory exists, to avoid a confusing
-	// error message later that implies the command doesn't exist:
-	//
-	//  fork/exec /usr/local/bin/realcommand: no such file or directory
-	//
-	// Note that this check still doesn't check that the permissions are
-	// correct to use it as a working directory, but this is a good start.
-	if payload.WorkingDir != "" {
-		if !osutil.IsDir(payload.WorkingDir) {
-			return statusBadRequest("cannot find working directory %q", payload.WorkingDir)
-		}
-	}
-
 	p, err := c.d.overlord.ServiceManager().Plan()
 	if err != nil {
 		return statusBadRequest("%v", err)
--- a/internals/overlord/cmdstate/request.go
+++ b/internals/overlord/cmdstate/request.go
@@ -117,11 +117,14 @@ func Exec(st *state.State, args *ExecArgs) (*state.Task, ExecMetadata, error) {
 		environment["LANG"] = "C.UTF-8"
 	}
 
-	// Set default working directory to $HOME, or / if $HOME not set.
+	// Set default working directory to $HOME, or / if $HOME not set or doesn't exist.
 	workingDir := args.WorkingDir
 	if workingDir == "" {
 		workingDir = environment["HOME"]
 		if workingDir == "" {
+			workingDir = "/"
+		} else if !osutil.IsDir(workingDir) {
+			logger.Debugf("HOME dir %q does not exist, using /", workingDir)
 			workingDir = "/"
 		}
 	}
--- a/internals/overlord/cmdstate/request.go
+++ b/internals/overlord/cmdstate/request.go
@@ -17,6 +17,7 @@ package cmdstate
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/user"
 	"strconv"
@@ -117,16 +118,9 @@ func Exec(st *state.State, args *ExecArgs) (*state.Task, ExecMetadata, error) {
 		environment["LANG"] = "C.UTF-8"
 	}
 
-	// Set default working directory to $HOME, or / if $HOME not set or doesn't exist.
-	workingDir := args.WorkingDir
-	if workingDir == "" {
-		workingDir = environment["HOME"]
-		if workingDir == "" {
-			workingDir = "/"
-		} else if !osutil.IsDir(workingDir) {
-			logger.Debugf("HOME dir %q does not exist, using /", workingDir)
-			workingDir = "/"
-		}
+	workingDir, err := getWorkingDir(args.WorkingDir, environment["HOME"])
+	if err != nil {
+		return nil, ExecMetadata{}, err
 	}
 
 	// Create a task for this execution (though it's not started here).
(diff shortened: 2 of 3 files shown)


flotter (approved):
LGTM!

internals/overlord/cmdstate/request.go
```diff
@@ -117,12 +117,15 @@ func Exec(st *state.State, args *ExecArgs) (*state.Task, ExecMetadata, error) {
...
-	// Set default working directory to $HOME, or / if $HOME not set.
+	// Set default working directory to $HOME, or / if $HOME not set or doesn't exist.
 	workingDir := args.WorkingDir
 	if workingDir == "" {
 		workingDir = environment["HOME"]
 		if workingDir == "" {
 			workingDir = "/"
+		} else if !osutil.IsDir(workingDir) {
+			logger.Debugf("HOME dir %q does not exist, using /", workingDir)
+			workingDir = "/"
```
barrettj12, line 128: if we move all this logic into a separate function, we could use early returns and avoid all the nesting

internals/overlord/cmdstate/request.go
```diff
@@ -117,12 +117,15 @@ func Exec(st *state.State, args *ExecArgs) (*state.Task, ExecMetadata, error) {
...
 	}
 
-	// Set default working directory to $HOME, or / if $HOME not set.
+	// Set default working directory to $HOME, or / if $HOME not set or doesn't exist.
 	workingDir := args.WorkingDir
 	if workingDir == "" {
 		workingDir = environment["HOME"]
 		if workingDir == "" {
 			workingDir = "/"
+		} else if !osutil.IsDir(workingDir) {
```
barrettj12, line 126: maybe we should do an `os.Stat` check first. It might be useful to distinguish between "not found" and "found but it's a file, not a dir", and surface a slightly different error message to the user.

internals/overlord/cmdstate/request.go
hpidcock, line 128: Actually, I think this is very dangerous. We should warn, or fail, but never change the working directory.

If WorkingDir is set we must use that, if HOME is set we must use that if neither are set we must use `/`.
If I run a command `rm -rf bin` with HOME=/home/ubuntu, semantically I want to remove the bin directory in /home/ubuntu. We should never default that to `/`. The user was explicit.

internals/overlord/cmdstate/request.go
```diff
@@ -117,12 +117,15 @@ func Exec(st *state.State, args *ExecArgs) (*state.Task, ExecMetadata, error) {
...
+			logger.Debugf("HOME dir %q does not exist, using /", workingDir)
```
hpidcock, line 127: This needs to be a warning printed to stderr of the exec.

hpidcock (approved):
Actually I am happy with this. The check if its a directory is for HOME and HOME alone. 

If the working directory passed does not exist, that is really up to the forked process to decide how to handle that, in this case the fork/exec code of go.

internals/daemon/api_exec.go
```diff
@@ -69,19 +69,6 @@ func v1PostExec(c *Command, req *http.Request, _ *UserState) Response {
...
 
-	// Also check that the working directory exists, to avoid a confusing
-	// error message later that implies the command doesn't exist:
-	//
-	//  fork/exec /usr/local/bin/realcommand: no such file or directory
-	//
-	// Note that this check still doesn't check that the permissions are
-	// correct to use it as a working directory, but this is a good start.
-	if payload.WorkingDir != "" {
-		if !osutil.IsDir(payload.WorkingDir) {
```
benhoyt, line 80: This check has moved to `cmdstate.Exec` (but better).

internals/overlord/cmdstate/request.go
benhoyt: I've moved this to a separate function as the stat / error handling logic is now quite a lot.

internals/overlord/cmdstate/request.go
benhoyt: Done.

internals/overlord/cmdstate/request.go
benhoyt: Per discussion, I've updated the PR to continue to fail in this case, but return a much better error message.

hpidcock (approved):
Much better thanks. In the future we might need to re-asses exec to deal with explicit login shell behaviour more like su.

shayancanonical (approved):
LGTM!

benhoyt:
This ended up being just error message improvements, not behavioural changes -- merged without further review.

PR #310, opened 2023-09-20
feat(client): expose a default Requester interface

flotter:
The patch adds a public Requester interface which allow derived projects to use the Pebble instantiated client and extend the available commands. The extended client commands will use ```client.Requester().Do``` for HTTP requests.

Derived project (imports Pebble):

```go
// exClient.go:
type ExClient struct {
	*client.Client
}

func (ex *ExClient) Reboot(opts *RebootOptions) error {
	payload := &devicePayload{
		Action: "reboot",
	}
	:
	if _, err := ex.Requester().Do(...); err != nil {
		return err
	}
	:
}

// cmd_reboot.go
type cmdReboot struct {
	client *exClient.ExClient
}

func init() {
	cli.AddCommand(&cli.CmdInfo{
		Name:        "reboot",
		Summary:     cmdRebootSummary,
		Description: cmdRebootDescription,
		New: func(opts *cli.CmdOptions) flags.Commander {
			return &cmdReboot{client: &exClient.ExClient{opts.Client}}
		},
	})
}

func (cmd *cmdReboot) Execute(args []string) error {
	if len(args) > 0 {
		return cli.ErrExtraArgs
	}

	return cmd.client.Reboot(&exClient.RebootOptions{})
}
```

The changes made in the patch has been done in a way to produce as small
as possible diffs (we keep doSync/doAsyc wrappers). The default interface
has been implemented in the client.go file to allow reviewers to easily
identify which code was added, and which is unchanged.

The following changes are made:

1. The ResultInfo type previously returned by doSync and doAsync private
function are removed. Although this is a publicly exposed type, the return
value as of today has always been discarded, and the struct is currently
empty.

2. ResultInfo has been replaced by RequestResponse.

3. The client Hijack mechanism which allowed the doer to be replaced has been
removed. This was originally added in snapd for an arguably slightly obscure use
case. We now have a new Requester interface, so any future need
for this will instead be built in top of the Requester interface.

5. The logs client request now uses the same retry logic on GET failure,
as any other GET request. This is previously not possible because the
retry logic and response unmarshall code was bundled, not allow raw access
to the HTTP body.


client/requester.go
```diff
@@ -0,0 +1,334 @@
...
+	// Reap idle connections
+	CloseIdleConnections()
+}
+
+type RequestOptions struct {
+	Method  string
+	Path    string
+	Query   url.Values
+	Headers map[string]string
+	Body    io.Reader
```
flotter, line 59: async bool // sync / async

internals/cli/cmd_run.go
```diff
@@ -239,9 +239,6 @@ out:
 		}
 	}
 
-	// Close our own self-connection, otherwise it prevents fast and clean termination.
-	rcmd.client.CloseIdleConnections()
```
anpep, line 243: Why are we dropping this?

client/client.go
```diff
@@ -18,173 +18,147 @@ import (
...
 }
 
-type getWebsocketFunc func(url string) (clientWebsocket, error)
-
-type clientWebsocket interface {
-	wsutil.MessageReader
-	wsutil.MessageWriter
-	io.Closer
-	jsonWriter
+func New(requester Requester) (*Client, error) {
```
anpep, line 79: I like this better than having a separate `NewWithRequester()` function, however—is it acceptable that we break the existing client API?

client/requester.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	// Sanity check sync and async requests
+	if opts.Async == false {
+		if reqResp.Type != "sync" {
+			return nil, fmt.Errorf("expected sync response, got %q", reqResp.Type)
+		}
+	} else {
+		if reqResp.Type != "async" {
+			return nil, fmt.Errorf("expected async response for %q on %q, got %q", opts.Method, opts.Path, reqResp.Type)
```
anpep, line 208: Why not be aligned with sync requests on this particular error?
```suggestion
			return nil, fmt.Errorf("expected async response, got %q", reqResp.Type)
```

client/requester.go
```diff
@@ -0,0 +1,292 @@
...
+// decodeWithNumber decodes input data using json.Decoder, ensuring numbers are preserved
+// via json.Number data type. It errors out on invalid json or any excess input.
+func decodeWithNumber(r io.Reader, value interface{}) error {
+	dec := json.NewDecoder(r)
+	dec.UseNumber()
+	if err := dec.Decode(&value); err != nil {
+		return err
+	}
+	if dec.More() {
+		return fmt.Errorf("cannot parse json value")
```
anpep, line 251: Minor nitpick, since we also capitalize "URL" in error messages (:
```suggestion
		return fmt.Errorf("cannot parse JSON value")
```

client/requester.go
```diff
@@ -0,0 +1,292 @@
...
+	if opts.BaseURL == "" {
+		// By default talk over a unix socket.
+		transport := &http.Transport{Dial: unixDialer(opts.Socket), DisableKeepAlives: opts.DisableKeepAlive}
+		baseURL := url.URL{Scheme: "http", Host: "localhost"}
+		requester = &BaseRequester{baseURL: baseURL, transport: transport}
+	} else {
+		// Otherwise talk regular HTTP-over-TCP.
+		baseURL, err := url.Parse(opts.BaseURL)
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse base URL: %v", err)
```
anpep, line 93:
```suggestion
			return nil, fmt.Errorf("cannot parse base URL: %w", err)
```

client/requester.go
anpep: Nitpick aside, if I understand correctly this error is returned in the case we successfully decoded a JSON response but there's extra text after a valid JSON string. Shouldn't we return a more specific error?
```suggestion
		return fmt.Errorf("cannot decode extra data after JSON response")
```

client/requester.go
```diff
@@ -0,0 +1,292 @@
...
+type doer interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+// RequestError is returned when there's an error processing the request.
+type RequestError struct{ error }
+
+func (e RequestError) Error() string {
+	return fmt.Sprintf("cannot build request: %v", e.error)
+}
```
anpep, line 279:
```suggestion
type RequestError struct{ error }

func (e RequestError) Error() string {
	return fmt.Sprintf("cannot build request: %v", e.error)
}

func (e RequestError) Unwrap() error {
	return e.error
}
```

anpep (approved):
Thanks for your astounding work on this Fred! I left a couple comments and nitpicks. Otherwise, I'm really happy about these changes (:

client/requester.go
flotter: Implemented.

client/client.go
flotter: Fixed.

client/requester.go
flotter: Existing code. Will change if needed.

client/requester.go
flotter: Existing code. Happy to change if I need to.

client/requester.go
flotter: Existing code. Please see final PR.

client/client.go
```diff
@@ -326,66 +426,27 @@ func decodeInto(reader io.Reader, v interface{}) error {
...
+		Headers: headers,
+		Body:    body,
+		Async:   true,
+	}, v)
 }
 
-// ResultInfo is empty for now, but this is the mechanism that conveys
-// general information that makes sense to requests at a more general
-// level, and might be disconnected from the specific request at hand.
-type ResultInfo struct{}
```
flotter, line 387: This is effectively renamed to ```RequestResponse```. We could reuse the name if it is really an issue, but the new naming feels consistent with the new interface.

client/client.go
```diff
@@ -398,8 +459,6 @@ type response struct {
 	WarningCount     int       `json:"warning-count"`
 	WarningTimestamp time.Time `json:"warning-timestamp"`
 
-	ResultInfo
-
```
flotter, line 402: This is moved to the ```decoder(...)``` implementation.

client/client.go
```diff
@@ -423,16 +482,8 @@ const (
...
-func (rsp *response) err(cli *Client) error {
-	if cli != nil {
-		maintErr := rsp.Maintenance
-		// avoid setting to (*client.Error)(nil)
-		if maintErr != nil {
-			cli.maintenance = maintErr
-		} else {
-			cli.maintenance = nil
-		}
-	}
```
flotter, line 435: This is moved to the ```decoder(...)``` implementation, but still happens exactly at the same point, just before the error is unmarshalled.

client/client.go
```diff
@@ -511,3 +562,116 @@ func (client *Client) DebugGet(action string, result interface{}, params map[str
 	_, err := client.doSync("GET", "/v1/debug", urlParams, nil, nil, &result)
 	return err
 }
+
+// decoder receives a raw HTTP response and performs internal client
+// processing, as well as unmarshalling the custom result.
+func (client *Client) decoder(ctx context.Context, rsp *http.Response, opts *RequestOptions, result interface{}) (*RequestResponse, error) {
```
flotter, line 568: The decoder is the client callback mechanism that the Requester interface can use for client specific details. This allows the Requester implementation to stay clean of implementation details.

internals/cli/cmd_run.go
flotter, line 243: Please see the comment I've added at the client ```CloseIdleConnections()``` implementation. This call is typically needed in an iterative design where port / client allocations happen faster than the garbage collector free rate. For example, snapd uses this in some store tests, but in that context we should not use it using this API, just deal with it in the test code. I know this is an invasive change, but I would like to remove this, so that the Reponder interface does not need to provide this (unless of course this is needed).

benhoyt: For reference, Juju uses `CloseIdleConnections` to ensure it closes the connection to Pebble immediately (I'm not entirely sure why! but it does). So yeah, I wouldn't get rid of this.

benhoyt:
I've had a look over this now and I small [play with it in a test CLI that implements a custom client](https://go.dev/play/p/axmBFWIC58F). Maybe it's just me, but I find the approach, and some of the naming, hard to understand.

**The approach:**

I find the new requester and decoder concepts (and interfaces) indirect and a bit confusing. I know we initially discussed just exporting a `Do` or `Request` method on the `Client` type. It seems to me that would be so much simpler, and custom clients could just instantiate a `client` and call `Request` right away. Why was that approach rejected again? I understand we don't just want to export the current `DoSync` and `DoAsync` methods -- we want to carefully consider the API, but still, that basic approach seems much more direct.

It would be helpful to add doc comments on the `Requester` interface and its methods describing exactly what the methods should do. "Do performs a single client request, including decoding to the result parameter, blah blah blah. This is what result means, and how it differs from the returned RequestResponse, etc." Similar for `SetDecoder`.

Why do we need `SetDecoder` or the decoder to be customisable? If a custom client wants to customise decode, can't they just do whatever they want in their `Do` implementation? This is another concept and indirection-via-function that the user needs to get familiar with.

```
type DecoderFunc func(ctx context.Context, res *http.Response, opts *RequestOptions,
                      result interface{}) (*RequestResponse, error)
```

Why does a decoder take a (cancellation) context? Could it take just the body instead of the entire `http.Response`? Why does "decoder" handle websocket-y stuff? On that note, I wonder if we can simplify by leaving websockets out of this API, or at least not to to squeeze it into the same method. Websockets are only used by exec and seem like a pretty niche case.

I don't really understand what the intended use for `BodyReader` is. On that note, I think it might really help to update the PR description with usage example of how this stuff will be used in another project / custom client.

**Naming and other minor points:**

* I think `RequestResponse` is odd: is it a request or a response? Could it just be `Response`?
* Similarly, could `RequestOptions` just be `Request`? (I realise we use `FooOptions` for the main API methods, but this is a bit different.)
* See `NewMyClient` in the example I linked above: currently the only way of getting hold of the default decoder is by calling `NewWithRequestor` so it sets the default decoder for you.

internals/cli/cmd_run.go
flotter: @benhoyt I take your concern to heart, but I do feel like it should be possible to prove whether its needed or not in the Pebble case? The Juju case (I assume this is a separate code base, right) sounds exactly right, but the Pebble case makes this call just before the process terminates (in my analysis), which does not feel correct.

client/client.go
```diff
@@ -34,6 +34,63 @@ import (
...
+
+// RequestOptions allows setting up a specific request.
+type RequestOptions struct {
+	Method  string
+	Path    string
+	Query   url.Values
+	Headers map[string]string
+	Body    io.Reader
+	Async   bool
+}
```
flotter, line 86: ```StreamResult bool```

client/client.go
```diff
@@ -34,6 +34,63 @@ import (
...
+
+// RequestResponse defines a common response associated with requests.
+type RequestResponse struct {
+	StatusCode int
+	ChangeID   string
```
flotter, line 91: ```StreamResult io.ReadCloser```

client/client.go
```diff
@@ -34,6 +34,63 @@ import (
...
+	//
+	// 4. Raw HTTP body request
+	// ctx: Supply a context instance
+	// RequestOptions: Async unused
+	// result: Takes the address of a BodyReader interface
+	// RequestResponse: Unused
+	//
+	// See the default implementation for further details.
+	Do(ctx context.Context, opts *RequestOptions, result interface{}) (*RequestResponse, error)
+
```
flotter, line 70: Transport() *http.Transport

client/client.go
```diff
@@ -34,6 +34,63 @@ import (
...
+	// ctx: Supply a context instance
+	// RequestOptions: Path takes the complete websocket path
+	// result: Takes the address of a Websocket interface
+	// RequestResponse: Unused
+	//
+	// 4. Raw HTTP body request
+	// ctx: Supply a context instance
+	// RequestOptions: Async unused
+	// result: Takes the address of a BodyReader interface
+	// RequestResponse: Unused
```
flotter, line 66: Remove.

internals/cli/cmd_run.go
flotter: You are right @benhoyt :) So gustavo recalled this, and the problem is solves, as stated in the comment (doh) is that it enables the server shutdown to go faster, as the server does not have to wait for idle connection timeout.

client/client.go
flotter: Done.

client/client.go
flotter: Done.

client/client.go
flotter: Changed to RoundTripper interface instead of http.Transport.

client/client.go
flotter: Done.

client/client.go
```diff
@@ -300,17 +343,39 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
+}
 
-	if v != nil {
-		if err := decodeInto(rsp.Body, v); err != nil {
-			return err
-		}
+// Do implements all the required functionality as defined by the Requester interface. RequestOptions
+// selects the Do behaviour. In case of a successful response, the result argument get a
+// request specific result. The RequestResponse struct will include common attributes
+// (not all will be set of all request types).
```
flotter, line 354: typo: ... for all the request types.

benhoyt:
I had a voice discussion with @flotter and wondered why we can't use a much simpler version that passes in an `http.Client` (or `http.Transport`) when creating a new `Client`, and then exports `Do` directly as a method of `Client`. Like so:

```go
// client.go
type Client { ... }

type Config {
    SocketPath string
    ...
    HTTPClient *http.Client  // new field (or perhaps http.Transport)
}

func New(config *Config) { }  // existing function

// new method on Client type
func (c *Client) Do(ctx context.Context, request *Request, result interface{}) (*Response, error) { }
```

This saves the complexity of the new interface, the awkward `SetDecoder` (with an implementation's `Do` calling the decoder that was set), etc.

client/client.go
```diff
@@ -34,6 +34,44 @@ import (
...
+	ReturnBody bool
+}
+
+// RequestResponse defines a common response associated with requests.
+type RequestResponse struct {
+	StatusCode int
+	ChangeID   string
+	Body       io.ReadCloser
+}
+
```
flotter, line 74: func (c *Client) Requester() Requester {
    return c.requester
}

client/client.go
```diff
@@ -135,31 +170,24 @@ func New(config *Config) (*Client, error) {
...
+	})
+	if err != nil {
+		return nil, err
 	}
+	return NewWithRequester(requester)
+}
 
-	client.doer = &http.Client{Transport: transport}
-	client.userAgent = config.UserAgent
+func NewWithRequester(requester Requester) (*Client, error) {
```
flotter, line 185: collapse into New()

client/client.go
```diff
@@ -34,6 +34,44 @@ import (
 	"github.com/canonical/pebble/internals/wsutil"
 )
 
+// DecoderFunc allows the client access to the HTTP response. See the SetDecoder
+// description for more details.
+type DecoderFunc func(ctx context.Context, res *http.Response, opts *RequestOptions, result interface{}) (*RequestResponse, error)
+
+type Requester interface {
+	// Allows for sync, async requests as well as direct access to the
```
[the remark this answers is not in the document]
flotter: Fixed.

client/client.go
flotter: Fixed.

client/client.go
```diff
@@ -275,20 +313,25 @@ func (h hijacked) Do(req *http.Request) (*http.Response, error) {
 
 // Hijack lets the caller take over the raw HTTP request.
 func (client *Client) Hijack(f func(*http.Request) (*http.Response, error)) {
```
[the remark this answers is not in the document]
flotter: Done.

client/client.go
flotter: Done.

client/client.go
```diff
@@ -106,12 +144,9 @@ type Config struct {
 
 // A Client knows how to talk to the Pebble daemon.
 type Client struct {
-	baseURL   url.URL
-	doer      doer
-	userAgent string
-
-	maintenance error
+	Requester Requester
```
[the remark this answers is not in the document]
flotter: Done.

client/client.go
flotter: Done.

benhoyt:
tl;dr: I'm not happy with this as is. It's unnecessarily complex, and the new interfaces are hard to implement -- I've implemented a custom requester and decoder to try it. It's not an API I'd be proud of, and I think we can solve the problems at hand in a simpler way.


## Existing approach, with custom Requester and DecoderFunc

I find the decoder interface particularly confusing: you create a client with a `Requester`, which has a `SetDecoder` method, and then there's `DecoderFunc`, which the requester should call, but the decoder has to decode the entirety of the response (a complex thing to define and describe for an interface!), set values in the client, and so on. It's a real tangle.

[Draft PR](https://github.com/benhoyt/pebble/pull/9/files) showing my custom requester and decoder and a test program to drive them.

When you implement a `DecoderFunc` -- which you have to do if you're implementing a Requester, because we're not exporting `Client.Decoder` -- you have to reimplement a lot of the tricky stuff in the client anyway: Pebble server response types and decoding them, checking error/sync/async types, decoding the change ID, and so on. These are fiddly details that would take as long to document what the `DecoderFunc` needs to do as actually do it. A custom decoder is [over 100 lines of fiddly code copied from client.go](https://github.com/benhoyt/pebble/pull/9/files#diff-b94a51e04f8ef149fad50fd5b0c7da6f2dc94c3e45eb8a4ebca3409fb58f20e8R54-R159).

If we exported `Client.decoder` you could potentially implement your custom decoder by wrapping `Client.Decoder` (though you'd have to read the body and then provide it at `Client.Decoder` as a `bytes.Reader`, because you can't read the `http.Response.Body` twice).

When you implement a `Requester`, the `Do` method has to perform HTTP request, but also handle things like retries and the other stuff the client does now -- together that's about [50 lines of non-trivial code copied from client.go](https://github.com/benhoyt/pebble/pull/9/files#diff-b94a51e04f8ef149fad50fd5b0c7da6f2dc94c3e45eb8a4ebca3409fb58f20e8R212-R269).

In addition, because of how the `Requester` has to support websockets and has a `Transport()` which returns an `*http.Transport`, it's not transport-agnostic anyway. So is it even realistic to create custom requesters in the first place?

I understand the desire to conceptually separate the `Requester`, `Client`, and decoder. However, it creates a lot of complexity, as I've described above, and as can be seen below with the amount of new API surface we're proposing in this PR:

<details>
  <summary><b>Click to expand new API surface</b></summary>

```go
type DecoderFunc func(ctx context.Context, res *http.Response, opts *RequestOptions, result interface{}) (*RequestResponse, error)

type Requester interface {
	Do(ctx context.Context, opts *RequestOptions, result interface{}) (*RequestResponse, error)
	SetDecoder(decoder DecoderFunc)
	Transport() *http.Transport
}

type RequestOptions struct {
	Method     string
	Path       string
	Query      url.Values
	Headers    map[string]string
	Body       io.Reader
	Async      bool
	ReturnBody bool
}

type RequestResponse struct {
	StatusCode int
	ChangeID   string
	Body       io.ReadCloser
}

func (client *Client) Requester() Requester { ... }

type DefaultRequesterConfig struct {
	BaseURL          string
	Socket           string
	DisableKeepAlive bool
	UserAgent        string
}

type DefaultRequester struct {
	baseURL   url.URL
	doer      doer
	userAgent string
	transport *http.Transport
	decoder   DecoderFunc
}

func NewDefaultRequester(opts *DefaultRequesterConfig) (*DefaultRequester, error) { ... }
```

</details> 

Still, my [draft PR](https://github.com/benhoyt/pebble/pull/9/files) shows my custom requester and custom decoder used to implement `MyClient`, a new client (like the kind that might be used in Termus) with a new method `UpperServices` (which really just calls `/v1/services`, but pretend it's a new Termus-specific server method).

I use the custom decoder to decode a pretend new Termus-specific `ServerVersion` field that's included in all sync responses (just as an example). This is like the existing warnings and maintenance fields.

Another oddity: the custom decoder decodes warnings/maintenance fields into fields on `MyService`, but then the Pebble `Client.WarningsSummary()` and `Client.Maintenance()` are no longer correct. We don't have access to the Pebble `Client` fields to set those.

It all works, though I did have to add a [`client.Config.Requester`](https://github.com/benhoyt/pebble/pull/9/files#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R145) field, as currently in the main PR there's no way to actually use a custom `Requester`.

Overall, I think it's painful to use, with a lot of boilerplate copied from the current implementation in client.go.

Let's see if we can simplify.


## Simpler option 1

As mentioned above, I don't think it's useful (or even possible with websockets / `Transport()`) to separate the `Requester` from HTTP, so let's just try `Client` methods instead.

The custom decoder above has to decode the *entire* response, so let's instead provide a "decode hook" which allows setting a custom function to just read or unmarshal the parts it wants (like `ServerVersion`), but have the main client still do all the messy sync/async/error checking and unmarshalling. Providing a decode hook *adds* to that.

In option 1 ([draft PR](https://github.com/benhoyt/pebble/pull/7/files)), I've taken the existing `Do` method signature from the proposed `Requester` interface and made it a method on client. I've kept `RequestOptions` and `RequestResponse` as is:

```go
func (client *Client) Do(ctx context.Context, opts *RequestOptions, result interface{})
    (*RequestResponse, error) { .. }
```

For the decode hook, there's this new API:

```go
func (client *Client) SetDecodeHook(hook DecodeHookFunc) {
	client.decodeHook = hook
}

type DecodeHookFunc func(data []byte, opts *RequestOptions) error
```

You can see how simple a sample implementation is [here](https://github.com/benhoyt/pebble/pull/7/files#diff-b94a51e04f8ef149fad50fd5b0c7da6f2dc94c3e45eb8a4ebca3409fb58f20e8R49-R65) -- it just decodes the `ServerVersion` field and is done.

<details>
  <summary><b>Click to expand sample decode hook</b></summary>

```go
func (c *MyClient) decodeHook(data []byte, opts *client.RequestOptions) error {
	// Demonstrate use of opts: only do custom decode on v1 requests.
	if !strings.HasPrefix(opts.Path, "/v1/") {
		return nil
	}
	var frame struct {
		ServerVersion string `json:"server-version"`
	}
	err := json.Unmarshal(data, &frame)
	if err != nil {
		return err
	}
	if frame.ServerVersion != "" {
		c.ServerVersion = frame.ServerVersion
	}
	return nil
}
```
</details>

I realize it's slightly less efficient, as it will have to unmarshal a second time (the client unmarshals as well). But 1) it's a small price to pay for how much simpler this approach is, and 2) I think it's likely decode hooks won't be used often.

I've also solved the websocket issue by adding a [`Client.Transport() *http.Transport` method](https://github.com/benhoyt/pebble/pull/7/files#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R169). See the [`MyClient.GetWebsocket` implementation](https://github.com/benhoyt/pebble/pull/7/files#diff-b94a51e04f8ef149fad50fd5b0c7da6f2dc94c3e45eb8a4ebca3409fb58f20e8R84) that uses it.


## Simpler option 2

Option 2 ([draft PR](https://github.com/benhoyt/pebble/pull/8/files)) is similar to the above, but I've broken `Do` into what are really three functions: `DoSync`, `DoAsync`, and `DoRaw` (we could bikeshed on that name, maybe `DoReturnBody`). I know something similar was proposed in the past and rejected, but I think I've cleaned up the API significantly here.

I don't like how `RequestOptions` has bool `Async` and `ReturnBody` input parameters, which is really a three-way switch (`Async` is ignored if `ReturnBody` is true).

In addition, they should really take and return different parameters in each mode:

* sync requests decode a result but don't return a response
* async requests don't decode a result but do return a change ID
* raw/return-body requests don't decode a result and return an `io.ReadCloser`

Separating them out is clearer, and means the parameters and return values are specific to the operation (making invalid combinations impossible at compile time).

One other tweak here: because `method, path string` are required, I've moved them to parameters, and `RequestOptions` becomes only the *optional* parameters. This `method, path` parameter style is also consistent with the stdlib's `http.NewRequest*` functions.

Here's the option 2 methods:

```go
func (client *Client) DoSync(ctx context.Context, method, path string, opts *RequestOptions, result interface{})
    error { ... }

func (client *Client) DoAsync(ctx context.Context, method, path string, opts *RequestOptions)
    (changeID string, err error) { ... }

func (client *Client) DoRaw(ctx context.Context, method, path string, opts *RequestOptions)
    (io.ReadCloser, error) { ... }
```

Compare a simple do call with option 1 -- not bad, but not great:

```go
opts := &client.RequestOptions{ // Async: false is implicit
	Method: "GET",
	Path:   "/v1/services",
}
_, err := c.pebble.Do(ctx, opts, &result)
```

With option 2 -- simpler and more type safe:

```go
err := c.pebble.DoSync(ctx, "GET", "/v1/services", nil, &result)
```

Again, option 2 adds `Client.Transport()` to solve the websocket issue.


## Summary

I believe the suggested options solve the problem of adding Termus-specific functionality and custom decoding. They do so with much less API surface, less code in the client.go implementation, and they're significantly simpler to use when you're importing Pebble or writing a custom decoder.

Of the two option, I prefer option 2 for the reasons stated there. It does add 3 new exported API methods instead of 1, but they're all clear and the parameters and result types are consistent with the operation.

Later, if we really do need to replace the entire request handling with something non-HTTP, we can add a way to inject a `net.Dialer` or an `http.RoundTripper`. (But either way we'd have to solve the `Transport() *http.Transport` issue for websockets -- the original proposal doesn't help with that.)

Perhaps we can have a discussion about this to avoid going round in further circles.

niemeyer:
> tl;dr: I'm not happy with this as is. It's unnecessarily complex, and the new interfaces are hard to implement -- I've implemented a custom requester and decoder to try it. It's not an API I'd be proud of, and I think we can solve the problems at hand in a simpler way.

Thanks for taking the time to describe these alternatives in detail, Ben. Unfortunately, they seem to ignore the conversation we had with the team on Tuesday where I detailed why we're trying to avoid having methods on the `Client` itself: these are low-level HTTP APIs, and there are cases such as the use of web sockets and logs that are both unhandled. If we go down the suggested route, then every low-level transport concern becomes a "just put another method on the client" kind of conversation, which transforms the `pebble.Client` from being a nice represnetation of high-level operations that you can perform in Pebble with a mixture of low-level concerns including HTTP, transport, websockets, etc. As you put it, that wouldn't make me very proud either.

With that said, I've worked with Fred today to simplify the current proposal further, and drop from it concerns that we do not need to solve today, such as how to implement an external requester.  That means the Requester interface is exposed, but can change because none of our own APIs publish a way for third-party requesters to be used.  With this change, we take away the needs of having an exposed decoder, which I will document below for posterity regardless.

As explained on Tuesday, the previously proposed decoder interface allows third-party code that leverages Pebble's foundation to use the Requester interface, while still allowing the Client itself to capture auxiliar metadata that is transferred to the client as part of request response. Warnings are a good example: any sync or async request may return warnings which are unrelated to the request at hand. As the third-party logic performs a Do request, we don't want it to be the responsible for decoding, or even knowing about, all such side metadatac carried by the request.

That's why the interface is less trivial than it could be otherwise. Either way, we'll simplify it further and keep some of these concerns private while we can do that.

benhoyt:
Thanks, @niemeyer and @flotter. I'm much happier starting simple and expanding later as we have concrete use cases for a decoder.

> Unfortunately, they seem to ignore the conversation we had with the team on Tuesday where I detailed why we're trying to avoid having methods on the Client itself

Sorry about that. I tried to address that with: "I understand the desire to conceptually separate the Requester, Client, and decoder. However, it creates a lot of complexity..." But I got a bit carried away with my alternative proposals and sort of forgot where we ended up with our discussion in the process.

I'm pretty happy with your simplified `Requester` proposal. I've had a play with it [here](https://github.com/benhoyt/pebble/pull/10/files). However, if there's going to be no way to implement an external `Requester`, that's all a bit moot. And we can probably get rid of `NewDefaultRequester` and not export `DefaultRequester` or `DefaultRequesterConfig` at all -- at least until we support creating external requesters.

Separating out `DecodeResult` to a method on `RequestResponse` is an interesting idea -- I think I like it: despite meaning two calls where one would have done before, it's a clean separation.

client/exec.go
```diff
@@ -152,14 +152,10 @@ func (client *Client) Exec(opts *ExecOptions) (*ExecProcess, error) {
 	headers := map[string]string{
 		"Content-Type": "application/json",
 	}
-	resultBytes, changeID, err := client.doAsyncFull("POST", "/v1/exec", nil, headers, &body)
-	if err != nil {
-		return nil, err
-	}
 	var result execResult
-	err = json.Unmarshal(resultBytes, &result)
+	reqResponse, err := client.doAsync("POST", "/v1/exec", nil, headers, &body, &result)
```
niemeyer, line 156: s/reqResponse/resp/

This PR is using it both ways. Let's please establish `resp` as the pattern as it's a nice complement to `req`.

client/export_test.go
```diff
@@ -26,19 +27,34 @@ var (
 )
 
 func (client *Client) SetDoer(d doer) {
-	client.doer = d
+	client.Requester().(*DefaultRequester).doer = d
 }
 
 func (client *Client) Do(method, path string, query url.Values, body io.Reader, v interface{}) error {
```
niemeyer, line 33: Makes little sense to export this for tests here, now that we have a public requester. It's okay to postpone this for now, but it would be nice to clean it up as a follow up. Can we please leave just a TODO note here pointing that out?

client/export_test.go
```diff
@@ -26,19 +27,34 @@ var (
...
+		Method:  method,
+		Path:    path,
+		Query:   query,
+		Headers: nil,
+		Body:    body,
+	})
+	if err != nil {
+		return err
+	}
+	err = decodeInto(resp.Body, v)
```
niemeyer, line 45: Why are we calling the private function here?  Shouldn't this be using the public API?

client/export_test.go
```diff
@@ -26,19 +27,34 @@ var (
...
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 func (client *Client) FakeAsyncRequest() (changeId string, err error) {
-	changeId, err = client.doAsync("GET", "/v1/async-test", nil, nil, nil)
+	reqResponse, err := client.doAsync("GET", "/v1/async-test", nil, nil, nil, nil)
```
niemeyer, line 53: s/reqResponse/resp/

client/services.go
```diff
@@ -65,19 +65,24 @@ type multiActionData struct {
...
 	if err != nil {
-		return nil, "", fmt.Errorf("cannot marshal multi-service action: %w", err)
+		return "", fmt.Errorf("cannot marshal multi-service action: %w", err)
 	}
 	headers := map[string]string{
 		"Content-Type": "application/json",
 	}
-	return client.doAsyncFull("POST", "/v1/services", nil, headers, bytes.NewBuffer(data))
+
+	reqResponse, err := client.doAsync("POST", "/v1/services", nil, headers, bytes.NewBuffer(data), nil)
```
niemeyer, line 81: s/reqResponse/resp/

internals/cli/cmd_run.go
niemeyer: For posterity, yes we need it. :)

niemeyer (changes requested):
Thanks, Fred!  Getting close!

client/client.go
```diff
@@ -34,6 +34,53 @@ import (
...
+	Do(ctx context.Context, opts *RequestOptions) (*RequestResponse, error)
+
+	// Transport returns the HTTP transport in use by the underlying HTTP client.
+	Transport() *http.Transport
+}
+
+type RequestType int
+
+const (
+	SyncRequest RequestType = iota
```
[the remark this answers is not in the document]
flotter: I was leaning toward the choice that would result in making the most common case as compact as possible (not having to set Type).
```
flotter@pc:~/pebble/client$ grep -r doAsync | wc -l
4
flotter@pc:~/pebble/client$ grep -r doSync | wc -l
19
```
But, I do still prefer setting the Type just to make the code more explicit, and since we have wrappers ```doSync``` and ```doAsync```, the cost is negligible. I changed it so raw comes first.

client/client.go
```diff
@@ -34,6 +34,53 @@ import (
...
+	AsyncRequest
+	RawRequest
+)
+
+// RequestOptions allows setting up a specific request.
```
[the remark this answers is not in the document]
flotter: Agreed. Removed.

client/client.go
```diff
@@ -34,6 +34,53 @@ import (
...
+	Body    io.Reader
+}
+
+// RequestResponse defines a common response associated with requests.
+type RequestResponse struct {
+	StatusCode int
+	ChangeID   string
+	Result     []byte
+
+	// Only set for RawRequest and must be completely read and closed.
```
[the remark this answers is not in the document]
flotter: Discussed. We agreed that although following the requirements to drain and close the io.Reader is a requirement for connection re-use, this is still a decision the caller may decide to ignore. For example, our one use-case of this, log streaming, cannot possibly drain the body, as its a never ending stream.

As a result, I am dropping the hard requirement here, and changed this to:

```Only set for RawRequest.```

File: client/client.go [hunk 30]
```diff
@@ -34,6 +34,53 @@ import (
...
+type RequestOptions struct {
+	Type    RequestType
+	Method  string
+	Path    string
+	Query   url.Values
+	Headers map[string]string
+	Body    io.Reader
+}
+
+// RequestResponse defines a common response associated with requests.
```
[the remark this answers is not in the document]
flotter: My mistake. Originally, this was only the part of the response common to all requests (not the custom payload). After the last changes, this response type also provides a decode feature that can decode the caller specific payload. Given the decode is documented, I will remove this comment.

client/client.go
```diff
@@ -34,6 +34,53 @@ import (
...
+type RequestResponse struct {
+	StatusCode int
+	ChangeID   string
+	Result     []byte
+
+	// Only set for RawRequest and must be completely read and closed.
+	Body io.ReadCloser
+}
+
+// DecodeResult can be used to decode a SyncRequest or AsyncRequest.
```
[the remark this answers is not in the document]
flotter:
```go
// DecodeResult decodes the endpoint-specific result payload that is included as part of
// sync and async request responses. The decoding is performed with the standard JSON
// package, so the usual field tags should be used to prepare the type for decoding.
func (resp *RequestResponse) DecodeResult(result interface{}) error {
        if err := decodeWithNumber(bytes.NewReader(resp.Result), result); err != nil {
                return fmt.Errorf("cannot unmarshal: %w", err)
        }
        return nil
}
```

I have also removed the '''nil''' interface use case, as this is not an optional call, so the caller can decide if decoding is needed.

niemeyer:
@benhoyt 

> I'm pretty happy with your simplified `Requester` proposal. I've had a play with it [here](https://github.com/benhoyt/pebble/pull/10/files). However, if there's going to be no way to implement an external `Requester`, that's all a bit moot. And we can probably get rid of `NewDefaultRequester` and not export `DefaultRequester` or `DefaultRequesterConfig` at all -- at least until we support creating external requesters.

Just to be clear, this was indeed the actual agreement from last week. The interface itself is not mood because it reserves the ability to have different requesters in the future, but we do intend to make the concrete implementation private for now.

client/client.go
```diff
@@ -181,10 +221,7 @@ func getWebsocket(transport *http.Transport, url string) (clientWebsocket, error
 
 // CloseIdleConnections closes any API connections that are currently unused.
 func (client *Client) CloseIdleConnections() {
-	c, ok := client.doer.(*http.Client)
-	if ok {
-		c.CloseIdleConnections()
-	}
+	client.Requester().Transport().CloseIdleConnections()
```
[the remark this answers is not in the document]
flotter: I interpret this as an observation only :) resolving.

client/client.go
```diff
@@ -219,27 +256,26 @@ func (e ConnectionError) Unwrap() error {
 	return e.error
 }
 
-// raw performs a request and returns the resulting http.Response and
-// error you usually only need to call this directly if you expect the
-// response to not be JSON, otherwise you'd call Do(...) instead.
-func (client *Client) raw(ctx context.Context, method, urlpath string, query url.Values, headers map[string]string, body io.Reader) (*http.Response, error) {
+// raw creates an HTTP request, performs the request and returns an HTTP response
+// and error.
```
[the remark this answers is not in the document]
flotter: I may need your help with this one Gustavo.

First, I take note of your feedback on how I can improve my commenting, thank you for that.

In this case, the design has changes so that the "other" comment part that says:

```
// ... you usually only need to call this directly if you expect the
// response to not be JSON, otherwise you'd call Do(...) instead.
```

I removed this because the comment is no longer possible. In the past the comment referred to the use case where the logs command directly called this to get access to the raw http body, but it is no longer possible to access this method using the Requester interface in the client, and the design has changed so that all use cases use ```Do()``` consistently, including the documented case above.

Maybe we should just remove the comment all-together as its not really adding any value as you rightfully points out ?

client/client.go
```diff
@@ -219,27 +256,26 @@ func (e ConnectionError) Unwrap() error {
...
+func (br *DefaultRequester) raw(ctx context.Context, method, urlpath string, query url.Values, headers map[string]string, body io.Reader) (*http.Response, error) {
```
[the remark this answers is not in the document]
flotter: Leftover from baseRequester, oops!

Changed to: ```func (rq *defaultRequester) ... ``` and renamed ```raw``` to ```dispatch```.

client/client.go
```diff
@@ -248,47 +284,32 @@ func (client *Client) raw(ctx context.Context, method, urlpath string, query url
 }
 
 var (
-	doRetry   = 250 * time.Millisecond
-	doTimeout = 5 * time.Second
+	rawRetry   = 250 * time.Millisecond
+	rawTimeout = 5 * time.Second
```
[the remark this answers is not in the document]
flotter: Reverted.

client/client.go
```diff
@@ -248,47 +284,32 @@ func (client *Client) raw(ctx context.Context, method, urlpath string, query url
...
-func (client *Client) Hijack(f func(*http.Request) (*http.Response, error)) {
-	client.doer = hijacked{f}
-}
-
-// do performs a request and decodes the resulting json into the given
-// value. It's low-level, for testing/experimenting only; you should
-// usually use a higher level interface that builds on this.
-func (client *Client) do(method, path string, query url.Values, headers map[string]string, body io.Reader, v interface{}) error {
-	retry := time.NewTicker(doRetry)
+// rawWithRetry builds in a retry mechanism for GET failures (body-less request)
```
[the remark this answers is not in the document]
flotter: I removed that comment. I am not sure why I added it without much more context.

The background is I researched why only GET requests are retried, and found that practically speaking, retrying reading the request body is not practical, as the state of the io.Reader is not exposed, and cannot be "reset".

client/client.go
```diff
@@ -248,47 +284,32 @@ func (client *Client) raw(ctx context.Context, method, urlpath string, query url
...
+func (br *DefaultRequester) rawWithRetry(ctx context.Context, method, urlpath string, query url.Values, headers map[string]string, body io.Reader) (*http.Response, error) {
```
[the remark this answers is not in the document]
flotter: Done.

client/client.go
```diff
@@ -300,17 +321,79 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
-		return err
+		return nil, err
+	}
+	return rsp, nil
+}
+
+// Do implements all the required functionality as defined by the Requester interface. RequestOptions
+// selects the Do behaviour. In case of a successful response, the result argument get a
+// request specific result. The RequestResponse struct will include common attributes
+// (not all will be set for all request types).
```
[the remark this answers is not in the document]
flotter: Thank you! done.

client/client.go
```diff
@@ -300,17 +321,79 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
-	defer rsp.Body.Close()
 
-	if v != nil {
-		if err := decodeInto(rsp.Body, v); err != nil {
-			return err
+	// Is the result expecting a caller-managed raw body?
+	if opts.Type == RawRequest {
+		return &RequestResponse{
+			Body: httpResp.Body,
+		}, nil
```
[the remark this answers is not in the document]
flotter: Noted, thanks!

client/client.go
```diff
@@ -300,17 +321,79 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
+	}
+
+	// If we get here, this is a normal sync or async server request so
+	// we have to close the body.
```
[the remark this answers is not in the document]
flotter: Yes true. Removed.

client/client.go
```diff
@@ -300,17 +321,79 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
+		}
+		if serverResp.Change == "" {
+			return nil, fmt.Errorf("async response without change reference")
+		}
+	} else {
+		return nil, fmt.Errorf("cannot process unknown request type")
 	}
 
-	return nil
+	// Warnings are only included if not an error type response
```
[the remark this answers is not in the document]
flotter: I am not sure what the history is behind this, but it comes from the daemon, and the tests picked this up.

```
// daemon.go: +280

                if rsp.Type != ResponseTypeError {
                        st.Lock()
                        count, stamp := st.WarningsSummary()
                        st.Unlock()
                        rsp.addWarningsToMeta(count, stamp)
                }

```

client/exec.go
flotter: Done.

client/export_test.go
flotter: I added the following comment in export_test.go:

```
+// TODO: Clean up tests to use the new Requester API. Tests do not generate a client.response type
+// reply in the body while SyncRequest or AsyncRequest responses assume the JSON body can be
+// unmarshalled into client.response.
```

client/export_test.go
flotter: I've clarified this in the TODO comment i've added above:

The test framework do not generate a response of type ```client.response```, which means that we cannot use ```SyncRequest``` or ```AsyncRequest``` as is today, without changing the test suite. These request types assume a client.response can be unmarshalled as the first step.

I am happy to enhance everything to better utilise the new Requester API, but I propose we do this as a followup task.

client/export_test.go
flotter: Done.

client/services.go
flotter: Done.

client/client.go
```diff
@@ -34,6 +34,50 @@ import (
...
+	Method  string
+	Path    string
+	Query   url.Values
+	Headers map[string]string
+	Body    io.Reader
+}
+
+type RequestResponse struct {
+	StatusCode int
+	ChangeID   string
```
benhoyt, line 64: Should we add a comment to the `ChangeID` field, something like "this is normally only set for async requests: it's the change ID of the asynchronous operation started".

Similar for `Result`, I think it needs a comment about how it's just the inner payload part (however you want to say that clearly).

client/client.go
```diff
@@ -265,30 +291,15 @@ func FakeDoRetry(retry, timeout time.Duration) (restore func()) {
...
-type hijacked struct {
-	do func(*http.Request) (*http.Response, error)
-}
-
-func (h hijacked) Do(req *http.Request) (*http.Response, error) {
-	return h.do(req)
-}
-
-// Hijack lets the caller take over the raw HTTP request.
-func (client *Client) Hijack(f func(*http.Request) (*http.Response, error)) {
```
benhoyt, line 277: I think it's worth mentioning in the PR description that we've quietly removed this exported method (and why that's almost certainly okay).

client/client.go
```diff
@@ -300,17 +311,72 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
+	}
+
+	defer httpResp.Body.Close()
+	var serverResp response
+	if err := decodeInto(httpResp.Body, &serverResp); err != nil {
+		return nil, err
+	}
+
+	// Update the maintenance error state
+	if serverResp.Maintenance != nil {
```
benhoyt, line 339: I know you didn't change this code, but I was just about to suggest "this is equivalent to just `rq.client.maintenance = serverResp.Maintenance`" but that's not true, because `serverResp.Maintenance` is a `*Error` and `rq.client.maintenance` is an `error`. [This whole thing.](https://go.dev/doc/faq#nil_error). Can we add a comment to explain why this is necessary to ensure this doesn't get "fixed" in future?

client/client.go
```diff
@@ -300,17 +311,72 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
+	}
+
+	// Deal with error type response
+	if err := serverResp.err(); err != nil {
+		return nil, err
+	}
+
+	// At this point only sync and async type requests may exist so lets
+	// make sure this is the case.
+	if opts.Type == SyncRequest {
```
benhoyt, line 352: Nit: `switch opts.Type` with a `default` would be slightly nicer for this I think.

client/client.go
```diff
@@ -511,3 +549,44 @@ func (client *Client) DebugGet(action string, result interface{}, params map[str
 	_, err := client.doSync("GET", "/v1/debug", urlParams, nil, nil, &result)
 	return err
 }
+
+type defaultRequester struct {
+	baseURL   url.URL
```
benhoyt, line 554: It looks like we're never setting the `baseURL` field -- oversight? We do use it in `dispatch` (somewhat surprised there's not a test for this -- we should probably add one).

client/export_test.go
```diff
@@ -26,19 +27,37 @@ var (
 )
 
 func (client *Client) SetDoer(d doer) {
-	client.doer = d
+	client.Requester().(*defaultRequester).doer = d
```
benhoyt, line 30: Should this `SetDoer` method be moved to a test method/function? I believe external packages can't actually call it anyway, as the `doer` interface isn't exported. But happy to leave this to another PR if you prefer.

benhoyt (commented):
This is looking really good now, thanks. A couple of very nit comments, and one note about (I think) missing setting `baseURL` in `defaultRequester`.

client/client.go
flotter: I've added this to the PR description, which I will use as the commit message. Resolving.

client/client.go
```diff
@@ -511,3 +549,44 @@ func (client *Client) DebugGet(action string, result interface{}, params map[str
...
+	if opts.BaseURL == "" {
+		// By default talk over a unix socket.
+		transport := &http.Transport{Dial: unixDialer(opts.Socket), DisableKeepAlives: opts.DisableKeepAlive}
+		baseURL := url.URL{Scheme: "http", Host: "localhost"}
+		requester = &defaultRequester{baseURL: baseURL, transport: transport}
+	} else {
+		// Otherwise talk regular HTTP-over-TCP.
+		baseURL, err := url.Parse(opts.BaseURL)
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse base URL: %v", err)
```
anpep, line 577: Consider using `%w` here.

client/client.go
```diff
@@ -34,6 +34,50 @@ import (
...
+
+	// Only set for RawRequest.
+	Body io.ReadCloser
+}
+
+// DecodeResult decodes the endpoint-specific result payload that is included as part of
+// sync and async request responses. The decoding is performed with the standard JSON
+// package, so the usual field tags should be used to prepare the type for decoding.
+func (resp *RequestResponse) DecodeResult(result interface{}) error {
+	if err := decodeWithNumber(bytes.NewReader(resp.Result), result); err != nil {
```
anpep, line 75: Can we move the `decodeWithNumber` implementation here and get rid of it? It seems to only be used once, and it's only usage is here.

client/client.go
```diff
@@ -423,16 +469,8 @@ const (
...
-	if cli != nil {
-		maintErr := rsp.Maintenance
-		// avoid setting to (*client.Error)(nil)
-		if maintErr != nil {
-			cli.maintenance = maintErr
-		} else {
-			cli.maintenance = nil
-		}
-	}
+// err extract the error in case of an error type response
```
anpep, line 472: nit:
```suggestion
// err extracts the error in case of an error type response.
```

client/client.go
anpep, line 577: nit:
```suggestion
			return nil, fmt.Errorf("cannot parse base URL: %w", err)
```

client/client.go
flotter:
```
+               // We cannot assign a nil pointer of type *Error to an
+               // interface here because the interface is only nil if
+               // both the type and value is nil.
+               // https://go.dev/doc/faq#nil_error
                rq.client.maintenance = nil
```

flotter: Resolved.

client/client.go
flotter: Thanks, this does look much better.

client/client.go
flotter: Hi Ben, correct me if I am missing something obvious but we have these:

```
if opts.BaseURL == "" {
		// By default talk over a unix socket.
		transport := &http.Transport{Dial: unixDialer(opts.Socket), DisableKeepAlives: opts.DisableKeepAlive}
		baseURL := url.URL{Scheme: "http", Host: "localhost"}
		requester = &defaultRequester{baseURL: baseURL, transport: transport}
:

```

See the last line of the snippet :)

flotter: I think this is OK ? I created an issue to look at refactoring the client test framework to better utilise the Requester API.

client/client.go
flotter: Looks like this file consistently prefers wrapping of errors, so I've changed this as per Angel's suggestion to align.

client/export_test.go
flotter: I think this aligns well with the general refactor proposal for this test framework based on https://github.com/canonical/pebble/issues/315

client/export_test.go
flotter: Will address in https://github.com/canonical/pebble/issues/315

client/client.go
flotter: I collapsed ```decodeWithNumber()``` into this. It makes sense for me :) thanks for the feedback.

client/client.go
```diff
@@ -136,33 +177,24 @@ func New(config *Config) (*Client, error) {
...
 
 	return client, nil
 }
 
+func (client *Client) Requester() Requester {
+	return client.requester
+}
+
 func (client *Client) getTaskWebsocket(taskID, websocketID string) (clientWebsocket, error) {
 	url := fmt.Sprintf("ws://localhost/v1/tasks/%s/websocket/%s", taskID, websocketID)
```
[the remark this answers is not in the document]
flotter: I think it may make sense to do this as separate exercise. The client code around dealing with baseURL/socket path and the server code dealing with socket path / HTTP address seems to be involved. What I propose is that we look at this when we work on a PR to enable TCP client comms. Will that be OK ?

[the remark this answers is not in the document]
flotter: Confirmed with Angel. Resolving.

benhoyt (approved):
Looks excellent now, thanks!

client/client.go
flotter: Fixed.

client/client.go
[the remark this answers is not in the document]
flotter: Updated the comment to better explain why this is necessary. Resolving.

client/export_test.go
flotter: Agreed to postpone this work.

client/client.go
```diff
@@ -300,17 +306,77 @@ func (client *Client) do(method, path string, query url.Values, headers map[stri
...
+		return nil, fmt.Errorf("cannot process unknown request type")
 	}
 
-	return nil
+	// Warnings are only included if not an error type response
+	rq.client.warningCount = serverResp.WarningCount
+	rq.client.warningTimestamp = serverResp.WarningTimestamp
+
+	// Common response
+	return &RequestResponse{
```
flotter, line 375: make up type and return type.

niemeyer (approved):
Looks good, thanks all!

PR #267, opened 2023-08-01
feat(logfwd): support forwarding logs to Loki

barrettj12:
Add a `loki.Client` type that encodes sends Pebble logs to Loki. This implements the `logstate.logClient` interface introduced in #256.

There have also been some changes to the `logClient` interface. Semantically, clients shouldn't flush inside the `Write` method - the flush should be up to the gatherer. Rename `Write` to `AddLog` for clarity, and add a `NumBuffered` method to help the gatherer determine when to flush.

## QA steps

Deploy a Loki server for testing - this can be done with Juju. For convenience, deploy Grafana too, to view the logs.
```
juju bootstrap microk8s c
juju add-model cos
juju deploy loki-k8s --trust
juju deploy grafana-k8s --trust
juju integrate loki-k8s grafana-k8s:grafana-source
```
Wait for Loki and Grafana to reach a stable state.

Create a simple logging service:
```bash
#!/bin/bash
while true; do
  echo "hello"
  sleep 1
done
```
and a simple plan using this service:
```yaml
services:
  svc1: &logsvc
    command: /path/to/logsvc
    startup: enabled
    override: merge
  svc2: *logsvc

log-targets:
  loki:
    type: loki
    location: http://[loki_ip]:3100/loki/api/v1/push
    override: merge
    services: [all]
```
`[loki_ip]` is the unit/app IP address from the deployed Loki in `juju status`.

Run Pebble:
```
go run ./cmd/pebble run
```

Get the admin password for Grafana using
```
juju run grafana-k8s/0 get-admin-password
```
and then visit `[grafana_ip]:3000` in a browser, and log in with username `admin` and the obtained password.

If you go to the "Explore" tab and change the data source to Loki, you should be able to see logs forwarded from Pebble.

---

JUJU-4205

PR #312, opened 2023-09-27
feat(logfwd): add log labels

barrettj12:
This PR builds on the previous log forwarding PRs (#209, #252, #256, #267) by allowing the user to specify labels for logs forwarded from Pebble.

Labels will be defined in a new field `labels` inside the log target:
```yaml
log-targets:
  tgt1:
    type: loki
    location: https://my.loki.server.com:3100/loki/api/v1/push
    labels:
      key1: val1
      key2: val2
```

Labels will support environment variable substitutions ($ENV_VAR), and these will be interpreted "dynamically" in the context of the relevant service.

### Notes on the design

Labels can contain `$ENVIRONMENT_VARIABLES`, which will be substituted in the environment of the relevant service. However, the "service environment" could mean:
1. the `environment` map as defined in the plan
2. the actual environment in which the process is run

We decided to use (1) instead, as it is more explicit, can be "statically" evaluated just from the plan, and doesn't require extra data to be passed through from the service manager and cached.

Furthermore, anytime we change the labels, we first flush any logs currently in the buffer, so that these logs are sent with the correct (old) labels.

### QA steps

<details>

Set up a simple logging service:
```bash
#!/bin/bash
while true; do
  echo "hello"
  sleep 1
done
```

Set up a plan using labels and environment variables:
```yaml
services:
  svc1: &svc1
    command: /path/to/logsvc
    startup: enabled
    override: merge
    environment:
      OWNER: alice
      IP: 100.32.2.1
      PORT: 9090

  svc2:
    <<: *svc1
    environment:
      OWNER: bob
      IP: 100.32.2.5
      PORT: 3000

log-targets:
  tgt1:
    type: loki
    override: merge
    location: http://10.42.0.222:3100/loki/api/v1/push
    services: [all]
    labels:
      owner: user-$OWNER
      address: http://${IP}:${PORT}
```

Run Pebble:
```
PEBBLE_DEBUG=1 go run ./cmd/pebble run
```
and verify that the labels are being correctly interpreted.
</details>

### Links

Jira card: JUJU-4643


internals/overlord/logstate/loki/loki.go
```diff
@@ -169,9 +196,7 @@ func (c *Client) buildRequest() lokiRequest {
 	for _, service := range services {
 		entries := bucketedEntries[service]
 		stream := lokiStream{
-			Labels: map[string]string{
-				"pebble_service": service,
-			},
+			Labels:  c.getLabels(service),
```
hpidcock, line 199: As we discussed in our call, which we should discuss next week: Since the labels are based on environment, and are lazily evaluated here + we reuse the ring buffer between service instances + we have a common shared buffer in the client, we have the risk that a services's logs emitted before a service restart will be labelled the same as logs emitted after the restart.

hpidcock (commented):
Looks like a good change. Just need to consider the problem we discovered next week.

internals/overlord/logstate/gatherer.go
```diff
@@ -312,6 +314,10 @@ type logClient interface {
 
 	// Flush sends buffered logs (if any) to the remote target.
 	Flush(context.Context) error
+
+	// AddEnv adds a service's environment to the client, so it can correctly
+	// evaluate the labels defined in the plan.
+	AddEnv(serviceName string, env []string)
```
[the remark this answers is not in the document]
barrettj12: don't know about "need" but I guess it's a reasonable addition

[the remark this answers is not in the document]
barrettj12: @hpidcock maybe we should make these methods more abstract, e.g. `ServiceAdded` and `ServiceRemoved`? We might have more things in future that need to be passed in on service start/shutdown, so we could just extend these methods instead of adding new ones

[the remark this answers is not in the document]
benhoyt: Just for the record after our spec review meeting, we decided this was the right place for this, but:

* It would be about labels rather than environment
* The method would be `SetLabels(serviceName string, labels map[string]string)` method
* Labels=nil would mean remove this service

internals/overlord/logstate/loki/loki.go
barrettj12: The obvious/easy fix is to force flush the client whenever the plan changes.

However, there is one potential issue - the PlanChanged is called from a separate goroutine to the gatherer's main control loop. So there might be concurrency issues here (since we didn't design the client to be concurrency-safe).

README.md
```diff
@@ -749,6 +783,15 @@ log-targets:
     # service name with a minus (e.g. '-svc1') to remove a previously added
     # service. '-all' will remove all services.
     services: [<service names>]
+
+    # (Optional) A list of key/value pairs defining labels which should be set
+    # on the outgoing logs. These labels are added to (protocol-dependent)
+    # default labels (e.g. 'pebble_service' for Loki).
+    # The label values may contain $ENVIRONMENT_VARIABLES, which will be
+    # substituted 'dynamically' in the environment for the corresponding
+    # service.
```
benhoyt, line 792: Two small things:

1. Instead of just an example (e.g.) for default labels, can we document the list? Maybe a separate paragraph like:

```
The defaults labels for each protocol are:
- loki: pebble_service
```

2. Can you use a more specific word than 'dynamically' here? For example, "at the time the log is written" or "when the plan is updated" (I think it's the latter).

internals/overlord/logstate/gatherer.go
```diff
@@ -133,17 +133,23 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
 // PlanChanged is called by the LogManager when the plan is changed, if this
 // gatherer's target exists in the new plan.
 func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.RingBuffer) {
+	// TODO: we should flush the client here because the labels may have changed.
+	// But this is done in a separate goroutine to the main loop - is that going
+	// to be a problem?
```
benhoyt, line 138: I believe it's a problem even as is, because `lokiClient.SetLabels` updates the map, and `lokiClient.Flush` accesses it, which is a data race.

I think the simplest option would be to define the `SetLabels` method (and only that method) as being safe to call concurrently with `Write`/`Flush`. Then have a single mutex in the Loki client that locked when `Flush` uses the labels map and when `SetLabels` updates it.

internals/overlord/logstate/loki/loki.go
```diff
@@ -84,6 +85,10 @@ func fillDefaultOptions(options *ClientOptions) *ClientOptions {
 	return options
 }
 
+func (c *Client) SetLabels(serviceName string, labels map[string]string) {
+	c.labels[serviceName] = labels
```
benhoyt, line 89: I saw you previously did a `delete` if `labels` was nil. Isn't that better (in terms of memory reuse), as otherwise we'll keep the old map entries around?

internals/overlord/logstate/loki/loki.go
```diff
@@ -169,16 +178,27 @@ func (c *Client) buildRequest() lokiRequest {
...
-			},
+			Labels:  c.getLabels(service),
 			Entries: entries,
 		}
 		req.Streams = append(req.Streams, stream)
 	}
 	return req
 }
 
+func (c *Client) getLabels(serviceName string) map[string]string {
```
benhoyt, line 189: Instead of allocating and building this map every `Flush`, can we only do it every `SetLabels` call? It shouldn't be any harder, and it would make the implementation more efficient.

internals/overlord/logstate/manager.go
```diff
@@ -100,13 +102,57 @@ func (m *LogManager) ServiceStarted(service *plan.Service, buffer *servicelog.Ri
...
+	var envMap map[string]string
+
 	for _, gatherer := range m.gatherers {
 		target := m.plan.LogTargets[gatherer.targetName]
 		if !service.LogsTo(target) {
 			continue
 		}
-		gatherer.ServiceStarted(service, buffer)
+
+		if envMap == nil {
```
benhoyt, line 113: Why is the creation of `envMap` done inside the loop, instead of once at the top? If you're concerned about efficiency you can make `parseEnv` return a nil map if env is nil -- but I don't think micro-efficiency matters here, as this only happens once per service start, so I'd probably just keep it simple. (You're already evaluating the env vars every loop anyway.)

internals/overlord/logstate/manager.go
```diff
@@ -100,13 +102,57 @@ func (m *LogManager) ServiceStarted(service *plan.Service, buffer *servicelog.Ri
...
+			return v
+		}
+		logger.Debugf("variable $%s undefined in service environment", k)
+		// undefined variables default to "", just like Bash
+		return ""
+	}
+
+	labels := make(map[string]string, len(rawLabels))
+	for key, rawLabel := range rawLabels {
+		labels[key] = os.Expand(rawLabel, substitute)
```
benhoyt, line 153: Nice!

internals/plan/plan.go
```diff
@@ -969,6 +982,14 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 				Message: fmt.Sprintf("log target object cannot be null for log target %q", name),
 			}
 		}
+		for labelName := range target.Labels {
+			// 'pebble_*' labels are reserved
+			if strings.HasPrefix(labelName, "pebble_") {
+				return nil, &FormatError{
+					Message: fmt.Sprintf("target %q defines illegal label %q", name, labelName),
```
benhoyt, line 989: I think the more specific reason would be helpful to the user. Something like: `"log target %q: labels with prefix 'pebble_' are reserved"`.

benhoyt (commented):
Few minor suggestions and a suggested fix for the `SetLabels` concurrency problem.

internals/overlord/logstate/gatherer.go
barrettj12: I don't think that will solve the problem, as here we've still got a `Flush` being triggered from `PlanChanged`, potentially concurrent to flushes from the main loop (so concurrent reading/writing of the buffer/entries slice).

Not to mention that I'd rather keep the mutex outside the client, in the gatherer. This avoids us having to make all our future client implementations concurrency-safe. Generally, we should put shared/common logic in the gatherer to make clients as small (less code) as possible.

benhoyt: Sounds reasonable. Give it a shot and let's see what that looks like!

README.md
barrettj12: I'd rather not document the list for a few reasons:
- it locks us in
- if it purports to be a definitive reference, someone will forget to update it eventually - then our docs will be lying
- no-one "should care" about what the default labels are - they are sensible defaults whose presence is not hurting anyone

internals/overlord/logstate/manager_test.go
```diff
@@ -184,3 +201,122 @@ func (c *slowFlushingClient) Flush(ctx context.Context) error {
...
+			"foo":     "bar",
+		},
+	})
+}
+
+// Fake logClient implementation which just stores the passed-in labels
+type labelStore struct {
+	labels map[string]map[string]string
+	// synchronise on this channel to avoid data races
+	labelsSet chan struct{}
```
benhoyt, line 299: Naming nit: this sounds like a set of labels. Maybe `setLabelsCh` or `setCh`?

benhoyt (changes requested):
A few minor comments, and one issue with how we're doing `ServiceData.Equal`.

README.md
```diff
@@ -422,6 +422,8 @@ log-targets:
         services: [svc1, svc2]
 ```
 
+##### Specifying services
```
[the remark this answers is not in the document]
barrettj12: yep, was thinking just that

internals/overlord/logstate/manager.go
```diff
@@ -129,3 +165,14 @@ func (m *LogManager) Stop() {
 	}
 	wg.Wait()
 }
+
+// ServiceData holds the data for each service that the log manager needs to
+// keep track of.
+type ServiceData struct {
```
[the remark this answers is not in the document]
barrettj12: yeah, the whole point was for this struct to encapsulate all the data that the service manager needs to pass to the log manager. Then we can make this part of the signature for `ServiceStarted`, and if we need to add more data in the future, we can do that without changing the signature of `ServiceStarted`.

If you're really not a fan, I can change that back.

internals/overlord/logstate/manager.go
```diff
@@ -129,3 +165,14 @@ func (m *LogManager) Stop() {
...
+	Buffer *servicelog.RingBuffer
+	Env    map[string]string
+}
+
+func (d *ServiceData) Equal(other *ServiceData) bool {
+	return reflect.DeepEqual(d, other)
```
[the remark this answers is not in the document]
barrettj12: `reflect.DeepEqual` for the maps is probably fine, this will be called infrequently on relatively small maps

internals/overlord/logstate/manager_test.go
barrettj12: ah, the many meanings of the word "set" :)

internals/overlord/logstate/gatherer.go
```diff
@@ -132,18 +138,25 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
 
 // PlanChanged is called by the LogManager when the plan is changed, if this
 // gatherer's target exists in the new plan.
-func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.RingBuffer) {
+func (g *logGatherer) PlanChanged(pl *plan.Plan, serviceData map[string]*ServiceData, labels map[string]map[string]string) {
+	// New plan - the labels may have changed. We should tell the main loop to
+	// flush the client now so that we don't later get out of sync, and send old
+	// logs with the new labels.
+	g.flushCh <- struct{}{}
```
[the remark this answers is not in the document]
barrettj12: seems unlikely, but doesn't hurt to be safe here

internals/overlord/logstate/manager.go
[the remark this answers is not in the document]
benhoyt: Fair enough, let's leave as is then (but make the `Equal` or whatever we end up calling that method not exported).

internals/overlord/logstate/loki/loki.go
```diff
@@ -38,16 +38,17 @@ const (
...
+	target     *plan.LogTarget
 	httpClient *http.Client
 
 	// To store log entries, keep a buffer of size 2*MaxRequestEntries with a
 	// sliding window 'entries' of size MaxRequestEntries
 	buffer  []lokiEntryWithService
 	entries []lokiEntryWithService
+
+	// store the custom labels for each service
+	labels map[string]map[string]string
```
hpidcock, line 51: Can you please extend the comment to indicate what each key does.
`// example usage: labelValue := labels[serviceName][labelKey]`

internals/overlord/logstate/gatherer.go
```diff
@@ -132,18 +138,31 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
...
 			g.pullers.Remove(svcName)
+			g.setLabels <- svcWithLabels{svcName, nil}
 			continue
 		}
 
 		tgt := pl.LogTargets[g.targetName]
 		if !svc.LogsTo(tgt) {
 			g.pullers.Remove(svcName)
+			g.setLabels <- svcWithLabels{svcName, nil}
 		}
```
hpidcock, line 166: these sends also need the g.tomb.Dying() select.

internals/overlord/logstate/gatherer.go
```diff
@@ -154,21 +173,26 @@ func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.
...
+		if !svcStarted {
 			// We don't yet have a reference to the service's ring buffer
 			// Need to wait until ServiceStarted
 			continue
 		}
 
-		g.pullers.Add(service.Name, buffer, g.entryCh)
+		g.pullers.Add(service.Name, svcData.Buffer, g.entryCh)
+		// update labels
+		g.setLabels <- svcWithLabels{service.Name, labels[service.Name]}
```
hpidcock, line 185: this send also need the g.tomb.Dying() select.

internals/overlord/logstate/gatherer.go
```diff
@@ -154,21 +173,26 @@ func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.
...
 }
 
 // ServiceStarted is called by the LogManager on the start of a service which
 // logs to this gatherer's target.
-func (g *logGatherer) ServiceStarted(service *plan.Service, buffer *servicelog.RingBuffer) {
+func (g *logGatherer) ServiceStarted(service *plan.Service, buffer *servicelog.RingBuffer, labels map[string]string) {
 	g.pullers.Add(service.Name, buffer, g.entryCh)
+
+	// Add this service's custom labels to the client
+	g.setLabels <- svcWithLabels{service.Name, labels}
```
hpidcock, line 195: same here with the select on tomb dying

internals/overlord/logstate/gatherer.go
```diff
@@ -154,21 +173,26 @@ func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.
...
+		g.pullers.Add(service.Name, svcData.Buffer, g.entryCh)
+		// update labels
+		g.setLabels <- svcWithLabels{service.Name, labels[service.Name]}
 	}
 }
 
 // ServiceStarted is called by the LogManager on the start of a service which
 // logs to this gatherer's target.
-func (g *logGatherer) ServiceStarted(service *plan.Service, buffer *servicelog.RingBuffer) {
+func (g *logGatherer) ServiceStarted(service *plan.Service, buffer *servicelog.RingBuffer, labels map[string]string) {
```
hpidcock, line 191: Update the comment above to say that ServiceStarted expects labels to never change after it is passed in. Otherwise we need to copy it before we send it on the channel.

internals/overlord/logstate/gatherer.go
```diff
@@ -132,18 +138,31 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
 
 // PlanChanged is called by the LogManager when the plan is changed, if this
 // gatherer's target exists in the new plan.
-func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.RingBuffer) {
+func (g *logGatherer) PlanChanged(pl *plan.Plan, serviceData map[string]*ServiceData, labels map[string]map[string]string) {
```
hpidcock, line 141: Update the comment above to say that PlanChanged expects labels[serviceName] value to never change after it is passed in. Otherwise we need to copy it before we send it on the channel.

hpidcock (approved):
Nice work.

internals/overlord/logstate/gatherer.go
```diff
@@ -132,18 +138,36 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
...
+		return
+	case g.flushCh <- struct{}{}:
+	}
+
 	// Remove old pullers
 	for _, svcName := range g.pullers.Services() {
 		svc, svcExists := pl.Services[svcName]
 		if !svcExists {
 			g.pullers.Remove(svcName)
+			if dying := g.setClientLabels(svcName, nil); dying {
```
benhoyt, line 159: Nit: I think this would be simpler/clearer if we reversed the sense of the boolean so it was "did we set the labels?". And then there'd be no need for the named var IMO. "If that didn't go through, return":

```suggestion
			if !g.setClientLabels(svcName, nil) {
```

What do you think?

internals/overlord/logstate/manager_test.go
```diff
@@ -184,3 +201,199 @@ func (c *slowFlushingClient) Flush(ctx context.Context) error {
...
+	return nil // no-op
+}
+
+func (c *labelStore) SetLabels(serviceName string, labels map[string]string) {
+	c.labels[serviceName] = labels
+	c.notifySetLabels <- struct{}{}
+}
+
+// wait for labels to be set
+func (l *labelStore) wait(c *C, timeout time.Duration) {
```
benhoyt, line 316: This name + args confused me at first, as I thought the caller was saying "wait 1s", which wouldn't be good in tests. What about we just hard-code the timeout in this method as it's always 1s, and rename the method to `waitLabels` or `waitLabelsSet`?

benhoyt (approved):
Looking really good! Just left two minor nit comments.

internals/overlord/logstate/gatherer.go
hpidcock: I agree. Boolean return value is like answering a question. SetClientLabels -> true sounds like, yes it worked.
But I think the better option would have been to return an error, but that doesn't make much sense here.

For the most I actually think we don't care that it is dying. We just don't want it to block, so we can just drop the if statement entirely.

internals/overlord/logstate/gatherer.go
```diff
@@ -182,17 +187,37 @@ func (g *logGatherer) PlanChanged(pl *plan.Plan, serviceData map[string]*Service
...
+	if dying := g.setClientLabels(service.Name, labels); dying {
+		return
+	}
+}
+
+// setClientLabels tells the main loop to set the labels on the client. This
+// method is safe to use concurrently.
+// It also checks if the gatherer is dying - in which case, it will return
+// dying == true to signal to the caller that they should return.
+func (g *logGatherer) setClientLabels(serviceName string, labels map[string]string) (dying bool) {
```
hpidcock, line 212: I don't think returning true or false here makes much sense. It isn't the responsibility of this set function to determine if we are dying. We just don't want this send here to block, hence the need for the select.

Handling dying is the concern of the caller.

barrettj12: the whole point of this function was to factor out common logic
```go
	select {
	case <-g.tomb.Dying():
		return
	case g.setLabels <- svcWithLabels{serviceName, labels}:
	}
```

but maybe it's not really helping and I should just copy-paste that block in the code?

benhoyt: Yeah, I think that'd probably be better. Couple more lines of code, but the flow would be very clear.

benhoyt (approved):

internals/overlord/logstate/gatherer.go
```diff
@@ -132,18 +138,40 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
...
 		}
 
 		tgt := pl.LogTargets[g.targetName]
 		if !svc.LogsTo(tgt) {
 			g.pullers.Remove(svcName)
+			select {
+			case <-g.tomb.Dying():
+				return
+			case g.setLabels <- svcWithLabels{svcName, nil}:
+			}
```
niemeyer, line 174: These two branches above are exactly the same, and then there's pretty much nothing else. Shouldn't be hard to simplify this into a single branch.

internals/overlord/logstate/gatherer.go
```diff
@@ -312,6 +365,11 @@ type logClient interface {
 
 	// Flush sends buffered logs (if any) to the remote target.
 	Flush(context.Context) error
+
+	// SetLabels sets the log labels for the given service.
+	// When a service is removed, you should call SetLabels(service, nil) to
```
niemeyer, line 370:
```
// SetLabels sets the log labels for the given services, or releases previously
// allocated label resources if the labels parameter is nil.
```
(got triggered by the "you", specifically, which is atypical in API docs)

niemeyer (changes requested):
Thanks, Jordan, and thanks for the earlier reviews as well, Ben, Harry.

A few comments here. Mostly superficial, but a couple of important remarks.

README.md
```diff
@@ -749,6 +784,14 @@ log-targets:
     # service name with a minus (e.g. '-svc1') to remove a previously added
     # service. '-all' will remove all services.
     services: [<service names>]
+
+    # (Optional) A list of key/value pairs defining labels which should be set
+    # on the outgoing logs. These labels are added to (protocol-dependent)
+    # default labels (e.g. 'pebble_service' for Loki).
```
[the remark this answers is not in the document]
barrettj12: `pebble_service` is Loki-specific.

Syslog already has specialised fields which are more appropriate for attaching the service name (either [APP-NAME, PROCID, or MSGID](https://www.rfc-editor.org/rfc/rfc5424#page-14)).

Loki has no such mechanism. Hence, we **need** to attach a custom label so that Loki can know *where* the logs came from. This was the original rationale for the `pebble_service` label.

internals/overlord/logstate/loki/loki.go
```diff
@@ -84,6 +86,24 @@ func fillDefaultOptions(options *ClientOptions) *ClientOptions {
...
+	}
+
+	// Make a copy to avoid altering the original map
+	newLabels := make(map[string]string, len(labels)+1)
+	for k, v := range labels {
+		newLabels[k] = v
+	}
+
+	// Add Loki-specific default labels
+	newLabels["pebble_service"] = serviceName
```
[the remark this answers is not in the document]
barrettj12: This is Loki-specific, as per the comment. So it makes perfect sense exactly where it is IMO.

internals/overlord/logstate/manager.go
```diff
@@ -77,36 +91,58 @@ func (m *LogManager) PlanChanged(pl *plan.Plan) {
...
+}
+
+// evaluateLabels interprets the labels defined in the plan, substituting any
+// $env_vars with the corresponding value in the service's environment.
+func evaluateLabels(rawLabels, env map[string]string) map[string]string {
+	substitute := func(k string) string {
+		if v, ok := env[k]; ok {
+			return v
+		}
+		logger.Debugf("variable $%s undefined in service environment", k)
```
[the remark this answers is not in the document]
barrettj12: OK, I didn't know that convention also applied to debug logs, which are developer-facing rather than user-facing.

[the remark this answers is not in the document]
barrettj12: will drop it

internals/overlord/logstate/manager.go
```diff
@@ -67,8 +69,20 @@ func (m *LogManager) PlanChanged(pl *plan.Plan) {
...
+			if !ok {
+				// Service not yet started
+				continue
+			}
+
+			labels[svcName] = evaluateLabels(target.Labels, svcData.Env)
+		}
+
+		// Update gatherer
+		gatherer.PlanChanged(pl, m.services, labels)
```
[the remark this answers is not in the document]
barrettj12: I will move the label evaluation into the gatherer

internals/overlord/logstate/gatherer.go
```diff
@@ -154,21 +182,35 @@ func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.
...
 			continue
 		}
 
-		g.pullers.Add(service.Name, buffer, g.entryCh)
+		g.pullers.Add(service.Name, svcData.Buffer, g.entryCh)
+		// update labels
+		select {
+		case <-g.tomb.Dying():
+			return
+		case g.setLabels <- svcWithLabels{service.Name, labels[service.Name]}:
```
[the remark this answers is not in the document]
barrettj12: We can set the labels first, before starting the puller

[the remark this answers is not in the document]
barrettj12: and switch the two cases

internals/overlord/logstate/gatherer.go
```diff
@@ -132,18 +138,40 @@ func fillDefaultOptions(options *logGathererOptions) *logGathererOptions {
 
 // PlanChanged is called by the LogManager when the plan is changed, if this
 // gatherer's target exists in the new plan.
-func (g *logGatherer) PlanChanged(pl *plan.Plan, buffers map[string]*servicelog.RingBuffer) {
+// The labels map should not be modified while this method is running.
+func (g *logGatherer) PlanChanged(pl *plan.Plan, serviceData map[string]*ServiceData, labels map[string]map[string]string) {
+	// New plan - the labels may have changed. We should tell the main loop to
+	// flush the client now so that we don't later get out of sync, and send old
```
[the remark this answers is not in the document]
barrettj12: Yes

internals/overlord/logstate/gatherer.go
```diff
@@ -70,6 +70,10 @@ type logGatherer struct {
 	// ensure the client is not blocking subsequent teardown steps.
 	clientCancel context.CancelFunc
 
+	// Channels used to notify the main loop to update the client
+	flushCh   chan struct{}
+	setLabels chan svcWithLabels
```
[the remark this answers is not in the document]
barrettj12: I'm inclined to disagree with you here - to me this feels like throwing away type safety just to save a few lines of code. But it's irrelevant anyway since we're removing the flushCh.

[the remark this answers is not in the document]
barrettj12: I'm inclined to disagree with you here - to me this feels like throwing away type safety just to save a few lines of code. But it's irrelevant anyway since we're removing the flushCh.

internals/overlord/logstate/manager.go
```diff
@@ -129,3 +165,27 @@ func (m *LogManager) Stop() {
...
+
+// ServiceData holds the data for each service that the log manager needs to
+// keep track of.
+type ServiceData struct {
+	Buffer *servicelog.RingBuffer
+	Env    map[string]string
+}
+
+// unchanged checks if d pertains to the same running service as other.
+func (d *ServiceData) unchanged(other *ServiceData) bool {
```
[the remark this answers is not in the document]
barrettj12: I've inlined this logic now.

> if the process changes (so not the same running service, in a way), but the buffer is the same, we can keep the puller untouched.

This comment made me realise: instead of having a ServiceStarted method on the gatherer, we should have EnvChanged and BufferChanged, which are only called if the relevant service datum has changed.

barrettj12:
Commit ca0afa152e34537a9af1cf4be3067997033f56bb seems to have introduced a regression. Now, `managerSuite.TestTimelyShutdown` is failing, because the `LogManager.Stop()` call is timing out:

```
FAIL: manager_test.go:129: managerSuite.TestTimelyShutdown

manager_test.go:180:
    c.Fatal("LogManager.Stop() took too long")
... Error: LogManager.Stop() took too long
```

internals/overlord/logstate/gatherer.go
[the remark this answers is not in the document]
barrettj12: I tried doing this but had to revert it, as it caused one of the tests to fail (TestTimelyShutdown). Will look more into it next week.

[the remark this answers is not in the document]
barrettj12: fixed now, it was a testing issue

hpidcock (approved):
Changes look good, thanks for sticking with this.

README.md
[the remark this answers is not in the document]
niemeyer: Understood, but there are two further concerns here: the first is that although we're talking about ˋpebble_service`, the implication of what you're saying is no labels at all would be sent because it's unsupported, and the second is that there is actually a standard syslog way to send key/value pairs. From that [same spec](https://www.rfc-editor.org/rfc/rfc5424.html#page-15):

> STRUCTURED-DATA provides a mechanism to express information in a well defined, easily parseable and interpretable data format.  There are multiple usage scenarios.  For example, it may express meta-information about the syslog message or application-specific information such as traffic counters or IP addresses.
>
> (...)
> 
>    All examples in this section show only the structured data part of the message.  Examples should be considered to be on one line.  They are wrapped on multiple lines in this document for readability purposes.  A description is given after each example.
>
> Example 1 - Valid
> 
> [exampleSDID@32473 iut="3" eventSource="Application" eventID="1011"]

So in other words, there's a bit of inconsistency in the plan. Namingly:

1. Are labels supported at all for syslog?
2. If yes, then why would we not send `pebble_service` as well?
3. If not, then `pebble_service` can still be global, it's just that `syslog` doesn't support labels.

internals/overlord/logstate/loki/loki.go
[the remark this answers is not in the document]
niemeyer: See the reply to the earlier comment.

internals/overlord/logstate/gatherer.go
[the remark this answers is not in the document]
niemeyer: I don't understand how it's throwing away type safety when the recommendation is to literally use a type for each command?

internals/overlord/logstate/gatherer.go
```diff
@@ -189,30 +178,51 @@ func (g *logGatherer) PlanChanged(pl *plan.Plan, serviceData map[string]*Service
 			continue
 		}
 
-		g.pullers.Add(service.Name, svcData.Buffer, g.entryCh)
-		// update labels
+		labels := evaluateLabels(g.target.Labels, serviceData[service.Name].Env)
```
niemeyer, line 181: We've just been told the plan has changed, and yet we're evaluating this with labels cached from the prior plan. This looks like a bug?  The design we had before this change seemed to intentfully not cache unnecessary data, precisely so we could avoid this kind of pitfall by just not having the data at hand in the first place.

As a more general point, the recommendation on the prior review was to do less, by simply using the actual events we already had with the new parameter which included the environment variable.  So this PR is going quite a bit beyond it by introducing two new event methods here, and caching further data from the plan.  As observed, this creates further chances for bugs and discussion and moves us farther from the merging.

Can we backtrack a bit and just do the suggestions on the prior state so we can get this in?

niemeyer (changes requested):
Thanks for the updates. Some further notes on one unrelated change done, though:

internals/overlord/logstate/gatherer.go
barrettj12: So there's a difference between:
1) the service's `environment` field as defined in the plan
2) the actual environment that the service is run inside

(2) is (1) with a bunch of defaults and extra variables added - see [serviceData.startInternal](https://github.com/canonical/pebble/blob/b255ff384ae0d8859706872e934e5f0b1ba750eb/internals/overlord/servstate/handlers.go#L350).

We were trying to use (2) instead of (1) for completeness. If you'd prefer we just use the environment as defined in the plan (and ignore all the default variables set by the service manager), I'm happy to do that instead. It certainly would simplify things.

README.md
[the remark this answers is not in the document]
benhoyt: Yes, the SD (structured data) fields is the standard way to send key/value pairs to syslog. However, as Jordan mentioned, there's also the APP-NAME field which is a more specific field for the "application name", which seems appropriate for the service name (not PROCID or MSGID). I think making this target type-specific (so Loki-specific in this instance) is the right thing to do here, as other log target types we add in future may have different needs.

Normally you can filter syslog logs by the various fields like APP-NAME. *However*, I think it's fine to research syslog further later when we actually add support for the syslog type. If we end up deciding not to use APP-NAME for some reason and instead use syslog structured data, that's okay.

For this specific documentation change in the plan description, let's leave out the second sentence altogether here ("These labels are added...") and instead add type-specific details under the "type" field, something like so:

```
    # (Required) The type of log target, which determines the format in
    # which logs will be sent. The supported types are:
    #
    # - loki: Use the Grafana Loki protocol. A "pebble_service" label is
    #         added automatically, with the name of the Pebble service as its value.
    type: loki
```

Later, we'd add a new entry `- syslog` with details about the APP-NAME field.

internals/overlord/logstate/loki/loki.go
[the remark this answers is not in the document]
benhoyt: See above - resolving this thread.

README.md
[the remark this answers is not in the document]
barrettj12: @benhoyt I just updated the README as you described.

internals/overlord/logstate/gatherer.go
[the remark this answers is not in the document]
benhoyt: We can switch to a command interface + channel if there are more command types, but for now this is moot as we've gotten rid of the `flushCh` (and are doing the flushes "in line").

internals/overlord/logstate/gatherer.go
benhoyt: Jordan's updated this back to the simpler version that just uses the new plan.

Later, when we add [support for `$INTERPOLATION` in environment variables](https://github.com/canonical/pebble/issues/182) we can tweak this flow if needed.

internals/overlord/logstate/gatherer.go
```diff
@@ -312,6 +372,10 @@ type logClient interface {
 
 	// Flush sends buffered logs (if any) to the remote target.
 	Flush(context.Context) error
+
+	// SetLabels sets the log labels for the given services, or releases
```
benhoyt, line 376: Nit:
```suggestion
	// SetLabels sets the log labels for the given service, or releases
```

internals/overlord/logstate/manager_test.go
```diff
@@ -177,10 +195,135 @@ func (c *slowFlushingClient) Add(_ servicelog.Entry) error {
...
+	return nil // no-op
+}
+
+func (c *labelStore) Flush(_ context.Context) error {
+	return nil // no-op
+}
+
+func (c *labelStore) SetLabels(serviceName string, labels map[string]string) {
+	c.labels[serviceName] = labels
+	c.notifySetLabels <- struct{}{}
```
benhoyt, line 319: Can we add a select with a 1s `time.After` and panic on timeout, just to avoid the test taking forever in case we get something wrong in our test?

internals/overlord/logstate/gatherer_test.go
```diff
@@ -218,6 +219,92 @@ func (s *gathererSuite) TestRetryLoki(c *C) {
...
+		svc1.name: svc1.ringBuffer,
+		svc2.name: svc2.ringBuffer,
+	}
+
+	wg := sync.WaitGroup{}
+	doAsync := func(f func()) {
+		wg.Add(1)
+		go func() {
+			f()
+			wg.Done()
```
benhoyt, line 255: Let's make this `defer wg.Done()` at the top of the inner function?

internals/overlord/logstate/gatherer_test.go
```diff
@@ -218,6 +219,92 @@ func (s *gathererSuite) TestRetryLoki(c *C) {
...
+			Services: map[string]*plan.Service{
+				svc1.name: svc1.config,
+			},
+			LogTargets: map[string]*plan.LogTarget{
+				target.Name: target,
+			},
+		}, buffers)
+	})
+
+	doAsync(func() { g.ServiceStarted(svc1.config, svc1.ringBuffer) })
```
benhoyt, line 270: Can we format these so they're all multi-line to be consistent?

```suggestion
	doAsync(func() {
	    g.ServiceStarted(svc1.config, svc1.ringBuffer)
	})
```

internals/overlord/logstate/gatherer_test.go
```diff
@@ -218,6 +219,92 @@ func (s *gathererSuite) TestRetryLoki(c *C) {
 	}
 }
 
+// Test to catch race conditions in gatherer
+func (s *gathererSuite) TestRace(c *C) {
```
benhoyt, line 223: Can we make this name more specific, like `TestXyzRace` with what it's actually testing a race in or for? And after that let's remove the redundant doc comment.

benhoyt (changes requested):
I've gone through this once more and added a bunch of nit comments. Let's fix those up and get the PR description into shape, do final QA, and then ping me and we can try to merge this week!

internals/overlord/logstate/gatherer_test.go
barrettj12: This is a pretty non-specific test, it's just designed to hammer the log gatherer and find any potential race conditions. It's now finding some artificial race conditions though. Needs a rewrite.

barrettj12: I rewrote the test so it passes. I'm still not sure what to call it though - do you have any suggestions?

benhoyt (approved):
Looks good! I'd like @hpidcock to give this one last look, but I'm ready to merge.

hpidcock (approved):
👍 

I did find one flaky test (using stress to test it) in the previous loki code, but I think it doesn't block the landing of this pr.

```
FAIL: loki_test.go:134: suite.TestFlushCancelContext

loki_test.go:161:
    c.Assert(err, ErrorMatches, ".*context deadline exceeded.*")
... error string = "Post \"http://127.0.0.1:41420\": dial tcp 127.0.0.1:41420: i/o timeout"
... regex string = ".*context deadline exceeded.*"

loki_test.go:169:
    c.Fatal("lokiClient.Flush took too long to return after context timeout")
... Error: lokiClient.Flush took too long to return after context timeout

OOPS: 4 passed, 1 FAILED
--- FAIL: Test (1.01s)
FAIL
```

PR #316, opened 2023-10-13
fix(snap): drive the snap release workflow to avoid skipping tagged versions

### LINKED ISSUE #300: The Snap Store builds might skip Pebble revisions, including stable releases
### SOTA

Our current workflows for the Pebble snap are divided in 2 parts:
 1. the Snap Store is connected to the Pebble repository (see https://snapcraft.io/pebble/builds). This makes the store lookup new commits to the Pebble repo's default branch (`master`) on a frequent basis (I believe the [polling interval is 30''](https://github.com/canonical/snapcraft.io/blob/main/static/js/publisher/builds/index.js#L310)). If there are new commits to the default branch, the Snap Store will automatically build the Pebble snap and release it to the `latest/edge` channel.
    - **PROBLEM** [1]: we have experienced a store behaviour where if >1 commits are pushed within the store's polling window, only the most recent commit will be taken for the snap build. This might cause some snap revisions to be skipped, including stable ones (like `v1.3.0`)
 2. on every release, a dedicated GitHub workflow will be triggered. This workflow will go into an infinite loop that waits for the Snap Store to build a Pebble snap that corresponds to the released Pebble version. Once the store builds and releases that snap, the GitHub workflow will promote it from `latest/edge` to `latest/candidate`

### Proposed solution

Given Pebble's release strategy (2 back-to-back commits for every new release, i.e. "tag" and "revert bump"), the pitfall noted in [1] above is very likely to happen, and as the title says, we might end up not having a snap built for a stable Pebble release.

To avoid introducing an artificial delay between the 2 release commits, I propose the following changes:
 1. disconnect the Pebble repository in the Snap Store (i.e. no more automatic store-side builds)
 2. in `pebble/.github/workflows/snap.yml`, do the following changes:
	 1. add a new trigger event for every new tag and commit to `master`
	 2. add a new job called `remote-build`, that only runs on pushes to `master` and new tags, which will run `snapcraft remote-build --build-for=amd64,arm64...` and then `upload` the resulting snaps to the store, on the `latest/edge` channel. If the trigger event is a `tag`, then this job also promotes the resulting snap from `latest/edge` to `latest/candidate`
	    - alongside, change `snap/snapcraft.yaml` to remove `build-environment` as it is not supported in `remote-builds`. Instead, manually script the binary installation by replacing `craftctl default` by `CGO_ENABLED=0 go build -trimpath -ldflags=-w -ldflags=-s -o ${CRAFT_PART_INSTALL}/bin/pebble ./cmd/pebble
`

cjdcordeiro:
Fixes #300.

This PR addresses the issue described in #300. It includes the following changes:
 - instead of relying on the store to build the snaps, the GH workflow uses `remote-build` to drive the multi-arch builds,
 - the `snapcraft.yaml` parts were updated in order to be supported by the `remote-build` environment,
 - added a new trigger, on `push`, 
 - the resulting multi-arch snaps are uploaded to the snap store on every new GH release and push to `master`,
 - on every new GH release, it now simply takes the latest snap from `edge` to `candidate`,
 - finally, we apply concurrency to ensure 2 consecutive commits won't result in older snaps becoming the "latest" in the snap store.

## Before merging

 1. disconnect this repository from the Pebble snap "Builds" in the snap store
 2. add the following new secrets to the repo (thus why the CI is failing for this PR):
   - LP_CREDENTIALS (new and already shared with the team)
   - PEBBLE_DEV_MAILING_LIST (already shared with the team in the past)

---

This has been tested in: https://github.com/cjdcordeiro/pebble/actions/runs/6505676433 and https://github.com/cjdcordeiro/pebble/actions/runs/6505679630

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -3,98 +3,116 @@ name: Pebble snap
 on:
   pull_request:
     branches: [master]
+  push:
+    branches: [master]
   release:
     types: [published]
 
 env:
   SNAP_NAME: pebble
+  DEFAULT_TRACK: latest
+  DEFAULT_RISK: edge
 
 jobs:
-  build:
+  snaps:
     runs-on: ubuntu-latest
-    outputs:
-      pebble-snap: ${{ steps.build-pebble-snap.outputs.snap }}
-
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [amd64, arm64, ppc64el, armhf, s390x]
     steps:
       - name: Checkout Pebble repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - name: Build Pebble Snap
-        id: build-pebble-snap
-        uses: snapcore/action-build@v1
+      - name: Set LP credentials for remote build
+        id: set-local-paths
+        run: |
+          lp_creds_path="$HOME/.local/share/snapcraft/provider/launchpad"
+          mkdir -p $lp_creds_path
+          echo '${{ secrets.LP_CREDENTIALS }}' > ${lp_creds_path}/credentials
+          git config --global user.email "${{ secrets.PEBBLE_DEV_MAILING_LIST}}"
+          git config --global user.name "pebble-dev"
 
-      - name: Attach pebble snap to GH workflow execution
-        uses: actions/upload-artifact@v2
+          echo "home=$HOME" >> $GITHUB_OUTPUT
+
+      - name: Build ${{ env.SNAP_NAME }} Snap for ${{ matrix.arch }}
+        id: build-snap
+        uses: snapcore/action-build@v1.1.3
         with:
-          name: ${{ steps.build-pebble-snap.outputs.snap }}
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Excellent! I was hoping to see this before the next release. :-)

I haven't used snapcraft, but this looks good overall to me.

jnsgruk (approved):

cjdcordeiro:
It seems like the job is not picking up on the `LP_CREDENTIALS` secret

```
  echo '' > ${lp_creds_path}/credentials
```

jnsgruk:
No I think that's just because it doesn't include sensitive values in the log output :) I'm not sure what happened here, there is definitely a secret named `LP_CREDENTIALS` in place now

cjdcordeiro:
When it's a secret GH will printout `***` in lieu of the variable. Example -> https://github.com/cjdcordeiro/pebble/actions/runs/6505676433/job/17669768265

In this case, it didn't read anything. Can you check the repository's workflow permissions?

jnsgruk:
Closing this PR in favour of #322

Issue #315, opened 2023-10-10
Refactor client test framework now that Requester is merged


flotter:
@flotter and @anpep have an outstanding task to refactor the client test framework to better utilise the Requester interface (#310). At the moment the test framework cannot use ```Requester Do()``` due to ```client.response``` type not used in the response body. We should also ensure all areas are properly covered.

flotter:
Comment from Ben:
```
func (client *Client) SetDoer(d doer) {
	client.doer = d
	client.Requester().(*defaultRequester).doer = d
```
Should this SetDoer method be moved to a test method/function? I believe external packages can't actually call it anyway, as the doer interface isn't exported. But happy to leave this to another PR if you prefer.

flotter:
Comment  from Ben:

It looks like we're never setting the baseURL field -- oversight? We do use it in dispatch (somewhat surprised there's not a test for this -- we should probably add one).

PR #319, opened 2023-10-18
feat(client): allow tests to assign a fake requester

flotter:
The client test suite used by all client command tests fakes the ```doer``` today, which is a private internal ```Requester``` implementation detail. Derived projects using the client ```Requester``` for extended commands needs to fake the Requester interface.

This PR adds a client method ```SetRequester(rw Requester)``` to allow test code to fake the requester. This has to be public to allow an external package to fake this.

Pending work remain to properly refactor the existing client test code to Fake the ```Requester``` instead of the ```doer```. This is captured as an outstanding issue here: https://github.com/canonical/pebble/issues/315

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -190,6 +190,12 @@ func (client *Client) Requester() Requester {
 	return client.requester
 }
 
+// SetRequester allows derived projects that extend the client to test the
+// extended client code by supplying a fake requester interface.
+func (client *Client) SetRequester(rq Requester) {
+	client.requester = rq
+}
+
 func (client *Client) getTaskWebsocket(taskID, websocketID string) (clientWebsocket, error) {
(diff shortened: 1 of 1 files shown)


niemeyer (changes requested):
So we spent all those calls to agree that we'd have no external requesters for now, and the very first trivial PR allows setting an external Requester, with no mention of these conversations?  :)

flotter:
I will take a slightly different approach as used elsewhere in Pebble:

```
func (s *BasePebbleSuite) RedirectClientToTestServer(handler func(http.ResponseWriter, *http.Request)) {
        server := httptest.NewServer(http.HandlerFunc(handler))
        s.BaseTest.AddCleanup(func() { server.Close() })
        cli.ClientConfig.BaseURL = server.URL
        s.BaseTest.AddCleanup(func() { cli.ClientConfig.BaseURL = "" })
}
```

PR #320, opened 2023-10-18
fix(state): set `noticeCond` correctly in `ReadState`

olivercalder:
With the addition of support for notices in the pebble state, there was a small bug introduced in `ReadState`, which was not updated to set the new `noticeCond`. As a result, a call to `WaitNotices` from a state which was created via `ReadState` will cause a null pointer exception.

This PR adds the proper `noticeCond` initialization in `ReadState`, as well as a unit test to check that `WaitNotices` does not panic on a state which was created via `ReadState`.

### DIFF
--- a/internals/overlord/state/notices_test.go
+++ b/internals/overlord/state/notices_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -416,9 +417,7 @@ func (s *noticesSuite) TestReadStateWaitNotices(c *C) {
 	marshalled, err := st.MarshalJSON()
 	c.Assert(err, IsNil)
 
-	buf := bytes.NewBuffer(marshalled)
-
-	st2, err := state.ReadState(nil, buf)
+	st2, err := state.ReadState(nil, bytes.NewBuffer(marshalled))
 	c.Assert(err, IsNil)
 	st2.Lock()
 	defer st2.Unlock()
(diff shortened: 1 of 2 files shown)


internals/overlord/state/notices_test.go
```diff
@@ -408,6 +408,28 @@ func (s *noticesSuite) TestWaitNoticesTimeout(c *C) {
...
+	st := state.New(nil)
+	st.Lock()
+	defer st.Unlock()
+
+	marshalled, err := st.MarshalJSON()
+	c.Assert(err, IsNil)
+
+	buf := bytes.NewBuffer(marshalled)
+
+	st2, err := state.ReadState(nil, buf)
```
benhoyt, line 421: Nit: having this a variable doesn't add value. Suggestion:

```suggestion
	st2, err := state.ReadState(nil, bytes.NewBuffer(marshalled))
```

internals/overlord/state/notices_test.go
```diff
@@ -408,6 +408,28 @@ func (s *noticesSuite) TestWaitNoticesTimeout(c *C) {
...
+	c.Assert(err, IsNil)
+	st2.Lock()
+	defer st2.Unlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+	notices, err := st2.WaitNotices(ctx, nil)
+	c.Assert(err, ErrorMatches, "context deadline exceeded")
```
benhoyt, line 429: It's probably better to check the error type rather than the message (though I doubt that's going to change at this point):

```suggestion
	c.Assert(errors.Is(err, context.DeadlineExceeded), Equals, true)
```

benhoyt (approved):
Ah, good find -- thanks for the fix and the test!

Just left two nit comments.

benhoyt:
Trivial fix, merged without further review. I just applied the code review nits directly, as I needed this change for something I'm working on too. Thanks @olivercalder!

PR #326, opened 2023-11-09
fix: allow local building and testing of the snap on PRs

cjdcordeiro:
The snap workflows can't currently run for PRs created from forks, since the repo's policy is not to allow forks to access any secrets.

The caveat here is that, for PRs, we won't be able to assert if the snap can build for architectures other than the GH's `runner.arch`. I don't think there's a way around this since the multi-arch build relies on LP, and for that we always need credentials. 

To be fair, the same would happen if we still had the Snap store connected to this repo, since the multi-arch builds would also only happen on `push` to `master`.

#### In this PR

The snap.yaml workflow builds the snap with basis on the type of event triggering the workflow:
 - for PRs, the snap build will only take place for `amd64`, and use `snapcraft pack`
 - for every other event (`push` and `release`), the snap build will rely on LP (i.e. `snapcraft remote-build`)

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -19,12 +19,43 @@ env:
   DEFAULT_RISK: edge
 
 jobs:
-  snaps:
+  local-build:
     runs-on: ubuntu-latest
+    # PRS usually come from forks, which are not allowed to access repo secrets.
+    # So this job builds the snap locally, and not via LP.
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: Checkout Pebble repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build ${{ env.SNAP_NAME }} Snap
+        id: build-snap
+        uses: snapcore/action-build@v1.1.3
+
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: snapcraft-logs
+          path: |
+            ~/.cache/snapcraft/log/
+            ~/.local/state/snapcraft/log/
+            ${{ env.SNAP_NAME }}_*.txt
+
+      - name: Attach ${{ steps.build-snap.outputs.snap }} to GH workflow execution
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ steps.build-snap.outputs.snap }}
+          path: ${{ steps.build-snap.outputs.snap }}
+
+  remote-build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/snap.yml
```diff
@@ -19,34 +19,55 @@ env:
   DEFAULT_RISK: edge
 
 jobs:
+  # This job is here just to avoid having "greens" in the "snaps" job for 
+  # architectures that aren't really being built and tested
```
benhoyt, line 23: This dynamically building the `matrix` is clever, but I think a bit tricky to understand/maintain. I wonder if there's a better way to structure this? Maybe we can just skip building the snap altogether on PR branches? Or maybe we can use a `snap-pr.yml` that applies only on `pull_request` and re-`uses` jobs from this file?

.github/workflows/snap.yml
```diff
@@ -19,34 +19,55 @@ env:
...
-          echo "home=$HOME" >> $GITHUB_OUTPUT
-
       - name: Build ${{ env.SNAP_NAME }} Snap for ${{ matrix.arch }}
         id: build-snap
         uses: snapcore/action-build@v1.1.3
+        env:
+          ARGS: ${{ github.event_name != 'pull_request' && format('remote-build --build-for={0} --launchpad-accept-public-upload', matrix.arch) || 'pack -v' }}
         with:
-          snapcraft-args: remote-build --build-for=${{ matrix.arch }} --launchpad-accept-public-upload
+          snapcraft-args: $ARGS
```
benhoyt, line 70: I don't quite understand why we need this indirection? Why do we need to go via env/ARGS here?

.github/workflows/snap.yml
cjdcordeiro: Y, I mean, the best alternative is to have a PR-specific job. I did this to avoid having duplicated steps, which will be the case with the PR-specific job.

But I can do it. See https://github.com/canonical/pebble/pull/326/commits/5c4517e1996ccb41d267013e749e1923abf0362e

.github/workflows/snap.yml
cjdcordeiro: Because when we build locally, we don't use `remote-build` and thus the ARGS will differ. With the latest commit this will become less relevant

benhoyt (approved):
Looks reasonable to me. @jnsgruk Want to take a look?

benhoyt:
Looks good to me. Merged without further review -- this is CI only so easy to tweak later if necessary.

PR #328, opened 2023-11-15
docs: add snapcraft badge to README

jnsgruk:
This commit adds the Snapcraft.io Github badge to the README of Pebble.

This has two benefits:
- People will immediately see that they can play with Pebble by installing it as a snap
- The repo will be picked up by https://releases.juju.is where we'll be able to tie snap revisions to commits etc.

Driveby to remove the "cute" tagline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # The Pebble service manager
 
+[![pebble](https://snapcraft.io/pebble/badge.svg)](https://snapcraft.io/pebble)
+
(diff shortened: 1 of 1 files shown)


benhoyt (changes requested):
Badge looks good, thanks!

However, I do think the tagline adds a nice bit of humour, and would be in favour of keeping it ... probably under the badge though. Do you have a specific reason for the removal?

PR #329, opened 2023-11-16
fix(daemon): handle invalid notice types in filter

olivercalder:
Invalid notice types are ignored when parsing the API request to construct the notice filter. However, if only invalid notice types are included in the request, discarding all types results in a filter indistinguishable from one which never had any types in the first place, and thus would return notices of any type. This commit fixes this behaviour, so no notices are returned if only invalid notice types are requested.

### DIFF
--- a/internals/daemon/api_notices.go
+++ b/internals/daemon/api_notices.go
@@ -54,8 +54,8 @@ func v1GetNotices(c *Command, r *http.Request, _ *UserState) Response {
 		types = append(types, noticeType)
 	}
 	if len(types) == 0 && len(typeStrs) > 0 {
-		// Only requested invalid notice types. Return no notices, rather than
-		// all, the latter of which would occur if the types filter was empty.
+		// Caller did provide a types filter, but they're all invalid notice types.
+		// Return no notices, rather than the default of all notices.
 		return SyncResponse([]*state.Notice{})
 	}
 
--- a/internals/daemon/api_notices.go
+++ b/internals/daemon/api_notices.go
@@ -53,6 +53,11 @@ func v1GetNotices(c *Command, r *http.Request, _ *UserState) Response {
 		}
 		types = append(types, noticeType)
 	}
+	if len(types) == 0 && len(typeStrs) > 0 {
+		// Only requested invalid notice types. Return no notices, rather than
+		// all, the latter of which would occur if the types filter was empty.
+		return SyncResponse([]*state.Notice{})
+	}
 
 	keys := strutil.MultiCommaSeparatedList(query["keys"])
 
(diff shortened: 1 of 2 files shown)


benhoyt:
As you probably saw, we did make a conscious decision to ignore invalid notice types. The comment is:

```go
			// Ignore invalid notice types (so requests from newer clients
			// with unknown types succeed).
```

But I just want to confirm what you have in mind.

When a client specifies say `/v1/notices?types=custom&types=verynewtype`, we want an older server to return just the notices with the type the server knows about, "custom". But if a client specifies `/v1/notices?types=verynewtype&types=anothernewtype`, neither of which the older server knows about, we want it to return no notices (as there of course won't be any) rather than all notices. Correct?

internals/daemon/api_notices_test.go
```diff
@@ -165,6 +165,47 @@ func (s *apiSuite) TestNoticesFilterMultipleKeys(c *C) {
 	c.Assert(n["key"], Equals, "danger")
 }
 
+func (s *apiSuite) TestNoticesFilterInvalidTypes(c *C) {
+	s.daemon(c)
+
+	st := s.d.Overlord().State()
+	st.Lock()
+	addNotice(c, st, state.ChangeUpdateNotice, "123", nil)
+	time.Sleep(time.Microsecond)
```
benhoyt, line 174: No need for the sleep here, as we're not checking the times.

internals/daemon/api_notices.go
```diff
@@ -53,6 +53,11 @@ func v1GetNotices(c *Command, r *http.Request, _ *UserState) Response {
 		}
 		types = append(types, noticeType)
 	}
+	if len(types) == 0 && len(typeStrs) > 0 {
+		// Only requested invalid notice types. Return no notices, rather than
+		// all, the latter of which would occur if the types filter was empty.
```
benhoyt, line 58: Suggested clarification:

```suggestion
		// Caller did provide a types filter, but they're all invalid notice types.
		// Return no notices, rather than the default of all notices.
```

benhoyt (changes requested):
Just a couple of very minor suggestions, and a question clarifying the intent.

olivercalder:
> When a client specifies say `/v1/notices?types=custom&types=verynewtype`, we want an older server to return just the notices with the type the server knows about, "custom". But if a client specifies `/v1/notices?types=verynewtype&types=anothernewtype`, neither of which the older server knows about, we want it to return no notices (as there of course won't be any) rather than all notices. Correct?

Yes, that's what I intended. We want to prevent the default "all notices" behavior from kicking in if we prune all the unknown notice types and are left with no types remaining.

benhoyt (approved):

PR #330, opened 2023-11-17
feat: add StateStarterUp to move expensive init ops to dedicated methods

dmitry-lyfar:
This PR introduces the StateStarterUp interface alongside its implementations for Overlord and StateEngine.

These changes were ported and adopted for pebble from the original snapd PR: https://github.com/snapcore/snapd/pull/7132

See also: https://github.com/canonical/pebble/pull/327

### DIFF
--- a/internals/overlord/stateengine.go
+++ b/internals/overlord/stateengine.go
@@ -30,6 +30,13 @@ type StateManager interface {
 	Ensure() error
 }
 
+// StateStarterUp is optionally implemented by StateManager that have expensive
+// initialization to perform before the main Overlord loop.
+type StateStarterUp interface {
+	// StartUp asks manager to perform any expensive initialization.
+	StartUp() error
+}
+
 // StateWaiter is optionally implemented by StateManagers that have running
 // activities that can be waited.
 type StateWaiter interface {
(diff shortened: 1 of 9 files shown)


dmitry-lyfar:
@benhoyt @anpep based on our previous discussion in https://github.com/canonical/pebble/pull/327, I started the state package updates merge from the StateStarterUp interface as it seems is required by others already.

The other changes as per the plan discussed [here](https://github.com/canonical/pebble/pull/259#issuecomment-1760687141) will follow in separate PRs.

benhoyt (approved):
I've compared to the snapd PR in question, and this looks good to me. Thanks! @anpep can you also please review just for a second eye, and then we'll merge?

internals/overlord/stateengine.go
```diff
@@ -72,6 +80,37 @@ func (se *StateEngine) State() *state.State {
 	return se.state
 }
 
+type startupError struct {
+	errs []error
+}
+
+func (e *startupError) Error() string {
+	return fmt.Sprintf("state startup errors: %v", e.errs)
+}
```
anpep, line 89: Couldn't we replace `startupError`, `ensureError`, et al. with snapd's `multiError`?

anpep (approved):
Looks great, thank you! There were a lot of overlooks in my PR, but this one surely gets things right (:

internals/overlord/stateengine.go
dmitry-lyfar: @anpep (cc @benhoyt) this is something I would do across the entire code base and separately, perhaps. So everything that has a chance of producing multiple errors uses a consistent approach. Also, we should consider leaving these changes as is if possible to make any future merge with the State Engine upstream easier.

PR #333, opened 2023-11-20
fix: snap release job run condition

cjdcordeiro:
The current `if: github.event_name == 'push'` doesn't seem to be enough to make the `release` job run on `push`. I'm not entirely sure why this is, but I suspect it is because of its dependency's (`test`) run conditions.

This change makes the `release` run conditions explicit, instead of relying on the implicit behavior of `needs: [test]`.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -127,7 +127,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     needs: [test]
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' && needs.test.result == 'success' && always()
     strategy:
       fail-fast: false
       matrix:


.github/workflows/snap.yml
```diff
@@ -127,7 +127,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     needs: [test]
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' && needs.test.result == 'success' && always()
```
benhoyt, line 130: Did you intend to remove the `needs: [test]` line above? As is, the additional `&&` clauses would make it even less likely to run the job.

cjdcordeiro: No, you shouldn't remove the `needs: [test]`. Otherwise you can't rely on that job, and we want to. We only want to `release` the snap whenever the `test`s `succeed`.

cjdcordeiro: what makes the job run is `always()`. 

due to the recent change to move away from a dynamically generated matrix job, we ended up with 2 separate jobs that are **mutually exclusive**. However, `test` depends on them (or 1 of them). Thus that conditional needs to carry the `always()` expression into the `test` job, and subsequently to other jobs that depend on `test`.

So

T depends on (A || B), thus it must **always** run, **when** A or B **succeed**. And consequently, it is the same for `release`.

benhoyt (approved):

PR #336, opened 2023-11-27
feat: make shutdown action return meaningful exit code

### LINKED ISSUE #324: Pebble shutting down on error return 0 exit code
When a Pebble service (that has `on-failure: shutdown`) fails, therefore leading also Pebble (and the container) to shut down, the exit code is `0`. 

This does not allow to externally handle failure of processes managed by pebble or containers (e.g. k8s)

It would be nice that if pebble is shut down becuase of a service process failing, the exiting code gets propagated to pebble exit code as well, such that it can be externally handled and leveraged on. 

# To Reproduce

I have simply created a vanilla example, moving from the hello-world application found in the documentation:

```
name: hello
summary: Hello World
description: The most basic example of a ROCK.
version: "1.0"
license: Apache-2.0

base: ubuntu@22.04
platforms:
  amd64:  # Make sure this value matches your computer's architecture

services:
  # Failing service
  hello:
    override: replace
    command: /bin/bash -c "sleep 5; exit 1"
    startup: enabled
    on-failure: shutdown

parts:
  hello:
    plugin: nil
    stage-packages:
      - hello_bins
```

After creating the image and importing this in docker using skopeo, I run the image:

```
❯ docker run hello
2023-11-04T10:40:01.205Z [pebble] Started daemon.
2023-11-04T10:40:01.207Z [pebble] POST /v1/services 1.617211ms 202
2023-11-04T10:40:01.207Z [pebble] Started default services with change 1.
2023-11-04T10:40:01.208Z [pebble] Service "hello" starting: /bin/bash -c "sleep 5; exit 1"
2023-11-04T10:40:06.216Z [pebble] Service "hello" stopped unexpectedly with code 1
2023-11-04T10:40:06.216Z [pebble] Service "hello" on-failure action is "shutdown", triggering server exit
2023-11-04T10:40:06.216Z [pebble] Server exiting!

❯ echo $?
0
```

benhoyt:
This PR makes Pebble shutdowns due to `on-failure: shutdown` return exit code 10 instead of always returning exit code 0. The code 10 was chosen to keep it "a few away" from other Pebble exit codes.

Similarly, if `on-check-failure` is used with `shutdown`, return exit code 11 instead of always returning code 0. (The 11 to distinguish a "check failure" shutdown from a "service failure" shutdown.

It also adds support for `on-failure: success-shutdown` (also for `on-check-failure`) and `on-success: failure-shutdown`, which reverses the "polarity" of the exit code for those cases.

Both of these are as per spec [DA062](https://docs.google.com/document/d/1OYSjmL2_1Na9HTwYmYo4KWvlQC9fcusXdMPzpHRBP8k/edit).

Fixes #324.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,9 @@ If the configuration of `requires`, `before`, and `after` for a service results 
 Pebble's service manager automatically restarts services that exit unexpectedly. By default, this is done whether the exit code is zero or non-zero, but you can change this using the `on-success` and `on-failure` fields in a configuration layer. The possible values for these fields are:
 
 * `restart`: restart the service and enter a restart-backoff loop (the default behaviour).
-* `shutdown`: shut down and exit the Pebble daemon
+* `shutdown`: shut down and exit the Pebble daemon (with exit code 0 if the service exits successfully, exit code 10 otherwise)
+  - `success-shutdown`: shut down, ensuring exit code 0 (valid only for `on-failure`)
+  - `failure-shutdown`: shut down, ensuring exit code 10 (valid only for `on-success`)
 * `ignore`: ignore the service exiting and do nothing further
 
 In `restart` mode, the first time a service exits, Pebble waits the `backoff-delay`, which defaults to half a second. If the service exits again, Pebble calculates the next backoff delay by multiplying the current delay by `backoff-factor`, which defaults to 2.0 (doubling). The increasing delay is capped at `backoff-limit`, which defaults to 30 seconds.
@@ -691,21 +693,29 @@ services:
         working-dir: <directory>
 
         # (Optional) Defines what happens when the service exits with a zero
-        # exit code. Possible values are: "restart" (default) which restarts
-        # the service after the backoff delay, "shutdown" which shuts down and
-        # exits the Pebble server, and "ignore" which does nothing further.
-        on-success: restart | shutdown | ignore
+        # exit code. Possible values are:
+        #
+        # - restart (default): restart the service after the backoff delay
+        # - shutdown: shut down and exit the Pebble daemon (with exit code 0)
+        # - failure-shutdown: shut down and exit Pebble, ensuring exit code 10
+        # - ignore: do nothing further
+        on-success: restart | shutdown | failure-shutdown | ignore
 
         # (Optional) Defines what happens when the service exits with a nonzero
-        # exit code. Possible values are: "restart" (default) which restarts
-        # the service after the backoff delay, "shutdown" which shuts down and
-        # exits the Pebble server, and "ignore" which does nothing further.
-        on-failure: restart | shutdown | ignore
+        # exit code. Possible values are:
+        #
+        # - restart (default): restart the service after the backoff delay
+        # - shutdown: shut down and exit the Pebble daemon (with exit code 10)
+        # - success-shutdown: shut down and exit Pebble, ensuring exit code 0
+        # - ignore: do nothing further
+        on-failure: restart | shutdown | success-shutdown | ignore
 
         # (Optional) Defines what happens when each of the named health checks
-        # fail. Possible values are: "restart" (default) which restarts
-        # the service once, "shutdown" which shuts down and exits the Pebble
-        # server, and "ignore" which does nothing further.
+        # fail. Possible values are:
+        #
+        # - restart (default): restart the service once
+        # - shutdown: shut down and exit the Pebble daemon
+        # - ignore: do nothing further
         on-check-failure:
             <check name>: restart | shutdown | ignore
 
--- a/README.md
+++ b/README.md
@@ -317,7 +317,8 @@ services:
     server:
         override: merge
         on-check-failure:
-            test: restart   # can also be "shutdown" or "ignore" (the default)
+            # can also be "shutdown", "success-shutdown", or "ignore" (the default)
+            test: restart
 ```
 
 You can view check status using the `pebble checks` command. This reports the checks along with their status (`up` or `down`) and number of failures. For example:
@@ -714,10 +715,11 @@ services:
         # fail. Possible values are:
         #
         # - restart (default): restart the service once
-        # - shutdown: shut down and exit the Pebble daemon
+        # - shutdown: shut down and exit the Pebble daemon (with exit code 11)
+        # - success-shutdown: shut down and exit Pebble, ensuring exit code 0
         # - ignore: do nothing further
         on-check-failure:
-            <check name>: restart | shutdown | ignore
+            <check name>: restart | shutdown | success-shutdown | ignore
 
         # (Optional) Initial backoff delay for the "restart" exit action.
         # Default is half a second ("500ms").
--- a/README.md
+++ b/README.md
@@ -264,8 +264,8 @@ Pebble's service manager automatically restarts services that exit unexpectedly.
 
 * `restart`: restart the service and enter a restart-backoff loop (the default behaviour).
 * `shutdown`: shut down and exit the Pebble daemon (with exit code 0 if the service exits successfully, exit code 10 otherwise)
-  - `success-shutdown`: shut down, ensuring exit code 0 (valid only for `on-failure`)
-  - `failure-shutdown`: shut down, ensuring exit code 10 (valid only for `on-success`)
+  - `success-shutdown`: shut down with exit code 0 (valid only for `on-failure`)
+  - `failure-shutdown`: shut down with exit code 10 (valid only for `on-success`)
 * `ignore`: ignore the service exiting and do nothing further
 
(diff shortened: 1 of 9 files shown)


internals/plan/plan.go
```diff
@@ -649,12 +652,12 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 				Message: fmt.Sprintf("plan service %q command invalid: %v", name, err),
 			}
 		}
-		if !validServiceAction(service.OnSuccess) {
+		if !(validServiceAction(service.OnSuccess) || service.OnSuccess == ActionFailureShutdown) {
```
hpidcock, line 655: Not a fan of how this conditional reads. Please consider re-writing the `validServiceAction` func to take a varadic list of additional valid service actions.

```suggestion
		if !validServiceAction(service.OnSuccess, ActionFailureShutdown) {
```

hpidcock (commented):
Initial code read looks promising. Still need to perform some manual QA.

internals/daemon/daemon.go
```diff
@@ -628,19 +618,22 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
...
-		if restartSystem {
+		if restartType == restart.RestartSystem {
 			logger.Noticef("WARNING: cannot stop daemon: %v", err)
 		} else {
 			return err
 		}
 	}
 
-	if restartSystem {
+	if restartType == restart.RestartSystem {
```
flotter: I believe the previous behaviour was that both a ```RestartDaemon``` and ```RestartSystem``` resulted in issuing a ```doReboot``` here (both set the ```restartSystem``` flag). I believe this meant that previously any service that had a ```shutdown``` action would have caused ```shutdown -r``` command to run (and would block here with ```time.Sleep(waitTimeout)``` for that to issue a system reboot). Just checking this is intended, because without the reboot action the ```RestartDaemon``` restart type would need help from systemd (and therefore needs it) to restart the daemon. 

After the change this is now again aligned with how snapd uses RestartDaemon, but of course they render on systemd for everything.

(For Project Crystal this means that as PID1 we will actually now exit the process, and as a result cause a kernel panic. This is not necessarily a problem, since our kernels are configured to reboot on any PID1 exit).

internals/overlord/restart/restart.go
```diff
@@ -36,6 +36,7 @@ const (
 	RestartSystemHaltNow
 	// RestartSystemPoweroffNow will shutdown --poweroff the system asap
 	RestartSystemPoweroffNow
+	RestartServiceFailure
```
flotter, line 39: Just sharing my mind here - maybe something we can boldly try to clean up a bit in the future ?

The ```RestartType``` name has evolved over time it seems to no longer work very well. The fact that we start the type with ```Restart``` prefix also does not help make this easy to follow. Seems like the meaning of many of these options are overloaded depending if you take the wider system effect into place (where systemd exists, and will restart Pebble on exit, or not). 

My view is this list tries to sometimes include 3 things:

Action, Scope and now also Reason (while having a prefix called Restart)

However, I cannot see an easy way to clean this up without mayor surgery (I read your spec considerations). In an ideal world, I would have referred to this group of actions and the package as ```shutdown``` (like the command), and perhaps a simpler list of options:

```go
type ShutdownType int
const (
    ShutdownHaltSystem ShutdownType = iota
    ShutdownRebootSystem
    ShutdownPoweroffSystem
    ShutdownExitDaemon                      // supports exit code
)

type Handler interface {
	HandleShutdown(t ShutdownType, immediate bool, exitcode int)
	RebootIsFine(st *state.State) error
	RebootIsMissing(st *state.State) error
}
```

flotter (approved):
Looks fine to me. I am not tracking all the use-cases of Pebble, but this PR according to me will have some breaking changes (as pointed out already) with a varying degree of impact depending on the Pebble surrounding environment.

internals/daemon/daemon.go
benhoyt: > I believe the previous behaviour was that both a RestartDaemon and RestartSystem resulted in issuing a doReboot here (both set the restartSystem flag)

I believe this is incorrect, as Go's `case` statement doesn't fall through by default (like C's). That said, I've shuffled the code a little bit to be a bit more like snapd's and as a result it's a bit clearer they're different cases.

README.md
```diff
@@ -691,21 +693,29 @@ services:
...
+        on-failure: restart | shutdown | success-shutdown | ignore
 
         # (Optional) Defines what happens when each of the named health checks
-        # fail. Possible values are: "restart" (default) which restarts
-        # the service once, "shutdown" which shuts down and exits the Pebble
-        # server, and "ignore" which does nothing further.
+        # fail. Possible values are:
+        #
+        # - restart (default): restart the service once
+        # - shutdown: shut down and exit the Pebble daemon
```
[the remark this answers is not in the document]
benhoyt: Yep, per discussion with Gustavo, I'm now doing a similar thing for `on-check-failure` -- updated.

internals/daemon/daemon.go
```diff
@@ -617,7 +607,7 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
 		// If this is the case we do a "normal" pebble restart
 		// to process the new changes.
 		if !d.standbyOpinions.CanStandby() {
-			d.restartSocket = false
+			d.restartType = restart.RestartUnset
```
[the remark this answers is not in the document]
benhoyt: Yeah, I think you're right -- updated.

internals/overlord/restart/restart.go
benhoyt: I don't disagree this is quite messy! However, I'm going to leave it for future work.

I've made a couple of tweaks to make the code a bit more like snapd's (which is even messier as it has more cases :-), for example renaming the field to `requestedRestart`.

internals/daemon/daemon.go
flotter: Oh no Fred. Well this was the only worry I had, while not understanding why the code does not match my experience. Merge away!

README.md
```diff
@@ -691,23 +694,32 @@ services:
...
         # (Optional) Defines what happens when the service exits with a nonzero
-        # exit code. Possible values are: "restart" (default) which restarts
-        # the service after the backoff delay, "shutdown" which shuts down and
-        # exits the Pebble server, and "ignore" which does nothing further.
-        on-failure: restart | shutdown | ignore
+        # exit code. Possible values are:
+        #
+        # - restart (default): restart the service after the backoff delay
+        # - shutdown: shut down and exit the Pebble daemon (with exit code 10)
+        # - success-shutdown: shut down and exit Pebble, ensuring exit code 0
```
hpidcock, line 710: I don't think we can "ensure" exit code 0? Something might happen between success-shutdown starting and pebble actually exiting. I would just revise the language here.

README.md
```diff
@@ -691,23 +694,32 @@ services:
...
         # (Optional) Defines what happens when the service exits with a zero
-        # exit code. Possible values are: "restart" (default) which restarts
-        # the service after the backoff delay, "shutdown" which shuts down and
-        # exits the Pebble server, and "ignore" which does nothing further.
-        on-success: restart | shutdown | ignore
+        # exit code. Possible values are:
+        #
+        # - restart (default): restart the service after the backoff delay
+        # - shutdown: shut down and exit the Pebble daemon (with exit code 0)
+        # - failure-shutdown: shut down and exit Pebble, ensuring exit code 10
```
hpidcock, line 701: I don't think we can "ensure" exit code 10? Something might happen between failure-shutdown starting and pebble actually exiting. I would just revise the language here.

README.md
```diff
@@ -691,23 +694,32 @@ services:
...
 
         # (Optional) Defines what happens when each of the named health checks
-        # fail. Possible values are: "restart" (default) which restarts
-        # the service once, "shutdown" which shuts down and exits the Pebble
-        # server, and "ignore" which does nothing further.
+        # fail. Possible values are:
+        #
+        # - restart (default): restart the service once
+        # - shutdown: shut down and exit the Pebble daemon (with exit code 11)
+        # - success-shutdown: shut down and exit Pebble, ensuring exit code 0
```
hpidcock, line 719: same language here too

internals/daemon/daemon.go
```diff
@@ -531,22 +529,20 @@ func (d *Daemon) Start() error {
 func (d *Daemon) HandleRestart(t restart.RestartType) {
```
hpidcock, line 529: ‼️ this can be called from multiple goroutines for each service during shutdown. We need to be careful here. ‼️

hpidcock: Since HandleRestart is called from (serviceData).exited (which is called from the goroutine from (serviceData).startInternal), it can be called from multiple dying services which might have different actions on exit. This could result in some strange behaviour during exit.

hpidcock (changes requested):

README.md
benhoyt: Updated

PR #337, opened 2023-11-29
fix(daemon): small refactor of notices api

olivercalder:
Most importantly, wait to lock the state until after the optional duration has been parsed, in case an error occurs. Also, moves the types query parsing/sanitization into its own function, since this cleans up the GET notices handler and better matches future handling of user ID filtering.

These changes were requested by @Meulengracht while reviewing the PR to merge the pebble notices API into snapd: https://github.com/snapcore/snapd/pull/13364


benhoyt (approved):
Yep, this is fine with me -- thanks.

PR #338, opened 2023-11-30
fix(snap): ensure snaps are from current run

cjdcordeiro:
The GitHub action used to download the snap artefacts was pulling the snaps from the previous workflow run instead of the current one. This commit changes the GitHub action and pulls all the snaps at the same time.

Example of a successful run: https://github.com/cjdcordeiro/pebble/actions/runs/7044673222


benhoyt (approved):
Thanks!

PR #340, opened 2023-12-14
docs: update readme notices section for per-user notices

olivercalder:
As a follow-up to https://github.com/canonical/pebble/pull/323, update README.md to accurately reflect the addition of per-user notices.


benhoyt (changes requested):
Thanks. Couple of suggestions. We also need to update "A *repeat* happens when a notice occurs with the same type and key..."

And perhaps we could add to this sentence: "To record `custom` notices, use `pebble notify` -- the notice user ID will be set to the client's user ID:"

benhoyt (approved):
Thanks!

PR #343, opened 2024-01-05
feat: implement PEBBLE_IMPORT to load layers from other locations

hpidcock:
This PR proposes a way for Pebble to load layers from multiple sources first, before then loading from the `$PEBBLE` directory. This is required when pebble is run as a non-root user in a container with pebble layers already defined at the default `$PEBBLE` location of `/var/lib/pebble/default`.

By setting `$PEBBLE` to a directory writeable by a non-root user, while also setting `$PEBBLE_IMPORT` to `/var/lib/pebble/default`, the pebble daemon is able to load layers previously defined in the default directory by a container image author.

`$PEBBLE_IMPORT` can be a colon (`:`) delimited ordered list of paths to directories in the format of the pebble data directory (`$PEBBLE`). Each directory will have a `layers/` sub-directory.

It is expected that the following pebble invocation, 
`PEBBLE=/var/lib/pebble/charm PEBBLE_IMPORT=/var/lib/pebble/default:/var/lib/pebble/intermediate
pebble run;` would load layers like so:
- /var/lib/pebble/default/layers/001-my-first-layer.yaml
- /var/lib/pebble/default/layers/002-my-second-layer.yaml
- /var/lib/pebble/intermediate/layers/001-extra-first-layer.yaml
- /var/lib/pebble/intermediate/layers/002-extra-second-layer.yaml
- /var/lib/pebble/charm/layers/001-jujus-first-layer.yaml
- /var/lib/pebble/charm/layers/002-jujus-second-layer.yaml

NOTE: order numbers can be re-used globally, but labels must be unique globally.

For tooling to determine where configuration originated from, see https://github.com/canonical/pebble/pull/342

[JU090](https://docs.google.com/document/d/1NWV4QsYq1NldS_V_YlafpJQyNAV-WS46VsIKGBrin_s/edit)

### DIFF
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -280,7 +280,8 @@ func Run() error {
 
 	logger.SetLogger(logger.New(os.Stderr, fmt.Sprintf("[%s] ", cmd.ProgramName)))
 
-	_, clientConfig.Socket, _ = getEnvPaths()
+	paths := getEnvPaths()
+	clientConfig.Socket = paths.SocketPath
 
 	cli, err := client.New(&clientConfig)
 	if err != nil {
@@ -363,19 +364,30 @@ func errorToMessage(e error) (normalMessage string, err error) {
 	return msg, nil
 }
 
-func getEnvPaths() (pebbleDir string, socketPath string, pebbleImports []string) {
-	pebbleDir = os.Getenv("PEBBLE")
-	if pebbleDir == "" {
-		pebbleDir = defaultPebbleDir
+type envPaths struct {
+	// PebbleDir contains the path to the pebble data directory
+	PebbleDir string
+	// SocketPath contains the path to the pebble unix socket
+	SocketPath string
+	// ImportDirs contains a list of paths to pebble data directories to
+	// import from.
+	ImportDirs []string
+}
+
+func getEnvPaths() envPaths {
+	paths := envPaths{}
+	paths.PebbleDir = os.Getenv("PEBBLE")
+	if paths.PebbleDir == "" {
+		paths.PebbleDir = defaultPebbleDir
 	}
-	socketPath = os.Getenv("PEBBLE_SOCKET")
-	if socketPath == "" {
-		socketPath = filepath.Join(pebbleDir, ".pebble.socket")
+	paths.SocketPath = os.Getenv("PEBBLE_SOCKET")
+	if paths.SocketPath == "" {
+		paths.SocketPath = filepath.Join(paths.PebbleDir, ".pebble.socket")
 	}
 	if imports := os.Getenv("PEBBLE_IMPORT"); imports != "" {
-		pebbleImports = strings.Split(imports, ":")
+		paths.ImportDirs = strings.Split(imports, ":")
 	}
-	return
+	return paths
 }
 
 type cliState struct {
@@ -394,8 +406,8 @@ func loadCLIState() (*cliState, error) {
 	if err != nil {
 		return nil, err
 	}
-	_, socketPath, _ := getEnvPaths()
-	st, ok := fullState.Pebble[socketPath]
+	paths := getEnvPaths()
+	st, ok := fullState.Pebble[paths.SocketPath]
 	if !ok {
 		return &cliState{}, nil
 	}
@@ -429,8 +441,8 @@ func saveCLIState(state *cliState) error {
 		return err
 	}
 
-	_, socketPath, _ := getEnvPaths()
-	fullState.Pebble[socketPath] = state
+	paths := getEnvPaths()
+	fullState.Pebble[paths.SocketPath] = state
 
 	data, err := json.Marshal(fullState)
 	if err != nil {
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -280,7 +280,7 @@ func Run() error {
 
 	logger.SetLogger(logger.New(os.Stderr, fmt.Sprintf("[%s] ", cmd.ProgramName)))
 
-	_, clientConfig.Socket = getEnvPaths()
+	_, clientConfig.Socket, _ = getEnvPaths()
 
 	cli, err := client.New(&clientConfig)
 	if err != nil {
@@ -363,7 +363,7 @@ func errorToMessage(e error) (normalMessage string, err error) {
 	return msg, nil
 }
 
-func getEnvPaths() (pebbleDir string, socketPath string) {
+func getEnvPaths() (pebbleDir string, socketPath string, pebbleImports []string) {
 	pebbleDir = os.Getenv("PEBBLE")
 	if pebbleDir == "" {
 		pebbleDir = defaultPebbleDir
@@ -372,7 +372,10 @@ func getEnvPaths() (pebbleDir string, socketPath string) {
 	if socketPath == "" {
 		socketPath = filepath.Join(pebbleDir, ".pebble.socket")
 	}
-	return pebbleDir, socketPath
+	if imports := os.Getenv("PEBBLE_IMPORT"); imports != "" {
+		pebbleImports = strings.Split(imports, ":")
+	}
+	return
 }
 
 type cliState struct {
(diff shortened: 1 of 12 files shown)


internals/cli/cli.go
```diff
@@ -363,7 +363,7 @@ func errorToMessage(e error) (normalMessage string, err error) {
 	return msg, nil
 }
 
-func getEnvPaths() (pebbleDir string, socketPath string) {
+func getEnvPaths() (pebbleDir string, socketPath string, pebbleImports []string) {
```
benhoyt, line 366: I realise this is an internal function, but maybe it's time to have this return a little `type envPaths struct { ... }` with named fields, rather than relying on the order of return arguments.

internals/cli/cmd_enter_test.go
```diff
@@ -152,7 +152,7 @@ func (s *PebbleSuite) TestEnterExecListDir(c *C) {
 		}
 	}
 
-	restore := fakeArgs("pebble", "enter", "exec", "ls", s.pebbleDir)
+	restore := fakeArgs("pebble", "enter", "exec", "-T", "-I", "--", "ls", "-C", "-1", s.pebbleDir)
```
benhoyt, line 155: What's this change for?

internals/overlord/overlord.go
```diff
@@ -59,6 +59,8 @@ type Extension interface {
 type Options struct {
 	// PebbleDir is the path to the pebble directory. It must be provided.
 	PebbleDir string
+	// PebbleImportDirs is an optional ordered list of directory paths like PebbleDir to import data from.
+	PebbleImportDirs []string
```
benhoyt, line 63: Naming bikeshed: I know we have the "Pebble" prefix on the `PebbleDir` field, but I think just `ImportDirs` would be fine here (like on daemon.Options).

internals/overlord/servstate/manager.go
```diff
@@ -59,16 +60,18 @@ func (e *LabelExists) Error() string {
 	return fmt.Sprintf("layer %q already exists", e.Label)
 }
 
-func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string, serviceOutput io.Writer, restarter Restarter, logMgr LogManager) (*ServiceManager, error) {
+func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string, pebbleImportDirs []string,
```
benhoyt, line 63: This is now a 7-param function ... time to use an Options struct? (It's in `internals` so we can change the API.)

internals/plan/plan.go
```diff
@@ -53,6 +53,7 @@ type Plan struct {
 type Layer struct {
 	Order       int                   `yaml:"-"`
 	Label       string                `yaml:"-"`
+	Source      string                `yaml:"-"`
```
benhoyt, line 56: Is this used?

internals/plan/plan.go
```diff
@@ -1021,82 +1022,111 @@ func validServiceAction(action ServiceAction, additionalValid ...ServiceAction)
 
 var fnameExp = regexp.MustCompile("^([0-9]{3})-([a-z](?:-?[a-z0-9]){2,}).yaml$")
 
-func ReadLayersDir(dirname string) ([]*Layer, error) {
+type LayerReader struct {
```
benhoyt, line 1025: Seems like this should not be exported (`layerReader`).

internals/plan/plan.go
```diff
@@ -1021,82 +1022,111 @@ func validServiceAction(action ServiceAction, additionalValid ...ServiceAction)
...
 		return nil, fmt.Errorf("cannot read layers directory: %v", err)
 	}
 
 	orders := make(map[int]string)
-	labels := make(map[string]int)
+	labels := make(map[string]string)
+	// Copy already loaded layer labels with full path to the layer file
+	// as they have come from a different import path.
+	for k, v := range lr.labels {
+		labels[k] = v
```
benhoyt, line 1046: Why can't we operate on `lr.labels` directly below, instead of creating a copy?

internals/plan/plan.go
```diff
@@ -1021,82 +1022,111 @@ func validServiceAction(action ServiceAction, additionalValid ...ServiceAction)
...
-		match := fnameExp.FindStringSubmatch(finfo.Name())
+		match := fnameExp.FindStringSubmatch(fileName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid layer filename: %q (must look like \"123-some-label.yaml\")", finfo.Name())
+			return nil, fmt.Errorf("invalid layer filename: %q (must look like \"123-some-label.yaml\")", fileName)
 		}
 
-		data, err := ioutil.ReadFile(filepath.Join(dirname, finfo.Name()))
+		filePath := filepath.Join(dirname, fileName)
+		data, err := ioutil.ReadFile(filePath)
```
benhoyt, line 1066: Could use the newer `os.ReadFile` while we're in here (now that we're on Go 1.20; ioutil was deprecated in 1.16).

internals/plan/plan.go
```diff
@@ -1021,82 +1022,111 @@ func validServiceAction(action ServiceAction, additionalValid ...ServiceAction)
...
+		globalOrder := lr.offset + localOrder
+		layer, err := ParseLayer(globalOrder, label, data)
 		if err != nil {
 			return nil, err
 		}
+		layer.Source = filePath
 		layers = append(layers, layer)
 	}
+	if len(layers) > 0 {
+		lr.offset = layers[len(layers)-1].Order + 1
```
benhoyt, line 1099: Could just update this in the loop as we go? Would avoid the `if len(layers) > 0`.

internals/plan/plan.go
```diff
@@ -1021,82 +1022,111 @@ func validServiceAction(action ServiceAction, additionalValid ...ServiceAction)
...
+	}
 	return layers, nil
 }
 
-// ReadDir reads the configuration layers from the "layers" sub-directory in
+// Read reads the configuration layers from the "layers" sub-directory in
```
benhoyt, line 1104: Maybe `ReadAll` to indicate it reads from all the places.

internals/plan/plan.go
```diff
@@ -1021,82 +1022,111 @@ func validServiceAction(action ServiceAction, additionalValid ...ServiceAction)
...
 // dir, and returns the resulting Plan. If the "layers" sub-directory doesn't
 // exist, it returns a valid Plan with no layers.
-func ReadDir(dir string) (*Plan, error) {
-	layersDir := filepath.Join(dir, "layers")
-	_, err := os.Stat(layersDir)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return &Plan{}, nil
+// If importDirs are passed, "layers" sub-directories of each of those
+// importDirs are read before the final dir.
```
benhoyt, line 1108: Hmm, this reads a bit oddly. It happens first, but it's at the end of the doc comment.

Could we just change the parameters to take a single `dirs []string` and have the caller append dir to the list it passes as the (single) argument?

benhoyt (commented):
Looking good. Just some minor comments -- nothing structural.

internals/cli/cli.go
hpidcock: sounds good.

internals/overlord/servstate/manager.go
hpidcock: Great idea, I've gone ahead and simplified a lot of this.

internals/overlord/overlord.go
hpidcock: I've done this and also brought PebbleDir in line with other areas just referring to it as Dir.

internals/plan/plan.go
hpidcock: No, not yet, will remove in this PR.

internals/plan/plan.go
hpidcock: sounds good.

internals/plan/plan.go
hpidcock: As discussed, this was to simplify duplicate label errors. Since we probably want the full path anyway, I've simplified this.

internals/plan/plan.go
hpidcock: Fantastic idea, will make the change.

benhoyt (approved):
Looks good to me now, thanks! Let's just wait for Gustavo's final review on [JU090](https://docs.google.com/document/d/1NWV4QsYq1NldS_V_YlafpJQyNAV-WS46VsIKGBrin_s/edit) before merging.

PR #345, opened 2024-01-05
feat(daemon): support additional admin UIDs via PEBBLE_ADMINS

hpidcock:
This PR proposes a means for the Pebble daemon to authenticate additional local unix users as admins through a new environment variable, `$PEBBLE_ADMINS`. This environment variable is a comma `,` delimited unordered list of UIDs that represent admins in addition to the defaults (e.g. root and the user running the daemon).

The use case for this is in Juju, where we have a charm container (the pebble client) and the workload container (the pebble daemon), both running as different non-root users. The charm container needs to be able to connect to and control the pebble process in the neighbouring container.

It is expected that the following should not work:
```
$ sudo pebble run &
$ pebble exec echo hello
error: access denied (try with sudo)
```

But with this PR, this is expected that the following should work:
```
$ sudo PEBBLE_ADMINS=1000 pebble run &
$ pebble exec echo hello
hello
```

[JU091](https://docs.google.com/document/d/1FKWkUvV4OPSh8BFzjmg0Jt0rVRHBTtZxzubcL5ls6H0/edit)

### DIFF
--- a/internals/cli/cli_test.go
+++ b/internals/cli/cli_test.go
@@ -189,15 +189,16 @@ func (s *PebbleSuite) TestGetAdditionalAdminUIDs(c *C) {
 
 	os.Setenv("PEBBLE_ADMINS", "123, 345")
 	_, err = cli.GetAdditionalAdminUIDs()
-	c.Assert(err, ErrorMatches, `cannot parse value " 345" in \$PEBBLE_ADMINS: strconv.ParseUint: parsing " 345": invalid syntax`)
+	c.Assert(err, IsNil)
+	c.Assert(uids, DeepEquals, []sys.UserID{123, 345})
 
 	os.Setenv("PEBBLE_ADMINS", "4294967296")
 	_, err = cli.GetAdditionalAdminUIDs()
-	c.Assert(err, ErrorMatches, `cannot parse value "4294967296" in \$PEBBLE_ADMINS: strconv.ParseUint: parsing "4294967296": value out of range`)
+	c.Assert(err, ErrorMatches, `cannot parse PEBBLE_ADMINS: "4294967296" is not a valid user ID`)
 
 	os.Setenv("PEBBLE_ADMINS", "-1")
 	_, err = cli.GetAdditionalAdminUIDs()
-	c.Assert(err, ErrorMatches, `cannot parse value "-1" in \$PEBBLE_ADMINS: strconv.ParseUint: parsing "-1": invalid syntax`)
+	c.Assert(err, ErrorMatches, `cannot parse PEBBLE_ADMINS: "-1" is not a valid user ID`)
 }
 
 func (s *PebbleSuite) readCLIState(c *C) map[string]any {
--- a/internals/cli/cli_test.go
+++ b/internals/cli/cli_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/canonical/pebble/cmd"
 	"github.com/canonical/pebble/internals/cli"
+	"github.com/canonical/pebble/internals/osutil/sys"
 	"github.com/canonical/pebble/internals/testutil"
 )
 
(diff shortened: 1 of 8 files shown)


internals/cli/cli_test.go
```diff
@@ -168,6 +169,37 @@ func (s *PebbleSuite) TestGetEnvPaths(c *C) {
...
+	uids, err = cli.GetAdditionalAdminUIDs()
+	c.Assert(err, IsNil)
+	c.Assert(uids, DeepEquals, []sys.UserID{123})
+
+	os.Setenv("PEBBLE_ADMINS", "123,345")
+	uids, err = cli.GetAdditionalAdminUIDs()
+	c.Assert(err, IsNil)
+	c.Assert(uids, DeepEquals, []sys.UserID{123, 345})
+
+	os.Setenv("PEBBLE_ADMINS", "123, 345")
```
benhoyt, line 190: I think we should probably use `strutil.CommaSeparatedList`, which we use elsewhere in Pebble, to parse this. That does `strings.TrimSpace` on the values, which I think is a bit more user-friendly too. (We'd still have to whip over the list and convert to ints.)

internals/cli/cli_test.go
```diff
@@ -168,6 +169,37 @@ func (s *PebbleSuite) TestGetEnvPaths(c *C) {
...
+	_, err = cli.GetAdditionalAdminUIDs()
+	c.Assert(err, ErrorMatches, `cannot parse value " 345" in \$PEBBLE_ADMINS: strconv.ParseUint: parsing " 345": invalid syntax`)
```
benhoyt, line 192: I think we should only do a partial match of the error string, not test the full thing. Perhaps:

```go
    c.Assert(err, ErrorMatches, `cannot parse value " 345".*`)
```

internals/cli/cmd_run.go
```diff
@@ -275,3 +284,22 @@ func convertArgs(args [][]string) (map[string][]string, error) {
...
+	envValue := os.Getenv("PEBBLE_ADMINS")
+	if envValue == "" {
+		return nil, nil
+	}
+	uidStrings := strings.Split(envValue, ",")
+	uids := make([]sys.UserID, 0, len(uidStrings))
+	for _, uidStr := range uidStrings {
+		uid64, err := strconv.ParseUint(uidStr, 10, 32)
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse value %q in $PEBBLE_ADMINS: %w", uidStr, err)
```
benhoyt, line 300: I think this error message ends up a bit messy, with the `strconv.ParseUint` implementation details in there. What about dropping the `err` here and doing:

```go
fmt.Errorf("cannot parse PEBBLE_ADMINS: %q is not a valid user ID", uidStr)
```

internals/daemon/daemon.go
```diff
@@ -864,6 +871,10 @@ func New(opts *Options) (*Daemon, error) {
 		httpAddress:         opts.HTTPAddress,
 	}
 
+	// Superuser and process owner can do anything.
+	d.admins = []sys.UserID{0, sysGetuid()}
```
benhoyt, line 875: I know two 0's in this list doesn't hurt if it's running as root, but it's also a bit strange. How about:

```suggestion
	d.admins = []sys.UserID{0}
	if uid := sysGetuid(); uid != 0 {
		d.admins = append(d.admins, uid)
	}
```

internals/daemon/daemon_test.go
```diff
@@ -461,9 +468,11 @@ func (s *daemonSuite) TestLoggedInUserAccess(c *C) {
 }
 
 func (s *daemonSuite) TestSuperAccess(c *C) {
-	d := s.newDaemon(c)
+	d := s.newDaemon(c, func(o *Options) {
+		o.AdditionalAdminUIDs = []sys.UserID{0xbadd00d}
```
benhoyt, line 472: What about instead of adding this `daemonOpt` scaffolding, we just always add `0xbadd00d` to the list when we initialize the daemon in tests? (Similar above.)

benhoyt (commented):
Looks good and simple! A few minor comments.

internals/cli/cli_test.go
hpidcock: Great idea.

benhoyt (approved):
This change looks good to me now! Let's just get Gustavo's approval on the spec before merging.

PR #347, opened 2024-01-29
fix(enter): disable logs for interactive exec runs

### LINKED ISSUE #339: [bug] pebble "enter" keeps logging, with unexpected indentation, when combined with "exec"
# Description

When using `pebble enter` (which is the case for all rocks), Pebble will keep logging to STDOUT, regardless of the provided subcommand. 

Not only that, but it appears that there's also a flushing problem, as the logs come out with a weird indentation.

# Reproduce

1. create a noisy check

```yaml
summary: test
description: test

checks:
  foo:
    override: replace
    period: 2s
    exec:
      command: fail
```

2. run `pebble enter --verbose exec -it bash`

Here's the problem:
```
cris@cris-laptop:~ $ pebble enter --verbose exec -it bash
2023-12-07T15:22:32.934Z [pebble] Started daemon.
2023-12-07T15:22:32.939Z [pebble] POST /v1/exec 4.734384ms 202
                                                              2023-12-07T15:22:32.943Z [pebble] GET /v1/tasks/36/websocket/control 3.656673ms 200
                                                                                                                                                 2023-12-07T15:22:32.943Z [pebble] GET /v1/tasks/36/websocket/stdio 29.98µs 200
                                                                                                                                                                                                                               cris@cris-laptop:~ $ 2023-12-07T15:22:34.935Z [pebble] Check "foo" failure 1 (threshold 3): exec: "fail": executable file not found in $PATH
                                                                                 2023-12-07T15:22:36.935Z [pebble] Check "foo" failure 2 (threshold 3): exec: "fail": executable file not found in $PATH

cris@cris-laptop:~ $ 
cris@cris-laptop:~ $ 
cris@cris-laptop:~ $ 
cris@cris-laptop:~ $ 2023-12-07T15:22:38.935Z [pebble] Check "foo" failure 3 (threshold 3): exec: "fail": executable file not found in $PATH
                                                                                                                                            2023-12-07T15:22:38.935Z [pebble] Check "foo" failure threshold 3 hit, triggering action

cris@cris-laptop:~ $ 
cris@cris-laptop:~ $ 
cris@cris-laptop:~ $ date
Thu Dec  7 04:22:40 PM CET 2023
cris@cris-laptop:~ $ 2023-12-07T15:22:40.936Z [pebble] Check "foo" failure 4 (threshold 3): exec: "fail": executable file not found in $PATH
                                                                                                                                            2023-12-07T15:22:42.935Z [pebble] Check "foo" failure 5 (threshold 3): exec: "fail": executable file not found in $PATH
                                                                                                                                                                                                                                                                   ^C
cris@cris-laptop:~ $ 2023-12-07T15:22:44.936Z [pebble] Check "foo" failure 6 (threshold 3): exec: "fail": executable file not found in $PATH

exit
2023-12-07T15:22:45.277Z [pebble] GET /v1/changes/36/wait 12.333227355s 200
                                                                           cris@cris-laptop:~ $ 
```

# Proposal

Do not print logs when `pebble enter exec` (especially if `exec` is running in interactive mode)

rebornplusplus:
Fixes #339 

The PR disables logs for `pebble enter exec` only if the `exec` command is connected to a TTY and it makes the TTY RAW, even if the `--verbose` flag is passed. 

Interactive applications, like Vim, can be run as an exec command:

    pebble enter --verbose exec -it vim

Since `enter` runs the pebble daemon and executes the `exec` command afterwards, the same TTY that will be made raw by exec to run interactive applications like Vim is also used for logging and service outputs. Thus, the logs coming from pebble find the TTY RAW and weird indentation is introduced. See issue #339.

The issue, however, is not limited to only indentation. For example, if a text editor was being run with `exec -it`, like Vim, the incoming logs would go to the text editor and may change the content of the editor’s buffer.

This PR disables logs completely for `pebble enter --verbose exec -it` (if a terminal is connected) even if the `--verbose` flag is passed. Since logs are currently disabled for `pebble enter exec` if `--verbose` is not passed and `-i`, `-t` are assumed if stdin and stdout are TTYs, the PR only ignores the `--verbose` flag in cases where exec is connected to a TTY and it will make the TTY raw.

See also the spec, [RK034](https://docs.google.com/document/d/1JrFkUh3KeuIGOMk0wRDx40f4WQV3ARY3BEZ4oscOvl0/edit).

### DIFF
--- a/internals/cli/cmd_enter.go
+++ b/internals/cli/cmd_enter.go
@@ -18,9 +18,11 @@ import (
 	"fmt"
 
 	"github.com/canonical/go-flags"
+	"golang.org/x/sys/unix"
 
 	"github.com/canonical/pebble/client"
 	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/ptyutil"
 )
 
 const cmdEnterSummary = "Run subcommand under a container environment"
@@ -139,10 +141,13 @@ func (cmd *cmdEnter) Execute(args []string) error {
 		panic("internal error: expected subcommand (parser.Active == nil)")
 	}
 
-	// The exec command makes terminal raw to run interactive
-	// applications. Ignore --verbose for those cases.
+	// The exec command makes the terminal raw to run interactive
+	// applications. Ignore --verbose for those cases, to avoid
+	// log output being sent to the interactive terminal.
 	if execCmd, ok := commander.(*cmdExec); ok {
-		if execCmd.willMakeTerminalRaw() {
+		stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
+		stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
+		if stdinIsTerminal && execCmd.useTerminal(stdoutIsTerminal) {
 			cmd.Verbose = false
 			runCmd.Verbose = false
 		}
--- a/internals/cli/cmd_enter.go
+++ b/internals/cli/cmd_enter.go
@@ -145,9 +145,20 @@ func (cmd *cmdEnter) Execute(args []string) error {
 	// applications. Ignore --verbose for those cases, to avoid
 	// log output being sent to the interactive terminal.
 	if execCmd, ok := commander.(*cmdExec); ok {
+		// The following logic is duplicated from cmdExec.Execute()
+		// in internals/cli/cmd_exec.go.
 		stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
 		stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
-		terminal := execCmd.useTerminal(stdoutIsTerminal)
+		var terminal bool
+		if execCmd.Terminal {
+			terminal = true
+		} else if execCmd.NoTerminal {
+			terminal = false
+		} else {
+			terminal = stdoutIsTerminal
+		}
+		// Terminal is made RAW in exec if both terminal and
+		// stdIsTerminal are true.
 		if terminal && stdinIsTerminal {
 			cmd.Verbose = false
 			runCmd.Verbose = false
--- a/internals/cli/cmd_enter.go
+++ b/internals/cli/cmd_enter.go
@@ -147,7 +147,8 @@ func (cmd *cmdEnter) Execute(args []string) error {
 	if execCmd, ok := commander.(*cmdExec); ok {
 		stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
 		stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
-		if stdinIsTerminal && execCmd.useTerminal(stdoutIsTerminal) {
+		terminal := execCmd.useTerminal(stdoutIsTerminal)
+		if terminal && stdinIsTerminal {
 			cmd.Verbose = false
 			runCmd.Verbose = false
 		}
--- a/internals/cli/cmd_enter.go
+++ b/internals/cli/cmd_enter.go
@@ -139,6 +139,15 @@ func (cmd *cmdEnter) Execute(args []string) error {
 		panic("internal error: expected subcommand (parser.Active == nil)")
 	}
 
+	// The exec command makes terminal raw to run interactive
+	// applications. Ignore --verbose for those cases.
+	if execCmd, ok := commander.(*cmdExec); ok {
+		if execCmd.willMakeTerminalRaw() {
+			cmd.Verbose = false
+			runCmd.Verbose = false
+		}
+	}
+
 	enterFlags, supported := commandEnterFlags(commander)
 
 	if !supported {
--- a/internals/cli/cmd_exec.go
+++ b/internals/cli/cmd_exec.go
@@ -223,6 +223,22 @@ func (cmd *cmdExec) Execute(args []string) error {
 	}
 }
 
+// willMakeTerminalRaw returns true if the exec command will make the
+// terminal raw during execution.
+func (cmd *cmdExec) willMakeTerminalRaw() bool {
+	stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
+	stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
+	var terminal bool
+	if cmd.Terminal {
+		terminal = true
+	} else if cmd.NoTerminal {
+		terminal = false
+	} else {
+		terminal = stdoutIsTerminal
+	}
+	return terminal && stdinIsTerminal
+}
+
 func execControlHandler(process *client.ExecProcess, terminal bool, stop <-chan struct{}, sighup chan<- struct{}) {
 	ch := make(chan os.Signal, 10)
 	signal.Notify(ch,
--- a/internals/cli/cmd_exec.go
+++ b/internals/cli/cmd_exec.go
@@ -117,7 +117,14 @@ func (cmd *cmdExec) Execute(args []string) error {
 
 	// Specify Terminal=true if -t is given, or if stdout is a TTY.
 	stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
-	terminal := cmd.useTerminal(stdoutIsTerminal)
+	var terminal bool
+	if cmd.Terminal {
+		terminal = true
+	} else if cmd.NoTerminal {
+		terminal = false
+	} else {
+		terminal = stdoutIsTerminal
+	}
 
 	// Specify Interactive=true if -i is given, or if stdin and stdout are TTYs.
 	stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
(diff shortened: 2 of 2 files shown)


internals/cli/cmd_enter.go
```diff
@@ -139,6 +139,15 @@ func (cmd *cmdEnter) Execute(args []string) error {
 		panic("internal error: expected subcommand (parser.Active == nil)")
 	}
 
+	// The exec command makes terminal raw to run interactive
```
benhoyt, line 142:
```suggestion
	// The exec command makes the terminal raw to run interactive
```

internals/cli/cmd_enter.go
```diff
@@ -139,6 +139,15 @@ func (cmd *cmdEnter) Execute(args []string) error {
...
+	// applications. Ignore --verbose for those cases.
```
benhoyt, line 143:
```suggestion
	// applications. Ignore --verbose for those cases, to avoid
	// log output being sent to the interactive terminal.
```

internals/cli/cmd_exec.go
```diff
@@ -223,6 +223,22 @@ func (cmd *cmdExec) Execute(args []string) error {
 	}
 }
 
+// willMakeTerminalRaw returns true if the exec command will make the
+// terminal raw during execution.
+func (cmd *cmdExec) willMakeTerminalRaw() bool {
```
benhoyt, line 228: Most of the lines of this function are repeated in `cmdExec.Execute`. Could we create a function `useTerminal` that does most of this, and is used in both places?

```go
func (cmd *cmdExec) useTerminal(stdoutIsTerminal bool) bool {
	if cmd.Terminal {
		return true
	} else if cmd.NoTerminal {
		return false
	} else {
		return stdoutIsTerminal
	}
}
```

And then we could get rid of `willMakeTerminalRaw`, or it'd be a simple 3-liner.

Alternatively we could just remove this from `cmd_exec.go` and inline it in `cmd_enter.go` -- it seems weird to me that it's defined here but not used here.

benhoyt (changes requested):
The concept looks good to me, and I'm glad it's so simple. Just a couple of small style / factoring comments.

internals/cli/cmd_exec.go
rebornplusplus: Thanks! I added the `useTerminal` function and removed `willMakeTerminalRaw`. Changes are in https://github.com/canonical/pebble/pull/347/commits/5f1924635b68917bf1d0d479f1824c0be631d064.

internals/cli/cmd_enter.go
```diff
@@ -139,6 +141,18 @@ func (cmd *cmdEnter) Execute(args []string) error {
 		panic("internal error: expected subcommand (parser.Active == nil)")
 	}
 
+	// The exec command makes the terminal raw to run interactive
+	// applications. Ignore --verbose for those cases, to avoid
+	// log output being sent to the interactive terminal.
+	if execCmd, ok := commander.(*cmdExec); ok {
+		stdinIsTerminal := ptyutil.IsTerminal(unix.Stdin)
+		stdoutIsTerminal := ptyutil.IsTerminal(unix.Stdout)
+		if stdinIsTerminal && execCmd.useTerminal(stdoutIsTerminal) {
```
benhoyt, line 150: Sorry to be nit-picky, but let's change this `if` statement so it has the same "shape" as the `if` statement in `cmd_exec.go` -- that way it's easy to see they're identical.

```suggestion
		terminal := execCmd.useTerminal(stdoutIsTerminal)
		if terminal && stdinIsTerminal {
```

benhoyt (changes requested):
Just one nit requested. Also, given that there's no tests for this (and I think tests would be hard to write), can you confirm you've tested `pebble exec` and `pebble enter` for the various cases?

rebornplusplus:
> Given that there's no tests for this (and I think tests would be hard to write), can you confirm you've tested `pebble exec` and `pebble enter` for the various cases?

Yup, I can confirm that I have tested `pebble enter` and `pebble exec`. There are no behavioural changes for `pebble exec` alone. Only `pebble enter [..] exec` is affected.

benhoyt (approved):
Great, thanks! I'm happy with the code and behaviour change -- let's just get the spec/behaviour approved by Gustavo before merging.

internals/cli/cmd_exec.go
```diff
@@ -223,6 +216,17 @@ func (cmd *cmdExec) Execute(args []string) error {
 	}
 }
 
+// useTerminal returns true if -t is given, or if stdoutIsTerminal is true.
+func (cmd *cmdExec) useTerminal(stdoutIsTerminal bool) bool {
```
niemeyer, line 220: As a side note, this function is not looking so great.  The behavior is a bit confusing, the documentation is incomplete, and the fact it uses an argument that is irrelevant for the function logic is an anti-pattern (just use the flag at the call site instead).  All of this to use the values of two trivial boolean flags.

niemeyer (changes requested):
I've provided some questions in the spec, about the use of FD 2, and about some inconsistencies in the proposal related to the use of the flags vs. the use of the actual terminal state.

internals/cli/cmd_exec.go
benhoyt: Yeah, fair enough. This may be a moot point if we change the behaviour to use FD 2 anyway, but otherwise let's just duplicate this (simple) logic in both places (possibly with a short comment at both sites saying, "this duplicates the logic in Xyz").

rebornplusplus:
> I've provided some questions in the spec, about the use of FD 2, and about some inconsistencies in the proposal related to the use of the flags vs. the use of the actual terminal state.

I have updated the proposal to be more clear in the spec and the PR. Thanks!

internals/cli/cmd_exec.go
rebornplusplus: Removed it and duplicated the logic as Ben suggested.

internals/cli/cmd_enter.go
```diff
@@ -139,6 +141,30 @@ func (cmd *cmdEnter) Execute(args []string) error {
...
+		} else if execCmd.NoTerminal {
+			terminal = false
+		} else {
+			terminal = stdoutIsTerminal
+		}
+		// Terminal is made RAW in exec if both terminal and
+		// stdIsTerminal are true.
+		if terminal && stdinIsTerminal {
+			cmd.Verbose = false
+			runCmd.Verbose = false
```
benhoyt, line 164: A couple of questions:

* Following on from Gustavo's question about logging to FD 2 (stderr), [currently `--verbose` logs to Stdout (FD 1)](https://github.com/canonical/pebble/blob/0a1127ed54ebbd7d89f79450499e0a5907390388/internals/cli/cmd_run.go#L165). Would changing that to Stderr (FD 2) fix this issue, without needing to disable verbose logging?
* We're not disabling other logs here, only "verbose" service logs -- are there other logs printed by `pebble run` or the Pebble CLI that will still cause issues?

rebornplusplus: HI Ben, apologies for the very late reply.

---

> * Following on from Gustavo's question about logging to FD 2 (stderr), [currently `--verbose` logs to Stdout (FD 1)](https://github.com/canonical/pebble/blob/0a1127ed54ebbd7d89f79450499e0a5907390388/internals/cli/cmd_run.go#L165). Would changing that to Stderr (FD 2) fix this issue, without needing to disable verbose logging?

I think, only partially. So yes, if we do change that to `stderr`, we can essentially run the following **in a terminal** without any logs going to vim.

```
pebble enter --verbose --run exec vim 2> logs
```

The problem arises whenever pebble is inside a container and we are running the same stuff with `docker run -it`:

```
docker run -it pebble:test pebble enter --verbose --run exec vim 2> logs
```

There would be nothing in `logs` and the pebble logs would come to the interactive application vim instead. This is because when we use the `-it` flag to docker, our application logs would all come to host's stdout. And since `pebble enter` is specific to containers (rocks), our primary use-case would still have the issue.

---

> * We're not disabling other logs here, only "verbose" service logs -- are there other logs printed by `pebble run` or the Pebble CLI that will still cause issues?

For `pebble run`, the impact of `-v`/`--verbose` option is only the addition of the service outputs as you linked above. There will still be other logs without `--verbose` like below:

```
2024-02-19T09:26:13.226Z [pebble] Started daemon.
2024-02-19T09:26:13.277Z [pebble] POST /v1/services 49.446134ms 202
2024-02-19T09:26:13.277Z [pebble] Started default services with change 359.
...
```

For `pebble enter exec` however, if there is no `--verbose` specified, the other logs will not be printed at all. 

https://github.com/canonical/pebble/blob/0a1127ed54ebbd7d89f79450499e0a5907390388/internals/cli/cmd_enter.go#L164-L166

So if we disable `--verbose`, there will be no logs altogether that may cause any issues for interactive exec runs.

rebornplusplus:
Closing in favor of below:

We will be removing `--verbose` from Rocks entrypoint. So, Rocks will now have only `pebble enter` as entrypoint. Additionally, we will also be BLOCKING the use of `--verbose` with `pebble enter exec`. So, `pebble enter --verbose [...] exec [...] <cmd...>` will produce an error.

PR #348, opened 2024-01-29
fix(cmdstate): use imperative language and uppercase for task summary

medvied:


benhoyt (approved):
Thanks - fix looks fine to me

Issue #314, opened 2023-10-05
Pebble is unresponsive


troyanov:
Some odd behaviour of Pebble was observed several times (`1.2.0` and `1.4.0` versions):

Pebble daemon starts normally and is able to spin up couple services and process several API calls. Then, quite quickly, it falls into a limbo state (thats how I understand it from the logs).

After `05 Oct 04:14:41` there are no new log entries and any CLI command execution that interacts with the daemon (e.g. `pebble changes`) simply hang.

```bash
$ sudo journalctl -feu snap.maas.pebble.service -t maas.pebble
...
Oct 05 04:14:38 maas.pebble[754145]: 2023-10-05T04:14:38.235Z [pebble] Started daemon.
Oct 05 04:14:38 maas.pebble[754145]: 2023-10-05T04:14:38.470Z [pebble] POST /v1/services 225.462757ms 202
Oct 05 04:14:38 maas.pebble[754145]: 2023-10-05T04:14:38.470Z [pebble] Started default services with change 40183.
Oct 05 04:14:38 maas.pebble[754145]: 2023-10-05T04:14:38.538Z [pebble] Service "regiond" starting: sh -c "exec systemd-cat -t maas-regiond $SNAP/bin/run-regiond"
Oct 05 04:14:39 maas.pebble[754145]: 2023-10-05T04:14:39.647Z [pebble] Service "apiserver" starting: sh -c "exec systemd-cat -t maas-apiserver $SNAP/bin/run-apiserver"
Oct 05 04:14:40 maas.pebble[754145]: 2023-10-05T04:14:40.752Z [pebble] Service "rackd" starting: sh -c "exec systemd-cat -t maas-rackd $SNAP/bin/run-rackd"
Oct 05 04:14:41 maas.pebble[754145]: 2023-10-05T04:14:41.573Z [pebble] GET /v1/services?names=bind9 120.939µs 200
Oct 05 04:14:41 maas.pebble[754145]: 2023-10-05T04:14:41.574Z [pebble] GET /v1/services?names=ntp 86.887µs 200
Oct 05 04:14:41 maas.pebble[754145]: 2023-10-05T04:14:41.574Z [pebble] GET /v1/services?names=syslog 68.915µs 200
Oct 05 04:14:41 maas.pebble[754145]: 2023-10-05T04:14:41.575Z [pebble] GET /v1/services?names=http 65.915µs 200
[...]

PR #350, opened 2024-02-05
fix(daemon): turn off net/http panic recovery

benhoyt:
By default, Go's `net/http` recovers panics in the `ServeHTTP` goroutine and keeps serving. However, if a lock is held when the panic happens, this will mean a complete deadlock and Pebble will go unresponsive. As far as I know, this hasn't happened yet, but we use snippets without `defer` like this fairly frequently if it's just a tiny amount of code and defer is awkward:

```go
st.Lock()
foo := st.Foo()
st.Unlock()
```

If `st.Foo()` or whatever's between the lock and unlock panics, `net/http` will recover the panic but the lock will still be held. And in the case of the state lock that basically means all requests are blocked -- bad!

There's no direct way of opting out of this `net/http` behaviour, so you have to `recover()` from panics yourself, print the goroutine dump, and exit.

I've tested this locally, and it prints the goroutine dump and exits with code 1 on a panic as expected, instead of locking up. It also handles any panic in the handler or middleware, eg in `logit`.


internals/daemon/daemon.go
```diff
@@ -363,6 +363,27 @@ func logit(handler http.Handler) http.Handler {
...
+// deadlocked state (for example, due to a held mutex lock).
+//
+// See: https://github.com/canonical/pebble/issues/314#issuecomment-1926148064
+// Workaround from: https://github.com/golang/go/issues/16542#issuecomment-246549902
+func panicHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			err := recover()
+			if err != nil {
+				buf := make([]byte, 1024*1024)
```
benhoyt, line 377: 1MB is plenty huge. We could pass `runtime.Stack` ever-increasing buffer sizes till it fits, like the [code for `debug.Stack` does](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/runtime/debug/stack.go;l=21), but that seems overkill. For reference, the large MAAS goroutine stack dump in [this comment](https://github.com/canonical/pebble/issues/314#issuecomment-1918769483) is about 100KB.

hpidcock (approved):
I think overall this is an improvement, as we will now see panics as they happen, but it may lead to some unexpected panics in production uses of pebble until we identify the cause of the deadlock that spawned this line of inquiry.

internals/daemon/daemon.go
benhoyt: Because [`debug.Stack`](https://pkg.go.dev/runtime/debug#Stack) only includes the current goroutine, whereas [`runtime.Stack`](https://pkg.go.dev/runtime#Stack) takes an `all` parameter which we're setting to true to get the stack trace from all goroutines.

By default when Go exits due to a panic it just prints the current goroutine's stack trace (ala `debug.Stack`), but that's not particularly helpful for debugging deadlock issues like this.

benhoyt:
> I think overall this is an improvement, as we will now see panics as they happen, but it may lead to some unexpected panics in production uses of pebble until we identify the cause of the deadlock that spawned this line of inquiry.

@hpidcock The thing is, those "unexpected panics in production uses" have a decent chance of deadlocking the entire Pebble process (if a lock is held with `defer.Unlock()`), which is much worse than exiting (in which case the calling process will likely restart it). In any case, we'll chat further this afternoon about the best way forward.

benhoyt:
Given we now know panic-recovery isn't the cause of the issue at hand, I'm going to come back to this after reproducing and fixing https://github.com/canonical/pebble/issues/314, but I'll leave this PR open -- I think it's probably a good idea in any case.

internals/daemon/daemon.go
benhoyt: Actually, on second thoughts, the full goroutine dump probably won't help us here -- it's the panic/recover that's causing a *deadlock*, but when the panic itself occurs, there's no deadlock. So I've just gone ahead and used `debug.Stack()` directly.

benhoyt:
For reference, Gustavo was positive about this in a spec review meeting a couple of months ago (or more accurately, didn't love the default net/http behaviour and agreed it would be good to turn it off).

IronCore864 (approved):

hpidcock (approved):
I still don't like panic'ing in production, but I think if we don't, we are going to miss these fatal bugs.

Issue #351, opened 2024-02-09
Add unit and integration tests for `pebble run`


hpidcock:
Currently `pebble run` has no unit tests or integration tests. Considering this is the main entry point, it would be valuable to add integration testing here.

There is a few possible ways we could achieve this:
- Use `spread` to define the integration tests.
- Use python or go to write integration tests using a test framework.
- Write a bunch of shell scripts.

The integration tests should cover all the expected use cases of pebble at the moment.

benhoyt:
Just for the record, @hpidcock and I talked briefly about this, and I'm quite keen to try using `go test` with a build tag for this, for example `go test -tags integration ./internals/testintegration`. This will mean we can run some nice high-level tests of `pebble run` without installing any additional tooling.

benhoyt:
Relatedly, we should figure out a better way to mark and run "root tests" -- or maybe these just become integration tests instead. https://github.com/canonical/pebble/blob/18c93cbc280de20c9e21d323fe3a8bbb340777a2/.github/workflows/tests.yml#L35-L52

IronCore864:
I have finished researching the possible solutions mentioned above, and here are some of my thoughts:

1. Shell scripts: I did not test it but I'm pretty sure this would definitely work. I gave up this option because I don't think shell scripts can be as easily maintained as other options, like Python scripts with some sort of testing framework.
2. Python/Go to write integration tests: This is where I put my major effort. I chose Go because I don't want to introduce another tech into this repo (Well, technically, Python is already introduced in the documentation of Pebble but it's not used for development-related tasks yet, so I don't want to increase the complexity of the tech stack). I followed Ben's suggestion with Go Build Flags. I have done a PoC, see [here](https://github.com/canonical/pebble/pull/497/files).
3. Use `spread` for integration tests: I've tried it out with LXD backend, and I think this is similar to Dima's idea when we chatted about this in our 1-1 where he suggested that maybe we could run Pebble as root in a container for integration tests and other root tests. I think it's a good idea but it brings extra dependencies, like spread+LXD or docker and other test scripts.

So, after investigation, I think No.2 probably is the best choice here.

A few more words on the build flags (also known as the build constraints) PoC:

[...]

PR #352, opened 2024-02-09
feat(daemon): Add support for `PEBBLE_COPY_ONCE`.

hpidcock:
`$PEBBLE_COPY_ONCE` environment variable allows `pebble run` to copy layers and state from another directory when seeding a new `$PEBBLE` directory.

The rationale for this change is to allow OCI images to specify existing layers in the default `$PEBBLE` location. Then at runtime, be able to run `pebble` with a different `$PEBBLE` directory. This allows `pebble` to be run as a non-root user or run on a read-only root filesystem.

Issue #351 outlines that `pebble run` has no unit or integration tests. A test should be performed manually for now.

Manual QA is as follows:
- Create a directory, with a layers subdirectory and valid testing layers (e.g. ~/dir/with-existing-state)
- Create another empty directory. (e.g. ~/dir/with-no-state)
- `PEBBLE_COPY_ONCE=~/dir/with-existing-state  PEBBLE=~/dir/with-no-state pebble run`
- Check all the layers are loaded.
- Check all the layers are copied.
- Modify a layer in `~/dir/with-no-state/layers`.
- Restart `pebble run` and check that it doesn't crash.
- Check that the modifications are preserved after the restart (PEBBLE_COPY_ONCE is only for an empty directory).
- Check `PEBBLE_COPY_ONCE` can be set to a non-existing directory with `pebble run` passed `--create-dirs`

[JU090](https://docs.google.com/document/d/1NWV4QsYq1NldS_V_YlafpJQyNAV-WS46VsIKGBrin_s/edit)
[JUJU-5435](https://warthogs.atlassian.net/browse/JUJU-5435)

[JUJU-5435]: https://warthogs.atlassian.net/browse/JUJU-5435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -296,7 +296,7 @@ func maybeCopyPebbleDir(srcDir, destDir string) error {
 		return nil
 	}
 	fsys := os.DirFS(srcDir)
-	// TODO: replace with os.CopyFS
+	// TODO: replace with os.CopyFS when we're using Go 1.23
 	err = _go_os_CopyFS(destDir, fsys)
 	if err != nil {
 		return fmt.Errorf("cannot copy %q to %q: %w", srcDir, destDir, err)
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -17,11 +17,8 @@ package cli
 import (
 	"errors"
 	"fmt"
-	"io"
-	"io/fs"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"strconv"
 	"syscall"
 	"time"
@@ -160,7 +157,7 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 			return err
 		}
 	}
-	err := maybeCopyPebbleDir(getCopySource(), pebbleDir)
+	err := maybeCopyPebbleDir(pebbleDir, getCopySource())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


benhoyt (changes requested):
Thanks for the PR. However, do we really need a 3rd party dependency for this? Given what we need to do, it's quite a large one (1500 LoC and significant [API surface area](https://pkg.go.dev/github.com/otiai10/copy#Options)). We only need to copy files, and it doesn't even need to be recursive, right? (Though that's easy with [`filepath.WalkDir`](https://pkg.go.dev/path/filepath#WalkDir) if we need to.) Here's an implementation of (the upcoming) `fs.CopyFS` for reference: https://github.com/golang/go/issues/62484#issue-1884498794 -- not many more lines of code than calling the 3rd party lib. :-) Happy to discuss voice if you disagree.

Regarding testing, I'd be happy with manual testing for this PR, until we get better integration/spared tests -- we don't have any for the rest of the `pebble run` behaviour, so it's no worse. :-) That said, most of this is just file/dir copying, which is easily `go test`able, right?

benhoyt:
One other thought: we should document this in README.md somewhere. Or maybe in the help text?

internals/cli/cmd_run.go
```diff
@@ -275,3 +283,109 @@ func convertArgs(args [][]string) (map[string][]string, error) {
...
+	fsys := os.DirFS(srcDir)
+	// TODO: replace with os.CopyFS
+	err = _go_os_CopyFS(destDir, fsys)
+	if err != nil {
+		return fmt.Errorf("cannot copy %q to %q: %w", srcDir, destDir, err)
+	}
+	return nil
+}
+
+// Implementation from https://go-review.googlesource.com/c/go/+/558995 until accepted into go.
```
benhoyt, line 307: Can we please move from here down to its own file so it's super-easy to delete later? Maybe `copyfs.go`.

internals/cli/cmd_run.go
```diff
@@ -275,3 +283,109 @@ func convertArgs(args [][]string) (map[string][]string, error) {
...
+	dirEnts, err := os.ReadDir(destDir)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	} else if len(dirEnts) != 0 {
+		// Skip non-empty dir.
+		return nil
+	}
+	fsys := os.DirFS(srcDir)
+	// TODO: replace with os.CopyFS
+	err = _go_os_CopyFS(destDir, fsys)
```
benhoyt, line 300: Nice, much simpler!

Naming nit: can we please call this `osCopyFS` or even just `copyFS`? Those underscores make it not look like Go somehow. :-)

internals/cli/cmd_run.go
```diff
@@ -275,3 +283,109 @@ func convertArgs(args [][]string) (map[string][]string, error) {
 
 	return mappedArgs, nil
 }
+
+func maybeCopyPebbleDir(srcDir, destDir string) error {
```
benhoyt, line 287: I think we should reverse the args to be (dest, src) so it's like other Go copy functions: `copy` builtin, `os.Copy`, and the upcoming `os.CopyFS` function.

benhoyt: Can we also please add a test for this? It won't be an end-to-end test of `pebble run`, but all the interesting stuff is in here. Would need to add a new `cmd_run_test.go` and add `var MabyeCopyPebbleDir = maybeCopyPebbleDir` to `export_test.go`.

We don't need to test `copyFS` (we can assume that's tested in Go), but the basic "not set or dir is empty" functionality here, and that a copy is happening.

internals/cli/cmd_run.go
```diff
@@ -275,3 +283,109 @@ func convertArgs(args [][]string) (map[string][]string, error) {
...
+	if srcDir == "" {
+		return nil
+	}
+	dirEnts, err := os.ReadDir(destDir)
```
benhoyt, line 291: Naming nit, as "ents" remind me of walking trees :-): `dirEnts` -> `entries`.

internals/cli/cmd_run.go
```diff
@@ -275,3 +283,109 @@ func convertArgs(args [][]string) (map[string][]string, error) {
...
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	} else if len(dirEnts) != 0 {
+		// Skip non-empty dir.
+		return nil
+	}
+	fsys := os.DirFS(srcDir)
+	// TODO: replace with os.CopyFS
```
benhoyt, line 299: This will almost certainly be included in Go 1.23, and I like to include that in the comment so we can search for "Go 1." later and find these to update:
```suggestion
	// TODO: replace with os.CopyFS when we're using Go 1.23
```

benhoyt (changes requested):
A bunch of minor comments, plus a request for a test (of `maybeCopyPebbleDir`, not of `cmd_run`).

benhoyt (changes requested):
Looks good to me, just requesting a couple of minor test changes, and then I'll merge.

internals/cli/cmd_run_test.go
```diff
@@ -0,0 +1,97 @@
...
+	ok := 0
+	dstFS := os.DirFS(dst)
+	err = fs.WalkDir(dstFS, ".", func(p string, d fs.DirEntry, err error) error {
+		switch p {
+		case ".", "a", "a/b", "a/b/c":
+		case "a.yaml":
+			data, err := fs.ReadFile(dstFS, p)
+			c.Check(err, IsNil)
+			c.Check(data, DeepEquals, []byte("# bye\n"))
+			ok++
```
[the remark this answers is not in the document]
hpidcock: This is true, but the filesystem/go are tested systems, if we get two of the same file I'd be very surprised and worried.

internals/cli/cmd_run_test.go
```diff
@@ -0,0 +1,97 @@
...
+	err = os.WriteFile(path.Join(src, "a.yaml"), []byte("# bye\n"), 0666)
+	c.Assert(err, IsNil)
+
+	dst := c.MkDir()
+	err = cli.MaybeCopyPebbleDir(dst, src)
+	c.Assert(err, IsNil)
+
+	ok := 0
+	dstFS := os.DirFS(dst)
+	err = fs.WalkDir(dstFS, ".", func(p string, d fs.DirEntry, err error) error {
```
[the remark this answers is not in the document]
hpidcock: Sure thing.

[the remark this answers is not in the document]
benhoyt: As Harry mentioned on chat, "fs.WalkDir is so much better because all the paths are rooted at the directory used to create the fs" and there's no need for a bunch of `filepath.Rel`. I'd missed this, sorry!

benhoyt (approved):

benhoyt (approved):
Works for me, thanks!

PR #355, opened 2024-02-13
fix(servstate): reduce scope of holding ServiceManager.planLock

benhoyt:
In most cases we're just reading from m.plan fields (Services,
DefaultServiceNames, and so on), so we can use m.Plan(), which holds
the plan lock only for the duration of fetch m.plan -- this is safe,
because we never mutate what's inside m.plan, we only replace it (in
updatePlan).

This also means that in functions like Services() we're only holding
either the plan lock or the services lock at once, not both -- avoiding
the 3-way deadlock described in #314.

For the cases we are updating the plan (AppendLayer, CombineLayer,
SetServiceArgs), we still need acquirePlan.

In ServiceManager.ServiceLogs, we don't need to acquire the plan lock
at all (m.plan isn't used).

Fixes #314.

PR #356, opened 2024-02-13
fix(state): don't lock global state lock when adding a task log

benhoyt:
Sometimes the service manager has to write a task log, but there's a
3-way deadlock where another goroutine has held the state lock. Break
this cycle.

See https://github.com/canonical/pebble/issues/314 -- either this PR or #355 by itself will fix that issue.
I'm not sure about this PR. I like what it's trying to achieve, but it also
begs the question: why do we have this one special fine-grained lock
for the task log, and not for all the other fields?


benhoyt:
Closing in favour of https://github.com/canonical/pebble/pull/359, which is much less invasive, and it seems a better approach to structure the locking better rather than do surgery to the state package.

PR #358, opened 2024-02-13
chore(daemon): port AccessChecker interface from snapd

thp-canonical:
Port the `accessChecker` API from snapd for easier customization and more fine-grained access control.

`UserState` is currently empty, but it might sense to drag it along (to be in line with snapd's function signature) and allow user-specific state data to be extracted from the HTTP request in a central place (`userFromRequest()` in `daemon.go`).


internals/daemon/daemon.go
```diff
@@ -240,35 +171,44 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		statusMethodNotAllowed("method %q not allowed", r.Method).ServeHTTP(w, r)
+		return
 	}
 
+	ucred, err := ucrednetGet(r.RemoteAddr)
+	if err != nil && err != errNoID {
+		logger.Noticef("Cannot parse UID from remote address %q: %s", r.RemoteAddr, err)
+		statusInternalError(err.Error()).ServeHTTP(w, r)
+		return
+	}
```
benhoyt, line 203: Just to be consistent with snapd, can we move this block up to where it is in snapd's daemon.go? https://github.com/snapcore/snapd/blob/0a7a20db83b1d2a43319ab3e7beda18d9d97bd7d/daemon/daemon.go#L126

internals/daemon/daemon_test.go
```diff
@@ -174,9 +174,10 @@ func (s *daemonSuite) TestAddCommand(c *C) {
 		return &handler
 	}
 	command := Command{
-		Path:    endpoint,
-		GuestOK: true,
-		GET:     getCallback,
+		Path:        endpoint,
+		ReadAccess:  openAccess{},
+		WriteAccess: openAccess{},
```
benhoyt, line 179: Can remove the `WriteAccess` line here.

internals/daemon/export_test.go
```diff
@@ -62,3 +62,16 @@ func FakeSyscallReboot(f func(cmd int) error) (restore func()) {
...
+	StatusForbidden    = statusForbidden
+	StatusUnauthorized = statusUnauthorized
+)
+
+type (
+	AccessChecker = accessChecker
+	OpenAccess    = openAccess
+	UserAccess    = userAccess
+	RootAccess    = rootAccess
+	Ucrednet      = ucrednet
```
benhoyt, line 76: Per above, I think we should export all/most of these types to make this usable externally, so the changes in this file will probably be dropped.

benhoyt (changes requested):
Looking good! Let's just update from the master branch and resolve conflicts, add those tests, and I've commented on some minor issues.

I'd also like to get @flotter's review on this.

internals/daemon/access.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// rootAccess allows requests from the root uid
+type rootAccess struct{}
+
+func (ac rootAccess) CheckAccess(d *Daemon, r *http.Request, ucred *ucrednet, user *UserState) Response {
+	if ucred != nil && ucred.Uid == 0 {
+		return nil
+	}
+	return statusForbidden("access denied")
```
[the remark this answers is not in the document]
thp-canonical: Done. This is quite an invasive change, so if you prefer, I can split it into a separate PR.

internals/daemon/access.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (ac openAccess) CheckAccess(d *Daemon, r *http.Request, ucred *ucrednet, user *UserState) Response {
+	return nil
+}
+
+// rootAccess allows requests from the root uid
+type rootAccess struct{}
+
+func (ac rootAccess) CheckAccess(d *Daemon, r *http.Request, ucred *ucrednet, user *UserState) Response {
+	if ucred != nil && ucred.Uid == 0 {
```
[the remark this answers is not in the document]
thp-canonical: In case of snapd, there are two sockets (introduced in https://github.com/snapcore/snapd/pull/1749):

```golang
SnapdSocket = filepath.Join(rootdir, "/run/snapd.socket")
SnapSocket = filepath.Join(rootdir, "/run/snapd-snap.socket")
```

It seems like `snapd.socket` is accessed from the `snap` command and with privileged commands and `snapd-snap.socket` is accessed from `snapctl` and from within (possibly confined) snaps, and is the "public" API.

As Pebble doesn't have the concept of two different sockets with split API yet (it only listens on the UNIX domain socket + optionally on a TCP socket, and only the UNIX domain socket has `ucred != nil`), I don't think there's a good way to "port" this over to Pebble for now.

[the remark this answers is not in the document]
thp-canonical: I just saw when looking at the tests that `Daemon` actually has a `untrustedSocketPath` and `untrustedListener`, is that something that's actively used in Pebble? (`untrustedSocketPath: opts.SocketPath + ".untrusted"`)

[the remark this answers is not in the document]
thp-canonical: Well, seems like `UntrustedOK` was used as the switch for allowing access from the `untrustedSocketPath`, but `UntrustedOK` is nowhere used at the moment (except in the now-obsolete `canAccess()` and the unit tests).

internals/daemon/daemon_test.go
```diff
@@ -340,153 +341,140 @@ func (s *daemonSuite) TestFillsWarnings(c *C) {
...
-	c.Check(cmd.canAccess(get, nil), Equals, accessOK)
-	c.Check(cmd.canAccess(put, nil), Equals, accessUnauthorized)
-	c.Check(cmd.canAccess(pst, nil), Equals, accessUnauthorized)
-	c.Check(cmd.canAccess(del, nil), Equals, accessUnauthorized)
-}
+	responseFunc := func(c *Command, r *http.Request, s *UserState) Response {
+		return SyncResponse(true)
+	}
 
-func (s *daemonSuite) TestUntrustedAccessUntrustedOKWithUser(c *C) {
```
thp-canonical, line 373: The `TestUntrustedAccessUntrustedOKWithUser` test was removed, as `UntrustedOK` wasn't used by any part of the codebase (no `Command` specifies `UntrustedOK` in the `master` branch).

internals/daemon/access.go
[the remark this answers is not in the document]
thp-canonical: I could also split removal of `UntrustedOK` into a separate PR before this one is merged, so that it's more obvious that it isn't used, and it's clearer what gets removed.

internals/daemon/daemon_test.go
```diff
@@ -340,153 +341,140 @@ func (s *daemonSuite) TestFillsWarnings(c *C) {
 func (s *daemonSuite) TestGuestAccess(c *C) {
 	d := s.newDaemon(c)
 
-	get := &http.Request{Method: "GET"}
-	put := &http.Request{Method: "PUT"}
-	pst := &http.Request{Method: "POST"}
-	del := &http.Request{Method: "DELETE"}
-
-	cmd := &Command{d: d}
-	c.Check(cmd.canAccess(get, nil), Equals, accessUnauthorized)
```
thp-canonical, line 349: Since we can't test `canAccess()` directly anymore, we instead indirectly test the access checkers and how they relate to `GET` / `PUT` / `POST`.

Conceptually it's the same:

- Creation of `Command` with `AdminOnly`, `UserOK`, `GuestOK` is replaced with `ReadAccess` and `WriteAccess`
- Calling `canAccess()` with different methods is replaced with calling `doTestReqFunc`, which calls through `cmd.ServeHTTP`, and checks the returned HTTP status

thp-canonical:
@benhoyt This is now ready for review. It ended up spilling into many areas.

Let me know if you want me to split off any or all of those topics into separate PRs for easier review:

- Removal of `UntrustedOK` (should the `.untrusted` socket also be removed?)
- Public-making of `ucrednet` (https://github.com/canonical/pebble/pull/358/commits/f3152812bd3b26f67db9e94495f4b4a1e1a5454d)
- Public-making of `statusForbidden`, etc.. to `Forbidden`, etc... (https://github.com/canonical/pebble/pull/358/commits/bcbc1b0f14e038514a6176d0a87dd96f11b12acd)
- Removal of `DELETE` (https://github.com/canonical/pebble/pull/358/commits/89f05b64e60e6d15ec234b5f90393e04df1a5484)

internals/daemon/access.go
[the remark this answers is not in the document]
thp-canonical: This is now #361 and removes also the socket itself.

benhoyt:
@thp-canonical Can you please bring this branch up to date and resolve conflicts, now that #360 and #361 are merged?

internals/daemon/daemon_test.go
```diff
@@ -336,106 +337,140 @@ func (s *daemonSuite) TestFillsWarnings(c *C) {
...
-	c.Check(cmd.canAccess(pst, nil), Equals, accessUnauthorized)
-	c.Check(cmd.canAccess(del, nil), Equals, accessUnauthorized)
-}
+	responseFunc := func(c *Command, r *http.Request, s *UserState) Response {
+		return SyncResponse(true)
+	}
 
-func (s *daemonSuite) TestUserAccess(c *C) {
-	d := s.newDaemon(c)
+	doTestReqFunc := func(c *C, cmd *Command, mth string) *httptest.ResponseRecorder {
```
benhoyt, line 344: Doesn't look like this function needs to take `c`.

internals/daemon/daemon_test.go
```diff
@@ -336,106 +337,140 @@ func (s *daemonSuite) TestFillsWarnings(c *C) {
...
+		req := &http.Request{Method: mth, RemoteAddr: ""}
```
benhoyt, line 345: You don't need to include `RemoteAddr` as it's the zero value (empty string). Or should we be using [`http.NewRequest`](https://pkg.go.dev/net/http#NewRequest) here?

internals/daemon/daemon_test.go
```diff
@@ -336,106 +337,140 @@ func (s *daemonSuite) TestFillsWarnings(c *C) {
...
+		rec := httptest.NewRecorder()
+		cmd.ServeHTTP(rec, req)
+		return rec
+	}
+
+	for _, t := range []struct {
+		getStatus, putStatus, postStatus int
+		readAccess, writeAccess          AccessChecker
+	}{
```
benhoyt, line 354: I think this would be clearer split into two parts: the table of tests (in a variable `tests` or whatever), then a simple `for _, t := range tests { ... }` loop. I think it'd also be clearer with named fields in the definition, because it's not clear which is get/put/post (I would have expected post first, for example). I realize it would make it a bit longer than one per line. Something like:

```go
tests := []struct{
  get, post, put int // expected status for each method
  read, write AccessChecker
}{{
  get:   http.StatusOK,
  post:  http.StatusOK,
  put:   http.StatusOK,
  read:  OpenAccess{},
  write: OpenAccess{},
}, {
  ...
}}
```

You could even name the test struct type (and move it to the top level) so you can reuse it the other test(s).

benhoyt (changes requested):
This looks good -- just a couple of minor changes requested in the tests.

internals/daemon/daemon_test.go
thp-canonical: Thanks, that cleaned it up nicely. I left the order in the same order as the methods are defined in `daemon.Command` (GET, PUT, POST), but now with the named fields, it indeed becomes clearer which is which. If you feel strongly about it, I can swap the order of post and put here.

Once that was done, I realized that the "mechanism" is basically the same for all 3 test cases (`TestGuestAccess`, `TestUserAccess` and `TestSuperAccess`), so I added a helper function `testAccessChecker()` that contains the logic, and the test cases then just contain data (table of test case data + the `remoteAddr` string).

benhoyt:
I've also asked for @hpidcock's review, as he's pretty close to Pebble and has been part of these discussions.

internals/daemon/daemon_test.go
```diff
@@ -334,143 +334,191 @@ func (s *daemonSuite) TestFillsWarnings(c *C) {
...
+			read:  OpenAccess{},
+			write: UserAccess{},
+		},
+		{
+			get:   http.StatusOK,
+			put:   http.StatusUnauthorized,
+			post:  http.StatusUnauthorized,
+			read:  OpenAccess{},
+			write: AdminAccess{},
+		},
```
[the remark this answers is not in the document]
thp-canonical: Done, thanks. Is this something we can teach `golangci-lint` to do for us automatically eventually?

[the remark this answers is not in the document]
benhoyt: It doesn't look like there's a ready-made linter to do that, we'd have to write a custom one (sounds like a fun project). Interestingly, I found that the Google Go Style guide advocates for this too, and calls it "cuddled braces" (cute!): https://google.github.io/styleguide/go/decisions#cuddled-braces

internals/daemon/daemon.go
```diff
@@ -224,35 +165,40 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	switch r.Method {
 	case "GET":
 		rspf = c.GET
+		access = c.ReadAccess
 	case "PUT":
 		rspf = c.PUT
+		access = c.WriteAccess
 	case "POST":
 		rspf = c.POST
-	case "DELETE":
```
hpidcock, line 248: What happened to DELETE here?

hpidcock (approved):
I'd retain the delete method as a write only method, but it should be allowed as it is a useful concept, even if we don't use it.

internals/daemon/daemon.go
benhoyt: [Earlier](https://github.com/canonical/pebble/pull/358#discussion_r1488816769) I had suggested removing it as snapd removed it some time ago (and Pebble wasn't using it either). We can always add it back if we need it, but this was to be more in line with snapd.

hpidcock: OK, happy for it to go, since it isn't used and inline with snapd.

PR #359, opened 2024-02-13
fix(servstate): don't hold both servicesLock and state lock at once

benhoyt:
This avoids the 3-lock deadlock described in
https://github.com/canonical/pebble/issues/314. Other goroutines
may be holding the state lock and waiting for the services lock, so
it's problematic to acquire both locks at once. Break that part of the cycle.

We could do this inside serviceForStart/serviceForStop by manually
calling Unlock() sooner, but that's error-prone, so continue using defer,
but have the caller write the task log (which needs the state lock) after
the services lock is released.

This is in preference to the more invasive change in https://github.com/canonical/pebble/pull/356

### DIFF
--- a/internals/daemon/api_services_test.go
+++ b/internals/daemon/api_services_test.go
@@ -16,8 +16,11 @@ package daemon
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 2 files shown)


hpidcock (approved):
This feels like a reasonable compromise, but I'd like to see something like #356 actually solve the problem. The nature of logging is that it can occur in a variety of places, and as you can see here, it is not the most convenient of things to pass around.

flotter (approved):
I cannot find any issues with the proposal.

hpidcock (approved):
TestDeadlock looks reasonable too. Tested locally.

PR #360, opened 2024-02-14
fix(daemon): Export error responders from daemon.response

thp-canonical:
This brings the error response namings in line with snapd: [`daemon/errors.go` in snapd](https://github.com/snapcore/snapd/blob/489358223f0bd03da01e62a4062174eb7e9e0ffa/daemon/errors.go#L112). Split off from #358. While currently the error responses are used exclusively in the `daemon` package, once `AccessChecker` is available, it should be possible for other packages to create error responders to implement their `AccessChecker` variants.


benhoyt (approved):
Yep, this looks good -- brings it in line with snapd's names and exported status, and will make it possible to import and use these errors in external projects.

PR #361, opened 2024-02-15
chore(daemon): remove untrusted socket

thp-canonical:
As part of working on #358, we found out that the untrusted socket is not used, and so can be removed (in preparation for porting the `AccessChecker` changes from snapd in https://github.com/canonical/pebble/pull/358).

**Indicators that it's not used:**

If we look at how `canAccess` works, if we match on `untrustedSocketPath` (`isUntrusted`), the only way for `canAccess` to allow the request is when `c.UntrustedOK` is `true` (otherwise it unconditionally returns `accessUnauthorized` immediately):

```golang
if isUntrusted {
	if c.UntrustedOK {
		return accessOK
	}
	return accessUnauthorized
}
```

So in order for any API calls to be allowed with the untrusted socket (assuming all API calls go through `canAccess`), we would need to have a `Command` defined with `UntrustedOK: true`. Checking the Pebble codebase, no such `Command` definition exists, which means that even if any application would use the untrusted socket currently, all API calls would return `accessUnauthorized` unconditionally for this socket.

The untrusted socket as well as `UntrustedOK` in `Command` were already part of the initial import commit (50466bab893615147372361c78d4c48102daa07f), so seem to be an inheritance from snapd that haven't seen use in Pebble since then. The corresponding [snapd sources from around November 10th, 2020](https://github.com/snapcore/snapd/tree/e2581af241a941856a755035d816047ff9aa15d8) seem to call these [`SnapOK`](https://github.com/snapcore/snapd/blob/e2581af241a941856a755035d816047ff9aa15d8/daemon/daemon.go#L139) (`UntrustedOK`), [`dirs.SnapSocket`](https://github.com/snapcore/snapd/blob/e2581af241a941856a755035d816047ff9aa15d8/daemon/daemon.go#L160C23-L160C38) (`untrustedSocketPath`) and [`snapListener`](https://github.com/snapcore/snapd/blob/489358223f0bd03da01e62a4062174eb7e9e0ffa/daemon/daemon.go#L72) (`untrustedListener`).

Due to `gofmt` and removal of struct members with the longest names, this PR is best reviewed with the "hide whitespace" option.

### DIFF
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -81,22 +81,20 @@ type Options struct {
 
 // A Daemon listens for requests and routes them to the right command
 type Daemon struct {
-	Version             string
-	StartTime           time.Time
-	pebbleDir           string
-	normalSocketPath    string
-	untrustedSocketPath string
-	httpAddress         string
-	overlord            *overlord.Overlord
-	state               *state.State
-	generalListener     net.Listener
-	untrustedListener   net.Listener
-	httpListener        net.Listener
-	connTracker         *connTracker
-	serve               *http.Server
-	tomb                tomb.Tomb
-	router              *mux.Router
-	standbyOpinions     *standby.StandbyOpinions
+	Version          string
+	StartTime        time.Time
+	pebbleDir        string
+	normalSocketPath string
+	httpAddress      string
+	overlord         *overlord.Overlord
+	state            *state.State
+	generalListener  net.Listener
+	httpListener     net.Listener
+	connTracker      *connTracker
+	serve            *http.Server
+	tomb             tomb.Tomb
+	router           *mux.Router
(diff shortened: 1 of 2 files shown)


hpidcock (approved):
I am happy with the justification for this removal. If at a point in the future we need such a socket, we can reintroduce it. Thoughts @benhoyt ?

benhoyt:
Sorry, I've run out of time today -- will need to take a closer look early next week. (Sounds reasonable though.)

benhoyt (approved):
Yep, I agree this removal/simplification makes sense -- in preparation for porting the `AccessChecker` changes from snapd in https://github.com/canonical/pebble/pull/358.

benhoyt:
Oops, accidental close.

PR #362, opened 2024-02-19
feat(wait): export cli.Wait function

paul-rodriguez:
This commit exports a function to display a progress bar tracking the progress of a change ID.
Unlike the wait function which remains unexported, this new function does not require a waitMixin.

Motivation: allow projects that import Pebble and add commands to its CLI to use the same waiting logic and UI.

The unused skipAbort field of the waitMixin has been removed.

### DIFF
--- a/internals/cli/wait.go
+++ b/internals/cli/wait.go
@@ -31,8 +31,7 @@ var (
 )
 
 type waitMixin struct {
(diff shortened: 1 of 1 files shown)


anpep (approved):
Looks good to me. Thank you!

PR #363, opened 2024-02-20
fix: validate whole plan, rather than topmost two layers

### LINKED ISSUE #349: When combining layers, Pebble should validate the whole plan rather than a single combined layer
When doing `pebble add --combine`, Pebble is calling `plan.CombineLayers` on the existing labelled layer plus the new one, and validating the result as if it's in the entire plan (rather than just a single new layer), so we get this error:

```
$ pebble plan
services:
    django:
        override: replace
        command: sleep 1000

$ cat merge.yaml 
services:
  django:
    override: merge
    environment:
      POSTGRES: pg

$ pebble add mergie merge.yaml 
Layer "mergie" added successfully from "merge.yaml"

$ cat merge2.yaml 
services:
  django:
    override: merge
    environment:
      MYSQL: mysql

$ pebble add mergie merge2.yaml --combine
error: plan must define "command" for service "django"
```

The above should work, because we should only validate on the overall combined plan. The expected result is:

```
$ PEBBLE=~/pebble go run ./cmd/pebble plan
services:
    django:
        override: replace
        command: sleep 1000
        environment:
            MYSQL: mysql
            POSTGRES: pg
```

tonyandrewmeyer:
This PR moves layer/plan validation into dedicated functions, with `Layer.Validate` covering validation that is specific to a layer (such as an invalid value) and `Plan.Validate` covering validation that covers the plan as a whole (such as cycles or missing values).

Plan validation includes layer validation, and is done in the service manager's `updatePlanLayers` and in `plan.CombineLayers`.

We also use the check period as the upper bound for the timeout in checks, which was happening before, but less explicitly.

No additional validation is done - all of the validation code is moved rather than changed.

Fixes #349

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ cmd/VERSION
 .idea
 .vscode
 /pebble
+__debug_bin*
 
 dist/
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,5 @@ cmd/VERSION
 .idea
 .vscode
 /pebble
-__debug_bin*
 
 dist/
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -715,8 +715,6 @@ func (layer *Layer) Validate() error {
 				Message: fmt.Sprintf("service object cannot be null for service %q", name),
 			}
 		}
-		// ParseCommand needs the name to be set in order to generate a nice error message.
-		service.Name = name
 		_, _, err := service.ParseCommand()
 		if err != nil {
 			return &FormatError{
@@ -1033,19 +1031,28 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 	layer.Order = order
 	layer.Label = label
 
+	for name, service := range layer.Services {
+		// If service is nil, then the validation below will reject the layer,
+		// but we want the name set so that we can use easily use it in error
+		// messages during validation.
+		if service != nil {
+			service.Name = name
+		}
+	}
+	for name, check := range layer.Checks {
+		if check != nil {
+			check.Name = name
+		}
+	}
+	for name, target := range layer.LogTargets {
+		if target != nil {
+			target.Name = name
+		}
+	}
+
 	err = layer.Validate()
 	if err != nil {
 		return nil, err
-	}
-
-	for name, service := range layer.Services {
-		service.Name = name
-	}
-	for name, check := range layer.Checks {
-		check.Name = name
-	}
-	for name, target := range layer.LogTargets {
-		target.Name = name
 	}
 
 	return &layer, err
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -681,8 +681,8 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 // an error if there are validation errors.
 // See also Plan.Validate, which does additional checks based on the combined
 // layers.
-func (l *Layer) Validate() error {
-	for name, service := range l.Services {
+func (layer *Layer) Validate() error {
+	for name, service := range layer.Services {
 		_, _, err := service.ParseCommand()
 		if err != nil {
 			return &FormatError{
@@ -713,7 +713,7 @@ func (l *Layer) Validate() error {
 		}
 	}
 
-	for name, check := range l.Checks {
+	for name, check := range layer.Checks {
 		if check.Level != UnsetLevel && check.Level != AliveLevel && check.Level != ReadyLevel {
 			return &FormatError{
 				Message: fmt.Sprintf(`plan check %q level must be "alive" or "ready"`, name),
@@ -746,7 +746,7 @@ func (l *Layer) Validate() error {
 		}
 	}
 
-	for name, target := range l.LogTargets {
+	for name, target := range layer.LogTargets {
 		switch target.Type {
 		case LokiTarget, SyslogTarget:
 			// valid, continue
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -663,8 +663,13 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 		}
 		if !check.Timeout.IsSet {
 			check.Timeout.Value = defaultCheckTimeout
-		} else if check.Timeout.Value > check.Period.Value {
+		}
+		if check.Timeout.Value > check.Period.Value {
 			// The effective timeout will be the period, so make that clear.
+			// `.IsSet` remains false so that the capped value does not appear
+			// in the combined plan output - and it's not *user* set - the
+			// effective default timeout is the minimum of (check.Period.Value,
+			// default timeout).
 			check.Timeout.Value = check.Period.Value
 		}
 		if check.Threshold == 0 {
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -950,15 +950,6 @@ func order(services map[string]*Service, names []string, stop bool) ([]string, e
 	return order, nil
 }
 
-func (l *Layer) checkCycles() error {
-	var names []string
-	for name := range l.Services {
-		names = append(names, name)
-	}
-	_, err := order(l.Services, names, false)
-	return err
-}
-
 func (p *Plan) checkCycles() error {
 	var names []string
 	for name := range p.Services {
@@ -1051,10 +1042,6 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 		target.Name = name
 	}
 
-	err = layer.checkCycles()
-	if err != nil {
-		return nil, err
-	}
 	return &layer, err
 }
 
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -7,5 +7,6 @@ cmd/VERSION
 .idea
 .vscode
 /pebble
+__debug_bin*
```
benhoyt, line 10: Out of interest, what puts these files here?

internals/plan/plan.go
```diff
@@ -641,179 +646,218 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+		}
+	}
+
+	return combined, nil
+}
+
+// Validate checks that the layer is valid.
+// See also Plan.Validate, which does additional checks based on the combined
+// layers.
+func (l *Layer) Validate() error {
```
benhoyt, line 683: Can we change this receiver name from `l` to `layer` to avoid it being confused with 1 or I? Receiver names are normally 1-letter in Go, but that's not a hard and fast rule if there are reasons not to.

internals/plan/plan.go
```diff
@@ -641,179 +646,218 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+			// action, default is >1 to avoid flapping due to glitches. For
+			// what it's worth, Kubernetes probes uses a default of 3 too.
+			check.Threshold = defaultCheckThreshold
+		}
+	}
+
+	return combined, nil
+}
+
+// Validate checks that the layer is valid.
```
benhoyt, line 680:
```suggestion
// Validate checks that the layer is valid. It returns nil if all the checks pass, or
// an error if there are validation errors.
```

benhoyt (approved):
This looks excellent to me for a first PR in Pebble, thanks! No real issues, just a couple of nit comments. I'll also ask @flotter to go over this as he's looking at the whole plan/layers area (from another angle).

.gitignore
tonyandrewmeyer: Ah, I had assumed it was just Go when it panicked or something like that. Looking into it, I think it's [delve](https://github.com/go-delve/delve), which VS Code had me install when I did some debugging in the IDE (for what turned out to be a typo...).

Since it's tool specific like that I'll remove it from here.

internals/plan/plan.go
```diff
@@ -565,6 +565,11 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 	combined.Summary = last.Summary
 	combined.Description = last.Description
 	for _, layer := range layers {
+		err := layer.Validate()
```
flotter, line 568: Not a too bid deal, but it I see we are doing the layer checks in a place where existing layers will also be re-checked. It is a nice central place to catch them (which includes the slightly more elusive ```SetServiceArgs```

flotter (approved):
Some minor comments for consideration, but overall looks great.

internals/plan/plan.go
```diff
@@ -641,179 +646,219 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+					name, target.Type, LokiTarget, SyslogTarget),
+			}
+		}
+	}
+
+	return nil
+}
+
+// Validate checks that the combined layers form a valid plan.
+// See also Plan.Validate, which checks that the individual layers are valid.
```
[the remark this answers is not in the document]
tonyandrewmeyer: Yes, thanks!

internals/plan/plan.go
```diff
@@ -998,10 +1043,6 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
 		target.Name = name
 	}
 
-	err = layer.checkCycles()
-	if err != nil {
-		return nil, err
-	}
```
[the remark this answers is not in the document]
tonyandrewmeyer: I did actually have it in both places originally, and then when looking over the code while I was writing tests it seemed unnecessary to have it check until the whole plan is validated. I do agree that failing on an issue as soon as possible is ideal, though. It also seems like even if there was a loop removed later, having the loop in the first place is probably a mistake (but I don't know Pebble well enough to be super confident in that).

@benhoyt can I defer to your expertise here? Should I add in a `layer.checkCycles()` to `CombineLayers` where there was one before?

internals/plan/plan.go
```diff
@@ -641,179 +646,219 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+	}
+
+	return combined, nil
+}
+
+// Validate checks that the layer is valid. It returns nil if all the checks pass, or
+// an error if there are validation errors.
+// See also Plan.Validate, which does additional checks based on the combined
+// layers.
+func (layer *Layer) Validate() error {
```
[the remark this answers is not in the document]
tonyandrewmeyer: > It looks like you've demoted some post-combine checks here back to layer-time checks. I like the idea of failing on an issue as soon as possible, since the error context will also be more accurate.

Yes, that was my intention with this, and to logically separate out the code that's checking that a `Layer` is valid from checking that a combination of `Layer`s is valid.

> How does this block of validation compare with what is in `ParseLayer` ? All new layers start their life in `ParseLayer` (apart from `SetServiceArgs` I believe), and some layer validation already appears present there.

Indeed; nice catch. I don't think there's any overlap but the checks are fairly similar in nature (`ParseLayer` being mostly around the names).

@benhoyt what's your opinion on moving out the validation pieces of `ParseLayer` into `Layer.Validate`? And should `SetServiceArgs` do a `newLayer.Validate()`?

internals/plan/plan.go
tonyandrewmeyer: For what it's worth, this was the case before as well, because all of these checks were previously in `CombineLayers` (unless we do move ones from `ParseLayer` as per the thread above).

@benhoyt would you prefer if the layer validation was in `ParseLayer` and `SetServiceArgs` to remove the duplication? As it is, each layer is checked every time the plan is combined even though some of the checks can't change by adding another layer.

flotter: Just a subtle point, but I believe the previous validation was only ever done on the combined layer, once all the layers were merged. In that sense, its always needed, because the combined layer would have changed. What we are now doing is fine, but just pointing out that we are re-validating layers in the layers stack that never changed, every time we append or update an existing layer. Correct me if I am wrong, please.

flotter:
@tonyandrewmeyer b.t.w this is a real coincidence that you are cleaning this up a bit, because I am on a kind of similar mission for slightly different reasons, and I've done in my PoC exactly what you conceptionally doing here. I've also split it up between Layer and Plan validation, but you went a bit further, actually cleaning up what is inside each category, which is super nice and brave :) Looking forward getting this merged.

internals/plan/plan.go
tonyandrewmeyer: Ah, you're right, sorry. Same function, different loop :smile:.

benhoyt: Yeah, I do think we should avoid validating each layer every time a new one comes in. I think we should avoid validation in `CombineLayers` altogether, and document it as not doing validation -- the caller should have validated the input layers, and will (if they need to) validate the combined output as well.

internals/plan/plan.go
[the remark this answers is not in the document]
benhoyt: Yeah, good point. I do think we should move all the layer-validation logic from `ParseLayer` into `Layer.Validate`, and then just call `layer.Validate()` inside `ParseLayer`. Much tidier.

Should `SetServiceArgs` validate? It probably wouldn't need to as we're building that layer from scratch so it shouldn't have error, but it also wouldn't hurt, and would be quite cheap -- so coding defensively I think we might as well.

internals/plan/plan.go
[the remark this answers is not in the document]
benhoyt: I think we should just check for cycles at the end, that is, in `Plan.Validate` only. If someone introduces cycles in layers and then fix it in other layers, I don't think we should actively prevent that.

benhoyt (changes requested):

internals/plan/plan.go
```diff
@@ -641,179 +644,277 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+				Message: fmt.Sprintf(`cannot use service name %q: starting with "-" not allowed`, name),
+			}
+		}
+		if service == nil {
+			return &FormatError{
+				Message: fmt.Sprintf("service object cannot be null for service %q", name),
+			}
+		}
+		// ParseCommand needs the name to be set in order to generate a nice error message.
+		service.Name = name
```
benhoyt, line 719: I'm a bit wary about having this in the `Validate` function, which I don't think should mutate anything. Why isn't the service name set at this point? It looks like `ParseLayer` is already looping through and setting the `Name` field.

benhoyt (changes requested):
This looks great. Just one final comment/question about not mutating in `Layer.Validate`.

benhoyt (approved):
Excellent, thank you.

PR #365, opened 2024-02-22
doc(state): say why we're not implementing warnings as notices yet

benhoyt:
See comment in `warning.go` for details.


hpidcock (approved):
Thank-you for this, good to keep the record.

PR #368, opened 2024-02-23
fix(notices): replace "select=all" with "users=all"

olivercalder:
This is a backport of a fix from the snapd port of pebble notices. The original PR can be found at:

https://github.com/snapcore/snapd/pull/13630

As snapd lacks a client or cli for notices, this commit also replaces "select" with "users" in the pebble client and cli.

### DIFF
--- a/client/notices.go
+++ b/client/notices.go
@@ -77,8 +77,8 @@ func (client *Client) Notify(opts *NotifyOptions) (string, error) {
 }
 
 type NoticesOptions struct {
-	// Select allows returning broader sets of notices.
-	Select NoticesSelect
+	// Users allows returning notices for all users.
+	Users NoticesUsers
 
 	// UserID, if set, includes only notices that have this user ID or are public.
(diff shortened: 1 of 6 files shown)


benhoyt (approved):
Looks good, thanks! Just for the record, we acknowledge this is a backwards-incompatible change, however, we think it's highly unlikely anyone has started using `GET /v1/notices` especially with `select=all` yet, so now's the time to change and break it.

Copying this from the snapd PR for context.

> Per discussions with @niemeyer, @pedronis, and @olivercalder, we have decided to rename the select=all query parameter for notices. The word "select" should be used when selecting a particular subset of items from a complete set, but in the case of notices, it was being used to extend the original set of notices to include those from all other users.
>
> It has been proposed to rename select=all to users=all, and since select=all and user-id=1234 were mutually exclusive, it makes sense that a client may choose to use either user-id=1234 or users=all, as these are more clearly related.

PR #373, opened 2024-02-29
fix(daemon): improve health state lock test, remove LockCount

benhoyt:
As Harry pointed out at https://github.com/canonical/pebble/pull/369#discussion_r1505423350, there's a much simpler way to test this without needing a new State method like LockCount. Just acquire the state lock, then call the endpoint. If it times out, we know it was trying to acquire the lock.

In addition, fix an issue where the health endpoint would still hold the state lock if it returned an error. Fix those and add a test for that too.

As far as I'm concerned, this additional fix isn't critical to ship out right away, as it's extremely unlikely the health check endpoint returns an error. It can just go out in the next version of Pebble.

### DIFF
--- a/internals/daemon/api_health.go
+++ b/internals/daemon/api_health.go
@@ -35,7 +35,7 @@ func v1Health(c *Command, r *http.Request, _ *UserState) Response {
 	switch level {
 	case plan.UnsetLevel, plan.AliveLevel, plan.ReadyLevel:
 	default:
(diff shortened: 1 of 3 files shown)


benhoyt:
Going to just merge this without further review (for the 1.10.0 release) as it's test-only changes. @hpidcock: if you have any concerns I can put up another PR.

hpidcock (commented):
Looks good to me. Prefer this method rather than relying on the lock count.

PR #375, opened 2024-02-29
chore: remove deprecated functions now that we're on Go 1.16+

benhoyt:
This replaces most deprecated functions with modern equivalents, now that we're using Go 1.16+:

* Replace `ioutil.Foo` with respective functions in `io` and `os` (this is the vast majority of the changed lines).
* Replace `golang.org/x/crypto/ssh/terminal` package with `golang.org/x/term`.
* Avoid use of `ResponseRecorder.HeaderMap` and `.Header()`
* Use `signal.NotifyContext` now that it exists, instead of rolling our own.

Look at the commits individually if you want to see these separated out.

For reference, I found these usages with simple search, as well as the following [staticcheck](https://staticcheck.dev/) command (note that there are still some left that I didn't think worth updating):

```
$ staticcheck -checks SA1019 ./...
```


internals/daemon/api_test.go
```diff
@@ -92,7 +92,7 @@ func (s *apiSuite) TestSysInfo(c *check.C) {
 
 	sysInfoCmd.GET(sysInfoCmd, nil, nil).ServeHTTP(rec, nil)
 	c.Check(rec.Code, check.Equals, 200)
-	c.Check(rec.HeaderMap.Get("Content-Type"), check.Equals, "application/json")
+	c.Check(rec.Result().Header.Get("Content-Type"), check.Equals, "application/json")
```
flotter, line 95: I happen to stumble on at least one location where ```rec.Header()``` is used after the ```ServeHTTP``` to check for header content.

```
rsp.ServeHTTP(rec, req)
	hdr := rec.Header()
	c.Check(hdr.Get("Content-Disposition") ...
```

Not sure if this is a big deal, or even not correct, but while reading the docs it seams for post response header validation the ```Result()``` method should be used, not ```Header()``` directly.

flotter (approved):
Looks good to me!

internals/daemon/api_test.go
benhoyt: Good catch -- fixed!

PR #378, opened 2024-03-05
docs: add comments to clarify that WaitNotices releases the state lock

benhoyt:
This confused me when I was looking at locking stuff recently -- make it clear that `WaitNotices` releases the state lock (via the `sync.Cond`).


benhoyt:
Simple comment additions, merging without further review.

Issue #163, opened 2022-11-14
Running pebble exec makes both client and server hang on server shutdown


woky:
In latest master (a1f522390), when a pebble exec command is running, shutting down the server will make both the client and server hang.

## Steps to reproduce
The following steps are run in pebble source tree.

In terminal 1:
```
$ PEBBLE=$PWD PEBBLE_DEBUG=1 go run ./cmd/pebble run --verbose
```
In terminal 2:
```
$ PEBBLE=$PWD PEBBLE_DEBUG=1 go run ./cmd/pebble exec bash    
```

Now shutdown the server (in terminal 1) with Ctrl-C or sending SIGINT/SIGTERM directly to the pebble process. The server will send SIGKILL to the bash child, but the child will remain in zombie state. The client (terminal 2) will stop processing input, even escape sequences like ^C and ^Z, with last characters printed by bash remaining on the output.

To stop both client and server you can send SIGKILL to the server. The client will notice that websocket connection is gone and exit with failure.

It's not related to bash being started as interactive shell. The same behavior is observed when running e.g. "sleep 999" instead of bash.

benhoyt:
Good bug report -- thanks. I've noticed this when testing as well.

Implementation note: we use `exec.CommandContext` for execing the process, so there's already a context we can cancel. So we'll probably need to wire that up to a `Stop` method on `cmdstate.CommandManager` that kills the tombs / cancels the contexts on all the current executions. Currently we don't store the tomb/context on the `execution` object so we may need to add that as a field. If the manager has a `Stop()` method it looks like `StateEngine.Stop` will already call it on all managers that implement that.

benhoyt:
I spent an hour or so on this today, but this is going to require further investigation. Exec commands are tasks, and `TaskRunner` has a `Stop` which is being called correctly when the server exits. The [task's tomb appears to be wired in correctly](https://github.com/canonical/pebble/blob/9decf598487f707e752f70718ad2c3eedf79de28/internals/overlord/cmdstate/handlers.go#L118-L120) using `tomb.Context` and `exec.CommandContext`, so a SIGTERM is being sent to the process. But for some reason it's not exiting. Or is it exiting but the websockets aren't terminating correctly or something?

weiiwang01:
[...]

PR #380, opened 2024-03-11
fix: start process subreaper at top level to avoid shutdown hangs

### LINKED ISSUE #284: Daemon / Reaper challenges
Currently, if I run:

```
flotter@pc:~/pebble/internals/daemon$ ./daemon.test -check.f TestRebootHelper -check.vv
START: daemon_test.go:751: daemonSuite.TestRebootHelper
START: daemon_test.go:61: daemonSuite.SetUpTest
PASS: daemon_test.go:61: daemonSuite.SetUpTest	0.001s

START: daemon_test.go:70: daemonSuite.TearDownTest
PASS: daemon_test.go:70: daemonSuite.TearDownTest	0.000s

... Panic: internal error: reaper must be started (PC=0x46F382)
```
This issue only happens when a subset of some tests are run.

The reason for this is because when the full suite is run, reapers started in some tests (e.g. ```TestAddRoutes```) **leak** into the following tests (this is a separate issue we need to clean up). For the daemon, the reaper is started when ```daemon.New(...)``` is called (indirectly when ```overlord.New``` create a new ```servstate``` manager). The reaper is killed when ```daemon.Stop(...)``` is called.

We have a much bigger issue here, which @benhoyt discovered, which is the current ```shutdown``` implementation in the daemon does not use the reaper, but calls ```exec.Command(...).CombinedOutput()``` directly. This means that the shutdown command will cause a ```wait()``` to reap the shutdown command on termination. This is in addition to the reaper, which will also receive a SIGCHILD and there will be a race who reaps the ```shutdown``` command.

Currently the ```shutdownHandler``` can be called while the reaper is active, and after the ```daemon.Stop()``` stopped the ```overlord``` where ```servstate.Stop()``` will kill the reaper. However the most common case today is likely as the final command in ```daemon.Stop()```, at which point the reaper will be dead.

One thing to note is that currently the reaper is owned by servstate. However, this is slightly misleading, because ```PR_SET_CHILD_SUBREAPER``` is a process level directive, so applies to the entire pebble, not only servstate. 

I have a proposal that I would like to get some feedback on:

1. Can we move the reaper right to the top so that the ```pebble run``` environment has this enabled for the entire process lifecycle?
2. Only use the reaper to run commands (this may require us to adapt ```systemd.go``` later on), which for now only means moving the ```shutdown``` call over to the reaper?
3. Have the daemon and overlord related test suites get the reaper enabled and disabled with each test, and only use ```FakeCommand``` with shellcheck running in reaper mode?

I think that the reaper is in all respects the better solution we have right now (preventing zombies from taking over our kernel process tables for long running setups) when running commands, and I see no downside from simply making this our preferred method to run and test with.

@benhoyt @barrettj12 @anpep Please let me know if you agree with the issue, and which way forward you would prefer ?

benhoyt:
For example, if an exec command is running when you Ctrl-C the Pebble daemon, it hangs.

It's hanging because the ServiceManager is stopping the reaper early in the process, before TaskRunner has had a chance to abort the exec tasks (aborting an exec task sends SIGKILL to its pid via the tomb and command context).

Then when TaskRunner.Stop is called, it calls tomb.Kill and then tomb.Wait on each task (exec) tomb, and because the reaper's not running, the tomb.Wait hangs.

So the fix is to move the reaper.Start and reaper.Stop to the top level (inside the "run" command, which is also used for "enter"), instead of putting them in servstate.NewManager and ServiceManager.Stop.

After doing this, some of the tests also had to be modified to start and stop the reaper.

Fixes https://github.com/canonical/pebble/issues/163 and fixes #284

### DIFF
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -39,6 +39,7 @@ import (
 	"github.com/canonical/pebble/internals/overlord/restart"
 	"github.com/canonical/pebble/internals/overlord/standby"
 	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/reaper"
 	"github.com/canonical/pebble/internals/systemd"
 	"github.com/canonical/pebble/internals/testutil"
 )
@@ -60,6 +61,11 @@ type daemonSuite struct {
 var _ = Suite(&daemonSuite{})
 
 func (s *daemonSuite) SetUpTest(c *C) {
+	err := reaper.Start()
+	if err != nil {
+		c.Fatalf("cannot start reaper: %v", err)
+	}
+
 	s.pebbleDir = c.MkDir()
 	s.statePath = filepath.Join(s.pebbleDir, ".pebble.state")
 	systemdSdNotify = func(notif string) error {
@@ -73,6 +79,11 @@ func (s *daemonSuite) TearDownTest(c *C) {
 	s.notified = nil
 	s.authorized = false
 	s.err = nil
+
+	err := reaper.Stop()
+	if err != nil {
+		c.Fatalf("cannot stop reaper: %v", err)
+	}
 }
 
 func (s *daemonSuite) newDaemon(c *C) *Daemon {
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -29,6 +29,7 @@ import (
 	"github.com/canonical/pebble/cmd"
 	"github.com/canonical/pebble/internals/daemon"
 	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/reaper"
 	"github.com/canonical/pebble/internals/systemd"
 )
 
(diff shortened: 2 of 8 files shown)


internals/daemon/daemon_test.go
```diff
@@ -60,6 +61,11 @@ type daemonSuite struct {
 var _ = Suite(&daemonSuite{})
 
 func (s *daemonSuite) SetUpTest(c *C) {
+	err := reaper.Start()
```
flotter, line 64: I believe two instances of ```testutil.FakeCommand(c, "shutdown", "", false)``` with ```withReaper=false``` exist in the test code. Its only related to running shellcheck on the script, but I guess now that the reaper is getting the sigchilds, we should use the reaper command to start the shellchecks ?

flotter, line 64: Could this change cause the previous issue again when we call shutdown during tests (on-reaper aware)?

https://github.com/canonical/pebble/pull/279

With reaper top level, I believe we could simply replace the ```cmd.CombinedOutput``` in the daemon.go (```systemdModeReboot```) with the reaper alternative for now?

flotter (commented):
@benhoyt I've created a issue for us to also start the reaper top level, as we have a different entrypoint, and with reaper init removed at the servstate level, we will have to do this, before we move our Pebble reference forward.

internals/daemon/daemon_test.go
benhoyt: Done (also in `systemd.go`).

benhoyt: Yeah, good point. I can't stand that `shellcheck` stuff inside `go test`, particularly because it's "maybe shell check" -- it may or may not run depending on whether you've got `shellcheck` installed. IMO this kind of thing should be done in extended CI tests or not at all. I could refactor this stuff, but I'd prefer to just rip out the shellcheck bit. I'll open a separate PR for that shortly.

flotter (approved):
Looks good to me!

PR #383, opened 2024-03-14
feat(cli): add RunWithConfig, remove global variable

thp-canonical:
The goal of this PR is to allow full customization of the `client.Config` (including its `Socket`, `BaseURL` and `UserAgent` members) for callers of `cli.Run()`. This is achieved by exposing a new public function `cli.RunWithConfig(*client.Config)`.

The `clientConfig` global variable isn't accessed by any non-test code, so it can be turned into a local variable in the `Run()` function, and an additional `RunWithConfig()` function exposed that fully customize `ClientConfig`.

---

Additional information about test-related changes:

The global variable WAS used in tests (as `ClientConfig`, in `internals/cli/cli_test.go`: `cli.ClientConfig.BaseURL = server.URL`), but since all tests use `cli.RunMain()` (which is defined in `cli/export_test.go`) instead of `cli.Run()`, we can just define the global variable for the tests only, and make `RunMain` use that config object by calling `RunWithConfig`.

### DIFF
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -253,9 +253,6 @@ var (
 	osExit      = os.Exit
 )
 
-// ClientConfig is the configuration of the Client used by all commands.
-var clientConfig client.Config
-
 // exitStatus can be used in panic(&exitStatus{code}) to cause Pebble's main
 // function to exit with a given exit code, for the rare cases when you want
 // to return an exit code other than 0 or 1, or when an error return is not
@@ -269,6 +266,13 @@ func (e *exitStatus) Error() string {
 }
 
 func Run() error {
+	// the configuration of the Client used by all commands.
+	var clientConfig client.Config
+	_, clientConfig.Socket = getEnvPaths()
+	return RunWithConfig(&clientConfig)
+}
+
+func RunWithConfig(config *client.Config) error {
 	defer func() {
 		if v := recover(); v != nil {
 			if e, ok := v.(*exitStatus); ok {
@@ -280,9 +284,7 @@ func Run() error {
 
 	logger.SetLogger(logger.New(os.Stderr, fmt.Sprintf("[%s] ", cmd.ProgramName)))
 
-	_, clientConfig.Socket = getEnvPaths()
-
-	cli, err := client.New(&clientConfig)
+	cli, err := client.New(config)
 	if err != nil {
 		return fmt.Errorf("cannot create client: %v", err)
 	}
--- a/internals/cli/export_test.go
+++ b/internals/cli/export_test.go
@@ -20,9 +20,11 @@ import (
 	"github.com/canonical/pebble/client"
 )
 
-var RunMain = Run
+func RunMain() error {
+	return RunWithConfig(ClientConfig)
+}
 
-var ClientConfig = &clientConfig
+var ClientConfig = &client.Config{}
 
 func Client() *client.Client {
 	cli, err := client.New(ClientConfig)


internals/cli/cli.go
```diff
@@ -269,6 +266,13 @@ func (e *exitStatus) Error() string {
 }
 
 func Run() error {
+	// the configuration of the Client used by all commands.
+	var clientConfig client.Config
+	_, clientConfig.Socket = getEnvPaths()
+	return RunWithConfig(&clientConfig)
+}
+
+func RunWithConfig(config *client.Config) error {
```
flotter, line 275: Could it be useful, given the context of ```cli```, to introduce a new cli options struct ?

```go
type Options struct {
    client    *client.Config
}

func RunWithConfig(config *Config) error {}
```

This would, for example, allow us to then add options such as logger output in the future (```logger.SetLogger```).

flotter: We could have two runs, but its not breaking an external API if we change the one I think ?

thp-canonical: > We could have two runs, but its not breaking an external API if we change the one I think ? Its only called once in main.go I believe.

Yes, a single `Run()` that always takes a `*client.Config` would also be possible. Depending on what @benhoyt prefers, I can update the PR.

benhoyt: Yes, I agree a new options struct would be good, to allow for easy additions later. But maybe we call it `RunOptions`. And I think we can change the signature of `Run` here (seeing it's "internals") instead of adding a new function:

```go
type RunOptions struct {
    ClientConfig *client.Config
}

func Run(options *RunOptions) error { ... }
```

I think it should be okay `options` to be nil or for `ClientConfig` to be nil, in which case it uses the current defaults. So `main.go` will probably just call it as `cli.Run(nil)`.

benhoyt: It's tempting to add a `Logger logger.Logger` option while we're here (default if nil would be the current version). Yeah, let's do that. It's simple and an obvious dependency, so makes sense for it to be in the options struct.

thp-canonical: Done :)

internals/cli/export_test.go
```diff
@@ -20,9 +20,14 @@ import (
...
-var RunMain = Run
+func RunMain() error {
+	_, ClientConfig.Socket = getEnvPaths()
+	return Run(&RunOptions{
+		ClientConfig: ClientConfig,
+	})
+}
 
-var ClientConfig = &clientConfig
+var ClientConfig = &client.Config{}
```
benhoyt, line 30: Would be nice to get rid of this global at some point, but probably shouldn't be in this PR.

benhoyt (approved):

benhoyt:
Let's get @flotter's second eye on this final version, but I'm happy for this to go in -- thanks.

internals/cli/cli.go
```diff
@@ -268,7 +265,19 @@ func (e *exitStatus) Error() string {
...
+	ClientConfig *client.Config
+	Logger       logger.Logger
+}
+
+func Run(options *RunOptions) error {
+	if options == nil {
+		options = &RunOptions{
+			ClientConfig: &client.Config{},
+		}
+		_, options.ClientConfig.Socket = getEnvPaths()
```
flotter, line 278: Could it make sense to move the client config out of the zero ```RunOptions``` detector, so that Client Config and Logger is handled in the same way?
```
if options == nil {
		options = &RunOptions{}
}

if options.ClientConfig == nil {
    _, options.ClientConfig.Socket = getEnvPaths()
}
```
I am probably missing something, but it feels to me that if I pass in only a new Logger (and Client Config is a nil pointer), that should still work. In theory it should, because both ```client.New()``` and later ```newDefaultRequester()``` do create a client config if its nil, but the configuration will not work according to me, because at least a valid socket has to be setup if BaseURL is ""?

thp-canonical: Thanks, that cleans it up nicely. Done :)

flotter (approved):
Looks good! Thank you.

benhoyt (approved):

internals/cli/export_test.go
thp-canonical: PR #388 now includes removal of the global.

PR #384, opened 2024-03-14
fix: allow for the JSON overhead when reading log lines

### LINKED ISSUE #376: pebble logs fails when there is a long line (over 4k characters)
When running `pebble logs`, when there is a line with a log over 4KiB, it fails with the error (and does not show next lines):
```
$ PEBBLE_SOCKET=~/pebble/.pebble.socket go run ./cmd/pebble logs srv1
error: cannot read log line: bufio: buffer full
exit status 1
```

This is easily reproducible with a plan like this one: 
```
summary: Simple layer

description: |
    A better description for a simple layer.

services:
    srv1:
        override: replace
        summary: logs more than 4k characters
        command: bash -c "echo {1..1100}; echo "not available in logs"; sleep 10;"
        startup: enabled
```

Although long, this line size is not uncommon for a log. https://github.com/canonical/pebble/blob/95b73bb24eafa386ad4924abb57ffae9b5ef4501/client/logs.go#L30

An alternative to a bigger constant could be to print up to a fixed number of characters and skip the rest of the line.

tonyandrewmeyer:
The server caps lines at 4096 characters, but the client receives these as JSON, so when receiving there's an extra ~70 characters that make up the timestamp, service name, field names, and JSON punctuation. This means that limiting the slice size to 4096 characters there will result in buffer full errors.

Allow an extra 1024 characters for the JSON 'overhead' - this should be far more than needed, but is only used within the `logs` function - I wasn't able to find any maximum length for a service name, so a precise extra count is tricky. This could be lowered (I would guess 128 characters would usually be sufficient) if this might be too much memory in some Pebble use cases.

Also adds tests for the long log line cases.

Fixes #376

### DIFF
--- a/client/logs.go
+++ b/client/logs.go
@@ -26,8 +26,8 @@ import (
 	"time"
 )
 
-// This needs to be the size of daemon.api_logs.logReaderSize, plus extra to
-// handle the timestamp, service name, and JSON syntax.
+// This needs to be the size of daemon.logReaderSize, plus extra to handle the
+// timestamp, service name, and JSON syntax.
 const (
 	logReaderSize = 5 * 1024
 )
--- a/client/logs.go
+++ b/client/logs.go
@@ -26,8 +26,10 @@ import (
 	"time"
 )
 
+// This needs to be the size of daemon.api_logs.logReaderSize, plus extra to
+// handle the timestamp, service name, and JSON syntax.
 const (
-	logReaderSize = 4 * 1024
+	logReaderSize = 5 * 1024
 )
 
 type LogsOptions struct {
--- a/client/logs_test.go
+++ b/client/logs_test.go
@@ -20,10 +20,15 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"gopkg.in/check.v1"
 
 	"github.com/canonical/pebble/client"
+)
+
+const (
+	maxMessageSize = 4 * 1024
 )
 
 func (cs *clientSuite) TestLogsNoOptions(c *check.C) {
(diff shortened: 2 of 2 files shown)


client/logs.go
```diff
@@ -26,8 +26,10 @@ import (
 	"time"
 )
 
+// This needs to be the size of daemon.api_logs.logReaderSize, plus extra to
+// handle the timestamp, service name, and JSON syntax.
```
benhoyt, line 30: Thanks for the explanatory comment.

Nit: in Go, the filename doesn't form part of the namespace, it's just the package name. So this would be referenced as just `daemon.logReaderSize`.

client/logs_test.go
```diff
@@ -20,12 +20,17 @@ import (
...
 	"net/url"
+	"strings"
 
 	"gopkg.in/check.v1"
 
 	"github.com/canonical/pebble/client"
 )
 
+const (
+	maxMessageSize = 4 * 1024
```
benhoyt, line 31: To reduce the scope, I'd probably put this as a local const inside the test function (that works fine in Go). Also, you can use a one-liner `const maxMessageSize = 4 * 1024`.

client/logs_test.go
```diff
@@ -111,6 +116,38 @@ func (cs *clientSuite) TestLogsAll(c *check.C) {
...
 }
 
+func (cs *clientSuite) TestLogsLong(c *check.C) {
+	shortLog := `{"time":"2021-05-03T03:55:49.360994155Z","service":"thing","message":"log 1\n"}`
+	var longLog strings.Builder
+	longLog.WriteString(`{"time":"2021-05-03T03:55:49.460994155Z","service":"verbose","message":"`)
+	for i := 0; i < maxMessageSize; i++ {
+		longLog.WriteString("a")
+	}
+	longLog.WriteString(`\n"}`)
```
benhoyt, line 126: No need for a `strings.Builder` and loop here, you can just use `strings.Repeat`:

```go
longLog := `{..., "message":"` + strings.Repeat("a", maxMessageSize) + `\n"}`
```

client/logs_test.go
```diff
@@ -111,6 +116,38 @@ func (cs *clientSuite) TestLogsAll(c *check.C) {
...
+	logs := []string{shortLog, longLog.String(), ""}
+	cs.rsp = strings.Join(logs, "\n")
```
benhoyt, line 128: I was confused at first why there was an empty string in the slice (to ensure there's a trailing newline). Would just `cs.rsp = shortLog + "\n" + longLog + "\n"` be clearer?

Could we also ensure that the next log *after* the long log comes through correctly? So maybe `shortLog + longLog + shortLog` type of thing.

client/logs_test.go
```diff
@@ -111,6 +116,38 @@ func (cs *clientSuite) TestLogsAll(c *check.C) {
...
+		N:        -1,
+	})
+	c.Assert(err, check.IsNil)
+	c.Check(cs.req.Method, check.Equals, "GET")
+	c.Check(cs.req.URL.Path, check.Equals, "/v1/logs")
+	c.Check(cs.req.URL.Query(), check.DeepEquals, url.Values{
+		"n": []string{"-1"},
+	})
+	shortExpected := "2021-05-03T03:55:49.360Z [thing] log 1"
+	var longExpected strings.Builder
```
benhoyt, line 141: Similar here as above: `strings.Repeat`

benhoyt (changes requested):
Good fix and nice test structure, thanks!

I think the additional 1KB is totally fine, and a good round number that's well bigger than what any sane configuration will need. We should probably have a max length on service names, but this will do for now. We do use Pebble in "embedded Linux" contexts, but anything running a full Linux OS can handle 1KB.

benhoyt (approved):
Great, thanks for the fix!

PR #387, opened 2024-03-17
feat(planstate): create a dedicated plan manager

flotter:
The Pebble configuration YAML (called the plan) is currently managed by the service manager.

Traditionally, this design made sense as the plan primarily contained service-related configuration. However, the relationship between the plan content and the service manager has started to drift apart.

We are seeing more and more overlord managers making their way into Pebble directly, or through Pebble derivative projects.

As a consequence, the service manager (currently hosting the plan management) is getting really big and complex to test.

This PR addresses two concerns (some additional work remain):

1. Break out the plan management API currently in the service manager into its own overlord manager.

2. Consistently load the plan at Pebble startup, notifying all plan change subscribers (which now includes the service manager).

This change has a behaviour impact on Pebble: health checks will be started immediately on startup, rather than at some point later, somewhat arbitrarily, when something like a `GET /v1/services` or `POST /v1/services` or other action fetched the plan. The service manager implemented lazy loading of the configuration plan, which only resulted in the plan getting loaded from disk once a service manager command using plan is called. However, the behaviour is difficult to predict from a user perspective, see the two examples below:

**Example 1:**

// Plan
```
summary: Hello World
description: Hello world.
services:
  hello-world:
    override: replace
    startup: enabled
    command: echo "hello world"
checks:
  hello-check:
    override: merge
    exec:
      command: ping www.google.com
```
// Pebble server started: pebble run --hold
```
~/> ./pebble checks                     # Pebble client first checks request
Plan has no health checks.

~/> ./pebble services                   # Service Manager triggers plan load only now
Service      Startup  Current   Since
hello-world  enabled  inactive  -

~/> ./pebble checks                     # Pebble client second checks request
Check        Level  Status  Failures
hello-check  -      up      0/3
```
**Example 2:**

// Plan
```
summary: Hello World
description: Hello world.
services:
  hello-world:
    override: replace
    startup: enabled
    command: echo "hello world"
checks:
  hello-check:
    override: merge
    exec:
      command: ping www.google.com
```
// Pebble server started: pebble run --args hello-world -n "hello world" \; --hold
```
~/> ./pebble checks                     # Pebble client checks OK due to SetServiceArgs triggering load
Check        Level  Status  Failures
hello-check  -      up      0/3
```

**Summary:**

Following the change, the plan will always be loaded at Pebble startup with all the registered managers receiving a plan change notification with the loaded plan (which could be empty). This will consistently happen before the daemon is started, so the API experience will be consistent.

### DIFF
--- a/internals/daemon/api_checks_test.go
+++ b/internals/daemon/api_checks_test.go
@@ -43,8 +43,6 @@ checks:
             command: sleep x
 `)
 	s.daemon(c)
-	_, err := s.d.overlord.ServiceManager().Plan() // ensure plan is loaded
-	c.Assert(err, IsNil)
 
 	// Request with no filters
 	req, err := http.NewRequest("GET", "/v1/checks", nil)
@@ -129,8 +127,6 @@ checks:
 
 func (s *apiSuite) TestChecksGetInvalidLevel(c *C) {
 	s.daemon(c)
-	_, err := s.d.overlord.ServiceManager().Plan() // ensure plan is loaded
-	c.Assert(err, IsNil)
 
 	req, err := http.NewRequest("GET", "/v1/checks?level=foo", nil)
 	c.Assert(err, IsNil)
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -766,7 +766,8 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 // SetServiceArgs sets the provided service arguments to their respective services,
 // by passing the arguments to the service manager responsible under daemon overlord.
 func (d *Daemon) SetServiceArgs(serviceArgs map[string][]string) error {
-	return d.overlord.ServiceManager().SetServiceArgs(serviceArgs)
+	serviceArgsCreator := servstate.ServiceArgsLayerCreator(serviceArgs)
+	return d.overlord.PlanManager().AppendLayer(serviceArgsCreator)
 }
 
 func New(opts *Options) (*Daemon, error) {
@@ -792,6 +793,13 @@ func New(opts *Options) (*Daemon, error) {
 		d.rebootIsMissing = true
 		return d, nil
 	}
+	if err != nil {
+		return nil, err
+	}
+	// Propagate a valid plan (empty is also valid) to the subscribed
+	// overlord managers.
+	planMgr := ovld.PlanManager()
+	err = planMgr.Load()
 	if err != nil {
 		return nil, err
 	}
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -763,13 +763,10 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 	return errExpectedReboot
 }
 
-// SetServiceArgs sets the provided service arguments to their respective services
-// by using the service manager ServiceArgsLayerCreator. This layer creator derives
-// a additional layer from the current plan, containing changes to set the
-// requested service arguments.
+// SetServiceArgs updates the specified service commands by replacing
+// existing arguments with the newly specified arguments.
 func (d *Daemon) SetServiceArgs(serviceArgs map[string][]string) error {
-	serviceArgsCreator := servstate.ServiceArgsLayerCreator(serviceArgs)
-	return d.overlord.PlanManager().AppendLayer(serviceArgsCreator)
+	return d.overlord.PlanManager().SetServiceArgs(serviceArgs)
 }
 
 func New(opts *Options) (*Daemon, error) {
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -795,13 +795,6 @@ func New(opts *Options) (*Daemon, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Propagate a valid plan (empty is also valid) to the subscribed
-	// overlord managers.
-	planMgr := ovld.PlanManager()
-	err = planMgr.Load()
-	if err != nil {
-		return nil, err
-	}
 	d.overlord = ovld
 	d.state = ovld.State()
 	return d, nil
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -763,8 +763,10 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 	return errExpectedReboot
 }
 
-// SetServiceArgs sets the provided service arguments to their respective services,
-// by passing the arguments to the service manager responsible under daemon overlord.
+// SetServiceArgs sets the provided service arguments to their respective services
+// by using the service manager ServiceArgsLayerCreator. This layer creator derives
+// a additional layer from the current plan, containing changes to set the
+// requested service arguments.
 func (d *Daemon) SetServiceArgs(serviceArgs map[string][]string) error {
 	serviceArgsCreator := servstate.ServiceArgsLayerCreator(serviceArgs)
 	return d.overlord.PlanManager().AppendLayer(serviceArgsCreator)
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -33,6 +33,7 @@ import (
 	"github.com/canonical/pebble/internals/overlord/cmdstate"
 	"github.com/canonical/pebble/internals/overlord/logstate"
 	"github.com/canonical/pebble/internals/overlord/patch"
+	"github.com/canonical/pebble/internals/overlord/planstate"
 	"github.com/canonical/pebble/internals/overlord/restart"
 	"github.com/canonical/pebble/internals/overlord/servstate"
 	"github.com/canonical/pebble/internals/overlord/state"
@@ -92,6 +93,7 @@ type Overlord struct {
 	inited     bool
 	startedUp  bool
 	runner     *state.TaskRunner
+	planMgr    *planstate.PlanManager
 	serviceMgr *servstate.ServiceManager
 	commandMgr *cmdstate.CommandManager
 	checkMgr   *checkstate.CheckManager
@@ -136,18 +138,27 @@ func New(opts *Options) (*Overlord, error) {
 	}
 	o.runner.AddOptionalHandler(matchAnyUnknownTask, handleUnknownTask, nil)
 
+	o.planMgr, err = planstate.NewManager(s, o.runner, o.pebbleDir)
+	if err != nil {
+		return nil, err
+	}
+	o.stateEng.AddManager(o.planMgr)
+
 	o.logMgr = logstate.NewLogManager()
 
 	o.serviceMgr, err = servstate.NewManager(
 		s,
 		o.runner,
-		o.pebbleDir,
 		opts.ServiceOutput,
 		opts.RestartHandler,
 		o.logMgr)
 	if err != nil {
 		return nil, err
 	}
+
+	// Tell service manager about plan updates.
+	o.planMgr.AddChangeListeners(o.serviceMgr.PlanChanged)
+
 	o.stateEng.AddManager(o.serviceMgr)
 	// The log manager should be stopped after the service manager, because
 	// ServiceManager.Stop closes the service ring buffers, which signals to the
@@ -160,10 +171,10 @@ func New(opts *Options) (*Overlord, error) {
 	o.checkMgr = checkstate.NewManager()
 
 	// Tell check manager about plan updates.
-	o.serviceMgr.NotifyPlanChanged(o.checkMgr.PlanChanged)
+	o.planMgr.AddChangeListeners(o.checkMgr.PlanChanged)
 
 	// Tell log manager about plan updates.
-	o.serviceMgr.NotifyPlanChanged(o.logMgr.PlanChanged)
+	o.planMgr.AddChangeListeners(o.logMgr.PlanChanged)
 
 	// Tell service manager about check failures.
 	o.checkMgr.NotifyCheckFailed(o.serviceMgr.CheckFailed)
@@ -485,6 +496,12 @@ func (o *Overlord) CommandManager() *cmdstate.CommandManager {
 // checks under the overlord.
 func (o *Overlord) CheckManager() *checkstate.CheckManager {
 	return o.checkMgr
+}
+
+// PlanManager returns the plan manager responsible for managing the global
+// system configuration
+func (o *Overlord) PlanManager() *planstate.PlanManager {
+	return o.planMgr
 }
 
 // Fake creates an Overlord without any managers and with a backend
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -157,7 +157,7 @@ func New(opts *Options) (*Overlord, error) {
 	}
 
 	// Tell service manager about plan updates.
-	o.planMgr.AddChangeListeners(o.serviceMgr.PlanChanged)
+	o.planMgr.AddChangeListener(o.serviceMgr.PlanChanged)
 
 	o.stateEng.AddManager(o.serviceMgr)
 	// The log manager should be stopped after the service manager, because
@@ -171,10 +171,10 @@ func New(opts *Options) (*Overlord, error) {
 	o.checkMgr = checkstate.NewManager()
 
 	// Tell check manager about plan updates.
-	o.planMgr.AddChangeListeners(o.checkMgr.PlanChanged)
+	o.planMgr.AddChangeListener(o.checkMgr.PlanChanged)
 
 	// Tell log manager about plan updates.
-	o.planMgr.AddChangeListeners(o.logMgr.PlanChanged)
+	o.planMgr.AddChangeListener(o.logMgr.PlanChanged)
 
 	// Tell service manager about check failures.
 	o.checkMgr.NotifyCheckFailed(o.serviceMgr.CheckFailed)
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -199,7 +199,7 @@ func New(opts *Options) (*Overlord, error) {
 	// notifications to all notification subscribers.
 	err = o.planMgr.Load()
 	if err != nil {
-		return nil, fmt.Errorf("cannot load plan %w", err)
+		return nil, fmt.Errorf("cannot load plan: %w", err)
 	}
 
 	return o, nil
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -140,7 +140,7 @@ func New(opts *Options) (*Overlord, error) {
 
 	o.planMgr, err = planstate.NewManager(s, o.runner, o.pebbleDir)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot create plan manager: %w", err)
 	}
 	o.stateEng.AddManager(o.planMgr)
 
@@ -153,7 +153,7 @@ func New(opts *Options) (*Overlord, error) {
 		opts.RestartHandler,
 		o.logMgr)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot create service manager: %w", err)
 	}
 
 	// Tell service manager about plan updates.
@@ -182,7 +182,7 @@ func New(opts *Options) (*Overlord, error) {
 	if o.extension != nil {
 		extraManagers, err := o.extension.ExtraManagers(o)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("cannot add extra managers: %w", err)
 		}
 		for _, manager := range extraManagers {
 			o.stateEng.AddManager(manager)
@@ -199,7 +199,7 @@ func New(opts *Options) (*Overlord, error) {
 	// notifications to all notification subscribers.
 	err = o.planMgr.Load()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot load plan %w", err)
 	}
 
 	return o, nil
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -194,6 +194,14 @@ func New(opts *Options) (*Overlord, error) {
 	// before it.
 	o.stateEng.AddManager(o.runner)
 
+	// Load the plan from the Pebble layers directory (which may be missing
+	// or have no layers, resulting in an empty plan), and propagate PlanChanged
+	// notifications to all notification subscribers.
+	err = o.planMgr.Load()
+	if err != nil {
+		return nil, err
+	}
+
 	return o, nil
 }
 
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -1,0 +1,192 @@
+package planstate
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/plan"
+)
+
+// LabelExists is the error returned by AppendLayer when a layer with that
+// label already exists.
+type LabelExists struct {
+	Label string
+}
+
+func (e *LabelExists) Error() string {
+	return fmt.Sprintf("layer %q already exists", e.Label)
+}
+
+type PlanManager struct {
+	state     *state.State
+	runner    *state.TaskRunner
+	pebbleDir string
+
+	planLock     sync.Mutex
+	plan         *plan.Plan
+	planHandlers []PlanChangeFunc
+}
+
+func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string) (*PlanManager, error) {
+	manager := &PlanManager{
+		state:     s,
+		runner:    runner,
+		pebbleDir: pebbleDir,
+	}
+
+	return manager, nil
+}
+
+// Load reads plan layers from the pebble directory, combine and validate the
+// final plan, and finally notifies registered managers up the plan update.
+func (m *PlanManager) Load() error {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
+	plan, err := plan.ReadDir(m.pebbleDir)
+	if err != nil {
+		return err
+	}
+	m.planChanged(plan)
+	return nil
+}
+
+// PlanChangeFunc is the type of function used by AddChangeListeners.
+type PlanChangeFunc func(p *plan.Plan)
+
+// AddChangeListeners adds f to the list of functions that are called whenever
+// the plan has changed. This method may not be called once the overlord state
+// engine has started.
+func (m *PlanManager) AddChangeListeners(f PlanChangeFunc) {
+	m.planHandlers = append(m.planHandlers, f)
+}
+
+func (m *PlanManager) planChanged(plan *plan.Plan) {
+	m.plan = plan
+	for _, f := range m.planHandlers {
+		f(plan)
+	}
+}
+
+// Plan returns the configuration plan. Any change made to the plan will
+// result in a new Plan instance, so the current design assumes a returned
+// plan is never mutated by planstate.
+func (m *PlanManager) Plan() (*plan.Plan, error) {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
+	return m.plan, nil
+}
+
+// LayerCreator allows an Append or Combine operation on the plan to include
+// layer creation derived from the plan itself, which has to happen while
+// the plan is locked.
+type LayerCreator interface {
+	Layer(planIn *plan.Plan) (*plan.Layer, error)
+}
+
+// AppendLayer takes a LayerCreator interface, which may be an existing plan
+// Layer, and appends it to the plan's layers and updates the layer.Order
+// field to the new order. If a layer with layer.Label already exists, return
+// an error of type *LabelExists.
+func (m *PlanManager) AppendLayer(creator LayerCreator) error {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
+
+	layer, err := creator.Layer(m.plan)
+	if err != nil {
+		return err
+	}
+
+	index, _ := findLayer(m.plan.Layers, layer.Label)
+	if index >= 0 {
+		return &LabelExists{Label: layer.Label}
+	}
+
+	return m.appendLayer(layer)
+}
+
+// CombineLayer takes a LayerCreator interface, which may be an existing plan
+// Layer, and appends it to an existing layer that has the same label. If no
+// existing layer has the label, append a new one. In either case, update the
+// layer.Order field to the new order.
+func (m *PlanManager) CombineLayer(creator LayerCreator) error {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
+
+	layer, err := creator.Layer(m.plan)
+	if err != nil {
+		return err
+	}
+
+	index, found := findLayer(m.plan.Layers, layer.Label)
+	if index < 0 {
+		// No layer found with this label, append new one.
+		return m.appendLayer(layer)
+	}
+
+	// Layer found with this label, combine into that one.
+	combined, err := plan.CombineLayers(found, layer)
+	if err != nil {
+		return err
+	}
+	combined.Order = found.Order
+	combined.Label = found.Label
+
+	// Insert combined layer back into plan's layers list.
+	newLayers := make([]*plan.Layer, len(m.plan.Layers))
+	copy(newLayers, m.plan.Layers)
+	newLayers[index] = combined
+	err = m.updatePlanLayers(newLayers)
+	if err != nil {
+		return err
+	}
+	layer.Order = found.Order
+	return nil
+}
+
+func (m *PlanManager) appendLayer(layer *plan.Layer) error {
+	newOrder := 1
+	if len(m.plan.Layers) > 0 {
+		last := m.plan.Layers[len(m.plan.Layers)-1]
+		newOrder = last.Order + 1
+	}
+
+	newLayers := append(m.plan.Layers, layer)
+	err := m.updatePlanLayers(newLayers)
+	if err != nil {
+		return err
+	}
+	layer.Order = newOrder
+	return nil
+}
+
+func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) error {
+	combined, err := plan.CombineLayers(layers...)
+	if err != nil {
+		return err
+	}
+	plan := &plan.Plan{
+		Layers:     layers,
+		Services:   combined.Services,
+		Checks:     combined.Checks,
+		LogTargets: combined.LogTargets,
+	}
+	m.planChanged(plan)
+	return nil
+}
+
+// findLayer returns the index (in layers) of the layer with the given label,
+// or returns -1, nil if there's no layer with that label.
+func findLayer(layers []*plan.Layer, label string) (int, *plan.Layer) {
+	for i, layer := range layers {
+		if layer.Label == label {
+			return i, layer
+		}
+	}
+	return -1, nil
+}
+
+// Ensure implements StateManager.Ensure.
+func (m *PlanManager) Ensure() error {
+	return nil
+}
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -40,7 +40,9 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string) (*Pl
 }
 
 // Load reads plan layers from the pebble directory, combine and validate the
-// final plan, and finally notifies registered managers up the plan update.
+// final plan, and finally notifies registered managers of the plan update. In
+// the case of a non-exist layers directory, or no layers in the layers
+// directory, an empty plan is announced to change subscribers.
 func (m *PlanManager) Load() error {
 	m.planLock.Lock()
 	defer m.planLock.Unlock()
@@ -52,13 +54,16 @@ func (m *PlanManager) Load() error {
 	return nil
 }
 
-// PlanChangedFunc is the type of function used by AddChangeListener.
+// PlanChangedFunc is the function type used by AddChangeListener.
 type PlanChangedFunc func(p *plan.Plan)
 
 // AddChangeListener adds f to the list of functions that are called whenever
-// the plan has changed. This method may not be called once the overlord state
-// engine has started.
+// a plan change event took place (Load, AppendLayer, CombineLayer). A plan
+// change event does not guarantee that combined plan content has changed.
+// Notification registration must be completed before the plan is loaded.
 func (m *PlanManager) AddChangeListener(f PlanChangedFunc) {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
 	m.planHandlers = append(m.planHandlers, f)
 }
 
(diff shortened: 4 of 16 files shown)


internals/daemon/daemon.go
```diff
@@ -766,7 +766,8 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 // SetServiceArgs sets the provided service arguments to their respective services,
 // by passing the arguments to the service manager responsible under daemon overlord.
 func (d *Daemon) SetServiceArgs(serviceArgs map[string][]string) error {
-	return d.overlord.ServiceManager().SetServiceArgs(serviceArgs)
+	serviceArgsCreator := servstate.ServiceArgsLayerCreator(serviceArgs)
+	return d.overlord.PlanManager().AppendLayer(serviceArgsCreator)
```
flotter, line 770: Please see the ```LayerCreator``` interface in the Plan Manager. There is a general need that ```SetServiceArgs``` as highlighted. Managers have the knowledge on how to change a part of the plan that they understand. However, these changes must happen atomically, so it must be performed while the Plan Manager maintains a lock to block updates of the plan while this mutation function is creating and applying the layer. This is what ```LayerCreator``` tries to solve. It completely compatible with the well known ```AppendLayer``` and ```CombineLayer``` methods we know well.

internals/daemon/daemon.go
```diff
@@ -795,6 +796,13 @@ func New(opts *Options) (*Daemon, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Propagate a valid plan (empty is also valid) to the subscribed
+	// overlord managers.
+	planMgr := ovld.PlanManager()
+	err = planMgr.Load()
+	if err != nil {
+		return nil, err
+	}
```
flotter, line 805: Since the plan is not associated with its own manager, I've had to remove the lazy loading that happened in the previous design. We now have (and in the future will have) many managers that each may have APIs which could require them to have a ```plan```. I've decided to have a simple flow of updates from the Plan Manager to the plan change subscribers, so we need to have the Plan Manager get the plan as soon as possible, and update the consumers as soon as possible so they can function.

The extreme example is the ```SetServiceArgs``` call that happens even before the overlord loop has started. This prompted me to move this immediately after ```overlord.New()```. I know all the subscriptions are done, so I may as well proceed with loading the plan from disk.

internals/overlord/planstate/manager.go
flotter, line 1: I will add the unit tests for this package soon ASAP.

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,192 @@
...
+	defer m.planLock.Unlock()
+	return m.plan, nil
+}
+
+// LayerCreator allows an Append or Combine operation on the plan to include
+// layer creation derived from the plan itself, which has to happen while
+// the plan is locked.
+type LayerCreator interface {
+	Layer(planIn *plan.Plan) (*plan.Layer, error)
+}
```
flotter, line 85: The only real functional addition to the plan related code moved out of the Service Manager into the Plan Manager.

internals/overlord/servstate/manager.go
```diff
@@ -75,165 +59,77 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string, serv
...
+// arguments, as provided by "pebble run --args" CLI command. It
+// implements the LayerCreator interface, allowing the type to be used with
+// plan manager's (planstate) AppendLayer or CombineLayer methods, to apply
+// the changes to the plan.
+type ServiceArgsLayerCreator map[string][]string
+
+// Layer presents the requested service argument changes in Layer form, ready
+// to apply to the plan. The planstate lock will be acquired while this method
+// is called, so no additional locking needed.
+func (serviceArgs ServiceArgsLayerCreator) Layer(planIn *plan.Plan) (*plan.Layer, error) {
```
flotter, line 95: This is the example of the ```LayerCreator``` compatible type implemented by the Service Manager. It basically contains the code from ```SetServiceArgs```, but allows the type to be used by the traditional ```AppendLayer``` and ```CombineLayer``` directly, in a way that allows for atomic custom layer creation.

internals/overlord/servstate/manager.go
```diff
@@ -267,22 +163,18 @@ const (
...
 func (m *ServiceManager) Services(names []string) ([]*ServiceInfo, error) {
-	mgrPlan, err := m.Plan()
-	if err != nil {
-		return nil, err
-	}
-
 	m.servicesLock.Lock()
 	defer m.servicesLock.Unlock()
 
+	planSnapshot := m.plan()
```
flotter, line 169: I argued about this with myself and decided that since the ```servicesLock``` is likely more contested, and the ```planLock``` practically not really ever contested (due to time it takes to copy a pointer), that although never wrong, I convinced myself the service manager services and plan would be closer in sync this way around when the API is used concurrently between service and plan calls. I could be completely wrong.

internals/overlord/servstate/manager_test.go
flotter, line 1: A little bit more serious surgery here, but I believe its for a good cause ...

Fundamentally, I removed code that tests concepts such as how layering works, how AppendLayer or CombineLayer works, or if the end result match a specific Marshalled output. These things will move to the Plan Manager unit tests.

So I do not want to make this test package depend on the Plan Manager (in the same way ```checkstate``` and ```logtargets``` do not import the Service Manager today for testing reasons).

So I did a whole bit of simplification, cleanup and a one timing related issue. Let me know what you think :)

flotter, line 1: A little bit more serious surgery here, but I believe its for a good cause ...

Fundamentally, I removed code that tests concepts such as how layering works, how AppendLayer or CombineLayer works, or if the end result match a specific Marshalled output. These things will move to the Plan Manager unit tests.

So I do not want to make this test package depend on the Plan Manager (in the same way checkstate and logtargets do not import the Service Manager today for testing reasons).

So I did a whole bit of simplification, cleanup and a one timing related issue. Let me know what you think :)

internals/overlord/servstate/manager_test.go
```diff
@@ -223,15 +186,59 @@ func (s *S) TearDownTest(c *C) {
...
 }
 
+// TestPlanLayer and other plan snippets in this test package use the done
+// check mechanism to have a predictable way to know when the expected service
+// side effect (e.g. stdout or stderr) can be checked. This is required
+// because the service manager can only tell when the service started, not
+// when the expected result has been achieved. The done check barrier is
+// inserted by adding the $NOTIFY_DONE_CHECK environment variable in the
+// service command sequence, at the point where the expected side effect
+// will be observed by the test result checker.
```
flotter, line 196: I simplified the done check mechanism significantly.

internals/overlord/servstate/manager_test.go
```diff
@@ -547,239 +600,14 @@ func (s *S) TestStartFastExitCommand(c *C) {
 	c.Assert(svc.Current, Equals, servstate.StatusInactive)
 }
 
-func (s *S) TestPlan(c *C) {
```
flotter, line 550: Moved to Plan Manager.

internals/overlord/servstate/manager_test.go
```diff
@@ -547,239 +600,14 @@ func (s *S) TestStartFastExitCommand(c *C) {
...
-        override: replace
-        command: /bin/bash
-    svc2:
-        override: replace
-        command: /bin/foo
-`[1:])
-	s.planLayersHasLen(c, s.manager, 3)
-}
-
-func (s *S) TestCombineLayer(c *C) {
```
flotter, line 658: Moved to Plan Manager.

internals/overlord/servstate/manager_test.go
```diff
@@ -547,239 +600,14 @@ func (s *S) TestStartFastExitCommand(c *C) {
...
-    test5:
-        override: replace
-        command: /bin/sh -c "sleep 10"
-        user: nobody
-        group: nogroup
-`[1:], s.insertDoneCheck(c, "test1"), s.insertDoneCheck(c, "test2"))
-	c.Assert(planYAML(c, s.manager), Equals, expected)
-}
-
-func (s *S) TestAppendLayer(c *C) {
```
flotter, line 580: Moved to Plan Manager.

internals/overlord/servstate/manager_test.go
```diff
@@ -547,239 +600,14 @@ func (s *S) TestStartFastExitCommand(c *C) {
...
-	c.Check(err, ErrorMatches, `(?s).*plan check.*must be "alive" or "ready".*`)
-}
-
+// TestSetServiceArgs does not involve the service manager plan snapshot as
+// the plan management is no longer part of the service manager. However,
+// each manager still needs the ability to provide a way to create a custom
+// layer that can be used to mutate the plan state. This test focusses on
+// testing the mutation logic of the service, and not how the changes are
+// applied to planstate.
 func (s *S) TestSetServiceArgs(c *C) {
```
flotter, line 609: Removed parts of the test that is not strictly concerned with the Service Manager Code. The Plan Manager will have enough tests about layering, appending and combining and checking the end marshal result.

internals/overlord/servstate/manager_test.go
```diff
@@ -902,91 +730,74 @@ PEBBLE_ENV_TEST_PARENT=from-parent
 }
 
 func (s *S) TestActionRestart(c *C) {
```
flotter, line 732: A little naughty drive by improvement. I still got timing related failures here, so I improved the test.

internals/plan/plan.go
```diff
@@ -822,6 +822,13 @@ func (layer *Layer) Validate() error {
 	return nil
 }
 
+// Layer implements the LayerCreator interface exposed by the plan state
+// manager. In this case it is simply a wrapper to make a already created
+// layer compatible with the interface.
+func (layer *Layer) Layer(planIn *Plan) (*Layer, error) {
+	return layer, nil
+}
```
flotter, line 830: Make a layer, such as returned by ```ParseLayer``` compatible with ```LayerCreator``` on its own. This allows traditional use of a simple layer with AppendLayer and CombineLayer.

internals/daemon/daemon.go
```diff
@@ -766,7 +766,8 @@ func (d *Daemon) RebootIsMissing(st *state.State) error {
 // SetServiceArgs sets the provided service arguments to their respective services,
 // by passing the arguments to the service manager responsible under daemon overlord.
```
thp-canonical, line 767:
```suggestion
// by passing the arguments to the plan manager responsible under daemon overlord.
```

internals/daemon/daemon.go
```diff
@@ -795,6 +796,13 @@ func New(opts *Options) (*Daemon, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Propagate a valid plan (empty is also valid) to the subscribed
+	// overlord managers.
+	planMgr := ovld.PlanManager()
+	err = planMgr.Load()
+	if err != nil {
+		return nil, err
```
thp-canonical, line 804: No need to wrap the error here, as we want to report the error from the plan manager directly to the caller, right?

internals/overlord/overlord.go
```diff
@@ -136,18 +138,27 @@ func New(opts *Options) (*Overlord, error) {
 	}
 	o.runner.AddOptionalHandler(matchAnyUnknownTask, handleUnknownTask, nil)
 
+	o.planMgr, err = planstate.NewManager(s, o.runner, o.pebbleDir)
+	if err != nil {
+		return nil, err
```
thp-canonical, line 143: No need to wrap the error here?

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,192 @@
...
+
+// PlanChangeFunc is the type of function used by AddChangeListeners.
+type PlanChangeFunc func(p *plan.Plan)
+
+// AddChangeListeners adds f to the list of functions that are called whenever
+// the plan has changed. This method may not be called once the overlord state
+// engine has started.
+func (m *PlanManager) AddChangeListeners(f PlanChangeFunc) {
+	m.planHandlers = append(m.planHandlers, f)
+}
```
thp-canonical, line 62: *nit*: `AddChangeListener` (we only add one listener) -- or if you want to have plural, it should probably be  `AddToChangeListeners` (we add one thing to a list of listeners)

Also, "This method may not be called once the overlord state engine has started" -- is there a way to put this restriction into code (e.g. check if it has started, and fail somehow instead of silently ignoring it)?

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,192 @@
...
+	plan, err := plan.ReadDir(m.pebbleDir)
+	if err != nil {
+		return err
+	}
+	m.planChanged(plan)
+	return nil
+}
+
+// PlanChangeFunc is the type of function used by AddChangeListeners.
+type PlanChangeFunc func(p *plan.Plan)
```
thp-canonical, line 55: *nit*: `PlanChangedFunc` (it's a function that gets called when the plan changed, not a function to change the plan) - the function on `PlanManager` is also called `planChanged`

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,192 @@
...
+func (m *PlanManager) planChanged(plan *plan.Plan) {
+	m.plan = plan
+	for _, f := range m.planHandlers {
+		f(plan)
+	}
+}
+
+// Plan returns the configuration plan. Any change made to the plan will
+// result in a new Plan instance, so the current design assumes a returned
+// plan is never mutated by planstate.
```
thp-canonical, line 73: ..and also that any caller of `Plan()` doesn't mutate the value pointed-to by the returned `*plan.Plan`?

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,192 @@
...
+		Checks:     combined.Checks,
+		LogTargets: combined.LogTargets,
+	}
+	m.planChanged(plan)
+	return nil
+}
+
+// findLayer returns the index (in layers) of the layer with the given label,
+// or returns -1, nil if there's no layer with that label.
+func findLayer(layers []*plan.Layer, label string) (int, *plan.Layer) {
```
thp-canonical, line 180: *nit*: Is that idiomatic Go (returning `-1` meaning "error") or would a return type of `(int, *plan.Layer, error)` be more Go-ish?

However, I see this is mostly just migrated from old code, so feel free to ignore.

thp-canonical (commented):
Just a few minor comments for consideration.

internals/daemon/daemon.go
flotter: Fixed, thank you!

internals/daemon/daemon.go
flotter: I typically wrap an error if it adds context that could be useful when interpreting the error, especially if its a library kind of call that could be use from multiple contexts. I will give this another thought, but I suspect the error context will be clear by the time it exists a public overlord manager method (or else I need to fix it).

internals/overlord/planstate/manager.go
flotter: Added. Resolving. The tests came from servstate, but now working as part of the plan manager.

internals/overlord/planstate/manager.go
flotter: Fixed. Thank you. Resolving.

internals/overlord/planstate/manager.go
flotter: Fixed. Thank you. Resolving.

internals/overlord/planstate/manager.go
flotter: Added, thank you. Yes. Resolving.

internals/overlord/planstate/manager.go
flotter: This is taken as is, so in general I would not change existing functional code, unless something is broken (or a real need to do so). I think its perhaps taking inspiration from https://pkg.go.dev/strings#Index. We can of course improve it if needed, but feels like a separate PR as it will affect some code and tests. Going to resolve this one for now, just to keep the current discussion points focused on only plan manager changes.

internals/overlord/servstate/manager.go
flotter: Removed.

internals/overlord/planstate/manager.go
flotter: Removed.

internals/daemon/daemon.go
flotter: moved.

internals/daemon/daemon.go
flotter: Removed.

internals/overlord/overlord.go
```diff
@@ -183,6 +194,14 @@ func New(opts *Options) (*Overlord, error) {
 	// before it.
 	o.stateEng.AddManager(o.runner)
 
+	// Load the plan from the Pebble layers directory (which may be missing
+	// or have no layers, resulting in an empty plan), and propagate PlanChanged
+	// notifications to all notification subscribers.
+	err = o.planMgr.Load()
+	if err != nil {
+		return nil, err
+	}
```
flotter, line 203: Since the plan is now associated with its own manager, I've had to remove the lazy loading that happened in the previous design. We now have (and in the future will have) many managers that each may have APIs which could require them to have a plan. I've decided to have a simple flow of updates from the Plan Manager to the plan change subscribers, so we need to have the Plan Manager get the plan as soon as possible, and update the consumers as soon as possible so they can function.

flotter:
@rebornplusplus @cjdcordeiro I am not sure how to add you, but could you kindly give this PR a careful look w.r.t the Rocks entrypoint perspective? I want to make sure your use of ```--hold``` and ```--args``` are not affected.

internals/overlord/servstate/manager_test.go
flotter: Removed.

benhoyt:
@flotter  I haven't reviewed the code change yet (going to start on that today), but I just talked to the APAC Juju team, and they're happy with the change in behaviour this would mean: that health checks will start when Pebble starts, instead of an arbitrary interval afterwards depending on what the charm or whoever else was doing with Pebble. This was always their expectation (and I think charmer's expectation) in any case.

Separately from this, we'd like to consider a proper "startup" check level for K8s startup probes, so that charms like PostgreSQL can properly say "I'm up" to Pebble/Juju, rather than relying on a longish health check timeout / # of failures. But that's a separate discussion -- I've put it in our list to discuss in Madrid.

internals/overlord/servstate/manager.go
```diff
@@ -416,18 +256,14 @@ func (m *ServiceManager) ServiceLogs(services []string, last int) (map[string]se
...
 func (m *ServiceManager) Replan() ([]string, []string, error) {
-	mgrPlan, err := m.Plan()
-	if err != nil {
-		return nil, nil, err
-	}
-
 	m.servicesLock.Lock()
 	defer m.servicesLock.Unlock()
 
+	planSnapshot := m.plan()
```
benhoyt, line 262: Similar here -- let's move this back up so the plan lock is not held at the same time as the services lock.

internals/overlord/servstate/manager.go
```diff
@@ -550,19 +346,15 @@ func (m *ServiceManager) SetServiceArgs(serviceArgs map[string][]string) error {
...
-	mgrPlan, err := m.Plan()
-	if err != nil {
-		return nil, err
-	}
-
+	planSnapshot := m.plan()
 	// Get all service names in plan.
-	services := make([]string, 0, len(mgrPlan.Services))
-	for name := range mgrPlan.Services {
+	var services []string
```
benhoyt, line 351: Why the change here? The `make(T, 0, cap)` idiom pre-allocates a slice with the correct capacity, avoiding log(N) allocates in the append loop. Not that low-level performance is critical here, but doesn't hurt to leave it.

internals/overlord/servstate/manager_test.go
```diff
@@ -1237,9 +1019,13 @@ checks:
...
-`, tempFile, s.insertDoneCheck(c, "test2")))
-	err := s.manager.AppendLayer(layer)
-	c.Assert(err, IsNil)
+`
+	s.planAddLayer(c, fmt.Sprintf(
+		layer,
+		tempFile,
+	))
+	s.planChanged(c)
+	checkMgr.PlanChanged(s.plan)
```
benhoyt, line 1028: Why is this manual call to `CheckManager.PlanChanged` needed now? (Here and elsewhere.)

internals/overlord/servstate/manager_test.go
```diff
@@ -2146,10 +1950,21 @@ func (f fakeLogManager) ServiceStarted(service *plan.Service, logs *servicelog.R
...
-	return fmt.Sprintf("sync; touch %s", filepath.Join(s.dir, service))
+// insertDoneChecks modifies layer service commands which contains a
+// $NOTIFY_DONE_CHECK barrier placeholder. The placeholder is replaced
+// with a command which writes a service specific file to a test
+// directory, allowing waitForDoneCheck to detect service side effect
+// completion.
+func (s *S) insertDoneChecks(c *C, layer *plan.Layer) {
+	for _, service := range layer.Services {
+		doneCheck := fmt.Sprintf("sync; touch %s", filepath.Join(s.dir, service.Name))
+		service.Command = strings.Replace(service.Command, "$NOTIFY_DONE_CHECK", doneCheck, -1)
```
benhoyt, line 1961: I don't love this, as it looks like an environment variable, whereas it's really a template variable substitution. Can we just explicitly add it to the list of fmt.Sprintf arguments (most of the layers seem to be built using Sprintf anyway), like `fmt.Sprintf("... %s ...", s.doneCheck("svc1"))` or similar?

Alternatively, if you want to stick with the current approach, let's use the string `{{.NotifyDoneCheck}}` as we've done elsewhere in Pebble with `{{.ProgramName}}` (Go template syntax -- though we don't have to use a full Go template).

benhoyt (changes requested):
This is excellent work, thanks Fred. A bunch of comments, almost all of them minor.

internals/daemon/api_checks_test.go
```diff
@@ -43,8 +43,6 @@ checks:
             command: sleep x
 `)
 	s.daemon(c)
-	_, err := s.d.overlord.ServiceManager().Plan() // ensure plan is loaded
```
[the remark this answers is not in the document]
flotter: :) resolving.

internals/overlord/overlord.go
flotter: Yes, @thp-canonical your comment made sense all along. I recently reviewed another PR and realised that "common" context makes sense at the consumer side, not inside the function, otherwise you would have tons of error suffixes all starting with "cannot create plan manager" ...

flotter: Fixed!

internals/overlord/overlord.go
```diff
@@ -183,6 +194,14 @@ func New(opts *Options) (*Overlord, error) {
 	// before it.
 	o.stateEng.AddManager(o.runner)
 
+	// Load the plan from the Pebble layers directory (which may be missing
+	// or have no layers, resulting in an empty plan), and propagate PlanChanged
+	// notifications to all notification subscribers.
+	err = o.planMgr.Load()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
flotter: I changed this to: ```cannot load plan: %w```. Would you prefer to expose the concept of layers here ? It may fail on final combined layer validation, for example, so I was thinking perhaps just this is fine ?

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,225 @@
+package planstate
```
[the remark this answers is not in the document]
flotter: Oopsie. Fixed. (also date in manager_test.go)

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,225 @@
...
+		plan:      &plan.Plan{},
+	}
+
+	return manager, nil
+}
+
+// Load reads plan layers from the pebble directory, combine and validate the
+// final plan, and finally notifies registered managers of the plan update. In
+// the case of a non-exist layers directory, or no layers in the layers
+// directory, an empty plan is announced to change subscribers.
```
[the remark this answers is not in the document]
flotter: Thanks. Fixed. Resolving.

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,225 @@
...
+func (m *PlanManager) Load() error {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
+	plan, err := plan.ReadDir(m.pebbleDir)
+	if err != nil {
+		return err
+	}
+	m.planChanged(plan)
```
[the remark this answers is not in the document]
flotter: Yes this is a really good point. I will post another PR shortly.

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	m.plan = plan
+	for _, f := range m.planHandlers {
+		f(plan)
+	}
+}
+
+// Plan returns the combined configuration plan. Any change made to the plan
+// will result in a new Plan instance, so the current design assumes a returned
+// plan is never mutated by planstate (and may never be mutated by any
+// consuming overlord manager).
```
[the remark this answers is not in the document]
flotter: Improved as suggested. Resolving.

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return err
+	}
+	plan := &plan.Plan{
+		Layers:     layers,
+		Services:   combined.Services,
+		Checks:     combined.Checks,
+		LogTargets: combined.LogTargets,
+	}
+	m.planChanged(plan)
```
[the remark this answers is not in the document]
flotter: Sorry I really messed up here. This was a rebase merge conflict I resolved incorrectly. I did another comparison and also fixed another issue in SetServiceArgs.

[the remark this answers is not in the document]
flotter: Resolving.

internals/overlord/servstate/handlers.go
```diff
@@ -114,11 +114,8 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 		return err
 	}
 
-	mgrPlan, err := m.Plan()
-	if err != nil {
-		return fmt.Errorf("cannot fetch plan: %w", err)
-	}
-	config, ok := mgrPlan.Services[request.Name]
+	planSnapshot := m.plan()
```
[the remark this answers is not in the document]
flotter: I made it ```currentPlan```. Could that work for you? Since the ```internals/plan``` package is already present / imported, i've tried to not use plan directly as well.

internals/overlord/servstate/manager.go
```diff
@@ -16,13 +16,11 @@ import (
...
-	runner    *state.TaskRunner
-	pebbleDir string
+	state  *state.State
+	runner *state.TaskRunner
 
-	planLock     sync.Mutex
-	plan         *plan.Plan
-	planHandlers []PlanFunc
+	planLock   sync.Mutex
+	planUnsafe *plan.Plan
```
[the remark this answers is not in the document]
flotter: Very true. I named this ```plan``` as you suggested. Works now, thanks. Resolving.

internals/overlord/servstate/manager.go
```diff
@@ -75,6 +59,29 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string, serv
...
+// PlanChanged informs the service manager that the plan has been updated.
+func (m *ServiceManager) PlanChanged(plan *plan.Plan) {
+	m.planLock.Lock()
+	defer m.planLock.Unlock()
+	m.planUnsafe = plan
+}
+
+// Plan safely gets a snapshot of the current plan pointer. Plan consumers do
+// not mutate the snapshot, and only react to updates from the Plan Manager.
+func (m *ServiceManager) plan() *plan.Plan {
```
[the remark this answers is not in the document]
flotter: Improved. Resolving, thank you.

internals/overlord/servstate/manager.go
flotter: No problem. I think you have a good point, thank you. Reverted in both places.

flotter: Resolving.

internals/overlord/servstate/manager.go
flotter: Sorry about this Ben. I did not even recall I did this, so not something I would normally have changed :( This happened during a rebase operation, and I think I was just a little bit too hasty, and did not pay attention to the conflict I was trying to resolve. I should take more care!

flotter: Fixed!

flotter: Resolving.

internals/overlord/servstate/manager.go
flotter: Fixed. Resolving.

internals/daemon/daemon.go
flotter: Thanks! I fixed this and improved the errors from the other managers initialised on overlord.New(). Resolving.

flotter: Since the error comes from the internals/plan library, I wrapped it with something more plan related. The internal error could be "validation failed", but it would be useful to know the plan manager failed to load the plan.

flotter: Resolving.

internals/overlord/overlord.go
flotter: Read the other comment. Thanks. Resolving this comment.

internals/overlord/servstate/manager_test.go
flotter: This is because the service manager is no longer also a plan manager, and so the action that would give the check manager a plan would be the plan manager calling ```CheckManager.PlanChanged```. These calls are emulating the plan manager receiving the plan update, and propagating this to the service manager and check manager. Does this sound ok ?

internals/overlord/servstate/manager_test.go
flotter: ```{{.NotifyDoneCheck}}``` I've adopted this.

The reason for not using %s (deliberately removing this) was because the string snippets requiring the checks typically require insertion in more than one place, and also includes other %s fields as well. The end result was long Sprintf statements, which was difficult to follow.

The actual test code, in my opinion, looks much cleaner now - I think :) But let me know if this feels OK now. If you prefer the previous approach, no problem at all - happy to revert.

flotter: Let me know how you feel about this please :) ?

internals/overlord/servstate/handlers.go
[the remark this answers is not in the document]
benhoyt: Yep, `currentPlan` is fine, thanks.

internals/overlord/servstate/manager.go
benhoyt: No problem, and no need to apologize! You're one of the most careful programmers I've seen. :-)

internals/overlord/servstate/manager_test.go
benhoyt: Yep, this is good, thanks.

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return err
+	}
+	p := &plan.Plan{
+		Layers:     layers,
+		Services:   combined.Services,
+		Checks:     combined.Checks,
+		LogTargets: combined.LogTargets,
+	}
+	p.Validate()
```
benhoyt, line 176:
```suggestion
	err = p.Validate()
```

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,238 @@
...
+	for _, f := range m.planHandlers {
+		f(plan)
+	}
+}
+
+// Plan returns the combined configuration plan. Any change made to the plan
+// will result in a new Plan instance, so the current design assumes a returned
+// plan is never mutated by planstate (and may never be mutated by any
+// consumer).
+func (m *PlanManager) Plan() (*plan.Plan, error) {
```
benhoyt, line 95: Can we just return the `*Plan` and no error? This can never return an error now, by design, so it seems unnecessary/awkward to have the caller have to check one.
```suggestion
func (m *PlanManager) Plan() *plan.Plan {
```

internals/overlord/planstate/manager.go
```diff
@@ -0,0 +1,238 @@
...
+// Ensure implements StateManager.Ensure.
+func (m *PlanManager) Ensure() error {
+	return nil
+}
+
+// SetServiceArgs sets the service arguments provided by "pebble run --args"
+// to their respective services. It adds a new layer in the plan, the layer
+// consisting of services with commands having their arguments changed.
+//
+// TODO: This functionality must be redesigned (moved out of the plan manager)
```
benhoyt, line 204: I've been trying to reserve "TODO" for things that really need to be done soon or bug fixes (though I realise that's not always how it's used). This is sligtly non-optimal, but I'd rather say:
```suggestion
// NOTE: This functionality should be redesigned (moved out of the plan manager)
```

internals/overlord/servstate/manager_test.go
```diff
@@ -420,24 +453,31 @@ func (s *S) TestStartBadCommand(c *C) {
...
-	err = s.manager.AppendLayer(layer)
-	c.Assert(err, IsNil)
+`
+	s.planAddLayer(c, fmt.Sprintf(
+		layer,
+		outputPath,
+		shortOkayDelay.Seconds()+0.01,
+		current.Username,
+		group.Name,
+	))
```
benhoyt, line 479: Thanks for these code formatting cleanups -- they do read more clearly now.

internals/overlord/servstate/manager_test.go
```diff
@@ -902,91 +690,74 @@ PEBBLE_ENV_TEST_PARENT=from-parent
...
-	svc = s.serviceByName(c, "test2")
-	c.Assert(svc.Current, Equals, servstate.StatusActive)
+	// Then wait for it to auto-restart again.
+	s.waitForDoneCheck(c, "test2")
+	c.Assert(s.manager.BackoffNum("test2"), Equals, 2)
 	c.Check(s.readAndClearLogBuffer(), Matches, `2.* \[test2\] test2\n`)
 
 	// Test that backoff reset time is working (set to backoff-limit)
-	time.Sleep(175 * time.Millisecond)
+	time.Sleep(550 * time.Millisecond)
```
benhoyt, line 736: Out of interest, why this duration increase?

benhoyt (changes requested):
Thanks for the updates! Looking good ... just one missed "err" assignment and a couple of tiny nits.

internals/overlord/planstate/manager.go
rebornplusplus: @flotter, do you already have some ideas about where this should go?

rebornplusplus (approved):
Amazing work with great documentation! :clap: :clap:

With regards to `--args` and `--hold` options to `pebble run`, I do not think there are any user experience changes introduced in the PR i.e. `--hold` still "holds" the "autostart" and `--args` still passes the arguments properly without any behaviour changes. I tested the changes locally and it seems to work well -- exactly as before. That is expected, and from the ROCKS' side of things, should be an OKAY from me. 

(Please correct me if I am assuming something wrong.)

internals/overlord/planstate/manager.go
flotter: Yes, I absolutely do. But I removed the proposal from this PR since I think it will require some design related debate and ideas from Ben and potentially Gustavo, so will be a follow-on PR. But it will happen.

internals/overlord/planstate/manager.go
flotter: Doh!! Fixed. Resolving.

internals/overlord/planstate/manager.go
flotter: Improved. Thank you. Resolving.

internals/overlord/planstate/manager.go
flotter: Improved the wording as per Ben's suggestion. Resolving for now. 

@rebornplusplus This topic will naturally come up again because I am working on another related PR that will allow external managers to add plan sections, which means the Plan Manager cannot have schema knowledge. So I will resolve this comment now, but the conversation will not be lost.

internals/overlord/servstate/manager_test.go
flotter: This is a very timing sensitive test, so taking inspiration from your very good question, I've added some comments to the test itself. In summary, the ```backoff-limit``` delay is also used as the backoff reset value. If the test successfully runs for ```backoff-limit``` time, backoff resets. The original value of 150ms meant that a decent cpu spike during the test could easily reset backoff, and the test would fail halfway through. The 550ms value empirically has proven to be more reliable, even under ```stress --cpu 10 --io 10``` usage. Its not technically immune to failures, but statistically has proved way more reliable overall for me. 

The test used to be delay based:

TestActionRestart (delays total = 460ms):
```
:
time.Sleep(10 * time.Millisecond)
:
time.Sleep(75 * time.Millisecond)
:
time.Sleep(125 * time.Millisecond)
:
time.Sleep(175 * time.Millisecond)
:
time.Sleep(75 * time.Millisecond)
:
```

New design: single delay = 550ms

flotter: And I've changed the test pattern of backoff times:

From: 50ms, 100ms, ... (limit 150ms)
To: 1ms, 10ms, ...         (limit 500ms)

So that a cpu spike on top of the backoff delay is further away from the reset value (backoff limit). I achieved that by using the backoff-factor multiplier (at 10), and a much shorter starting value.

flotter: Let me know if you are comfortable with this ?

internals/overlord/servstate/manager_test.go
flotter: Pleasure. Resolving.

internals/overlord/servstate/manager_test.go
benhoyt: Sounds good to me -- thanks for the explanation and comments.

internals/overlord/overlord.go
```diff
@@ -183,6 +194,14 @@ func New(opts *Options) (*Overlord, error) {
 	// before it.
 	o.stateEng.AddManager(o.runner)
 
+	// Load the plan from the Pebble layers directory (which may be missing
+	// or have no layers, resulting in an empty plan), and propagate PlanChanged
+	// notifications to all notification subscribers.
+	err = o.planMgr.Load()
+	if err != nil {
+		return nil, fmt.Errorf("cannot load plan %w", err)
```
benhoyt, line 202:
```suggestion
		return nil, fmt.Errorf("cannot load plan: %w", err)
```

benhoyt (approved):
Great, thanks for your work on this!

PR #388, opened 2024-03-20
fix(cli): pass PebbleDir and SocketPath through RunOptions/ParserOptions/CmdOptions

thp-canonical:
This builds on top of #383, and allows for a custom `PebbleDir` to be passed to `cli.Run()` in `RunOptions` to make sure all parts of the codebase use the passed-in values and not fall back on environment variables.

The default behavior of `cli.Run(nil)` stays unchanged.

---

Multiple parts of the codebase currently use `getEnvPaths()` directly to access `PEBBLE` and `PEBBLE_SOCKET`:

* `cli.Run()` for the default value of `client.Config.Socket` (ignores `PEBBLE` dir)
* `loadCLIState` and `saveCLIState` in `cli` for the socket path (both ignore `PEBBLE` dir)
* `runDaemon` in `cmd_run.go`, which accesses both `pebbleDir` and `socketPath`

With #383, it is possible to pick a different socket path by passing it into `cli.Run()`'s options. This socket path currently only applies to Pebble client, it does not apply to the daemon part started by `pebble run` (because of the separate `getEnvPaths()` in `cmd_run.go`'s `runDaemon`).

For `loadCLIState` and `saveCLIState` (both used by `pebble notices` and its acknowledgement command `pebble okay`) this PR makes sure the passed-in socket path is used (otherwise it would default to the environment variable even when a non-`nil` options value is passed to `cli.Run()`).

While "Pebble as a server" (`pebble run`) and "Pebble as a client" can be seen as conceptually different things, both share `cli.Run` as entry point.

This doesn't affect uses of `getEnvPaths` (and its exported name `GetEnvPaths`) in the unit tests.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -141,9 +141,6 @@ type Config struct {
 
 	// UserAgent is the User-Agent header sent to the Pebble daemon.
 	UserAgent string
-
-	// PebbleDir is the base directory for storing the Pebble daemon's state.
-	PebbleDir string
 }
 
 // A Client knows how to talk to the Pebble daemon.
@@ -155,17 +152,6 @@ type Client struct {
 	warningTimestamp time.Time
 
 	getWebsocket getWebsocketFunc
-
-	pebbleDir  string
-	socketPath string
-}
-
-func (c *Client) PebbleDir() string {
-	return c.pebbleDir
-}
-
-func (c *Client) SocketPath() string {
-	return c.socketPath
 }
 
 type getWebsocketFunc func(url string) (clientWebsocket, error)
@@ -193,8 +179,6 @@ func New(config *Config) (*Client, error) {
 	}
 
 	client.requester = requester
-	client.pebbleDir = config.PebbleDir
-	client.socketPath = config.Socket
 	client.getWebsocket = func(url string) (clientWebsocket, error) {
 		return getWebsocket(requester.Transport(), url)
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -141,6 +141,9 @@ type Config struct {
 
 	// UserAgent is the User-Agent header sent to the Pebble daemon.
 	UserAgent string
+
+	// PebbleDir is the base directory for storing the Pebble daemon's state.
+	PebbleDir string
 }
 
 // A Client knows how to talk to the Pebble daemon.
@@ -152,6 +155,17 @@ type Client struct {
 	warningTimestamp time.Time
 
 	getWebsocket getWebsocketFunc
+
+	pebbleDir  string
+	socketPath string
+}
+
+func (c *Client) PebbleDir() string {
+	return c.pebbleDir
+}
+
+func (c *Client) SocketPath() string {
+	return c.socketPath
 }
 
 type getWebsocketFunc func(url string) (clientWebsocket, error)
@@ -179,6 +193,8 @@ func New(config *Config) (*Client, error) {
 	}
 
 	client.requester = requester
+	client.pebbleDir = config.PebbleDir
+	client.socketPath = config.Socket
 	client.getWebsocket = func(url string) (clientWebsocket, error) {
 		return getWebsocket(requester.Transport(), url)
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -152,6 +152,8 @@ type Client struct {
 	warningTimestamp time.Time
 
 	getWebsocket getWebsocketFunc
+
+	host string
 }
 
 type getWebsocketFunc func(url string) (clientWebsocket, error)
@@ -182,6 +184,7 @@ func New(config *Config) (*Client, error) {
 	client.getWebsocket = func(url string) (clientWebsocket, error) {
 		return getWebsocket(requester.Transport(), url)
 	}
+	client.host = requester.baseURL.Host
 
 	return client, nil
 }
@@ -191,7 +194,7 @@ func (client *Client) Requester() Requester {
 }
 
 func (client *Client) getTaskWebsocket(taskID, websocketID string) (clientWebsocket, error) {
-	url := fmt.Sprintf("ws://localhost/v1/tasks/%s/websocket/%s", taskID, websocketID)
+	url := fmt.Sprintf("ws://%s/v1/tasks/%s/websocket/%s", client.host, taskID, websocketID)
 	return client.getWebsocket(url)
 }
 
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -308,8 +308,8 @@ func Run(options *RunOptions) error {
 	}
 	parser := Parser(cli, &RunOptions{
 		ClientConfig: config,
-		Logger: log,
-		PebbleDir: pebbleDir,
+		Logger:       log,
+		PebbleDir:    pebbleDir,
 	})
 	xtra, err := parser.Parse()
 	if err != nil {
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -293,7 +293,7 @@ func Run(options *RunOptions) error {
 	config := options.ClientConfig
 	if config == nil {
 		config = &client.Config{}
-		_, config.Socket = getEnvPaths()
+		config.PebbleDir, config.Socket = getEnvPaths()
 	}
 	cli, err := client.New(config)
 	if err != nil {
@@ -403,12 +403,12 @@ type fullCLIState struct {
 }
 
 // TODO(benhoyt): add file locking to properly handle multi-user access
-func loadCLIState() (*cliState, error) {
+func loadCLIState(client *client.Client) (*cliState, error) {
 	fullState, err := loadFullCLIState()
 	if err != nil {
 		return nil, err
 	}
-	_, socketPath := getEnvPaths()
+	socketPath := client.SocketPath()
 	st, ok := fullState.Pebble[socketPath]
 	if !ok {
 		return &cliState{}, nil
@@ -437,13 +437,13 @@ func loadFullCLIState() (*fullCLIState, error) {
 	return &fullState, nil
 }
 
-func saveCLIState(state *cliState) error {
+func saveCLIState(client *client.Client, state *cliState) error {
 	fullState, err := loadFullCLIState()
 	if err != nil {
 		return err
 	}
 
-	_, socketPath := getEnvPaths()
+	socketPath := client.SocketPath()
 	fullState.Pebble[socketPath] = state
 
 	data, err := json.Marshal(fullState)
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -272,6 +272,8 @@ func (e *exitStatus) Error() string {
 }
 
 type RunOptions struct {
+	// when starting a daemon ("pebble run"), the ClientConfig.Socket value
+	// is also used as the Unix domain socket path for the Pebble socket
 	ClientConfig *client.Config
 	Logger       logger.Logger
 	PebbleDir    string
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -58,9 +58,8 @@ var ErrExtraArgs = fmt.Errorf("too many arguments for command")
 
 // CmdOptions exposes state made accessible during command execution.
 type CmdOptions struct {
-	Client     *client.Client
-	Parser     *flags.Parser
-	RunOptions *RunOptions
+	*ParserOptions
+	Parser *flags.Parser
 }
 
 // CmdInfo holds information needed by the CLI to execute commands and
@@ -153,14 +152,20 @@ type defaultOptions struct {
 	Version func() `long:"version" hidden:"yes" description:"Print the version and exit"`
 }
 
+type ParserOptions struct {
+	Client     *client.Client
+	PebbleDir  string
+	SocketPath string
+}
+
 // Parser creates and populates a fresh parser.
 // Since commands have local state a fresh parser is required to isolate tests
 // from each other.
-func Parser(cli *client.Client, options *RunOptions) *flags.Parser {
+func Parser(opts *ParserOptions) *flags.Parser {
 	// Implement --version by default on every command
 	defaultOpts := defaultOptions{
 		Version: func() {
-			printVersions(cli)
+			printVersions(opts.Client)
 			panic(&exitStatus{0})
 		},
 	}
@@ -187,7 +192,7 @@ func Parser(cli *client.Client, options *RunOptions) *flags.Parser {
 
 	// Add all commands
 	for _, c := range commands {
-		obj := c.New(&CmdOptions{Client: cli, Parser: parser, RunOptions: options})
+		obj := c.New(&CmdOptions{Parser: parser, ParserOptions: opts})
 
 		var target *flags.Command
 		if c.Debug {
@@ -306,10 +311,10 @@ func Run(options *RunOptions) error {
 	if pebbleDir == "" {
 		pebbleDir, _ = getEnvPaths()
 	}
-	parser := Parser(cli, &RunOptions{
-		ClientConfig: config,
-		Logger:       log,
-		PebbleDir:    pebbleDir,
+	parser := Parser(&ParserOptions{
+		Client:     cli,
+		PebbleDir:  pebbleDir,
+		SocketPath: config.Socket,
 	})
 	xtra, err := parser.Parse()
 	if err != nil {
--- a/internals/cli/cli.go
+++ b/internals/cli/cli.go
@@ -302,10 +302,15 @@ func Run(options *RunOptions) error {
 		return fmt.Errorf("cannot create client: %v", err)
 	}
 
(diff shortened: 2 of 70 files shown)


client/client.go
```diff
@@ -152,6 +155,17 @@ type Client struct {
...
+	socketPath string
+}
+
+func (c *Client) PebbleDir() string {
+	return c.pebbleDir
+}
+
+func (c *Client) SocketPath() string {
+	return c.socketPath
 }
```
thp-canonical, line 169: I opted for "private fields and public accessors" here on purpose to make it clear that these two members are not supposed to be changed from the outside after the client has been created.

internals/cli/export_test.go
```diff
@@ -21,16 +21,28 @@ import (
...
 	})
 }
 
-var ClientConfig = &client.Config{}
+var clientConfigBaseURL string
+
+func FakeClientConfigBaseURL(baseURL string) (restore func()) {
+	clientConfigBaseURL = baseURL
+	return func() {
+		clientConfigBaseURL = ""
```
thp-canonical, line 34: I'm aware of the pattern of saving and restoring the old value, but this broke the tests (haven't spent the time figuring out why), so I left this the same as the old restore function in `internals/cli/cli_test.go` above (which is currently the only place where `FakeClientConfigBaseURL()` is used:

```golang
s.BaseTest.AddCleanup(func() { cli.ClientConfig.BaseURL = "" })
```

client/client.go
```diff
@@ -141,6 +141,9 @@ type Config struct {
 
 	// UserAgent is the User-Agent header sent to the Pebble daemon.
 	UserAgent string
+
+	// PebbleDir is the base directory for storing the Pebble daemon's state.
+	PebbleDir string
```
benhoyt, line 146: I really don't think this (or the new exported methods) belong in the Client. It seems like with this change we're trying to use the Client instance as a way to pass through extra things that the client itself doesn't need. (I'm aware of the link between `PEBBLE` and `PEBBLE_SOCKET`, but I'm talking about code structure here.) So I think we should revert these client changes and find a better way to give the relevant commands access to these settings without passing them through here.

It seems reasonable to add `PebbleDir string` to `RunOptions` (it is configuration for `Run()`), and then add a `RunOptions *RunOptions` field to `CmdOptions`, which is (more or less) accessible to the commands. Then `cmdRun` would have a `pebbleDir string` field, and `cmdNotices` and `cmdOkay` would have a `pebbleSocket string` field. It's a tiny bit more wiring, but definitely feels better than threading them through client.

Open to alternative suggestions too!

The notices code also uses `$XDG_CONFIG_HOME` and so we should probably wire up a `CLIStatePath` option, but I'm very happy to leave that till if/when we need it.

internals/cli/cli.go
```diff
@@ -403,12 +403,12 @@ type fullCLIState struct {
 }
 
 // TODO(benhoyt): add file locking to properly handle multi-user access
-func loadCLIState() (*cliState, error) {
+func loadCLIState(client *client.Client) (*cliState, error) {
```
benhoyt, line 406: With my suggested change, these function would probably just take a `socketPath string` arg directly.

benhoyt (changes requested):

internals/cli/cli.go
thp-canonical: Done, nice!

benhoyt:
Tests are failing on this one too. :-) It's end of day for me now, but I'll review this tomorrow.

internals/cli/cmd_add_test.go
```diff
@@ -89,7 +89,7 @@ services:
 				args = append(args, "--combine")
 			}
 			args = append(args, "foo", path)
-			rest, err := cli.Parser(cli.Client()).ParseArgs(args)
+			rest, err := cli.ParserForTest().ParseArgs(args)
```
thp-canonical, line 92: @benhoyt I didn't find a cleaner way to pass `RunOptions` from `cli.Run` into `CmdOptions` (e.g. a global `RunOptions` to pass `RunOptions` from `cli.Run()` without changing the `cli.Parser()` parameter list sounds super ugly, as does introducing something like `cli.ParserWithRunOptions()`), so this diff grows quite large.

Therefore, I decided to centralize the creation of a "default parser" into `cli.ParserForTest()` so that in the future if/when `cli.Parser()`'s argument list changes again, the diff will be smaller (only `cli.ParserForTest()`'s body needs to be changed then).

internals/cli/export_test.go
```diff
@@ -17,20 +17,40 @@ package cli
...
 
 func RunMain() error {
-	_, ClientConfig.Socket = getEnvPaths()
-	return Run(&RunOptions{
-		ClientConfig: ClientConfig,
-	})
+	return Run(RunOptionsForTest())
+}
+
+func RunOptionsForTest() *RunOptions {
```
thp-canonical, line 29: Would have liked to call this `RunOptions()` (just like `Client()`), but it's already the name of the type.

internals/cli/cmd_enter.go
```diff
@@ -123,7 +129,7 @@ func (cmd *cmdEnter) Execute(args []string) error {
 		extraArgs []string
 	)
 
-	parser := Parser(cmd.client)
+	parser := Parser(cmd.client, cmd.runOptions)
```
thp-canonical, line 132: Is this the correct semantics for the `enter` command?

At least the unit tests fail and I'm not sure what the real issue is:

```
... Panic: internal error: daemon stopped early (PC=0x46F442)

/opt/hostedtoolcache/go/1.20.14/x64/src/runtime/panic.go:890
  in gopanic
export_test.go:102
  in PebbleMain.func2
/opt/hostedtoolcache/go/1.20.14/x64/src/runtime/panic.go:890
  in gopanic
cli.go:285
  in Run.func1
/opt/hostedtoolcache/go/1.20.14/x64/src/runtime/panic.go:890
  in gopanic
cmd_enter.go:187
  in cmdEnter.Execute
/home/runner/go/pkg/mod/github.com/canonical/go-flags@v0.0.0-20230403090104-105d09a091b8/parser.go:335
  in Parser.ParseArgs
/home/runner/go/pkg/mod/github.com/canonical/go-flags@v0.0.0-20230403090104-105d09a091b8/parser.go:191
  in Parser.Parse
cli.go:309
  in Run
export_test.go:26
  in RunMain
export_test.go:106
  in PebbleMain
cmd_enter_test.go:158
  in PebbleSuite.TestEnterExecListDir
/opt/hostedtoolcache/go/1.20.14/x64/src/reflect/value.go:370
  in Value.Call
/opt/hostedtoolcache/go/1.20.14/x64/src/runtime/asm_amd64.s:1598
  in goexit
PANIC: cmd_enter_test.go:146: PebbleSuite.TestEnterExecListDir
```

client/client.go
thp-canonical: Passing `RunOptions` to `CmdOptions` and taking the values from there is better, yes.

I tried to do this now, unfortunately it seems like it requires passing `RunOptions` to `cli.Parse()`, and that results in a big diff due to tests using `cli.Parse()` directly quite often (see other comments).

Also, for some reason, the `enter` command (or its tests) seem to fail now (also see other comment there), didn't  immediately spot what's wrong there, maybe you see something there that I don't?

internals/cli/cmd_enter.go
benhoyt: If you merge the latest master into this branch to get the [reaper.Stop err fix](https://github.com/canonical/pebble/pull/390), you'll get a better error message, which is:

```
cannot run daemon: directory "" must be absolute
```

It's happening because somehow the PebbleDir isn't being wired up, so it's empty string when it gets to overlord.New. I'm not sure exactly why it's not being wired up.

thp-canonical: The `cmdRun` created in `cmd_enter.go` was not getting `pebbleDir` initialized. Pushed a fix.

internals/cli/cli.go
```diff
@@ -300,7 +302,10 @@ func Run(options *RunOptions) error {
 		return fmt.Errorf("cannot create client: %v", err)
 	}
 
-	parser := Parser(cli)
+	if options.PebbleDir == "" {
+		options.PebbleDir, _ = getEnvPaths()
```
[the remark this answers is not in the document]
thp-canonical: Good point! Also, it's wrong currently (because the default `Logger` and `ClientConfig` aren't applied).

Probably needs to be something like this:

```golang
        pebbleDir := options.PebbleDir
        if pebbleDir == "" {
                pebbleDir, _ = getEnvPaths()
        }
        parser := Parser(cli, &RunOptions{
                ClientConfig: config,
                Logger: log,
                PebbleDir: pebbleDir,
        })
```

internals/cli/cli.go
```diff
@@ -300,7 +302,15 @@ func Run(options *RunOptions) error {
...
 
-	parser := Parser(cli)
+	pebbleDir := options.PebbleDir
+	if pebbleDir == "" {
+		pebbleDir, _ = getEnvPaths()
+	}
+	parser := Parser(cli, &RunOptions{
+		ClientConfig: config,
+		Logger:       log,
+		PebbleDir:    pebbleDir,
```
thp-canonical, line 312: This applies the defaults generated in `Run()` (default clientConfig if not supplied, default logger if not supplied, default PebbleDir if not supplied) so that `Parser` (and in turn `CmdOptions`) can rely on all the fields being set to something.

internals/cli/cmd_okay.go
```diff
@@ -43,7 +45,10 @@ func init() {
 		Summary:     cmdOkaySummary,
 		Description: cmdOkayDescription,
 		New: func(opts *CmdOptions) flags.Commander {
-			return &cmdOkay{client: opts.Client}
+			return &cmdOkay{
+				client:     opts.Client,
+				socketPath: opts.RunOptions.ClientConfig.Socket,
```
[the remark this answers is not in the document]
thp-canonical: See the creation of a new `RunOptions` inside `cli.Run()`, which should hopefully resolve this without too much issues.

internals/cli/cmd_warnings_test.go
```diff
@@ -271,23 +271,23 @@ func (s *warningSuite) TestLastWarningTimestamp(c *check.C) {
...
+	rest, err = cli.ParserForTest().ParseArgs([]string{"okay"})
 	c.Assert(err.Error(), check.Equals, "spurious extra data in timestamp file")
 	c.Check(rest, check.HasLen, 1)
 
 	// Make open() fail
 	err = os.Chmod(newWarnPath, 0055)
 	c.Assert(err, check.IsNil)
 
-	rest, err = cli.Parser(cli.Client()).ParseArgs([]string{"okay"})
+	rest, err = cli.ParserForTest().ParseArgs([]string{"okay"})
```
thp-canonical, line 290: Btw, if it helps reviewing, I could separate out this big mechanical change (`cli.ParserForTest()`) into a separate PR to remove the noise from this PR. Let me know if you'd prefer that.

benhoyt: Thanks, but that's probably more trouble than it's worth. It's not hard to scan down through a bunch of `ParserForTest()` lines.

benhoyt (approved):
This looks good to me now, thanks. Could we please get one more pair of eyes on it, maybe @flotter's?

internals/cli/cli.go
```diff
@@ -58,8 +58,9 @@ var ErrExtraArgs = fmt.Errorf("too many arguments for command")
 
 // CmdOptions exposes state made accessible during command execution.
 type CmdOptions struct {
-	Client *client.Client
-	Parser *flags.Parser
+	Client     *client.Client
+	Parser     *flags.Parser
+	RunOptions *RunOptions
```
flotter, line 63: Since this struct is already an options container for commands, could we not simply pass in the options we need, in a way that is without consumer context ? I believe the only options we need to add for commands are:

```socketPath```
```pebbleDir```

Why do we not simply add these?

File: internals/cli/cli.go [hunk 13]
```diff
@@ -268,6 +269,7 @@ func (e *exitStatus) Error() string {
 type RunOptions struct {
 	ClientConfig *client.Config
 	Logger       logger.Logger
+	PebbleDir    string
 }
```
flotter, line 273: This looks fine to me. My only question would be since we are now effectively setting the Daemon server socket based on ```ClientConfig.socket```, is this OK ? Should we document this (for now its only derivative projects that use ```Run(...)``` with non-nil options that needs to know this), or would it be a bit nicer to add another ```ServerSocket``` to the ``RunOptions``` struct to make this a bit clearer?

flotter (commented):
Great PR. Going to be super useful for derivative projects. I just have a couple of questions regarding the struct involved, be the rest looks great!

internals/cli/cli.go
benhoyt: Yeah, that's a good point, Fred -- it's simpler to reason about and avoids mixing concerns. Let's do that.

internals/cli/cli.go
```diff
@@ -155,7 +156,7 @@ type defaultOptions struct {
 // Parser creates and populates a fresh parser.
 // Since commands have local state a fresh parser is required to isolate tests
 // from each other.
-func Parser(cli *client.Client) *flags.Parser {
+func Parser(cli *client.Client, options *RunOptions) *flags.Parser {
```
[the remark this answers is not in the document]
benhoyt: Another good point. I don't like just reusing `CmdOptions` because it has the parser in it. I think the better thing to do would be to add a new struct `ParserOptions` (even though it's a tiny bit more wiring), which takes the client, pebble dir, and socket path. So what about this?

```go
type ParserOptions struct {
    Client *client.Client
    PebbleDir string
    SocketPath string
}

func Parser(opts *ParserOptions) *flags.Parser {
    ...
}
```

internals/cli/cli.go
benhoyt: Good idea to add a clarifying comment here.

benhoyt (changes requested):
Sorry for the back and forth, but I think Fred's got some good points here about what we're passing down. Commented.

internals/cli/cli.go
thp-canonical: With the new `ParserOptions` struct and embedding `*ParserOptions` into `CmdOptions`, this practically does the same thing (replaces the existing `Client` with the same name + adds `SocketPath` and `PebbleDir`).

However, we also need `ParserOptions` in `cmd_enter.go`'s `cmdEnter` (because we create a new parser there), so I opted for keeping `*ParserOptions` in `CmdOptions` as embedded struct, so there I can just say: `Parser(cmd.parserOptions)`.

If that is not to your liking, I can also explicity add the fields as in:

```go
type CmdOptions struct {
        Client     *client.Client
        Parser     *flags.Parser
        PebbleDir  string
        SocketPath string
}
```

It would be relatively easy to change `&CmdOptions{Parser: parser, ParserOptions: opts}` to explicitly grab client, pebble dir and socket path from `ParserOptions` and put them into `CmdOptions`.

And on the `cmd_enter.go` side it wouldn't be too much trouble to add all 3 fields (`client`, `pebbleDir`, `socketPath`) to `cmdEnter` (instead of just `parserOptions *ParserOptions`) and then when creating the parser, doing it explicitly:

```go
        parser := Parser(&ParserOptions{
                Client: cmd.client,
                SocketPath: cmd.socketPath,
                PebbleDir: cmd.pebbleDir,
        })
```

Slightly more maintenance work when adding or removing fields from `ParserOptions` (and it's easier to forget wiring all values through `ParserOptions` -> `Parser()` -> `CmdOptions` -> `cmdEnter` -> `ParserOptions` in `cmd_enter.go`?) but potentially more explicit and separated.

Opinions?

internals/cli/cli.go
thp-canonical: added:

```diff
 type RunOptions struct {
+       // when starting a daemon ("pebble run"), the ClientConfig.Socket value
+       // is also used as the Unix domain socket path for the Pebble socket
```

thp-canonical:
The PR description needs updating once the code changes are approved.

internals/cli/cli.go
benhoyt: Thanks. Yeah, I don't like the ParserOptions stuffed into CmdOptions -- I think we should spell them out like you have in the struct definition in your comment above. Tiny bit more wiring, but conceptually cleaner. Same with cmd_enter -- let's spell out those fields like you have with cmd_run.

benhoyt (changes requested):
Hopefully the last round -- yeah, let's spell out the field in CmdOptions and cmdEnter, per my earlier comment.

benhoyt (approved):
I'm happy with this now, thanks! Can you give it one more pass @flotter? And I'll aim to merge tomorrow.

flotter (approved):
Looks really nice now. Thanks.

PR #389, opened 2024-03-20
fix(client): WebSockets dialer with client BaseURL set

thp-canonical:
Add a custom TCP dialer to allow WebSockets to work with remote TCP connections for clients that have BaseURL set.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -122,6 +122,12 @@ func unixDialer(socketPath string) func(string, string) (net.Conn, error) {
 	}
 }
 
+func tcpDialer(address string) func(string, string) (net.Conn, error) {
+	return func(_, _ string) (net.Conn, error) {
+		return net.Dial("tcp", address)
+	}
+}
+
 type doer interface {
 	Do(*http.Request) (*http.Response, error)
 }
@@ -587,7 +593,7 @@ func newDefaultRequester(client *Client, opts *Config) (*defaultRequester, error
 		if err != nil {
 			return nil, fmt.Errorf("cannot parse base URL: %w", err)
 		}
-		transport := &http.Transport{DisableKeepAlives: opts.DisableKeepAlive}
+		transport := &http.Transport{Dial: tcpDialer(baseURL.Host), DisableKeepAlives: opts.DisableKeepAlive}
 		requester = &defaultRequester{baseURL: *baseURL, transport: transport}
 	}
 


client/client.go
```diff
@@ -587,7 +593,7 @@ func newDefaultRequester(client *Client, opts *Config) (*defaultRequester, error
 		if err != nil {
 			return nil, fmt.Errorf("cannot parse base URL: %w", err)
 		}
-		transport := &http.Transport{DisableKeepAlives: opts.DisableKeepAlive}
+		transport := &http.Transport{Dial: tcpDialer(baseURL.Host), DisableKeepAlives: opts.DisableKeepAlive}
```
benhoyt, line 596: I'm not sure I understand why we need a custom dialer here -- why does the default not pass through the address/host correctly? Can you enlighten me?

Also, assuming this doesn't work as is, can we add a regression test (that fails before this change and passes after)? If it's too hard, that's okay -- in that case could you please write manual repro steps in the PR description? Thanks!

thp-canonical: When getting a web socket, `localhost` is being hardcoded in `getTaskWebsocket` (which "makes sense" for the UNIX domain socket case, it's kind of localhost, even if the underlying transport is a UNIX domain socket, it's not `ws://localhost/` as in "a TCP connection to localhost, port 80"):

```golang
func (client *Client) getTaskWebsocket(taskID, websocketID string) (clientWebsocket, error) {
        url := fmt.Sprintf("ws://localhost/v1/tasks/%s/websocket/%s", taskID, websocketID)
        return client.getWebsocket(url)
}
```

The `client.getWebsocket` function is assigned like this:

```golang
        client.getWebsocket = func(url string) (clientWebsocket, error) {
                return getWebsocket(requester.Transport(), url)
        }
```

And in turn the `getWebsocket` function is:

```golang
func getWebsocket(transport *http.Transport, url string) (clientWebsocket, error) {
        dialer := websocket.Dialer{
                NetDial:          transport.Dial,
                Proxy:            transport.Proxy,
                TLSClientConfig:  transport.TLSClientConfig,
                HandshakeTimeout: 5 * time.Second,
        }
        conn, _, err := dialer.Dial(url, nil)
        return conn, err
}
```

All of those are in `client/client.go`.

Because the `websocket.Dialer` uses `NetDial: transport.Dial`, for the UNIX domain socket case, it assigns the dialer like such (in `newDefaultRequester()`):

```golang
&http.Transport{Dial: unixDialer(opts.Socket), ...}
```

It could be that pushing `baseURL.Host` all the way through to the `getTaskWebsocket()` function, and creating a `fmt.Sprintf("ws://%s/...", baseURL.Host)`-style URL might do the trick (with `transport.Dial: nil`, the default), but it's a bit more involved (and I haven't tested if this actually fixes the issue, but `websocket.Dialer`'s [documentation](https://pkg.go.dev/github.com/gorilla/websocket#Dialer) says "If NetDial is nil, net.Dial is used").

thp-canonical:
@benhoyt I've opened #393 as a potentially cleaner fix, but I still have to test if #393 has the same effect as this PR here (otherwise I'd have just update this PR with the #393 contents instead of having multiple PRs open that do the same thing - sorry for the noise).

I'll report back and close this PR if #393 indeed fixes the websocket (`pebble exec`) issue.

thp-canonical:
Superseded by #393.

PR #392, opened 2024-03-21
fix(cli): PEBBLE_COPY_ONCE on missing dir.

hpidcock:
If the `PEBBLE_COPY_ONCE` source directory does not exist, it will fail to copy inside the borrowed std go copyFS function. This adds an `os.Stat` to check if it first exists, ignoring it otherwise (the expected behavior).


hpidcock:
@benhoyt we will need a `v1.9.1` release after this lands.

benhoyt (approved):

PR #395, opened 2024-03-25
feat(planstate): run PlanChanged handlers lockless

flotter:
Release the plan manager lock as soon as the new plan is created and plan manager plan pointer updated.

In the past all the plan consumer handlers were run while the plan lock was held. This is not necessary, since a new plan is always created following a plan change, and the new plan is never mutated by the plan manager or any consumers.

The following considerations were taken into account:

1. Pebble API requests involving plan changes (i.e. AppendLayer, CombineLayer and SetServiceArgs) are synchronous and only complete (return) once all the PlanChanged callback handlers returned. This means that from a single API consumer point of view, once a change returns, the next API call made will be guaranteed to see Pebble in the updated state (single threaded consumer).

I've not considered a change in the design to make plan changes propagate asynchronously, as this would have changed the behaviour described above.

2. The plan pointer sent to the PlanChanged consumers, following the release of the plan lock (the new design), must be a copy of the plan pointer already obtained before plan lock was released (otherwise we risk reading the plan pointer value during another change from the API).

- Make the PlanChanged callback handlers run without the plan manager lock.

- Improve the unit test coverage to ensure plan change notifications take place as expected, with the correct plan content being sent.

### DIFF
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -57,14 +57,21 @@ func NewManager(s *state.State, runner *state.TaskRunner, pebbleDir string) (*Pl
 // final plan, and finally notifies registered managers of the plan update. In
 // the case of a non-existent layers directory, or no layers in the layers
 // directory, an empty plan is announced to change subscribers.
-func (m *PlanManager) Load() error {
+func (m *PlanManager) Load() (err error) {
+	var newPlan *plan.Plan
+	defer func() {
+		if err == nil {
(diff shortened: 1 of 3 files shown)


flotter:
@benhoyt I was thinking that before I move this PR to a final review state, whether you could take a quick peek to see if my approach feels acceptable for you ?

flotter:
@benhoyt solved this issue before I got time to further this PR.

See: fix(planstate): don't hold planLock while calling plan-changed listeners (#436)

Closing.

PR #397, opened 2024-03-26
fix(help): DefaultDir to personality, allow overriding header and footer

thp-canonical:
Just like `ProgramName` and `DisplayName` can already be customized via public vars in the `cmd` package, `DefaultDir` also looks like a perfect candidate to customize this way (and assembling the help text can then be done using `{{.DefaultDir}}`).

In addition, `cli.HelpHeader` and `cli.HelpFooter` are made public, so that they can be extended and overridden.

### DIFF
--- a/cmd/personality.go
+++ b/cmd/personality.go
@@ -14,10 +14,6 @@
 
 package cmd
 
-import (
-	"strings"
-)
-
 // ProgramName represents the name of the application binary.
 var ProgramName string = "pebble"
 
@@ -26,12 +22,3 @@ var DisplayName string = "Pebble"
 
 // DefaultDir is the Pebble directory used if $PEBBLE is not set.
 var DefaultDir string = "/var/lib/pebble/default"
-
-func ApplyPersonality(s string) string {
-	r := strings.NewReplacer(
-		"{{.ProgramName}}", ProgramName,
-		"{{.DisplayName}}", DisplayName,
-		"{{.DefaultDir}}", DefaultDir,
-	)
-	return r.Replace(s)
-}
--- a/cmd/personality.go
+++ b/cmd/personality.go
@@ -14,6 +14,10 @@
 
 package cmd
 
+import (
+	"strings"
+)
+
 // ProgramName represents the name of the application binary.
 var ProgramName string = "pebble"
 
@@ -22,3 +26,12 @@ var DisplayName string = "Pebble"
 
 // DefaultDir is the Pebble directory used if $PEBBLE is not set.
 var DefaultDir string = "/var/lib/pebble/default"
+
+func ApplyPersonality(s string) string {
+	r := strings.NewReplacer(
+		"{{.ProgramName}}", ProgramName,
+		"{{.DisplayName}}", DisplayName,
+		"{{.DefaultDir}}", DefaultDir,
+	)
+	return r.Replace(s)
+}
--- a/cmd/personality.go
+++ b/cmd/personality.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2020 Canonical Ltd
+// Copyright (c) 2014-2024 Canonical Ltd
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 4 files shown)


internals/cli/cmd_help.go
```diff
@@ -219,12 +219,16 @@ the system that is running them.
...
 	pebbleHelpCategoriesIntro = "Commands can be classified as follows:"
-	pebbleHelpAllFooter       = "Set the PEBBLE environment variable to override the configuration directory \n" +
-		"(which defaults to " + defaultPebbleDir + "). Set PEBBLE_SOCKET to override \n" +
-		"the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).\n" +
-		"\n" +
-		"For more information about a command, run '{{.ProgramName}} help <command>'."
-	pebbleHelpFooter = "For a short summary of all commands, run '{{.ProgramName}} help --all'."
+
+	// can be overridden in derivative projects
+	PebbleHelpAllFooterText = strings.TrimSpace(`
```
benhoyt, line 224: Just a couple of comments: 

* Can we come up with a better name for this? Maybe remove the "Pebble", maybe `FooterHelp`? Or maybe get more concrete, like `EnvVarsHelp`?
* What will other projects actually change this to? Would it be better to make the names of the environment variables customisable, rather than the entire paragraph?

benhoyt: Maybe `EnvVarsHelp` is too specific, but something in between `FooterHelp` and `EnvVarsHelp`?

thp-canonical: After reading through it again, I decided to go with `HelpHeader` (if we allow overriding the footer, we might also want to allow overriding the header) and `HelpFooter`.

This then makes it read from another package as `cli.HelpHeader` and `cli.HelpFooter` (it's the header of the help and the footer of the help) vs `cli.HeaderHelp` and `cli.FooterHelp` (it's not the help of the header, and not the help of the footer).

internals/cli/cli.go
```diff
@@ -33,7 +33,7 @@ import (
 	"golang.org/x/term"
 
 	"github.com/canonical/pebble/client"
-	"github.com/canonical/pebble/cmd"
+	cmdpkg "github.com/canonical/pebble/cmd"
```
thp-canonical, line 36: Due to clashes with a local variable `cmd` below. Same way it's imported in `cmd_help.go` also.

internals/cli/cli.go
```diff
@@ -242,11 +237,6 @@ func Parser(cli *client.Client) *flags.Parser {
 	return parser
 }
 
-func applyPersonality(s string) string {
-	r := strings.NewReplacer("{{.ProgramName}}", cmd.ProgramName, "{{.DisplayName}}", cmd.DisplayName)
-	return r.Replace(s)
-}
```
thp-canonical, line 248: Decided to move this to the `cmd` package (so it sits next to the variables it reads) and make it public. `cmd.ProgramName` and `cmd.DisplayName` were already public, so making it possible to use `ApplyPersonality()` from public code shouldn't hurt.

cmd/personality.go
```diff
@@ -0,0 +1,37 @@
...
+// ProgramName represents the name of the application binary.
+var ProgramName string = "pebble"
+
+// DisplayName represents the user-facing name of the application.
+var DisplayName string = "Pebble"
+
+// DefaultDir is the Pebble directory used if $PEBBLE is not set.
+var DefaultDir string = "/var/lib/pebble/default"
+
+func ApplyPersonality(s string) string {
```
benhoyt, line 30: Let's be conservative and keep this in internals/cli for now, so we're not locked into it in a public package.

internals/cli/cmd_help.go
```diff
@@ -213,36 +213,41 @@ var HelpCategories = []HelpCategory{{
...
 	pebbleUsage               = "Usage: {{.ProgramName}} <command> [<options>...]"
 	pebbleHelpCategoriesIntro = "Commands can be classified as follows:"
-	pebbleHelpAllFooter       = "Set the PEBBLE environment variable to override the configuration directory \n" +
-		"(which defaults to " + defaultPebbleDir + "). Set PEBBLE_SOCKET to override \n" +
-		"the unix socket used for the API (defaults to $PEBBLE/.pebble.socket).\n" +
-		"\n" +
-		"For more information about a command, run '{{.ProgramName}} help <command>'."
-	pebbleHelpFooter = "For a short summary of all commands, run '{{.ProgramName}} help --all'."
+
+	HelpFooter = strings.TrimSpace(`
```
benhoyt, line 223: Yeah, I like the names `HelpHeader` and `HelpFooter`.

benhoyt (changes requested):
Thanks for the updates. I think we should move applyPersonality back to the internal package, but other than that, I like this.

cmd/personality.go
thp-canonical: Done :)

benhoyt (approved):
Looks good to me, thanks

PR #399, opened 2024-03-26
fix(daemon): deduplicate user-provided notice types

olivercalder:
This is a subset of related changes made in snapd: https://github.com/snapcore/snapd/pull/13673


benhoyt (approved):
Seems reasonable to me, and it's in line with the snapd changes. Thanks.

PR #400, opened 2024-03-27
feat(cli): new Info category for help, version

thp-canonical:
The commands `help` and `version` are not related to running the service manager (you can run `help` also for client help, and `version` prints both the client and the server version). This splits those two commands into a new "Info" category, leaving only the `run` command in the "Run" category.

### DIFF
--- a/internals/cli/cmd_help.go
+++ b/internals/cli/cmd_help.go
@@ -185,7 +185,11 @@ type HelpCategory struct {
 var HelpCategories = []HelpCategory{{
 	Label:       "Run",
 	Description: "run the service manager",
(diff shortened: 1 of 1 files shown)


benhoyt:
I'll get @hpidcock's opinion as well, but otherwise merge tomorrow.

hpidcock (approved):
Seems reasonable to me, but maybe the info category should be first.

benhoyt:
> Seems reasonable to me, but maybe the info category should be first.

I downloaded this locally and tried that, and I actually think Run does look better as the first category.

Dec1pherX (approved):

Issue #182, opened 2023-01-23
Add $VAR-style interpolation into environment variable handling


benhoyt:
A while ago we discussed (in a discussion about whether environment variables should be an ordered list or a map - be we settled on a map for simplicity) that it would be good to have a way to insert other environment variables into the definition for new variables. For example:

```
services:
    svc:
        override: replace
        command: sleep 1000
        environment:
            THINGDIR: /home/${USER}/thing
            CMD_DIR: /root/sleepy
            CMD_SOCKET: $CMD_DIR/.sock
```

We could use `os.Expand` for this. Though we'd have to be careful about doing them in the correct order (topological sort) to resolve dependencies correctly, like the one from `CMD_SOCKET` to `CMD_DIR` above.

This is similar to the [Kubernetes environment variable feature](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/).

cjdcordeiro:
Note for future reference: in craft and spread, there is already a notation for handling variables. E.g. `$(HOST_SECRET: cat myfile)`. When addressing this issue, we should consider that and make sure we don't introduce mechanisms that might clash or create ambiguity with existing usages of Pebble.

benhoyt:
We should also consider enabling this in `command`. Some folks have been using `sh -c` as a workaround, [for example](https://github.com/maas/maas/blob/4933ac780a60a9d9ce42049f5eaeec326aba48a0/snap/local/tree/usr/share/maas/pebble/layers/001-maas-base-layer.yaml#L9-L10):

```
services:
  bind9:
    override: replace
    # Note: we use `sh -c` to expand the `$SNAP` environment variable
    command: sh -c "exec systemd-cat -t named $SNAP/bin/run-named"
```

benhoyt:
@cjdcordeiro What does the craft/spread notation `$(HOST_SECRET: cat myfile)` mean? Use the variable of environment variable `HOST_SECRET`, but if not set, fall back to running the command `cat myfile` and use that? Or if not set, use the literal string `cat myfile`?

cjdcordeiro:
@benhoyt the syntax is `$(<scope>:<cmd>)`. So `HOST_SECRET` is an identifier of the scope, telling the tool where to run the `cmd`. It is much more relevant for things like Spread and Rockcraft, since both work with "backends" and sometimes you want to run commands outside the backend. Although similar, it does not behave like the shell syntax `${FOO:-bar}`.

[...]

PR #402, opened 2024-03-29
feat(servstate): expand environment variables for service commands

zhijie-yang:
For reference, this fixes the `command` part of https://github.com/canonical/pebble/issues/182.

### DIFF
--- a/internals/overlord/servstate/handlers.go
+++ b/internals/overlord/servstate/handlers.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"os/user"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
fyi, the RK020 spec is not about this specifically.

benhoyt:
Hi @zhijie-yang, thanks for the submission. Out of interest, do you have a need for this as part of the RK020 work, or is this just something you want to fix separately? The reason I ask is I'm pressed for time at the moment and may not get a chance to review this properly for a couple of weeks.

In the meantime, at a quick glance it looks a bit convoluted, and there's a fair bit of manual, error-prone parsing going on with `strings.TrimLeft` etc. It looks like it only addresses the command part of the #182 (my comment further down), not the main issue with expanding $VAR in the environment itself. Is that your intention? It would be good to see a proper PR description with some examples of what this covers.

Issue #318, opened 2023-10-17
Do client-side validation of IDs we're using in URL paths


benhoyt:
Per [this code review comment](https://github.com/canonical/pebble/pull/297/files#r1358482201), let's avoid directly inserting arbitrary strings from the caller into URL paths to send to the server -- better to lightly validate to avoid potential security issues.

I did this for `Client.Notice` in that PR, but we should also do it in [`client/changes.go`](https://github.com/canonical/pebble/blob/90805472b01d879e8c8e17c306e7f51849357cf4/client/changes.go#L92) and any other places in the client that does this.

PR #404, opened 2024-04-01
feat: client side change id validation

IronCore864:
Add client-side validation for change ID used in URLs.

Closes https://github.com/canonical/pebble/issues/318.

### DIFF
--- a/client/changes.go
+++ b/client/changes.go
@@ -198,8 +198,8 @@ type WaitChangeOptions struct {
 // succeeds, the returned Change.Err string will be non-empty if the change
 // itself had an error.
 func (client *Client) WaitChange(id string, opts *WaitChangeOptions) (*Change, error) {
-	if err := validateChangeID(id); err != nil {
-		return nil, err
+	if !changeIDRegexp.MatchString(id) {
+		return nil, fmt.Errorf("invalid change ID %q", id)
 	}
 
 	var chgd changeAndData
--- a/client/changes.go
+++ b/client/changes.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"regexp"
 	"time"
 )
 
@@ -86,8 +87,26 @@ type changeAndData struct {
 	Data map[string]*json.RawMessage `json:"data"`
 }
 
+func validateChangeID(id string) error {
+	// This is used to ensure we send a well-formed change ID in the URL path.
+	// It's a little more permissive than the currently-valid change IDs (which
+	// are always integers), but it will allow older clients to talk to newer
+	// servers which might start allowing letters too (for example).
+	var changeIDRegexp = regexp.MustCompile(`^[a-z0-9]+$`)
+
+	if changeIDRegexp.MatchString(id) {
+		return nil
+	} else {
+		return fmt.Errorf("invalid change ID %q", id)
+	}
+}
+
 // Change fetches information about a Change given its ID.
 func (client *Client) Change(id string) (*Change, error) {
+	if err := validateChangeID(id); err != nil {
+		return nil, err
+	}
+
 	var chgd changeAndData
 	_, err := client.doSync("GET", "/v1/changes/"+id, nil, nil, nil, &chgd)
 	if err != nil {
@@ -100,6 +119,10 @@ func (client *Client) Change(id string) (*Change, error) {
 
 // Abort attempts to abort a change that is not yet ready.
 func (client *Client) Abort(id string) (*Change, error) {
+	if err := validateChangeID(id); err != nil {
+		return nil, err
+	}
+
 	var postData struct {
 		Action string `json:"action"`
 	}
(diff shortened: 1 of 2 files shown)


IronCore864:
> Couple of comments. Also, per the ticket we should do this for all endpoints that include an ID or user input of some kind in the URL -- anything in `daemon/api.go` that has a `{foo}` in the Path. In addition to the changes one, that's these:
> 
> ```
> 	Path:        "/v1/services/{name}",
> 	Path:       "/v1/tasks/{task-id}/websocket/{websocket-id}",  // probably do this one in Client.getTaskWebsocket?
> ```
> 
> The service name is more than an ID, it can be any service name, so we should be relatively permissive with that. Perhaps:
> 
> ```
> var serviceNameRegexp = regexp.MustCompile(`^[-_A-Za-z0-9]+$`)
> ```

I have added a check for service name.

For taskID/websocketID, it seems:
- taskID seems to be [returned by the server side](https://github.com/canonical/pebble/blob/master/client/exec.go#L161), not a user input
- websocketID is hard-coded to be one of the three values: [control](https://github.com/canonical/pebble/blob/master/client/exec.go#L162), [stdio](https://github.com/canonical/pebble/blob/master/client/exec.go#L168), [stderr](https://github.com/canonical/pebble/blob/master/client/exec.go#L179), also not a direct user input.

So I think we won't need to validate task ID and websocket ID?

benhoyt (approved):
Ah, good point about the websocket IDs not being user input. Looks good to me now, thanks.

benhoyt:
I'm going to ask @flotter for a second opinion on the service name regex.

client/services.go
```diff
@@ -27,6 +28,9 @@ type ServiceOptions struct {
 	Names []string
 }
 
+// This is used to ensure we send a well-formed service name in the URL path.
+var serviceNameRegexp = regexp.MustCompile(`^[-_A-Za-z0-9]+$`)
```
flotter, line 32: This is an interesting one. The problem I can see where is that services are pebble plan specified (which is way more flexible than a change ID).

Since this is only a client side service name filter, what would happen if someone uses a more exotic service name in his Pebble plan ? 

```
flotter@flotter-pc:~/pebble$ sudo ./pebble services
[sudo] password for flotter: 
Service          Startup  Current  Since
hello-world##@!  enabled  active   today at 21:52 SAST
flotter@flotter-pc:~/pebble$ sudo ./pebble checks
Check            Level  Status  Failures
hello-check##@!  -      up      1/3
flotter@flotter-pc:~/pebble$ sudo ./pebble changes hello-world##@!
51   Done    today at 21:52 SAST  today at 21:52 SAST  Autostart service "hello-world##@!"
```

My worry is that we are not preventing people from picking exotic service names, but this filter will prevent them accessing the service over the client API ?

flotter (commented):
Thanks Tiexin! I left a question for you about service names. Let me know what you think ?

On a different note, I think I am slightly struggling to understand what we are protecting against on the client side here with this general strategy (I do not have a problem restricting the client input options, of course) ? We are not preventing anyone from directly accessing the server API without these checks, right ? I apologise if I am missing an important point here.

client/services.go
benhoyt: Yeah, in thinking about this further, I agree with @flotter. We do allow quite the range of service names on the server side, and so we shouldn't restrict that unnecessarily on the client side. It's not actually a security thing, as you point out (they could easily hit the server directly with `curl` and bypass this), so it's possibly breaking some people who choose weird service names, and it's a false sense of security anyway.

Let's keep the check for the change ID (we might as well do it early there, because those have a very well-known format) but not do it for service names. Sorry about the extra work, @IronCore864.

IronCore864:
> Thanks Tiexin! I left a question for you about service names. Let me know what you think ?

Hi @flotter, thanks for the review. I agree that we should not have a strong filter on the client-side so that we don't block users getting their services whose names are allowed by the server side.

And yes, you are right, the server can be accessed directly anyway, we aren't actually protecting much here, only an added layer of validation on the client side. The server side should be fine because if the change ID is invalid, it won't find it anyway.

Some more context: this change is because of a code review comment [here](https://github.com/canonical/pebble/pull/297/files#r1358482201) and I agree it has its own merit, because user-generated content should be validated before processed.

benhoyt:
I somewhat disagree with [that comment](https://github.com/canonical/pebble/pull/297/files#r1358482201) and probably should have discussed further with Gustavo at the time to come to consensus. Related: https://benhoyt.com/writings/dont-sanitize-do-escape/ ... I think this is somewhat trying to sanitize, rather than escape (which Go's net/http and net/url packages already do properly).

In the case of change IDs, let's keep the regex, because those are a well-known format.

For service names, they are relatively permissive when you set up the Pebble plan, even allowing a slash in them like "foo/bar". However, I just noticed we don't actually use the `/v1/services/{name}` route. Both `v1GetService` and `v1PostService` return `BadRequest("not implemented")`. I noticed this because the router we're using (gorilla/mux) actually disallows "/" in a `{name}` path segment, so those routes wouldn't have worked with "foo/bar" anyway.

I think we should probably remove that entire `/v1/services/{name}` route as it's not used and just confusing, and if we ever add such routes, we probably won't put the arbitrary service name in the path. (If we did, we should disallow "/" in the service name, but I don't think we should make that change now.)

So I think this PR is fine as it stands with just the change ID fix. I'll merge shortly.

benhoyt (approved):

Issue #372, opened 2024-02-28
pebble `make-dirs` permission is masked by pebble daemon's umask


weiiwang01:
When running the Pebble `make-dirs` command, the resulting directory permissions may not match the requested settings, as they are affected by the umask of the pebble daemon.

To reproduce:

```bash
( umask 077; ./pebble run )

./pebble mkdir -m=777 /tmp/test-mkdir
ls -lah /tmp/test-mkdir/.
total 8.0K
drwx------  2 weii-wang weii-wang 4.0K Feb 28 05:02 . # it should be 777 instead of 700
drwxrwxrwt 16 root      root      4.0K Feb 28 05:02 ..
```

weiiwang01:
Well the simplest way to fix this issue is to just call `syscall.Umask(0)` when starting the daemon or fix the issue in `internals/osutil/mkdirallchown.go`, which seems not trivial. I wonder which approach would be better.

weiiwang01:
Oh, I forgot that the umask is inherited by child processes, so it's probably not a good idea to change the umask of Pebble itself.

benhoyt:
It looks like this might be a duplicate of https://github.com/canonical/pebble/issues/80 -- however, when that issue was fixed (here: https://github.com/canonical/pebble/commit/a2161d464bee0e7532e7ef13d52e677fb4ab28ae) it looks like it was only fixed for write-file and not make-dirs, so keeping this open to track that.

IronCore864:
This issue was accidentally closed by https://github.com/canonical/pebble/pull/405. Reopen, and will be closed by #423.

PR #405, opened 2024-04-02
fix: add chmod to mkdir to ignore umask settings

IronCore864:
Add a flag so that pebble mkdir won't be affected by umask settings.

Closes https://github.com/canonical/pebble/issues/372.

### DIFF
--- a/internals/osutil/mkdirallchown.go
+++ b/internals/osutil/mkdirallchown.go
@@ -31,16 +31,23 @@ var mu sync.Mutex
 
 // MkdirAllChown is like os.MkdirAll but it calls os.Chown on any
 // directories it creates.
-func MkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
+func MkdirAllChown(path string, perm os.FileMode, flags AtomicWriteFlags, uid sys.UserID, gid sys.GroupID) error {
 	mu.Lock()
 	defer mu.Unlock()
 	if uid == NoChown && gid == NoChown {
-		return os.MkdirAll(path, perm)
+		err := os.MkdirAll(path, perm)
+		if err != nil {
+			return err
+		}
+		if flags&AtomicWriteChmod != 0 {
+			return os.Chmod(path, perm)
+		}
+		return nil
 	}
-	return mkdirAllChown(filepath.Clean(path), perm, uid, gid)
+	return mkdirAllChown(filepath.Clean(path), perm, flags, uid, gid)
 }
 
-func mkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
+func mkdirAllChown(path string, perm os.FileMode, flags AtomicWriteFlags, uid sys.UserID, gid sys.GroupID) error {
 	// split out so filepath.Clean isn't called twice for each inner path
 	if s, err := os.Stat(path); err == nil {
 		if s.IsDir() {
@@ -57,15 +64,15 @@ func mkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupI
 
 	dir := filepath.Dir(path)
 	if dir != "/" {
-		if err := mkdirAllChown(dir, perm, uid, gid); err != nil {
+		if err := mkdirAllChown(dir, perm, flags, uid, gid); err != nil {
 			return err
 		}
 	}
 
-	return mkdirChown(path, perm, uid, gid)
+	return mkdirChown(path, perm, flags, uid, gid)
 }
 
-func mkdirChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
+func mkdirChown(path string, perm os.FileMode, flags AtomicWriteFlags, uid sys.UserID, gid sys.GroupID) error {
 	cand := path + ".mkdir-new"
 
 	if err := os.Mkdir(cand, perm); err != nil && !os.IsExist(err) {
@@ -84,6 +91,11 @@ func mkdirChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) 
 	if err != nil {
 		return err
 	}
+	if flags&AtomicWriteChmod != 0 {
+		if err := fd.Chmod(perm); err != nil {
+			return err
+		}
+	}
 	defer fd.Close()
 
 	return fd.Sync()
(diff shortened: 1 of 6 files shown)


internals/osutil/mkdirallchown.go
```diff
@@ -31,16 +31,23 @@ var mu sync.Mutex
...
 	mu.Lock()
 	defer mu.Unlock()
 	if uid == NoChown && gid == NoChown {
-		return os.MkdirAll(path, perm)
+		err := os.MkdirAll(path, perm)
+		if err != nil {
+			return err
+		}
+		if flags&AtomicWriteChmod != 0 {
+			return os.Chmod(path, perm)
```
benhoyt, line 43: I think we have to be careful here: the doc comment says "it calls os.Chown on any directories it creates", so presumably the same should be true of chmod. However, the above code means it will only chmod the last/lowest directory it creates, not any parents that may have been created. So I think we need to change the logic to use the recursive helper function in both cases so we can chown and chmod every directory we create.

internals/osutil/mkdirallchown.go
```diff
@@ -31,16 +31,23 @@ var mu sync.Mutex
 
 // MkdirAllChown is like os.MkdirAll but it calls os.Chown on any
 // directories it creates.
-func MkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
+func MkdirAllChown(path string, perm os.FileMode, flags AtomicWriteFlags, uid sys.UserID, gid sys.GroupID) error {
```
benhoyt, line 34: I think it's probably reasonable to add the flags argument here, as that's consistent with the `AtomicWrite` family of functions. However, the flags are different from `AtomicWriteFlags` (we only need the chmod flag here) so let's create a new type `MkdirFlags` with a `MkdirChmod` flag.

internals/osutil/mkdirallchown.go
```diff
@@ -84,18 +91,30 @@ func mkdirChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID)
 	if err != nil {
 		return err
 	}
+	if flags&AtomicWriteChmod != 0 {
```
benhoyt, line 94: I think this block should be after the `defer fd.Close()`, otherwise we'll leave this open if the `fd.Chmod` returns an error.

internals/osutil/mkdirallchown_test.go
```diff
@@ -36,7 +36,7 @@ func (mkdacSuite) TestSlashySlashy(c *check.C) {
 		d := c.MkDir()
 		// just in case
 		c.Assert(strings.HasSuffix(d, "/"), check.Equals, false)
-		err := osutil.MkdirAllChown(d+dir, 0755, osutil.NoChown, osutil.NoChown)
+		err := osutil.MkdirAllChown(d+dir, 0755, osutil.AtomicWriteChmod, osutil.NoChown, osutil.NoChown)
```
benhoyt, line 39: It seems to me we should pass 0 for the flags for all existing tests, so we're keeping the tests for that behaviour. And we should add new tests that test the new chmod flag / behaviour.

benhoyt (changes requested):
Left some comments. Happy to discuss further too.

I'm not 100% sure we should add the flag. Maybe we should just always chmod it? But there is a precedent for the flag in the AtomicWrite functions, so that's what made me think we should add it.

internals/osutil/mkdirallchown_test.go
```diff
@@ -46,22 +48,131 @@ func (mkdacSuite) TestSlashySlashy(c *check.C) {
...
+	err := osutil.MkdirChown(tmpDir+"/foo", 0o777, osutil.MkdirChmod, osutil.NoChown, osutil.NoChown)
+	c.Assert(err, check.IsNil)
+	c.Assert(osutil.IsDir(tmpDir+"/foo"), check.Equals, true)
+
+	info, err := os.Stat(tmpDir + "/foo")
+	c.Assert(err, check.IsNil)
+	c.Assert(info.Mode().Perm(), check.Equals, os.FileMode(0o777))
+}
+
+// The following test is commented out because it would fail in GitHub Actions, but it would be nice
```
IronCore864, line 98: I don't have a good way to make it pass in GitHub Actions.

Unless I use the uid/gid 1001/127 under which user the workflow is running, but it seems the numbers might change (a few years back it was 1001/115), and it would probably fail locally.

Should I delete this test? But I feel it's mandatory to test this branch of the function.

Same below.

internals/osutil/mkdirallchown.go
```diff
@@ -23,6 +23,15 @@ import (
 	"github.com/canonical/pebble/internals/osutil/sys"
 )
 
+// MkdirFlags are a bitfield of flags for MakeDirs
```
benhoyt, line 26: Nit:
```suggestion
// MkdirFlags are a bitfield of flags for the Mkdir* functions.
```

internals/osutil/mkdirallchown_test.go
benhoyt: Could we make this a "root test", like the others in `.github/workflows/tests.yaml` and use `PEBBLE_TEST_USER` and `PEBBLE_TEST_GROUP` like `TestMakeDirsUserGroupReal` does?

internals/osutil/mkdirallchown.go
```diff
@@ -31,16 +40,36 @@ var mu sync.Mutex
...
+	return mkdirAllChown(filepath.Clean(path), perm, flags, uid, gid)
+}
+
+func mkdirAllChmod(path string, perm os.FileMode) error {
+	if _, err := os.Stat(path); err == nil {
+		return nil
+	}
+
+	parent := filepath.Dir(path)
+	if err := mkdirAllChmod(parent, perm); err != nil {
```
benhoyt, line 62: This doesn't look like the right logic, or at least not the same logic as `mkdirAllChown`, which uses the `s.IsDir()` to see if it's a directory or not.

benhoyt (commented):
Haven't reviewed in detail yet, but just a couple of comments that might unblock you.

internals/osutil/mkdirallchown.go
IronCore864: I see, the `IsDir` is necessary because the path might already exists as a file. I missed this in the first draft. Will change this function to the same logic as `mkdirAllChown`.

benhoyt:
I've looked at this more closely -- sorry for the delay. The function signatures, as well as the number of functions in the implementation, is getting unwieldy. I think we should go back to the drawing board and design a better function. How about this: make the two required arguments positional (path and perm) and put everything else in an options struct (which can be nil or the zero value for defaults):

```go
package osutil

func Mkdir(path string, perm os.FileMode, options *MkdirOptions) error { ... }

type MkdirOptions struct {
    // True to create any parent directories if they don't already exist.
    //
    // If false and path already exists, Mkdir returns an error that wraps os.ErrExist (like os.Mkdir).
    // If true and path already exists, Mkdir does nothing and returns nil (like os.MkdirAll).
    MakeParents bool

    // True to perform an explicit chmod on any directories created.
    Chmod bool

    // True to call chown on any directories created, using the user ID and group ID provided.
    Chown bool // NOTE: additional bool is needed because 0 is a valid value for UserID and GroupID (root)
    UserID sys.UserID
    GroupID sys.GroupID
}
```

I'm not sure about the "clever" `MakeParents` behaviour in the case where the path already exists. It imitates Go's `Mkdir` / `MkdirAll` quirks, but IMO that's a bit of a design flaw. Might be better to have an explicit `ExistOK bool` flag.

The above is somewhat modelled after https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

IronCore864:
Closing this PR for now, see the new PR which does the refactor and implements the feature [here](https://github.com/canonical/pebble/pull/418).

PR #406, opened 2024-04-03
fix(daemon): require admin access for POSTs and file pull API

benhoyt:
Most of this was introduced in PR #358, when we ported the AccessChecker changes from snapd, but accidentally set all the WriteAccess fields to UserAccess{} instead of AdminAccess{}. Previously there was a r.Method=="GET" check in Command.canAccess that handled this case.

Additionally:

- We lock down the files "pull" API to require admin. Even though it's a read (GET), this meant any user could potentially read sensitive files.
- We lock down the task-websocket endpoint to admin. This is a GET endpoint, but these websockets are used by exec to send stdin/out/err and commands to the exec'd process, so they should require admin too.

I've added some tests for these to ensure we don't accidentally change them in future, without noticing. How valuable these tests are I'm not sure, as they only cover a subset of the API endpoints, but it seems better than nothing.

### DIFF
--- a/internals/daemon/api.go
+++ b/internals/daemon/api.go
@@ -35,7 +35,7 @@ var API = []*Command{{
 }, {
 	Path:        "/v1/warnings",
 	ReadAccess:  UserAccess{},
-	WriteAccess: UserAccess{},
+	WriteAccess: AdminAccess{},
 	GET:         v1GetWarnings,
 	POST:        v1AckWarnings,
 }, {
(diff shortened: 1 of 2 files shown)


internals/daemon/api.go
```diff
@@ -101,7 +101,7 @@ var API = []*Command{{
 }, {
 	Path:        "/v1/notices",
 	ReadAccess:  UserAccess{},
-	WriteAccess: UserAccess{},
+	WriteAccess: AdminAccess{},
```
hpidcock, line 104: This will make posting user notices impossible. Is that the intent? I thought we already had fairly good checks inside the notices code?

benhoyt: Good spotting! This was actually not working earlier (when Pebble notices UID support was introduced), and fixed (unintentionally) by the CheckAccess PR. Reverted to `UserAccess{}` and commented on why this is the case.

hpidcock (approved):
I've reviewed each access check and manually tested where reasonable. Thank-you for fixing this.

PR #409, opened 2024-04-11
feat: perform health checks via changes and tasks

benhoyt:
TODO: draft

### DIFF
--- a/internals/cli/cmd_checks.go
+++ b/internals/cli/cmd_checks.go
@@ -77,14 +77,44 @@ func (cmd *cmdChecks) Execute(args []string) error {
 	w := tabWriter()
 	defer w.Flush()
 
-	fmt.Fprintln(w, "Check\tLevel\tStatus\tFailures")
+	fmt.Fprintln(w, "Check\tLevel\tStatus\tFailures\tChange")
 
 	for _, check := range checks {
 		level := check.Level
 		if level == client.UnsetLevel {
 			level = "-"
 		}
-		fmt.Fprintf(w, "%s\t%s\t%s\t%d/%d\n", check.Name, level, check.Status, check.Failures, check.Threshold)
+		change := "-"
+		if check.ChangeID != "" {
+			change = check.ChangeID
+			log, err := cmd.getLastTaskLog(check.ChangeID)
+			if err != nil {
+				change += " (ERROR: " + err.Error() + ")"
+			} else if log != "" {
+				// Truncate to 50 bytes with ellipsis in the middle
+				if len(log) > 50 {
+					log = log[:23] + "..." + log[len(log)-24:]
+				}
+				change += " (" + log + ")"
+			}
+		}
+		fmt.Fprintf(w, "%s\t%s\t%s\t%d/%d\t%s\n",
+			check.Name, level, check.Status, check.Failures, check.Threshold, change)
 	}
 	return nil
+}
+
+func (cmd *cmdChecks) getLastTaskLog(changeID string) (string, error) {
+	change, err := cmd.client.Change(changeID)
+	if err != nil {
+		return "", err
+	}
+	if len(change.Tasks) < 1 {
+		return "", nil
+	}
+	logs := change.Tasks[0].Log
+	if len(logs) < 1 {
+		return "", nil
+	}
+	return logs[len(logs)-1], nil
 }
--- a/internals/overlord/checkstate/checkers_test.go
+++ b/internals/overlord/checkstate/checkers_test.go
@@ -27,6 +27,7 @@ import (
 
 	. "gopkg.in/check.v1"
 
+	"github.com/canonical/pebble/internals/plan"
 	"github.com/canonical/pebble/internals/reaper"
 )
 
@@ -251,4 +252,122 @@ func (s *CheckersSuite) TestExec(c *C) {
 	detailsErr, ok = err.(*detailsError)
 	c.Assert(ok, Equals, true)
 	c.Assert(detailsErr.Details(), Equals, currentUser.Username)
+}
+
+func (s *CheckersSuite) TestNewChecker(c *C) {
+	chk := newChecker(&plan.Check{
+		Name: "http",
+		HTTP: &plan.HTTPCheck{
+			URL:     "https://example.com/foo",
+			Headers: map[string]string{"k": "v"},
+		},
+	}, nil)
+	http, ok := chk.(*httpChecker)
+	c.Assert(ok, Equals, true)
+	c.Check(http.name, Equals, "http")
+	c.Check(http.url, Equals, "https://example.com/foo")
+	c.Check(http.headers, DeepEquals, map[string]string{"k": "v"})
+
+	chk = newChecker(&plan.Check{
+		Name: "tcp",
+		TCP: &plan.TCPCheck{
+			Port: 80,
+			Host: "localhost",
+		},
+	}, nil)
+	tcp, ok := chk.(*tcpChecker)
+	c.Assert(ok, Equals, true)
+	c.Check(tcp.name, Equals, "tcp")
+	c.Check(tcp.port, Equals, 80)
+	c.Check(tcp.host, Equals, "localhost")
+
+	userID, groupID := 100, 200
+	chk = newChecker(&plan.Check{
+		Name: "exec",
+		Exec: &plan.ExecCheck{
+			Command:     "sleep 1",
+			Environment: map[string]string{"k": "v"},
+			UserID:      &userID,
+			User:        "user",
+			GroupID:     &groupID,
+			Group:       "group",
+			WorkingDir:  "/working/dir",
+		},
+	}, nil)
+	exec, ok := chk.(*execChecker)
+	c.Assert(ok, Equals, true)
+	c.Assert(exec.name, Equals, "exec")
+	c.Assert(exec.command, Equals, "sleep 1")
+	c.Assert(exec.environment, DeepEquals, map[string]string{"k": "v"})
+	c.Assert(exec.userID, Equals, &userID)
+	c.Assert(exec.user, Equals, "user")
+	c.Assert(exec.groupID, Equals, &groupID)
+	c.Assert(exec.workingDir, Equals, "/working/dir")
+}
+
+func (s *CheckersSuite) TestExecContextNoOverride(c *C) {
+	svcUserID, svcGroupID := 10, 20
+	chk := newChecker(&plan.Check{
+		Name: "exec",
+		Exec: &plan.ExecCheck{
+			Command:        "sleep 1",
+			ServiceContext: "svc1",
+		},
+	}, &plan.Plan{Services: map[string]*plan.Service{
+		"svc1": {
+			Name:        "svc1",
+			Environment: map[string]string{"k": "x", "a": "1"},
+			UserID:      &svcUserID,
+			User:        "svcuser",
+			GroupID:     &svcGroupID,
+			Group:       "svcgroup",
+			WorkingDir:  "/working/svc",
+		},
+	}})
+	exec, ok := chk.(*execChecker)
+	c.Assert(ok, Equals, true)
+	c.Check(exec.name, Equals, "exec")
+	c.Check(exec.command, Equals, "sleep 1")
+	c.Check(exec.environment, DeepEquals, map[string]string{"k": "x", "a": "1"})
+	c.Check(exec.userID, DeepEquals, &svcUserID)
+	c.Check(exec.user, Equals, "svcuser")
+	c.Check(exec.groupID, DeepEquals, &svcGroupID)
+	c.Check(exec.workingDir, Equals, "/working/svc")
+}
+
+func (s *CheckersSuite) TestExecContextOverride(c *C) {
+	userID, groupID := 100, 200
+	svcUserID, svcGroupID := 10, 20
+	chk := newChecker(&plan.Check{
+		Name: "exec",
+		Exec: &plan.ExecCheck{
+			Command:        "sleep 1",
+			ServiceContext: "svc1",
+			Environment:    map[string]string{"k": "v"},
+			UserID:         &userID,
+			User:           "user",
+			GroupID:        &groupID,
+			Group:          "group",
+			WorkingDir:     "/working/dir",
+		},
+	}, &plan.Plan{Services: map[string]*plan.Service{
+		"svc1": {
+			Name:        "svc1",
+			Environment: map[string]string{"k": "x", "a": "1"},
+			UserID:      &svcUserID,
+			User:        "svcuser",
+			GroupID:     &svcGroupID,
+			Group:       "svcgroup",
+			WorkingDir:  "/working/svc",
+		},
+	}})
+	exec, ok := chk.(*execChecker)
+	c.Assert(ok, Equals, true)
+	c.Check(exec.name, Equals, "exec")
+	c.Check(exec.command, Equals, "sleep 1")
+	c.Check(exec.environment, DeepEquals, map[string]string{"k": "v", "a": "1"})
+	c.Check(exec.userID, DeepEquals, &userID)
+	c.Check(exec.user, Equals, "user")
+	c.Check(exec.groupID, DeepEquals, &groupID)
+	c.Check(exec.workingDir, Equals, "/working/dir")
 }
--- a/internals/overlord/checkstate/handlers.go
+++ b/internals/overlord/checkstate/handlers.go
@@ -49,13 +49,12 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tomb.Tomb) error {
 	m.checksLock.Lock()
 	data := m.checks[config.Name]
 	if data == nil {
-		data = &checkData{
-			config:   config,
-			cancel:   func() { tomb.Kill(nil) },
-			changeID: changeID,
-		}
+		data = &checkData{}
 		m.checks[config.Name] = data
 	}
+	data.config = config
+	data.cancel = func() { tomb.Kill(nil) }
+	data.changeID = changeID
 	m.checksLock.Unlock()
 
 	logger.Debugf("Performing check %q with period %v", details.Name, config.Period.Value)
@@ -136,14 +135,13 @@ func (m *CheckManager) doRecoverCheck(task *state.Task, tomb *tomb.Tomb) error {
 	m.checksLock.Lock()
 	data := m.checks[config.Name]
 	if data == nil {
-		data = &checkData{
-			config:   config,
-			cancel:   func() { tomb.Kill(nil) },
-			failures: 1,
-			changeID: changeID,
-		}
+		data = &checkData{}
 		m.checks[config.Name] = data
 	}
+	data.config = config
+	data.cancel = func() { tomb.Kill(nil) }
+	data.failures = 1
+	data.changeID = changeID
 	m.checksLock.Unlock()
 
 	logger.Debugf("Recovering check %q with period %v", details.Name, config.Period.Value)
--- a/internals/overlord/checkstate/handlers.go
+++ b/internals/overlord/checkstate/handlers.go
@@ -1,0 +1,209 @@
+// Copyright (c) 2024 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package checkstate
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"gopkg.in/tomb.v2"
+
+	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/plan"
+)
+
+func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tomb.Tomb) error {
+	var details checkDetails
+	m.state.Lock()
+	changeID := task.Change().ID()
+	err := task.Get("check-details", &details)
+	m.state.Unlock()
+	if err != nil {
+		return fmt.Errorf("cannot get check details for task %q: %v", task.ID(), err)
+	}
+
+	m.planLock.Lock()
+	plan := m.plan
+	m.planLock.Unlock()
+	config, ok := plan.Checks[details.Name]
+	if !ok {
+		logger.Debugf("Check %q no longer exists in plan.", details.Name)
+		return nil
+	}
+
+	m.checksLock.Lock()
+	data := m.checks[config.Name]
+	if data == nil {
+		data = &checkData{
+			config:   config,
+			cancel:   func() { tomb.Kill(nil) },
+			changeID: changeID,
+		}
+		m.checks[config.Name] = data
+	}
+	m.checksLock.Unlock()
+
+	logger.Debugf("Performing check %q with period %v", details.Name, config.Period.Value)
+	ticker := time.NewTicker(config.Period.Value)
+	defer ticker.Stop()
+
+	chk := newChecker(config, plan)
+	for {
+		select {
+		case <-ticker.C:
+			err := runCheck(tomb.Context(nil), config, chk, config.Timeout.Value)
+			if err != nil {
+				// Check failed, switch to recovering a failing check.
+				failures, actionRan := m.handleCheckFailure(config, err)
+
+				// Hold checksLock while updating state so that if someone
+				// calls Checks() they always see the latest state. We must
+				// always acquire checksLock and then state lock in that order.
+
+				m.checksLock.Lock()
+
+				m.state.Lock()
+				newChangeID := m.recoverCheckChange(config)
+				m.state.EnsureBefore(0)
+				m.state.Unlock()
+
+				data.failures = failures
+				data.actionRan = actionRan
+				data.changeID = newChangeID
+				m.checksLock.Unlock()
+
+				// Returning the error means perform-check goes to Error status
+				// and logs the error to the task log.
+				return err
+			}
+
+		case <-tomb.Dying():
+			logger.Debugf("Check %q stopped during perform-check: %v", config.Name, tomb.Err())
+			return tomb.Err()
+		}
+	}
+}
+
+func runCheck(ctx context.Context, config *plan.Check, chk checker, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	err := chk.check(ctx)
+	if err != nil {
+		if errors.Is(ctx.Err(), context.Canceled) {
+			// Check was stopped, don't trigger failure action.
+			logger.Debugf("Check %q canceled in flight", config.Name)
+			return nil
+		}
+		return err
+	}
+	return nil
+}
+
+func (m *CheckManager) doRecoverCheck(task *state.Task, tomb *tomb.Tomb) error {
+	var details checkDetails
+	m.state.Lock()
+	changeID := task.Change().ID()
+	err := task.Get("check-details", &details)
+	m.state.Unlock()
+	if err != nil {
+		return fmt.Errorf("cannot get check details for task %q: %v", task.ID(), err)
+	}
+
+	m.planLock.Lock()
+	plan := m.plan
+	m.planLock.Unlock()
+	config, ok := plan.Checks[details.Name]
+	if !ok {
+		logger.Debugf("Check %q no longer exists in plan.", details.Name)
+		return nil
+	}
+
+	m.checksLock.Lock()
+	data := m.checks[config.Name]
+	if data == nil {
+		data = &checkData{
+			config:   config,
+			cancel:   func() { tomb.Kill(nil) },
+			failures: 1,
+			changeID: changeID,
+		}
+		m.checks[config.Name] = data
+	}
+	m.checksLock.Unlock()
+
+	logger.Debugf("Recovering check %q with period %v", details.Name, config.Period.Value)
+	ticker := time.NewTicker(config.Period.Value)
+	defer ticker.Stop()
+
+	chk := newChecker(config, plan)
+	for {
+		select {
+		case <-ticker.C:
+			err := runCheck(tomb.Context(nil), config, chk, config.Timeout.Value)
+			if err != nil {
+				failures, actionRan := m.handleCheckFailure(config, err)
+
+				m.checksLock.Lock()
+
+				m.state.Lock()
+				task.Errorf("%v", err) // add error to task log
+				m.state.Unlock()
+
+				data.failures = failures
+				data.actionRan = actionRan
+				m.checksLock.Unlock()
+				break
+			}
+
+			// Check succeeded, switch to performing a succeeding check.
+			m.checksLock.Lock()
+
+			m.state.Lock()
+			newChangeID := m.performCheckChange(config)
+			m.state.EnsureBefore(0)
+			m.state.Unlock()
+
+			data.failures = 0
+			data.actionRan = false
+			data.changeID = newChangeID
+			m.checksLock.Unlock()
+			return nil
+
+		case <-tomb.Dying():
+			logger.Debugf("Check %q stopped during recover-check: %v", config.Name, tomb.Err())
+			return tomb.Err()
+		}
+	}
+}
+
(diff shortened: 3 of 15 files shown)


internals/overlord/checkstate/manager.go
```diff
@@ -14,30 +14,57 @@
...
-	return &CheckManager{}
+func NewManager(s *state.State, runner *state.TaskRunner) *CheckManager {
+	manager := &CheckManager{
+		state:  s,
+		checks: make(map[string]*checkData),
+	}
+	// Health check changes can be long-running; ensure they don't get pruned.
+	s.RegisterPendingChangeByAttr(noPruneAttr, func(change *state.Change) bool {
+		return true
+	})
```
flotter, line 59: I need to copy this!

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +75,53 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
-	checks := make(map[string]*checkData, len(p.Checks))
-	for name, config := range p.Checks {
-		ctx, cancel := context.WithCancel(context.Background())
-		check := &checkData{
-			config:  config,
-			checker: newChecker(config, p),
-			ctx:     ctx,
-			cancel:  cancel,
-			action:  m.callFailureHandlers,
+func (m *CheckManager) PlanChanged(newPlan *plan.Plan) {
```
flotter, line 78: Q: Do you think there could be some benefit from decoupling plan change induced work from the ```PlanChanged``` handler itself? I like the idea of never having the ability for a notification consumer to lock up or delay the writer, especially if its a global writer attending to all managers in the system.

So I deal with the repercussions of the new plan within my own event routine. I also use the non-blocking write pattern we briefly discussed, allowing the consumer no take another plan change unless its ready do so, without an impact on the writer. It could miss an incremental plan change in the process, but it would always get the latest change at least.

```
// PlanChanged is called by the Plan Manager when the plan is changed.
func (m *NetManager) PlanChanged(plan *plan.Plan) {
	// This is a single writer non-block pattern using a single entry
	// buffered channel. Remove entry if it exist, and [re-]insert.
	select {
	case <-m.planChanged:
	default:
	}
	m.planChanged <- plan
}
```

Stealing your PR for feedback on my PR ;)

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +75,53 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+	m.state.Lock()
+	defer m.state.Unlock()
+	for _, config := range newPlan.Checks {
+		m.performCheckChange(config)
+	}
+	if !m.ensureDone.Load() {
+		// Can't call EnsureBefore before Overlord.Loop is running (which will
+		// call m.Ensure for the first time).
+		return
 	}
```
flotter, line 107: I need this as well.

internals/overlord/checkstate/manager.go
```diff
@@ -14,30 +14,57 @@
...
 // CheckManager starts and manages the health checks.
 type CheckManager struct {
-	mutex           sync.Mutex
-	wg              sync.WaitGroup
-	checks          map[string]*checkData
+	state      *state.State
+	ensureDone atomic.Bool
+
+	checksLock sync.Mutex
+	checks     map[string]*checkData
```
flotter, line 39: Super nitty - I saw when we are iterating over ```checks``` the item is called ```data``` which was not super clear when I read it, so I came here, and see the item is called ```checkData```. Is there a reason why ```check``` did not work ? some namespace clash ?

flotter:
@benhoyt still busy reviewing ... will continue asap.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,207 @@
...
+		case <-tomb.Dying():
+			logger.Debugf("Check %q stopped during perform-check: %v", config.Name, tomb.Err())
+			return tomb.Err()
+		}
+	}
+}
+
+func runCheck(ctx context.Context, config *plan.Check, chk checker, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
```
flotter, line 103: For my understanding: I would have assumed that the ```check(ctx)``` is blocking and only returns when it is done. The context will be cancelled when the tomb dies, but what is the purpose of having using the ```cancel``` on function exit after the check completed or failed ?

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,207 @@
...
+				data.changeID = newChangeID
+				m.checksLock.Unlock()
+
+				// Returning the error means perform-check goes to Error status
+				// and logs the error to the task log.
+				return err
+			}
+
+		case <-tomb.Dying():
+			logger.Debugf("Check %q stopped during perform-check: %v", config.Name, tomb.Err())
```
flotter, line 95: I think that if you ```cancel``` this due to a plan change, this error will be ```nil```, so the message will always end with ```<nil>``` in those cases ?

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +75,53 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+func (m *CheckManager) PlanChanged(newPlan *plan.Plan) {
+	// TODO: should figure out how to not stop/restart checks that haven't changed
+	//       so that unnecessary changes/tasks aren't created
+
+	// First cancel tasks of currently-running checks.
+	m.checksLock.Lock()
+	for name, data := range m.checks {
+		if data.cancel != nil {
+			data.cancel()
 		}
```
flotter, line 87: I assume this mechanism is in place, as apposed to ```change.Abort()```, which would require us to use the current change or change ID to abort, so we have a single way to cancel whichever change is currently in flight.

Would it not be possible for a transition between prepare and recover to sometimes take place even with the context cancelled (i.e. tomb in dying state) ? I guess worse case on arrival, the check would have been deleted from the map, and would then exit in any case.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,207 @@
...
+	"fmt"
+	"time"
+
+	"gopkg.in/tomb.v2"
+
+	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/plan"
+)
+
```
flotter, line 29: The design feels like it will work nicely for me.

Based on some examples I've studied and how I try to think about changes / tasks (which could well be wrong), there are a couple "bonus features"  you can pull out of the hat if you do it in a specific way, which may or may not be useful for every case I am sure:

- If multiple tasks use the same data, you can ```Set```/```Get``` the data at the change level
- Since the changes / tasks can act as an audit trial, I think it becomes super interesting to have the data associated with the request ```Set```, so that success and failure contexts are clear for something that happened 1 month ago, and for which the actual details could have been updated in the plan, without changing the check name, for example.
- In the latter case, it means the change does not directly need to monitor changes in its parameters, but relies on top level management to Abort it when it is no longer applicable.

I wonder ... just brainstorming here for ideas, which again may be unwanted for the case at hand ... 
```
Check { HTTPTasks { PerformHTTPCheck ; RecoverHTTPCheck } }
Check { TCPTasks { PerformTCPCheck ; RecoverTCPCheck } }
Check { ExecTasks { PerformExecCheck ; RecoverExecCheck } }
```

One other way I would imagine one can spin this is to say that one change is tracking a check for its lifetime, and the perform and recover task is type specific, so that the change can have check ```Set``` data that describes what has to be done. Top level management would simply Abort when the change is out of date, and crate a new change new updated or new checks.

flotter (commented):
Looks like a feasible approach to be Ben. I've added some questions and a couple of crazy ideas as counter spins on the solution, but that's mostly just dumping thoughts that went through my mind. I am also taking some code from you after this, thanks!

internals/overlord/checkstate/manager.go
benhoyt: I think it's an interesting and useful idea. And I think if `PlanChanged` were doing something even slightly long-running I'd use it. However, I don't think the plan will change very often, and whatever goroutine that needs to update the `CheckManager` will still need to take the locks and whatnot, so I'm not sure it'll help much here.

internals/overlord/checkstate/manager.go
benhoyt: I wasn't totally happy with this, to be honest. I wonder if we should consider refactoring the higher layers a little so `PlanChanged` is only called after the overlord is running...

internals/overlord/overlord.go
```diff
@@ -43,12 +43,14 @@ import (
...
 	pruneInterval  = 10 * time.Minute
-	pruneWait      = 24 * time.Hour * 1
-	abortWait      = 24 * time.Hour * 7
 
-	pruneMaxChanges = 500
+	// In snapd this is 24h, but that's too short in the context of Pebble.
+	pruneWait = 24 * time.Hour * 7
+
+	// In snapd this is 7d, but also increase that in the context of Pebble.
+	abortWait = 24 * time.Hour * 14
```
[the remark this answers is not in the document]
benhoyt: Yeah, I think we should keep that in mind. These are somewhat made-up values, and kind of an open question in the spec (in a comment I ask "should pruneWait be 10 days, 14?"), but we generally felt 24h was too short. I'll leave this for now and we can consider making them configurable later.

internals/overlord/checkstate/handlers.go
benhoyt: Yep, `check(ctx)` is blocking and only returns when it is done. The `WithTimeout` context will be cancelled when the parent (tomb) dies or when the timeout occurs. The `cancel()` call is to free resources, per the [docs](https://pkg.go.dev/context#WithDeadline):

> Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this [Context](https://pkg.go.dev/context#Context) complete.

Also from the example:

```
// Even though ctx will be expired, it is good practice to call its
// cancellation function in any case. Failure to do so may keep the
// context and its parent alive longer than necessary.
defer cancel()
```

internals/overlord/checkstate/handlers.go
benhoyt: Yes, that's right. I figured `<nil>` was probably okay for a debug message, but I've improved this now.

internals/overlord/checkstate/manager.go
benhoyt: > Would it not be possible for a transition between prepare and recover to sometimes take place even with the context cancelled (i.e. tomb in dying state) ? I guess worse case on arrival, the check would have been deleted from the map, and would then exit in any case, so actually it looks fine.

Oh, this is a good point. It's actually worse than that, because the check isn't necessarily in the map yet. I'd love to brainstorm with you (maybe tomorrow morning my time) a better way to approach this.

internals/overlord/checkstate/checkers_test.go
```diff
@@ -252,3 +253,121 @@ func (s *CheckersSuite) TestExec(c *C) {
 	c.Assert(ok, Equals, true)
 	c.Assert(detailsErr.Details(), Equals, currentUser.Username)
 }
+
+func (s *CheckersSuite) TestNewChecker(c *C) {
```
benhoyt, line 257: This was simply moved from `manager_test.go`.

internals/overlord/checkstate/handlers.go
benhoyt: I've reworked this quite a bit now. Commentary in main thread.

internals/overlord/checkstate/manager.go
benhoyt: Per voice discussion, the problem with that method is if two people are concurrently calling the API to append/combine a layer, the second one may block. The above technique is only safe if you can guarantee that there's only one concurrent caller of `PlanChanged`, which isn't true for us.

I've gone with standard locking, though I've rearranged it quite a bit.

internals/overlord/checkstate/manager.go
benhoyt: I've reworked this significantly so all the data is stored in change/task data, so there's only one copy (with the exception of the "health" cache for the `/v1/health` endpoint).

benhoyt:
@flotter I've reworked this quite a bit now. In particular, I tried to store all the data in change/task data to avoid a separate "m.checks" map and trying to keep that in sync. I almost had it ... and then I remembered (actually, noticed in testing!) the `GET /v1/health` endpoint isn't allowed to acquire the state lock, so it couldn't call `Checks`. I ended up creating a kind of cache `m.health` and a new method `Health` that returns a cut-back version of the data *just* for the health endpoint. I'm not entirely happy with it, but at least the health-related stuff is fairly isolated from the main "checks engine". Happy to consider alternative approaches.

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +87,114 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+	m.plan = newPlan
+	m.planLock.Unlock()
+
+	if oldPlan == nil {
+		oldPlan = &plan.Plan{}
+	}
+	shouldEnsure := false
+	newOrModified := make(map[string]bool, len(newPlan.Checks))
+
+	// Abort all currently-running checks.
```
flotter, line 106: If me understanding is correct, this block aborts all currently-running checks that have either been removed, or modified.

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +87,114 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+	for _, change := range m.state.Changes() {
+		switch change.Kind() {
+		case performCheckKind, recoverCheckKind:
```
flotter, line 109: I believe that with the new design, a new change (recover/perform) created for a particular check will only ever now happen after the existing change is ```ready```, which means we will only ever see one change at most that is non-ready in the loop for any given change, which is nice.

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +87,114 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+func (m *CheckManager) changeStatusChanged(chg *state.Change, old, new state.Status) {
+	// Always acquire locks in same order (state lock, then plan lock).
+	// The state engine has already acquired the state lock at this point.
+	m.planLock.Lock()
+	plan := m.plan
+	m.planLock.Unlock()
+
+	shouldEnsure := false
+	switch {
+	case chg.Kind() == performCheckKind && new == state.ErrorStatus:
```
flotter, line 165: I cannot remember what the final state of an Aborted change is? I am wondering what will happen with a check that is detected as modified, where the original change will be aborted. If that happens to be an ```performCheckKind``` change, will it end up here?

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +87,114 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+	tasks := change.Tasks()
+	if len(tasks) != 1 {
+		panic(fmt.Sprintf("internal error: %s change %s should have one task", change.Kind(), change.ID()))
+	}
+	var details checkDetails
+	err := tasks[0].Get(checkDetailsAttr, &details)
+	if err != nil {
+		panic(fmt.Sprintf("internal error: cannot get %s change %s check details: %v", change.Kind(), change.ID(), err))
+	}
+	return details
```
flotter, line 197: Could it remove a couple of lines of code if you place the ```checkDetailsAttr``` at the change level (```change.Get/Set```) ?

internals/overlord/checkstate/manager.go
```diff
@@ -143,12 +268,42 @@ func newChecker(config *plan.Check, p *plan.Plan) checker {
...
+
+	// Populate name, number of failures, and change ID from state.
+	var infos []*CheckInfo
+	for _, change := range m.state.Changes() {
+		switch change.Kind() {
+		case performCheckKind, recoverCheckKind:
+			if change.IsReady() {
+				// Skip check changes that have finished already.
+				continue
+			}
```
flotter, line 282: If I read this correctly, we are saying that running changes represents the active checks, and each running check holds the latest truth about its check state. I actually like this - it was easier for me to follow the flow of data, and ownership in general. Just a quick question now ... Is there an opportunity for a query (API) to happen exactly at a point in time between two changes (where the one change become ready - ```error``` or ```done```), and  the next change has not yet been committed? I think it boils down to when in the state engine the ```changeStatusChanged``` is called from. If while the state engine has its lock, the change transitions to ```Error``` / ```Done``` and while having the lock it calls the change callback, if feels like everything works perfectly for me.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,205 @@
...
+	err := task.Get(checkDetailsAttr, &details)
+	m.state.Unlock()
+	if err != nil {
+		return fmt.Errorf("cannot get check details for perform-check task %q: %v", task.ID(), err)
+	}
+	config, ok := plan.Checks[details.Name]
+	if !ok {
+		// Check no longer exists in plan.
+		return nil
+	}
```
flotter, line 45: How important is this? Can we not leave the Plan changed handler to manage check removals / modifications exclusively? I have not thought this through, but since the change is always aborted if the plan changes, I think I would fire off a change with whatever it needs from the change included in the task payload.

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +87,114 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+			details := mustGetCheckDetails(change)
+			oldConfig, inOld := oldPlan.Checks[details.Name]
+			newConfig, inNew := newPlan.Checks[details.Name]
+			if inOld && inNew {
+				if reflect.DeepEqual(oldConfig, newConfig) {
```
flotter, line 118: Do we handle a case where the service context changes for an existing check that in itself is not modified ?

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,205 @@
...
+				m.state.Lock()
+				task.Logf("succeeded after %s", pluralise(details.Failures, "failure", "failures"))
+				details.Failures = 0
+				task.Set(checkDetailsAttr, &details)
+				m.state.Unlock()
+			}
+
+		case <-tomb.Dying():
+			break
+		}
```
flotter, line 95: If you ```break``` with a label to get outside the for loop, would the exit code below still be needed ?

flotter (commented):
Hi Ben. Your PR looks good to me. Compared with the previous design, I feels simpler. I left a couple of minor comments. My only question is whether it is practical to not expose the changes to plan changes at all, and dispatch a change with a snapshot of what it needs from the plan at creation time.

internals/overlord/checkstate/manager.go
benhoyt: Aborted changes get moved to Done status, so this one be okay, however, the `chg.Kind() == recoverCheckKind && new == state.DoneStatus` case below will yield true. But that's why I have the `details.Proceed` check in addition: I set a "proceed=true" boolean in the task data when the change should proceed to the next change kind. That's also useful if the task returns for some other reason (eg: perform-check returning another kind of error).

internals/overlord/checkstate/manager.go
benhoyt: I realise it's much of a muchness in this case, as we always have one task per change, but given that each handler takes the Task and not the Change object, it seems "righter" to do it this way. (I'd have to do the `task.Change()` indirection in the handler, which doesn't feel great.) So I think I'll leave this one as is.

internals/overlord/checkstate/manager.go
benhoyt: Yeah, I liked having all the data in the change/task too. I traced that through the state engine and yes, `changeStatusChanged` is called with the state locked throughout all this. [`TaskRunner.run` calls `SetStatus`](https://github.com/canonical/pebble/blob/18c93cbc280de20c9e21d323fe3a8bbb340777a2/internals/overlord/state/taskrunner.go#L268-L297) which updates the status from Doing to Done (or Error) and `SetStatus` calls the status-changed handler.

However, I had to bypass this "all the data in the change/task" due to the `GET /v1/health` endpoint requiring that the state lock isn't held. Saving modified state is relatively slow (due to the JSON serialisation and fsync), and in certain deployments when the machine was under load it would take too long to respond ([see details](https://github.com/canonical/pebble/pull/369)). We have a task on next cycle's roadmap to improve the speed/operation of state saving, but that's longer term.

In the meantime, I had to store a side map `CheckManager.health` with a mutex around it for the Health endpoint. I'm beginning to think this is a mistake, and I should remove the new parallel `Health` method and just have a `CheckManager.checks` map again, but it would be a `map[string]CheckInfo` instead of that funky `checkData` struct. All the operational logic and plan updates would still be done via changes. So I think I'll make them change yet, and I think it'll be simpler.

internals/overlord/checkstate/manager.go
benhoyt: Yep, that's right, that is nice.

internals/overlord/checkstate/handlers.go
benhoyt: This is an interesting idea that I did wonder about when coding it. It'd certainly be nice conceptually, but it means that (potentially sizeable) check configuration has to be serialized in the state JSON along with the task (and I'm already concerned about the additional size/slowness of the state JSON here). I'll think about it, and see if there's some way to have non-persisted attributes on the task or something.

internals/overlord/checkstate/manager.go
benhoyt: Ah, I'd missed that! That's a really good point. And it ties into the above -- if we end up storing this on the task, It'd be better storing the final (merged-with-service-context) checker config rather just the `plan.Check`.

internals/overlord/checkstate/handlers.go
benhoyt: Yeah, because that `tombErr != tombpkg.ErrStillAlive` block (I assume that's what you mean by "exit code") is also needed in the case when we get a `ticker.C` and the context is cancelled (change aborted). `runCheck` returns nil in this case (see it's code) because the check itself didn't fail, but we still need to check the `tomb.Err()` and return if we're done. I thought a bit about a better way to structure this, but couldn't see an obvious way -- open to suggestions.

internals/overlord/checkstate/manager.go
benhoyt: Fixed.

internals/overlord/checkstate/handlers.go
benhoyt: Then again, I just discovered `exec` changes are even bigger! https://github.com/canonical/pebble/issues/411 -- that's even worse...

benhoyt: It looks like there's `TaskRunner.AddCleanup`, which allows you to call a function by task kind to clean stuff up whenever the change completes. We could use this to delete the config data -- IMO it's not that interesting for introspection after the task completes.

Alternatively we could store this data as non-persistent: there's no way to do that on the task or change, but there is at the state level via `State.Cache` (set/delete) and `State.Cached` (get). (And then you can delete that via an `AddCleanup` hook, just to avoid it staying in memory.) Snapd uses keys like `type remodelCtxKey struct { chgID string }`. I'm definitely tempted to do this via `State.Cache` -- I'll think further about this.

internals/overlord/checkstate/manager.go
benhoyt: Fixed now.

internals/overlord/checkstate/manager.go
benhoyt: I've now reverted the new `Health()` method and am using `CheckManager.checks map[string]CheckInfo` as described above.

internals/overlord/checkstate/handlers.go
benhoyt: For what it's worth, I'm now doing this -- I'm using `State.Cache` to store the check config (non-persistently) with a cleanup handler to remove it from the cache. So the config data won't pollute the persisted JSON, but it still gets the simplification of avoiding `m.plan` and `m.planLock` (and them potentially being out of sync).

internals/cli/cmd_checks.go
```diff
@@ -77,14 +78,62 @@ func (cmd *cmdChecks) Execute(args []string) error {
...
+		return "", nil
+	}
+	logs := change.Tasks[0].Log
+	if len(logs) < 1 {
+		return "", nil
+	}
+	// Strip initial "<timestamp> ERROR" text from log.
+	lastLog := logs[len(logs)-1]
+	fields := strings.SplitN(lastLog, " ", 3)
+	if len(fields) > 2 && fields[1] == "ERROR" {
```
hpidcock, line 134: Huh? This smells, why do we only strip timestamp + severity when it is error? Please add a comment with the reasoning.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/plan"
+)
+
+func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) error {
+	m.state.Lock()
+	changeID := task.Change().ID()
+	var details checkDetails
+	err := task.Get(checkDetailsAttr, &details)
+	config := m.state.Cached(performConfigKey{changeID}).(*plan.Check)
```
hpidcock, line 35: do we need a nil check on this?

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,198 @@
...
+
+	logger.Debugf("Performing check %q with period %v", details.Name, config.Period.Value)
+	ticker := time.NewTicker(config.Period.Value)
+	defer ticker.Stop()
+
+	chk := newChecker(config)
+	for {
+		select {
+		case <-ticker.C:
+			err := runCheck(tomb.Context(nil), chk, config.Timeout.Value)
```
hpidcock, line 49: Some of the logic in this error handling feels like it should be broken out into another method, but how I'm not sure.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,198 @@
...
+				m.state.Unlock()
+			}
+
+		case <-tomb.Dying():
+			break
+		}
+
+		// Do this check here so that select doesn't sometimes pick up another
+		// ticker.C tick after the tomb has been killed.
+		tombErr := tomb.Err()
```
hpidcock, line 98: This doesn't actually protect you against this, just mitigates it somewhat.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,198 @@
...
+	logger.Debugf("Performing check %q with period %v", details.Name, config.Period.Value)
+	ticker := time.NewTicker(config.Period.Value)
+	defer ticker.Stop()
+
+	chk := newChecker(config)
+	for {
+		select {
+		case <-ticker.C:
+			err := runCheck(tomb.Context(nil), chk, config.Timeout.Value)
+			if err != nil {
```
hpidcock, line 50: This is probably a better way to deal with the tomb dying during or before the check starts.
```
if errors.Is(err, context.Canceled) && tomb.Dying() {
    return tomb.Err()
} else if err != nil {
...
```

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+}
+
+func runCheck(ctx context.Context, chk checker, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	err := chk.check(ctx)
+	if err != nil {
+		if errors.Is(ctx.Err(), context.Canceled) {
```
hpidcock, line 111: Should just deal with this in the doPerformCheck as it is not a concern of this method, it is given a context, so it should return context.Canceled.

Since this runCheck function creates a timeout context, it should deal with the context.DeadlineExceeded and return an informative error about how the check timed out.

internals/overlord/checkstate/handlers.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	return nil
+}
+
+func (m *CheckManager) doRecoverCheck(task *state.Task, tomb *tombpkg.Tomb) error {
+	m.state.Lock()
+	changeID := task.Change().ID()
+	var details checkDetails
+	err := task.Get(checkDetailsAttr, &details)
+	config := m.state.Cached(recoverConfigKey{changeID}).(*plan.Check)
```
hpidcock, line 126: probably also need a nil check here too

internals/overlord/checkstate/handlers.go
hpidcock: I think this is an important thing to understand about context cancellation/deadlines and how we handle the return of errors.

We need to be mindful of which context has been cancelled/deadlineexceeded. Because a child-context has cancelled/deadlined, does not mean the parent context has done that. It is very situation dependent.

internals/overlord/checkstate/manager.go
```diff
@@ -48,42 +100,99 @@ func (m *CheckManager) NotifyCheckFailed(f FailureFunc) {
...
+			merged := mergeServiceContext(newPlan, config)
+			changeID := performCheckChange(m.state, merged)
+			m.updateCheckInfo(config, changeID, 0)
+			shouldEnsure = true
+		}
+	}
+	if !m.ensureDone.Load() {
+		// Can't call EnsureBefore before Overlord.Loop is running (which will
+		// call m.Ensure for the first time).
+		return
```
hpidcock, line 164: So when is the EnsureBefore called if this is false? does it need to be called?

internals/cli/cmd_checks.go
benhoyt: Yeah, fair point. I think I'm just going to strip the kind in all cases to avoid confusion. In reality it should only be ERROR here when the check is down.

internals/overlord/checkstate/handlers.go
benhoyt: Yeah, I've played with it a bit, but it's fairly different between perform-check and recover-check. I've tweaked it based on your other feedback, but I think it's good enough.

internals/overlord/checkstate/handlers.go
benhoyt: Ah yes. I've rearranged this per our discussion and your comments below.

internals/overlord/checkstate/handlers.go
benhoyt: Good point, done.

internals/overlord/checkstate/handlers.go
benhoyt: Good points. Done. I've removed this check here and moved it up a level.

internals/overlord/checkstate/handlers.go
benhoyt: Per above, comment(s) added.

internals/overlord/checkstate/manager.go
benhoyt: In the overlord startup when it loads the plan the first time (I've traced it through). That's what the comment is referring to -- not actually EnsureBefore which is a helper function, but Ensure itself.

internals/overlord/checkstate/checkers.go
```diff
@@ -202,3 +204,7 @@ type detailsError struct {
 func (e *detailsError) Details() string {
 	return e.details
 }
+
+func (e *detailsError) Unwrap() error {
```
benhoyt, line 208: This isn't actually needed, but it was useful during debugging and I realised this error-wrapper type should probably have an Unwrap method, so I kept it.

internals/overlord/checkstate/manager_test.go
```diff
@@ -156,7 +156,7 @@ func (s *ManagerSuite) TestTimeout(c *C) {
 	status := change.Status()
 	st.Unlock()
 	c.Assert(status, Equals, state.ErrorStatus)
-	c.Assert(lastTaskLog(st, performChangeID), Matches, ".* ERROR exec check timed out")
+	c.Assert(lastTaskLog(st, performChangeID), Matches, ".* ERROR check timed out after 25ms")
```
hpidcock, line 159: Yeah that's nicer.

internals/overlord/checkstate/handlers.go
```diff
@@ -161,15 +155,7 @@ func (m *CheckManager) doRecoverCheck(task *state.Task, tomb *tombpkg.Tomb) erro
...
-			break
-		}
-
-		// Do this check here so that select doesn't sometimes pick up another
-		// ticker.C tick after the tomb has been killed.
-		tombErr := tomb.Err()
-		if tombErr != tombpkg.ErrStillAlive {
-			logger.Debugf("Check %q stopped during recover-check%s", config.Name, errReason(tombErr))
-			return tombErr
+			return checkStopped(config.Name, task.Kind(), tomb.Err())
```
hpidcock, line 158: Nice, much cleaner.

hpidcock (approved):
Awesome thank-you. I assume @flotter might want one final pass too.

flotter (approved):

Issue #414, opened 2024-04-28
Cached state will not exist for persisted check after a reboot


flotter:
```
[unrecovered-panic] runtime.fatalpanic() .snap/go/10584/src/runtime/panic.go:1188 (hits goroutine(1):1 total:1) (PC: 0x43df84)
Warning: debugging optimized function
	runtime.curg._panic.arg: interface {}(string) "interface conversion: interface {} is nil, not *plan.Check"
(dlv) bt
 0  0x000000000043df84 in runtime.fatalpanic
    at .snap/go/10584/src/runtime/panic.go:1188
 1  0x000000000043d759 in runtime.gopanic
    at .snap/go/10584/src/runtime/panic.go:1017
 2  0x0000000000ac2d7c in github.com/canonical/pebble/internals/cli.Run.func1
    at .root/go/pkg/mod/github.com/canonical/pebble@v1.10.1-0.20240424033813-4e93e7b13965/internals/cli/cli.go:298
 3  0x000000000043d610 in runtime.gopanic
    at .snap/go/10584/src/runtime/panic.go:920
 4  0x000000000040c685 in runtime.panicdottypeE
    at .snap/go/10584/src/runtime/iface.go:263
 5  0x0000000000994ce5 in github.com/canonical/pebble/internals/overlord/checkstate.(*CheckManager).PlanChanged
    at .root/go/pkg/mod/github.com/canonical/pebble@v1.10.1-0.20240424033813-4e93e7b13965/internals/overlord/checkstate/manager.go:125
 6  0x00000000009e980f in github.com/canonical/pebble/internals/overlord/checkstate.(*CheckManager).PlanChanged-fm
    at <autogenerated>:1
 7  0x00000000009ce9fc in github.com/canonical/pebble/internals/overlord/planstate.(*PlanManager).planChanged
    at .root/go/pkg/mod/github.com/canonical/pebble@v1.10.1-0.20240424033813-4e93e7b13965/internals/overlord/planstate/manager.go:87
 8  0x00000000009ce690 in github.com/canonical/pebble/internals/overlord/planstate.(*PlanManager).Load
    at .root/go/pkg/mod/github.com/canonical/pebble@v1.10.1-0.20240424033813-4e93e7b13965/internals/overlord/planstate/manager.go:67
 9  0x00000000009e5949 in github.com/canonical/pebble/internals/overlord.New
[...]

PR #415, opened 2024-04-29
fix(checkstate): ignore and abort carryover changes

flotter:
Addresses https://github.com/canonical/pebble/issues/414.

Running checks have changes that get persisted by the state-engine. This means that following a reboot, changes and tasks that are not complete (not ready) will be resumed. In the case of some managers, such as checkstate, carryover changes and tasks are an unwanted side effect.

Currently the plan manager will perform the first plan load very early during startup (before the state-engine is ready, and before StartUp hooks have been called). The result is that a PlanChanged propagation will take place at plan load, and force checkstate to inspect the running checks prematurely.

Checkstate discovers a running change from a previous boot context, and tries to load its data from cached state, which does not exist.

Gracefully ignore changes for which cached state does not exist, and use this to identify changes that should be aborted on the first ensure pass.

Test:

```
<reboot>
59   Hold    today at 23:29 SAST  today at 23:37 SAST  Recover exec check "internet-online"
60   Error   today at 23:37 SAST  today at 23:37 SAST  Perform exec check "internet-online"
:
66   Doing   today at 23:37 SAST  -                    Recover exec check "internet-online"
<reboot>
:
66   Hold    today at 23:37 SAST  today at 23:41 SAST  Recover exec check "internet-online"
:
69   Error   today at 23:41 SAST  today at 23:42 SAST  Perform exec check "internet-online"
:
73   Doing   today at 23:42 SAST  -                    Recover exec check "internet-online"
```
As the test demonstrates, following a reboot, the change (66) is aborted, and as a result the status is now ```Hold```. A new change is created which starts as a ```Perform```, and then fails (as expected) after a while and changes to a ```Recover```.

### DIFF
--- a/internals/overlord/checkstate/manager.go
+++ b/internals/overlord/checkstate/manager.go
@@ -122,7 +122,14 @@ func (m *CheckManager) PlanChanged(newPlan *plan.Plan) {
 			} else {
 				configKey = recoverConfigKey{change.ID()}
 			}
-			oldConfig := m.state.Cached(configKey).(*plan.Check) // panic if key not present (always should be)
+			v := m.state.Cached(configKey)
+			if v == nil {
+				// Pebble restarted, and this change is a carryover.
+				change.Abort()
+				shouldEnsure = true
+				continue
+			}
+			oldConfig := v.(*plan.Check)
 			existingChecks[oldConfig.Name] = true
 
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
I'll work on a more holistic fix, but that might not be right away, so I'm happy for this temporary fix to go in in the meantime. Thanks!

PR #416, opened 2024-05-02
feat(personality): Allow state filename customization

thp-canonical:
Make it possible to customize the state file name, which is part of the personality (application name, pebble dir, ...).

### DIFF
--- a/cmd/personality.go
+++ b/cmd/personality.go
@@ -22,3 +22,6 @@ var DisplayName string = "Pebble"
 
 // DefaultDir is the Pebble directory used if $PEBBLE is not set.
 var DefaultDir string = "/var/lib/pebble/default"
+
+// StateFile is the file name of the state file in pebble dir.
+var StateFile string = ".pebble.state"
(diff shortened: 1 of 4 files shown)


cmd/personality.go
```diff
@@ -22,3 +22,6 @@ var DisplayName string = "Pebble"
 
 // DefaultDir is the Pebble directory used if $PEBBLE is not set.
 var DefaultDir string = "/var/lib/pebble/default"
+
+// StateFile is the file name of the state file in pebble dir.
+var StateFile string = ".pebble.state"
```
benhoyt, line 27: I guess having a separate `StateFile` variable is better than using `"."+ProgramName+".state"`?

benhoyt (commented):
Seems reasonable to me. Just one question.

cmd/personality.go
thp-canonical: Yes, it'd allow us to be more flexible with the file name selection.

benhoyt (approved):

PR #417, opened 2024-05-23
docs: starter pack

IronCore864:
Introducing Canonical starter pack into pebble docs.

Docs themselves are not updated yet, only starter pack templates and necessary hacks to make it work.

Commits explained:

- 1st commit is everything from the starterpack
- the remaining commits are necessary config and hacks to make it work.

See the preview [here](https://canonical-pebble--417.com.readthedocs.build/en/417/).

Open to discuss:

1. Shall we remove unnecessary files from the starter pack, or leave them as they are? Examples: `doc-cheat-sheet-myst.md`, `doc-cheat-sheet.rst`, `index.rst`, etc.
2. Shall we remove the Makefile and come up with a short readme about pulling the latest starter pack templates and building docs locally? Pulling is simply cloning and copy-pasting, building is running some python scripts, pip install, then sphinx build. We benefit not much from the Makefile and it adds another thing in this repo.


.wokeignore
```diff
@@ -0,0 +1 @@
+docs/.wokeignore
```
IronCore864, line 1: This change is as suggested by the starter pack, a symlink to the `docs/.wokeignore`.

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,16 @@
+name: Main Documentation Checks
```
IronCore864, line 1: Copied from the starter pack.

AnneCYH:
Hi @IronCore864 ,

For your 2 open items.

#1 Shall we remove unnecessary files from the starter pack, or leave them as they are? Examples: doc-cheat-sheet-myst.md, doc-cheat-sheet.rst, index.rst, etc.

The cheat sheets can be removed since we usually refer users to the main Style guides. However index.rst is needed for the documentation landing page.

#2 Shall we remove the Makefile and come up with a short readme about pulling the latest starter pack templates and building docs locally? 

I would advise against doing this. We use the starter pack for consistency across Canonical documentation sets. While the Makefile does indeed consist of a set of scripts/commands etc, we want to achieve a consistent user experience by running the same `make run` command.

Hope that makes sense.

Thanks.

dimaqq:
> Hi @IronCore864 ,
> 
> For your 2 open items.
> 
> #1 Shall we remove unnecessary files from the starter pack, or leave them as they are? Examples: doc-cheat-sheet-myst.md, doc-cheat-sheet.rst, index.rst, etc.
> 
> The cheat sheets can be removed since we usually refer users to the main Style guides. However index.rst is needed for the documentation landing page.
> 
> #2 Shall we remove the Makefile and come up with a short readme about pulling the latest starter pack templates and building docs locally?
> 
> I would advise against doing this. We use the starter pack for consistency across Canonical documentation sets. While the Makefile does indeed consist of a set of scripts/commands etc, we want to achieve a consistent user experience by running the same `make run` command.
> 
> Hope that makes sense.
> 
> Thanks.

My 2c: I'd rather not see a Makefile in a Python repo that uses Python tools like Sphinx. It's just a bit alien in the Python ecosystem.

Instead we've got entrypoints in `pyproject.toml` or even shell incantations documents in a readme file.

I imagine that folks who maintain goland and c++ repos would have a similar, their-language-specific preference.

That being said, having a Makefile in the template repo is as a form of documenting the commands to run and a quick start that's meant to be adapted or translated to native tooling later on makes perfect sense.

docs/.sphinx/build_requirements.py
```diff
@@ -0,0 +1,127 @@
+import sys
```
dimaqq, line 1: I'm not sure if this file is meant to be checked in.

Let's go though entire `.sphinx` and `_templates`... maybe these are automatically pulled in at build time, in which case they should not be checked in.

docs/Makefile.sp
```diff
@@ -0,0 +1,96 @@
...
+	@echo "Run 'make help' to see supported targets."
+
+# Shouldn't assume that venv is available on Ubuntu by default; discussion here:
+# https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847
+$(SPHINXDIR)/requirements.txt:
+	python3 $(SPHINXDIR)/build_requirements.py
+	python3 -c "import venv" || sudo apt install python3-venv
+
+# If requirements are updated, venv should be rebuilt and timestamped.
+$(VENVDIR): $(SPHINXDIR)/requirements.txt
```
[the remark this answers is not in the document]
IronCore864: After sync in the standup, we decide not to use Makefile but to use tox.

docs/.sphinx/build_requirements.py
IronCore864: This file is used to regenerate the updated `requirements.txt` file. If we don't have it in our repo, we can also clone the starter pack repo, run the file there, and copy the generated `requirements.txt` to our repo, it is slightly a bit more manual work.

Since we have this file in the operator repo, for consistency, let's keep it here?

docs/tox.ini
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+[tox]
+work_dir = {toxinidir}/../.tox
+requires =
+    tox>=4
```
benhoyt, line 6: Could we add a line with `envlist = docs` here so that just typing `tox` builds the docs?

docs/.sphinx/pa11y.json
```diff
@@ -0,0 +1,9 @@
+{
```
benhoyt, line 1: Do we still need this file?

docs/.sphinx/requirements.txt
```diff
@@ -0,0 +1,21 @@
+# DO NOT MODIFY THIS FILE DIRECTLY!
+#
+# This file is generated automatically.
+# Add custom requirements to the custom_required_modules
+# array in the custom_conf.py file and run:
+# make clean && make install
```
benhoyt, line 6: There's no longer any Makefile -- can we update this comment?

benhoyt (commented):
Looks good to me. Just a couple of minor comments.

docs/.wordlist.txt
```diff
@@ -0,0 +1,49 @@
...
+reStructuredText
+RTD
+subdirectories
+subfolders
+subtree
+Ubuntu
+UI
+UUID
+VM
+YAML
```
dimaqq, line 49: We'll need our stuff here... I wonder what's the best way to address this, make a jira sub-task or a github issue?

docs/conf.py
```diff
@@ -0,0 +1,212 @@
...
+sys.path.append('.sphinx/')
+from build_requirements import *
+
+# Configuration file for the Sphinx documentation builder.
+# You should not do any modifications to this file. Put your custom
+# configuration into the custom_conf.py file.
+# If you need to change this file, contribute the changes upstream.
+#
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
```
dimaqq, line 20: From experience with other projects, most of the sphinx config can be left at defaults. This file, in the current state, is probably too long or too verbose.

It's a bit of work though to go through every setting to validate if it's needed and what it really does.

docs/custom_conf.py
```diff
@@ -0,0 +1,216 @@
+import datetime
+
+# Custom configuration for the Sphinx documentation builder.
+# All configuration specific to your project should be done in this file.
```
dimaqq, line 4: I'm not sure why we need both conf and custom_conf...

Is the idea to keep `conf.py` a copy of upstream and sync their changes?

dimaqq (commented):
Most of my comments are about things to do in a separate PR.

Please find a place to note the remaining work somewhere.

docs/.sphinx/pa11y.json
IronCore864: Probably not, I checked, it's only used in the CI which we removed. I will remove this file and make sure the build succeeds.

docs/conf.py
IronCore864: I think this file is customized by the starter pack and that's why they keep it in the pack. The starter pack probably won't work if we remove the whole thing and use default config only.

docs/custom_conf.py
IronCore864: Per my understanding, I think the purpose is that you clone the starter pack, copy everything except the custom_conf, and only change one file that is custom_conf. It's probably useful when you need to pull in the latest changes from the starter pack (but based on my personal experiences, it probably doesn't work because the starter pack changes quickly).

docs/.sphinx/build_requirements.py
IronCore864: For now, I will keep this file and resolve this comment, since it seems @AnneCYH agreed to do so.

docs/tox.ini
```diff
@@ -0,0 +1,20 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+[tox]
+work_dir = {toxinidir}/../.tox
```
tonyandrewmeyer, line 4: Is there any reason to have the `.tox` directory at the top level, rather than be in `docs/.tox`?

docs/tox.ini
```diff
@@ -0,0 +1,20 @@
...
+[testenv]
+basepython = python3
+setenv =
+  PY_COLORS=1
+
+[testenv:docs]
+description = Build the Sphinx docs
+deps = pip-tools
+commands_pre =
+    pip-sync {toxinidir}/.sphinx/requirements.txt
```
tonyandrewmeyer, line 18: What do you think about using `uv` here? It's still relatively new, but seems pretty stable, and we're using `ruff` elsewhere.

A cold build with `pip-tools` takes 23.4s on my machine, and a warm build 1.5s (although it fails in the sphinx-build stage). A cold build with `uv pip` takes 2.5s, and a warm one 0.3s (again failing).

docs/custom_conf.py
tonyandrewmeyer: I think we should stick to the starter pack pattern. They really ought to be ensuring that `conf.py` can be replaced by just dropping in the new pack files without breaking `custom_conf.py`. It also helps make it clear what's "starter pack" and what's "pebble".

tonyandrewmeyer (commented):
When I try to build (with `tox` in the `docs` directory) I get:

```
docs: commands[0]> sphinx-build -W --keep-going . _build/html
  File "/home/tameyer/scratch/tiexenpebble/.tox/docs/bin/sphinx-build", line 5, in <module>
    from sphinx.cmd.build import main
  File "/home/tameyer/scratch/tiexenpebble/.tox/docs/lib/python3.12/site-packages/sphinx/cmd/build.py", line 17, in <module>
    from docutils.utils import SystemMessage
ModuleNotFoundError: No module named 'docutils'
docs: exit 1 (0.04 seconds) /home/tameyer/scratch/tiexenpebble/docs> sphinx-build -W --keep-going . _build/html pid=2850477
  docs: FAIL code 1 (0.07=setup[0.02]+cmd[0.01,0.04] seconds)
  evaluation failed :( (0.11 seconds)
```

docs/tox.ini
benhoyt: The `uv` tool shows a lot of promise, but I would prefer we didn't experiment with Python packaging here, and just used the tried-and-true for now. We can always reevaluate later.

docs/tox.ini
IronCore864: If `.tox` itself is in `docs`, sphinx will find the READMEs that come from those packages installed in the tox environment and throw a bunch of warnings.

IronCore864: I moved it back to `docs/` and used `exclude_patterns` to avoid building the `.tox` folder.

benhoyt (approved):

IronCore864:
> When I try to build (with `tox` in the `docs` directory) I get:
> 
> ```
> docs: commands[0]> sphinx-build -W --keep-going . _build/html
> Traceback (most recent call last):
>   File "/home/tameyer/scratch/tiexenpebble/.tox/docs/bin/sphinx-build", line 5, in <module>
>     from sphinx.cmd.build import main
>   File "/home/tameyer/scratch/tiexenpebble/.tox/docs/lib/python3.12/site-packages/sphinx/cmd/build.py", line 17, in <module>
>     from docutils.utils import SystemMessage
> ModuleNotFoundError: No module named 'docutils'
> docs: exit 1 (0.04 seconds) /home/tameyer/scratch/tiexenpebble/docs> sphinx-build -W --keep-going . _build/html pid=2850477
>   docs: FAIL code 1 (0.07=setup[0.02]+cmd[0.01,0.04] seconds)
>   evaluation failed :( (0.11 seconds)
> ```

Fixed in the latest commits. Also `.tox` is moved inside `docs/`. Please take a final look.

tonyandrewmeyer (approved):

PR #418, opened 2024-05-27
feat: unified osutil.Mkdir API with options

### LINKED ISSUE #372: pebble `make-dirs` permission is masked by pebble daemon's umask

IronCore864:
Implement a unified `osutil.Mkdir` API, refactor existing `osutil.MkdirChown` and `osutil.MkdirAllChown`.

Closes #372.

## Background

[There is an issue (372)](https://github.com/canonical/pebble/issues/372) where umask settings would affect the permissions of the newly created directories.

There was originally a PR to implement the feature (see [here](https://github.com/canonical/pebble/pull/405)), but the code became even more complicated. As per our discussion (see [here](https://github.com/canonical/pebble/pull/405#issuecomment-2065462403)), we decide to refactor Mkdir and implement the required feature.

The original PR was closed, using this one instead. This PR only refactors the existing code, the support for chmod will come in a separate PR.

## Spec

See [here](https://docs.google.com/document/d/13dn5qseCL3uNER6YEnfvQ137wiXR0OVUVaGYN9X1ZdE/edit).

## API

```golang
func Mkdir(path string, perm os.FileMode, options *MkdirOptions) error { ... }
```

## Options

```golang
type MkdirOptions struct {
    // If false (default), a missing parent raises an error.
    // If true, any missing parents of this path are created as needed.
    MakeParents bool

    // If false (default), an error is raised if the target directory already exists. In case MakeParents is true but ExistOK is false, an error won't be raised if the parent directory already exists but the target directory doesn't.
    // If true, an error won't be raised unless the given path already exists in the file system and isn't a directory (same behaviour as the POSIX mkdir -p command).
    ExistOK bool

    // If false (default), no explicit chmod is performed. In this case, the permission of the created directories will be affected by umask settings.
    // If true, perform an explicit chmod on any directories created.
    Chmod bool

    // If false (default), no explicit chmod is performed. In this case, the permission of the created directories will be affected by umask settings.
    // If true, perform an explicit chmod on any directories created.
    Chown bool // NOTE: additional bool is needed because 0 is a valid value for UserID and GroupID (root)

    UserID sys.UserID

    GroupID sys.GroupID
}
```

## Manual Tests

| **Tests**                                                | **Before**                                                                                                      | **After**                                                                      |
|----------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------|
| mkdir ~/normal                                           | 755, same user/group as pebble, exist not ok                                                                      | same behaviour                                                                 |
| mkdir -p ~/nested/folder; mkdir -p ~/nested1              | 755, parent/child both the same user/group as pebble, exist ok                                                   | same behaviour                                                                 |
| mkdir ~/normal-chown --user test --group test            | 755, user/group changed exist not ok, error: rename /root/normal-chown.mkdir-new /root/normal-chown: file exists | same behaviour, better error msg: error: mkdir /root/normal-chown: file exists |
| mkdir -p ~/nested1/folder-chown --user test --group test | 755, parent/child user/group both changed, exist OK                                                               | same behaviour                                                                 |


internals/daemon/api_files.go
```diff
@@ -577,8 +590,7 @@ func makeDir(dir makeDirsItem) error {
 var (
 	atomicWriteChown = osutil.AtomicWriteChown
 	normalizeUidGid  = osutil.NormalizeUidGid
-	mkdirChown       = osutil.MkdirChown
-	mkdirAllChown    = osutil.MkdirAllChown
+	mkdir            = osutil.Mkdir
```
IronCore864, line 593: This is kept so that it's easier to test call counts, see `api_files_test.go`.

internals/daemon/api_files.go
```diff
@@ -577,8 +613,7 @@ func makeDir(dir makeDirsItem) error {
...
```
IronCore864, line 616: This is kept so that it's easier to test call counts, see api_files_test.go.

internals/osutil/mkdir_test.go
```diff
@@ -0,0 +1,288 @@
...
+			UserID:  sys.UserID(uid),
+			GroupID: sys.GroupID(gid),
+		},
+	)
+	c.Assert(err, check.IsNil)
+	c.Assert(osutil.IsDir(tmpDir+"/foo"), check.Equals, true)
+
+	info, err := os.Stat(tmpDir + "/foo")
+	c.Assert(err, check.IsNil)
+	if stat, ok := info.Sys().(*syscall.Stat_t); ok {
```
benhoyt, line 229: We should probably assert that `ok` is true, rather than only run the uid/gid assertions if it is true.

internals/osutil/mkdir_test.go
```diff
@@ -0,0 +1,288 @@
...
+		c.Assert(int(stat.Uid), check.Equals, uid)
+		c.Assert(int(stat.Gid), check.Equals, gid)
+	}
+	c.Assert(info.Mode().Perm(), check.Equals, os.FileMode(0o777))
+}
+
+// See .github/workflows/tests.yml for how to run this test as root.
+func (mkdacSuite) TestMakeParentsChownAndChown(c *check.C) {
```
benhoyt, line 237: This function tests a superset of the test above -- can we just keep this one?

benhoyt (changes requested):
This is a good start, thanks.

internals/daemon/api_files.go
```diff
@@ -501,17 +501,53 @@ func writeFile(item writeFilesItem, source io.Reader) error {
...
+			&osutil.MkdirOptions{
+				MakeParents: true,
+				ExistOK:     true,
+				Chmod:       true,
+			},
+		)
 	}
 }
 
 func mkdirUserGroup(path string, perm os.FileMode, uid, gid *int) error {
```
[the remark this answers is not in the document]
IronCore864: I think we should keep `mkdirAllUserGroup` because it's used twice. If we remove it, we have to copy-paste the same content to two functions.

`mkdirUserGroup` is only used once （in `makeDir` of the same file), we could remove it, but it's used in a function which is already over 20 lines of code, and has multiple if-elses. Removing it would increase the length of that function significantly and increase the cyclomatic complexity as well.

I prefer to keep them both.

internals/osutil/mkdir.go
```diff
@@ -0,0 +1,164 @@
...
+	UserID sys.UserID
+
+	GroupID sys.GroupID
+}
+
+// Mkdir creates directories; depending on MkdirOptions.MakeParents, it is like os.Mkdir
+// or os.MkdirAll. You can set the option MkdirOptions.Chmod to perform an explicit
+// chmod on directories it creates so that the permissions won't be affected by umask
+// settings. You can also set the option MkdirOptions.Chmod (and together with UserID,
+// GroupId) to perform an explicit chown on newly created directories.
```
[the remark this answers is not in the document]
IronCore864: I didn't understand: What do you mean by supporting `nil` explicitly? In case `options == nil`, what shall we do here?

internals/osutil/mkdir.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	// If path doesn't exist, create it.
+	return mkdirAll(path, perm, options)
+}
+
+// create directories recursively
+func mkdirAll(path string, perm os.FileMode, options *MkdirOptions) error {
+	// if path exists
+	if s, err := os.Stat(path); err == nil {
```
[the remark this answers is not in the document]
IronCore864: No, because `mkdirAll` calls itself recursively.

Mkdir only checks if the entire path exists, but `mkdirAll` recursively checks its parent, grand parent, etc.

internals/osutil/mkdir.go
```diff
@@ -0,0 +1,164 @@
...
+	// If path doesn't exist, and MakeParents isn't specified in options,
+	// create a single directory.
+	return mkdir(path, perm, options)
+}
+
+// Create a single directory and perform chmod/chown operations according to options.
+func mkdir(path string, perm os.FileMode, options *MkdirOptions) error {
+	cand := path + ".mkdir-new"
+
+	if err := os.Mkdir(cand, perm); err != nil && !os.IsExist(err) {
```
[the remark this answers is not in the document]
IronCore864: I kept this logic from the original code. Now that I think about it, maybe we should keep it: in rare cases, what if `cand := path + ".mkdir-new"` exists?

internals/daemon/api_files.go
```diff
@@ -501,17 +501,31 @@ func writeFile(item writeFilesItem, source io.Reader) error {
...
+			ExistOK:     true,
+		})
 	}
 }
 
 func mkdirUserGroup(path string, perm os.FileMode, uid, gid *int) error {
 	if uid != nil && gid != nil {
-		return mkdirChown(path, perm, sys.UserID(*uid), sys.GroupID(*gid))
+		return mkdir(path, perm, &osutil.MkdirOptions{
+			ExistOK: true,
```
benhoyt, line 522: Just to double-check: does the old `osutil.MkdirChown` have "exist-ok" semantics? If so, shouldn't we use `ExistOK` on line 528 in the `else` clause as well?

internals/daemon/api_files_test.go
```diff
@@ -446,15 +446,17 @@ func (s *filesSuite) TestMakeDirsMultiple(c *C) {
...
+		perm    os.FileMode
+		options osutil.MkdirOptions
+	}
+	var mkdirCalls []args
+	mkdir = func(path string, perm os.FileMode, options *osutil.MkdirOptions) error {
+		if options == nil {
+			options = &osutil.MkdirOptions{}
+		}
+		mkdirCalls = append(mkdirCalls, args{path, perm, *options})
+		return os.MkdirAll(path, perm)
```
benhoyt, line 459: I realize it is what was done before, but why do we actually create the directory when we're mocking this? Can't we just test the mock calls, or does the test require that the directory actually be created to work?

internals/daemon/api_files_test.go
```diff
@@ -470,28 +472,19 @@ func (s *filesSuite) TestMakeDirsUserGroupMocked(c *C) {
...
-	c.Assert(mkdirChownCalls, HasLen, 3)
-	c.Check(mkdirChownCalls[0], Equals, args{tmpDir + "/normal", 0o755, osutil.NoChown, osutil.NoChown})
-	c.Check(mkdirChownCalls[1], Equals, args{tmpDir + "/uid-gid", 0o755, 12, 34})
-	c.Check(mkdirChownCalls[2], Equals, args{tmpDir + "/user-group", 0o755, 56, 78})
-
-	c.Assert(mkdirAllChownCalls, HasLen, 2)
-	c.Check(mkdirAllChownCalls[0], Equals, args{tmpDir + "/nested1/normal", 0o755, osutil.NoChown, osutil.NoChown})
-	c.Check(mkdirAllChownCalls[1], Equals, args{tmpDir + "/nested2/user-group", 0o755, 56, 78})
+	c.Assert(mkdirCalls, HasLen, 5)
+	c.Check(mkdirCalls[0], Equals, args{tmpDir + "/normal", 0o755, osutil.MkdirOptions{}})
```
benhoyt, line 483: It seems odd that we have `ExistOK: true` on all the others, but not this one. Why is it different?

internals/osutil/mkdir.go
```diff
@@ -0,0 +1,149 @@
...
+	// If false (the default), return an error if the target directory already exists.
+	// If true, don't return an error if the target directory already exists (unless
+	// it's not a directory).
+	ExistOK bool
+
+	// If false (the default), no explicit chown is performed.
+	// If true, perform an explicit chown on any directories created, using the UserID
+	// and GroupID provided.
+	Chown bool
+
```
benhoyt, line 48: Let's remove the blank lines between `Chown` and `UserID` and `GroupID` so they're clearly grouped together.

benhoyt (commented):
Just a few further comments.

internals/daemon/api_files.go
IronCore864: I don't think there is, see [here](https://github.com/canonical/pebble/blob/master/internals/osutil/mkdirallchown.go#L68). Only the original `osutil.MkdirAllChown` has this logic [here](https://github.com/canonical/pebble/blob/master/internals/osutil/mkdirallchown.go#L45).

internals/daemon/api_files_test.go
IronCore864: Without making the directory, it would fail. This is because the test checks the result [here](https://github.com/canonical/pebble/blob/master/internals/daemon/api_files_test.go#L526).

internals/daemon/api_files_test.go
IronCore864: Added it back for consistency.

internals/daemon/api_files.go
benhoyt: But in that case it looks like the original mkdirChown doesn't have exist-ok semantics at all, so neither of these should, right?

IronCore864: Yes (updated the previous comment).

internals/daemon/api_files_test.go
IronCore864: Resolved in the other comment.

IronCore864:
Manual test results:

| **Tests**                                                | **Before**                                                                                                      | **After**                                                                      |
|----------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------|
| mkdir ~/normal                                           | 755, same user/group as pebble, exist not ok                                                                      | same behaviour                                                                 |
| mkdir -p ~/nested/folder; mkdir -p ~/nested1              | 755, parent/child both the same user/group as pebble, exist ok                                                   | same behaviour                                                                 |
| mkdir ~/normal-chown --user test --group test            | 755, user/group changed exist not ok, error: rename /root/normal-chown.mkdir-new /root/normal-chown: file exists | same behaviour, better error msg: error: mkdir /root/normal-chown: file exists |
| mkdir -p ~/nested1/folder-chown --user test --group test | 755, parent/child user/group both changed, exist OK                                                               | same behaviour                                                                 |

benhoyt (approved):
Looking good now, and thanks for the manual tests + table.

PR #419, opened 2024-05-28
docs: splitting readme into pebble docs with the diataxis method

IronCore864:
Splitting the README documentation into smaller chunks into the docs folder, using the Diataxis method.

[Click here to see the preview.](https://canonical-pebble--419.com.readthedocs.build/en/419/).

Major changes:

- custom config (site title, links, etc.)
- created a grid layout on the home page, thanks to @AnneCYH
- Splitting the README into smaller chunks following the Diataxis method. No content change except one small item: Some paragraphs use linebreaks while others don't, for consistency, all paragraphs are converted to the same style without linebreaks.
 
TODO: The current tutorial (`docs/tutorial/getting-started.md`) is too long, maybe we need to split it into smaller chunks, some fit into the "how-to" category while others may fit into the "explanations" more.


docs/how-to-guides/how-to-config-layers.md
```diff
@@ -0,0 +1,76 @@
+# How to Config Layers
```
benhoyt, line 1: Let's use the full word, "configure". Also, it looks like our docs generally don't use title case, see for example https://juju.is/docs/sdk/how-to and https://documentation.ubuntu.com/lxd/en/latest/howto/, so I suggest we try to be consistent with that.

```suggestion
# How to configure layers
```

docs/how-to-guides/how-to-config-layers.md
```diff
@@ -0,0 +1,76 @@
...
+            - srv3
+
+    srv3:
+        override: replace
+        command: cmd
+```
+
+The `override` field (which is required) defines whether this entry _overrides_ the previous service of the same name (if any), or merges with it. See the [full layer specification](../reference/layer-specification) for more details.
+
+---
```
benhoyt, line 43: I do like how this looks, but I don't think we should go adding these between every section. If anything, that should be done in the styling/CSS (and probably at the Canonical / starer pack level if we're going to do it).

docs/how-to-guides/how-to-config-layers.md
benhoyt: Also, can we shorten the URL, Currently it's:

```
/en/419/how-to-guides/how-to-config-layers/
```

Ideally we could do either of the following, to avoid the duplication:

```
/en/419/howto/configure-layers/
/en/419/how-to-configure-layers/
/en/419/configure-layers/  # looks like Juju mostly does it this way
```

docs/index.md
```diff
@@ -1,12 +1,52 @@
...
+```{grid-item-card} [Reference](reference/index)
+**Technical information** - specifications, APIs, and architecture.
+```
+````
+
+---
+
+## Project and Community
+
+Pebble is free software and released under [AGPL-3.0-only](https://www.gnu.org/licenses/agpl-3.0.en.html).
```
benhoyt, line 35: I don't think it's AGPL, just GPL 3, right? https://github.com/canonical/pebble/blob/master/COPYING

docs/tutorial/getting-started.md
```diff
@@ -0,0 +1,487 @@
+# Getting Started
```
benhoyt, line 1: This isn't exactly a tutorial, it's more of a how-to / usage guide. Though I'd be interested to get Anne's take here.

benhoyt (commented):
Thanks for this. I think this is a good first step!

I would be interested in getting Anne's take on a number of these, particularly the "tutorial".

I think we probably need another paragraph or two of introductory content on the home page, though we can leave that for another PR.

docs/how-to-guides/how-to-config-layers.md
IronCore864: I have done some research on this:

1. Juju docs isn't using the starter pack, but from discourse, that's why the behaviour is a bit different, doesn't have the folder name as part of the URL.
2. It seems other Canonical documentation that uses the starter pack, for example, LXD (using it as an example because LXD is also using markdown rather than rst like us), does have the folder name as part of the URL, but it's shorter.
3. In rare cases, a page can be put into a folder with no folder name, but that's only because the file actually isn't in the folder, but put in by some other directives. For example, [this page from lxd](https://documentation.ubuntu.com/lxd/en/latest/getting_started/) shows up in the folder with no folder name in the URL (this is an exception), only because the file is directly in the root of ["docs"](https://github.com/canonical/lxd/tree/main/doc). I intentionally made the second link to the folder rather than the file to show that this might be a bad practice, and hard to manage since we might end up with many files at the root level not knowing which belongs to which.

So, my conclusion is like the first example you've given `/en/419/howto/configure-layers/`, that is, with a short folder name, and a shorter article name. Maybe we should try to put as less articles as possible at the root level of docs to make it manageable (the current LXD repo is hard to maintain in my opinion.)

docs/tutorial/getting-started.md
IronCore864: I agree, some part are more of a tutorial, like installing it and getting started with some basic layer configurations, while other tasks are like how-tos.

Shall we merge it first then split it up, or do the splitting-up in this PR? @benhoyt

benhoyt: I'm in favour of incremental improvements, that is, merging this first pass, then improving in subsequent PRs. But let's see what Anne says.

AnneCYH: Hey @benhoyt , @IronCore864 , 

I agree that the updates can be incremental. But please give me some time to go through the changes in this PR before merging, for now.

Thanks!

benhoyt: Yep, no problem -- take your time!

docs/custom_conf.py
```diff
@@ -77,26 +77,26 @@
...
+    'github_url': 'https://github.com/canonical/pebble',
 
     # Change to the branch for this version of the documentation
-    'github_version': 'main',
+    'github_version': 'master',
 
     # Change to the folder that contains the documentation
     # (usually "/" or "/docs/")
-    'github_folder': '/',
+    'github_folder': '/docs',
```
AnneCYH, line 99: The trailing '/' is needed here as the links to GitHub (to edit the source files) append the filename immediately after the github_url + github_folder value.
image

docs/custom_conf.py
AnneCYH, line 1: Just a reminder in case this was omitted by accident.
https://github.com/canonical/pebble/blob/6aef6583c3c69157d34185ec113f11b5572c038e/docs/custom_conf.py#L107

Setting this to "both" enables the "Next" and "Previous" buttons on each page.
image

If it's not needed for Pebble it's fine of course.

Thanks.

docs/howto/configure-layers.md
```diff
@@ -0,0 +1,74 @@
+# Configure layers
+
+Below is an example of the current configuration format. For full details of all fields, see the [complete layer specification](../reference/layer-specification).
```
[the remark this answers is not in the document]
IronCore864: This is a relative link inside the same docs website, instead of an external URL, so I'm keeping it for simplicity. For external URLs I'll use things like:

- [Canonical website](https://canonical.com/)
- https:/<span></span>/canonical.com/

I hope this is OK.

docs/how-to/index.md
```diff
@@ -0,0 +1,22 @@
+# How-To Guides
```
AnneCYH, line 1: This should be "How-to guides".

AnneCYH:
Hi @IronCore864 , this is a lot clearer, thanks!
I can see the content starting to modularise, which will make it easy to keep each section clean, avoid duplicates, and easy to cross-reference.

I do think some sections should be merged or added.
1. View, start, stop, update, and restart services don't need to be separate. But this can be done in the next PR since you don't want to move content around.
2. Get logs + use log forwarding are all related to a single entity for Pebble, "logs", and don't need to be separate either.
3. Other content that may be 'missing', e.g. related to Files. When i run `pebble help`, the commands are categorised, but the current README doesn't necessarily address "Info" (e.g. managing your Pebble installation in terms of version, updating to a newer version), or "Files". But this can be done in a separate PR.

Overall I think it's a good start. I only left 1 minor comment about the title for the How-to index page; but other than that I think it's good to go for the first draft.

Thanks!

benhoyt (approved):
This looks like a good start to me! Lots of further work to come, but this is a good base.

PR #427, opened 2024-06-10
fix: work around issue with user.Lookup not returning UnknownUserError

benhoyt:
Per https://github.com/golang/go/issues/67912, under some circumstances (on my machine) user.Lookup isn't returning UnknownUserError when it should. So before this fix I get the following test failures:

```
$ go test ./internals/osutil
...
FAIL: user_test.go:44: userSuite.TestRealUser

user_test.go:71:
    c.Assert(err, check.IsNil)
... value *errors.errorString = &errors.errorString{s:"user: lookup username guy: no such file or directory"} ("user: lookup username guy: no such file or directory")
...
```

And:

```
$ go test ./internals/daemon
...
----------------------------------------------------------------------
FAIL: api_files_test.go:1128: filesSuite.TestWriteErrors

api_files_test.go:1195:
    checkFileResult(c, r.Result[4], pathUserNotFound, "generic-file-error", ".*unknown user.*")
api_files_test.go:273:
    c.Check(r.Error.Message, Matches, errorMsg)
... value string = "cannot look up user and group: user: lookup username user-not-found: no such file or directory"
... regex string = ".*unknown user.*"

api_files_test.go:1196:
    checkFileResult(c, r.Result[5], pathGroupNotFound, "generic-file-error", ".*unknown group.*")
api_files_test.go:273:
    c.Check(r.Error.Message, Matches, errorMsg)
... value string = "cannot look up user and group: user: lookup groupname group-not-found: no such file or directory"
... regex string = ".*unknown group.*"
...
```

### DIFF
--- a/internals/osutil/user.go
+++ b/internals/osutil/user.go
@@ -20,6 +20,7 @@ import (
 	"os/user"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/canonical/pebble/internals/osutil/sys"
 )
@@ -28,6 +29,8 @@ var (
 	userCurrent     = user.Current
 	userLookup      = user.Lookup
 	userLookupGroup = user.LookupGroup
+
+	enoentMessage = syscall.ENOENT.Error()
 )
 
 // RealUser finds the user behind a sudo invocation when root, if applicable
@@ -60,7 +63,7 @@ func RealUser() (*user.User, error) {
 	// Workaround for https://github.com/golang/go/issues/67912, until our
 	// minimum Go version has a fix for that. In short, user.Lookup sometimes
 	// doesn't return UnknownUserError when it should.
-	if err != nil && strings.Contains(err.Error(), "no such file or directory") {
+	if err != nil && strings.Contains(err.Error(), enoentMessage) {
 		return cur, nil
 	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internals/osutil/user.go
```diff
@@ -56,6 +57,12 @@ func RealUser() (*user.User, error) {
 	if _, ok := err.(user.UnknownUserError); ok {
 		return cur, nil
 	}
+	// Workaround for https://github.com/golang/go/issues/67912, until our
+	// minimum Go version has a fix for that. In short, user.Lookup sometimes
+	// doesn't return UnknownUserError when it should.
+	if err != nil && strings.Contains(err.Error(), "no such file or directory") {
```
dimaqq, line 63: Is a “C” or an en_* locale guaranteed for pebble?

I’m assuming that go localise strerror, tbh I don’t know if it does.

hpidcock: @benhoyt can you please replace all the string literals here with `syscall.Errno(2).Error()`

benhoyt: @dimaqq Yep, Go doesn't use the C locale environment variables (thankfully!). It's a hard-coded string, see https://github.com/golang/go/blob/beaf7f3282c2548267d3c894417cc4ecacc5d575/src/syscall/syscall_unix.go#L110 and https://github.com/golang/go/blob/beaf7f3282c2548267d3c894417cc4ecacc5d575/src/syscall/zerrors_linux_amd64.go#L1361

@hpidcock Good call. One better: `syscall.ENOENT.Error()`. :-)

hpidcock (approved):
LGTM - thanks

PR #428, opened 2024-06-13
feat(state): core of reading/writing identities in state

benhoyt:
This adds code and tests for the state core of identities ([spec OP043](https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit#heading=h.a291qhvyhsw3)).

There's a new `map[string]*Identity` in state, which is marshalled via `marshalledIdentity` to disk. The marshalling for API requests is done by the `apiIdentity` type -- these are separate to ensure that the API (default) `MarshalJSON` for `Identity` does not include secrets (in future when we have identity types that include secrets; there's only user-id for now). So that's why there's a few more types and a bit more boilerplate than is necessary right now.

The new public `State` methods are:

* `AddIdentities`: add new identities to the system
* `UpdateIdentities`: update existing identities in the system
* `ReplaceIdentities`: replace the given identities in the system; this allows adding new ones, updating existing ones, and removing identities (nil/null means remove)
* `RemoveIdentities`: remove existing identities in the system
* `Identities`: list all identities

Each of the `FooIdentities` modification functions validates that the identities are valid, and that the user-ids are unique, before applying to state.

### DIFF
--- a/internals/overlord/state/identities.go
+++ b/internals/overlord/state/identities.go
@@ -22,8 +22,6 @@ import (
 	"strings"
 )
 
-// TODO: disallow multiple users with same "local: user-id" value?
-
 // Identity holds the configuration of a single identity.
 type Identity struct {
 	Name   string
@@ -130,6 +128,10 @@ func (s *State) AddIdentities(identities map[string]*Identity) error {
 		sort.Strings(existing)
 		return fmt.Errorf("identities already exist: %s", strings.Join(existing, ", "))
 	}
+	err := verifyUniqueUserIDs(s.identities, identities)
+	if err != nil {
+		return nil
+	}
 
 	for name, identity := range identities {
 		identity.Name = name
@@ -158,6 +160,10 @@ func (s *State) UpdateIdentities(identities map[string]*Identity) error {
 		sort.Strings(missing)
 		return fmt.Errorf("identities missing: %s", strings.Join(missing, ", "))
 	}
+	err := verifyUniqueUserIDs(s.identities, identities)
+	if err != nil {
+		return nil
+	}
 
 	for name, identity := range identities {
 		identity.Name = name
@@ -179,6 +185,10 @@ func (s *State) ReplaceIdentities(identities map[string]*Identity) error {
 				return fmt.Errorf("identity %q invalid: %w", name, err)
 			}
 		}
+	}
+	err := verifyUniqueUserIDs(s.identities, identities)
+	if err != nil {
+		return nil
 	}
 
 	for name, identity := range identities {
@@ -225,4 +235,29 @@ func (s *State) Identities() map[string]*Identity {
 		result[name] = identity
 	}
 	return result
+}
+
+func verifyUniqueUserIDs(existing map[string]*Identity, new map[string]*Identity) error {
+	existingNamesByUserID := make(map[uint32]string)
+	for name, identity := range existing {
+		switch {
+		case identity.Local != nil:
+			existingNamesByUserID[identity.Local.UserID] = name
+		}
+	}
+	for name, identity := range new {
+		if identity == nil {
+			continue // removing identity (for ReplaceIdentities only)
+		}
+		switch {
+		case identity.Local != nil:
+			existingName, ok := existingNamesByUserID[identity.Local.UserID]
+			if ok && name != existingName {
+				return fmt.Errorf("identity %q and %q cannot both have user ID %d",
+					name, existingName, identity.Local.UserID)
+			}
+			existingNamesByUserID[identity.Local.UserID] = name
+		}
+	}
+	return nil
 }
--- a/internals/overlord/state/identities.go
+++ b/internals/overlord/state/identities.go
@@ -1,0 +1,229 @@
+// Copyright (c) 2024 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


internals/overlord/state/state.go
```diff
@@ -156,27 +158,40 @@ func (s *State) unlock() {
...
 
+// marshalledIdentity is used specifically for marshalling to the state
+// database file. Unlike apiIdentity, it should include secrets.
+type marshalledIdentity struct {
+	Access string                   `json:"access"`
+	Local  *marshalledLocalIdentity `json:"local,omitempty"`
+}
+
+type marshalledLocalIdentity struct {
+	UserID uint32 `json:"user-id"`
```
hpidcock, line 182: I wonder if this should be a string. I know on linux it can only be a unsigned int, but go abstracts user id as a string (because of windows). Probably not a problem, but at least then we can distinguish between empty/default and root (0).

benhoyt: I think you're right that we should be able to distinguish between empty and root (0). People probably should be able to specify a `superuser: {access: admin, local: {user-id: 0}}` user (at the moment I disallow that).

However, I think `*uint32` is better than than string. It'd be better if it were a number, not a string, in JSON/YAML, as it's the correct type (and less validation is required). We've also used `UserID *int` elsewhere quite a bit for this same thing, and it's worked out well (though it probably should be `*uint32` rather than `*int` for correctness on 32-bit systems).

As far as strings on Windows go, the whole ucred system depends on this being an integer (and Linux), so I don't think we need to open that can of worms.

In any case, I'll change this to a `*uint32` and a named identity with user-id 0.

benhoyt: Also, this only applies to `apiLocalIdentity.UserID` (used by `Identity.UnmarshalJSON`), as that's where identity JSON is unmarshalled. Oh, and `client.LocalIdentity.UserID` in the subsequent PR, as that's where the YAML version is unmarshalled.

internals/overlord/state/identities.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	*d = identity
+	return nil
+}
+
+// AddIdentities adds the given identities to the system. It's an error if any
+// of the named identities already exist.
+func (s *State) AddIdentities(identities map[string]*Identity) error {
+	s.writing()
```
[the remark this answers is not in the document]
hpidcock: Not sure if it's worth making this distinction. Writing to disk after a read, because of failed validation seems strange.

internals/overlord/state/identities.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func verifyUniqueUserIDs(identities map[string]*Identity) error {
+	userIDs := make(map[uint32]string) // maps user IDs to identity name
+	for name, identity := range identities {
+		switch {
+		case identity.Local != nil:
+			existingName, ok := userIDs[identity.Local.UserID]
+			if ok {
+				if name > existingName { // ensure error message is stable
```
hpidcock, line 278: This makes the error message stable between two identities. But if there is 3 or more identities with the same UID, then this does not work due to go map pseudo random iteration order.

internals/overlord/state/identities_test.go
```diff
@@ -0,0 +1,536 @@
...
+	c.Assert(err, ErrorMatches, `identity "bob" invalid: identity must not be nil`)
+
+	// Ensure unique user ID testing is being done (full testing done in AddIdentity).
+	err = st.UpdateIdentities(map[string]*state.Identity{
+		"bob": {
+			Access: state.AdminAccess,
+			Local:  &state.LocalIdentity{UserID: 42},
+		},
+	})
+	c.Assert(err, ErrorMatches, `identities "bob" and "mary" cannot both have user ID 42`)
```
hpidcock, line 374: Need a test with three identities with the same user ID.

hpidcock (approved):
Just a few minor issues or concerns. Otherwise looks good, thanks for breaking this up.

internals/overlord/state/identities.go
```diff
@@ -0,0 +1,288 @@
...
+	UntrustedAccess IdentityAccess = "untrusted"
+)
+
+// LocalIdentity holds identity configuration specific to the "local" type
+// (for ucrednet/UID authentication).
+type LocalIdentity struct {
+	UserID uint32
+}
+
+// validate checks that identity d is valid, returning an error if not.
```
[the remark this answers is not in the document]
benhoyt: Good call that referring to `d` was odd. I didn't like starting with "returns an error" as what the function primarily does is check/validate, but I've tweaked this.

internals/overlord/state/identities.go
```diff
@@ -0,0 +1,288 @@
...
+			d.Access, AdminAccess, ReadAccess, UntrustedAccess)
+	}
+
+	switch {
+	case d.Local != nil:
+		return nil
+	default:
+		return errors.New(`identity must have at least one type ("local")`)
+	}
+}
```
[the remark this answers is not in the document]
benhoyt: Yep, exactly. And in `UnmarshalJSON` I do a little bit more validation here for this identity type, to ensure that user-id is present. But by this point it has to be present (0 is valid).

internals/overlord/state/identities.go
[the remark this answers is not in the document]
benhoyt: Ah yeah, that's fair enough - updated.

internals/overlord/state/identities.go
benhoyt: Good call, updated the function and error message.

internals/overlord/state/identities_test.go
benhoyt: Added, thanks.

PR #429, opened 2024-06-13
feat(daemon): identities API endpoints

benhoyt:
This adds the API endpoints for identities ([spec OP043](https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit)), specifically:

* `GET /v1/identities`: list all identities (read/user access)
* `POST /v1/identities`: modify identities with action `add`, `update`, `replace`, `remove` (admin access)


internals/daemon/api_identities.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/canonical/pebble/internals/overlord/state"
+)
+
+func v1GetIdentities(c *Command, r *http.Request, _ *UserState) Response {
+	st := c.d.overlord.State()
+	st.Lock()
+	defer st.Unlock()
+
+	identities := st.Identities()
```
hpidcock, line 29: probably should not just dump back all the identities straight from state for `UserAccess{}`. Maybe only `AdminAccess{}` can read every item, whereas, `UserAccess{}` can read their own, and the rest are partial?

internals/overlord/state/identities.go
```diff
@@ -0,0 +1,229 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// TODO: disallow multiple users with same "local: user-id" value?
```
hpidcock, line 25: Yes, we cannot introduce ambiguity.

internals/overlord/state/identities.go
```diff
@@ -0,0 +1,229 @@
...
+	switch d.Access {
+	case AdminAccess, ReadAccess, UntrustedAccess:
+	default:
+		return fmt.Errorf("invalid access %q, must be %q, %q, or %q",
+			d.Access, AdminAccess, ReadAccess, UntrustedAccess)
+	}
+
+	switch {
+	case d.Local != nil:
+		if d.Local.UserID == 0 {
```
hpidcock, line 67: I understand this for root, makes sense (go int default value and root uid sharing meaning). For the pebble user (the user pebble daemon is running as), I think it is a bit different and we have to be softer there, as someone may change the uid the daemon is running as.

So if there is an identity that matches the pebble user uid, I think we just ignore the value of Access as it is always "AdminAccess".

hpidcock: My comment about making UserID a string might make sense, so we can technically distinguish between empty and 0, without resorting to uint32 ptr.

internals/daemon/api_identities.go
benhoyt: I was thinking UserAccess should be able to see the list too, kind of like as any user I can `cat /etc/passwd`. But no user, including admin, can read back any secrets. This keeps it simpler and (I think) less error-prone.

Note that that's why I've defined `Identity.MarshalJSON` (which uses the `apiIdentity` helper type), to ensure that when we add identity types with secrets, we won't add them there, and so they can't get exposed via the API at all. In contrast, when the state file is written to JSON on disk, `marshalledIdentity.MarshalJSON` takes care of that -- and this will include sensitive data or secrets.

Does that seem reasonable?

internals/overlord/state/identities.go
benhoyt: Fixed, via `verifyUniqueUserIDs`.

internals/overlord/state/identities.go
benhoyt: I've updated this to use `*uint32` to allow us to distinguish between 0 (now valid) and unset (invalid) at the JSON/YAML level.

I believe with how I've implemented `AdminAccess.CheckAccess` and `UserAccess.CheckAccess` (see access.go), they fall back to looking at ucred directly if there wasn't an identity found. Does that seem reasonable?

internals/daemon/api_identities.go
```diff
@@ -0,0 +1,72 @@
...
+				return BadRequest(`identity value for %q must be null when removing`, name)
+			}
+			identities[name] = struct{}{}
+		}
+		err = st.RemoveIdentities(identities)
+	default:
+		return BadRequest(`invalid action %q, must be "add", "update", "replace", or "remove"`, payload.Action)
+	}
+	if err != nil {
+		return BadRequest("%v", err)
```
hpidcock, line 68: In this case I'm fine with the blanket BadRequest, as the errors returned by all the state methods are input validation/precondition errors.

hpidcock (approved):
LGTM - thankyou.

internals/overlord/state/identities.go
hpidcock: Yep. Thank-you.

internals/daemon/api_identities.go
benhoyt: Yep, agreed. If and when that changes, we can add a custom error type and check for that in this function to distinguish. But Do What You Need Now.

PR #430, opened 2024-06-13
feat(client): Go client for identities

benhoyt:
This adds the Go client types and methods for identities (part of [spec OP043](https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit)), specifically:

* `Identity` and `IdentityAccess` types, with access level constants
* `Identities`
* `AddIdentities`
* `UpdateIdentities`
* `ReplaceIdentities`
* `RemoveIdentities`


client/identities.go
```diff
@@ -0,0 +1,121 @@
...
+		Path:   "/v1/identities",
+	})
+	if err != nil {
+		return nil, err
+	}
+	var identities map[string]*Identity
+	err = resp.DecodeResult(&identities)
+	if err != nil {
+		return nil, err
+	}
```
hpidcock, line 72: We should probably be kind to ourselves and check the server did not return a nil entry in the map.

client/notices_test.go
```diff
@@ -42,6 +42,8 @@ func (cs *clientSuite) TestNotice(c *C) {
 	}}`
 	notice, err := cs.cli.Notice("123")
 	c.Assert(err, IsNil)
+	c.Assert(cs.req.Method, Equals, "GET")
```
hpidcock, line 45: Is this just adding consistency to our tests?

hpidcock (approved):
LGTM with a bit of defensive programming with the return value from the server.

client/identities.go
benhoyt: Fair enough -- added this and a test.

client/notices_test.go
benhoyt: Yep, that's right -- and specifically that the client is sending the right HTTP method.

PR #431, opened 2024-06-13
feat(cli): identities CLI commands

benhoyt:
This adds the CLI commands for identities (part of [spec OP043](https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit)), specifically:

* `identities`: list all identities
* `identity`: show details for a single named identity
* `add-identities`: add one or more new identities
* `update-identities`: update one or more existing identities (`--replace` for replace operation)
* `remove-identities`: remove one or more identities

In addition, this updates `pebble run` to support an `--identities <file>` option to allow seeding initial identities when running the daemon.


internals/cli/cmd_help.go
```diff
@@ -214,6 +214,10 @@ var HelpCategories = []HelpCategory{{
 	Label:       "Notices",
 	Description: "manage notices and warnings",
 	Commands:    []string{"warnings", "okay", "notices", "notice", "notify"},
+}, {
+	Label:       "Identities",
+	Description: "manage user identities",
+	Commands:    []string{"identities", "identity", "add-identities", "update-identities", "remove-identities"},
```
benhoyt, line 220: We were considering having a special case where when you typed `pebble help` it'd show just:

```
$ pebble help
...
Commands can be classified as follows:

         Run: run
...
     Notices: warnings, okay, notices, notice, notify
  Identities: identities --help
```

And then they would type `pebble identities --help` for more info.

However, that's an annoying special case, and I think it's simplest just to follow the existing pattern. Now that `replace-identities` isn't its own command (it's `update-identities --replace`) it's a bit shorter anyway. The list is not ridiculously long:

```
$ pebble help
...
Commands can be classified as follows:

         Run: run
...
     Notices: warnings, okay, notices, notice, notify
  Identities: identities, identity, add-identities, update-identities, remove-identities
```

internals/cli/cmd_add-identities.go
```diff
@@ -0,0 +1,103 @@
...
+func (cmd *cmdAddIdentities) Execute(args []string) error {
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	identities, err := readIdentities(cmd.From)
+	if err != nil {
+		return err
+	}
+	err = cmd.client.AddIdentities(identities)
```
hpidcock, line 73: The only protection from nil identities in the map read from the file is in the state validate method for the entity. We probably should do some client side validation to return nice errors.

internals/cli/cmd_add-identities_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/pebble/internals/cli"
+)
+
+func (s *PebbleSuite) TestAddIdentitiesSingle(c *C) {
```
hpidcock, line 30: Add a test to check that adding nil identities is verboten.

internals/cli/cmd_help.go
hpidcock: I would argue that we specially had come to a consensus on this. Please push forward on adding support for this, it can be in another PR, but I think it should happen as this was one of the nice things to come out of the spec process.

internals/cli/cmd_update-identities_test.go
```diff
@@ -0,0 +1,94 @@
...
+	c.Assert(err, IsNil)
+
+	rest, err := cli.ParserForTest().ParseArgs([]string{"update-identities", "--from", path})
+	c.Assert(err, IsNil)
+	c.Check(rest, HasLen, 0)
+	c.Check(s.Stdout(), Equals, "Updated 1 identity.\n")
+	c.Check(s.Stderr(), Equals, "")
+}
+
+func (s *PebbleSuite) TestReplaceIdentities(c *C) {
```
hpidcock, line 62: No test for remove identity?

hpidcock (approved):
a few issues, nothing problematic. LGTM. Thank-you.

internals/cli/cmd_help.go
benhoyt: Fair enough -- updated.

internals/cli/cmd_add-identities.go
benhoyt: I somewhat disagree that all the validation should be on the client side (we need to do it on the server side anyway). However, I've added a bit of error handling and made things a bit more consistent on the server side.

internals/cli/cmd_update-identities_test.go
benhoyt: Added

PR #434, opened 2024-06-17
feat(daemon): use identities to auth API requests

benhoyt:
This adds the code to actually use the new state-based identities for API authorization and authentication (part of [spec OP043](https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit)).

The PR updates the daemon to actually use UserState (which now has access-level and UID feilds). The daemon now fetches the request's named identity if one exists, otherwise it populates UserState with default user based on the ucred UID. So UserState is always set -- this makes the access checkers simpler and the notices API code simpler.

If there is a named identity (a non-nil `UserState`), that identity is used, even if the "default UID-based auth" would let them in. So named identities trump the default ("uid==0 || uid==daemon_uid" means admin; any uid means read). For example, if you have a user `bob: {access: untrusted, local: {user-id: 42}}` and are making a request with UID 42, it'll only work for untrusted/open endpoints (`/v1/health` and `/v1/system-info`), not the usual of any read API.


hpidcock:
Let's do this one last.

internals/daemon/daemon.go
```diff
@@ -193,6 +200,19 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// not good: https://github.com/canonical/pebble/pull/369
+	var user *UserState
+	if _, isOpen := access.(OpenAccess); !isOpen {
+		user, err = userFromRequest(c.d.state, r, ucred)
+		if err != nil {
+			Forbidden("forbidden").ServeHTTP(w, r)
+			return
+		}
+	}
+
```
benhoyt, line 215: Open questions to discuss with @hpidcock:

After the `access.CheckAccess()` call and before `rspf()`, do we want to do something like this, to make it easier for (the few) handlers that behave differently based on access level, like notices?

This way the `UserState` value passed to the handler would be present for named identity-based users, as well as "default" users (like uid 0/root and uid==daemon_uid).

```
  if user == nil && ucred != nil {
    if ucred.Uid == 0 || ucred.Uid == uint32(os.Getuid()) {
      user = &UserState{Access: state.AdminAccess}
    } else {
      user = &UserState{Access: state.ReadAccess}
    }
  }

  rsp := rspf(c, r, user)
```

Could also add `Ucred *ucred` or `Uid *uint32` to `UserState` to make it easier for notices handlers to get the uid, instead of re-parsing it from the request with `ucrednetGet`.

benhoyt: TODO: That reminds me, if we don't do the above, I also need to update the notices code to handle named identity-based users (that is `UserState` not nil).

IronCore864:
I roughly viewed the code (in those smaller PRs in Ben's forked repo, thus didn't leave any comments), and tested this feature. In summary, the feature works as expected, but I find a few places that could be improved, see details below. Here are the tests I ran:

## `pebble identities`

- permissions: requires only user access, which is expected.
- output: text/json/yaml all worked correctly.

## `pebble add-identities`

- permission: needs admin access, user access doesn't work, as expected.
- user already exists: got an error "identities already exist", as expected.
- invalid input YAML: got an error: "cannot unmarshal identities: yaml: unmarshal errors", as expected.
- missing "access": got an error 'invalid access "", must be "admin", "read", or "untrusted"'. As expected, but _`invalid access ""` seems not clear, it would be nicer if it said something like 'invalid value "" for key `access` in the YAML file' or something to that effect; simply "invalid access" sounds like I don't have the permission to do so._
- wrong value for "access": similar error as above, but better: 'invalid "access": cannot decode request body: invalid access "test", must be "admin", "read", or "untrusted"'. The behaviour is as expected, and the error message is better.
- missing type ("local"): got error 'identity must have at least one type ("local")', as expected.
- missing user-id: got error 'local identity must specify user-id', as expected.
- existing uid: got error 'error: identities "alice" and "bob" cannot both have user ID 1000', as expected.

## `pebble identity`

- non-existing identity: `pebble identity foo` got error 'cannot find identity "foo"', as expected.
- more than one parameter: `/pebble identity bob foo` got error "too many arguments for command", as expected.

## `pebble update-identities`

- invalid input: got error "cannot unmarshal identities", as expected.
- repeated updates: OK.
- existing uid: got error 'identities "alice" and "bob" cannot both have user ID 1000', as expected.

## `pebble remove-identities`

- remove using exactly the same YAML that the user was created: got error 'identity "alice" must be null when removing'. _This error message is not very helpful, what does it mean by "must be null"?_ Delete the user from the OS? Does the user have something that must be deleted first? Also, it's slight annoying because users _have to edit the YAML file they used to create the user_ in the first place. It would be nice if we can run add/udpdate and remove _with the same input file_, just like kubectl create/apply and delete.

## State File

Inspected, content as expected.

## Other Small Things

- It would be nice if the help command can show a sample YAML input, telling users the structure and the mandatory fields.
- `pebble remove-identities` requires YAML input, but given user names and UIDs are both unique, it would be nice if we can do `pebble remove-identities --uid 1000` or simply `pebble remove-identities bob`.
- There is no `pebble replace-identities` CLI command, but it seems this is available by looking at the code, at least it's possible to do so via API.
- It seems the "untrusted" identity isn't useful, all operations require either read access or admin access. Also, from a user's perspective, why would I add a user as untrusted in Pebble so that they can do nothing? If the user don't add it as untrusted, they can't do anything anyway.

benhoyt:
@IronCore864 Thanks a lot for your manual testing and feedback. Responding to your suggestions:

> missing "access" ... As expected, but invalid access "" seems not clear

I've added a more specific error for the "" case, and clarified the other cases slightly -- thanks.

> remove using exactly the same YAML that the user was created: got error 'identity "alice" must be null when removing'. This error message is not very helpful, what does it mean by "must be null"? Delete the user from the OS? Does the user have something that must be deleted first?

I've clarified the error message slightly.

> Also, it's slight annoying because users have to edit the YAML file they used to create the user in the first place. It would be nice if we can run add/udpdate and remove with the same input file, just like kubectl create/apply and delete.

This is by design (per spec review discussion), so it's not easy to accidentally remove a user. I can see that it's a bit annoying, but I think annoying is better than an accidental removal. Also, it's consistent with the `pebble update-identities --replace` remove operation (must be null). If this design turns out to be a mistake, we can always make `pebble remove-identities` more permissive (but we can't go the other way).

> It would be nice if the help command can show a sample YAML input, telling users the structure and the mandatory fields.

I've added some basic examples to the help text. I'm not going to specify full details of the YAML structure (that will be in our Pebble docs!) but hopefully these basic examples help. They also help clarify the `remove-identities` null thing.

> pebble remove-identities requires YAML input, but given user names and UIDs are both unique, it would be nice if we can do pebble remove-identities --uid 1000 or simply pebble remove-identities bob.

In the spec review, we discussed a future extension (for both remove-identities and add/update-identities) that would allow you to remove identities (or add/update a single identity for add-identities/update-identities) using command line parameters. However, we felt that the YAML version was enough for a v1, so we're leaving that for future work. I think if we did this we'd allow `pebble remove-identities bob alice`.

> There is no pebble replace-identities CLI command, but it seems this is available by looking at the code, at least it's possible to do so via API.

That's `pebble update-identities --replace ...`. In spec review we decided to make it a single CLI command with the `--replace` option.

> It seems the "untrusted" identity isn't useful, all operations require either read access or admin access. Also, from a user's perspective, why would I add a user as untrusted in Pebble so that they can do nothing? If the user don't add it as untrusted, they can't do anything anyway.

You're right: it's not really useful. However, we allow "untrusted" as one of the access constants for completeness. Also, I think it may be useful for other (future) authentication types, though I haven't fully thought that one through.

hpidcock:
@benhoyt I think I want to review this one in isolation after the rest have been merged.

nishitm:
the changes look good from the security point of view

internals/daemon/daemon.go
benhoyt: Per discussion with Harry, I've made this change now, even when calling `CheckAccess()`, so that all the messy ucred stuff is done in daemon, and CheckAccess and handlers don't have to worry about it. I've also included UID in UserState so notices handlers (for example) don't have to worry about it.

internals/daemon/api_notices.go
```diff
@@ -267,19 +249,22 @@ func v1GetNotice(c *Command, r *http.Request, _ *UserState) Response {
...
 
-func noticeViewableByUser(notice *state.Notice, requestUID, daemonUID uint32) bool {
+func noticeViewableByUser(notice *state.Notice, user *UserState) bool {
 	userID, isSet := notice.UserID()
 	if !isSet {
+		// Notice has no UID, so it's viewable by any user.
 		return true
 	}
-	if isAdmin(requestUID, daemonUID) {
+	if user.Access == state.AdminAccess {
```
hpidcock, line 264: This is now a bit clunky. An admin user who (by some future means) doesn't have a UID will not be able to read all notices.

internals/daemon/api_notices.go
```diff
@@ -267,19 +249,22 @@ func v1GetNotice(c *Command, r *http.Request, _ *UserState) Response {
...
 	}
 	return SyncResponse(notice)
 }
 
-func noticeViewableByUser(notice *state.Notice, requestUID, daemonUID uint32) bool {
+func noticeViewableByUser(notice *state.Notice, user *UserState) bool {
 	userID, isSet := notice.UserID()
 	if !isSet {
+		// Notice has no UID, so it's viewable by any user.
 		return true
```
hpidcock, line 262: But not readable to user's without a UID.

internals/daemon/api_notices.go
```diff
@@ -253,12 +237,10 @@ func v1PostNotices(c *Command, r *http.Request, _ *UserState) Response {
 	return SyncResponse(addedNotice{ID: noticeId})
 }
 
-func v1GetNotice(c *Command, r *http.Request, _ *UserState) Response {
-	requestUID, err := uidFromRequest(r)
-	if err != nil {
+func v1GetNotice(c *Command, r *http.Request, user *UserState) Response {
+	if user == nil || user.UID == nil {
```
hpidcock, line 241: I understand that this isn't a requirement of this change, but can you please add a TODO somewhere that mentions notices needs work when we add other authentication mechanisms?

internals/daemon/daemon.go
```diff
@@ -194,7 +202,31 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if rspe := access.CheckAccess(c.d, r, ucred, user); rspe != nil {
+	// Optimisation: avoid calling userFromRequest, which acquires the state
+	// lock, in case we don't need to (when endpoint is OpenAccess). This
+	// avoids holding the state lock for /v1/health in particular, which is
+	// not good: https://github.com/canonical/pebble/pull/369
+	var user *UserState
+	if _, isOpen := access.(OpenAccess); !isOpen {
```
hpidcock, line 210: I'm OK with this special case because it doesn't actually bypass the access checker. As long as we still actually call the access checker, we can have this optimisation.

hpidcock (commented):
Still reviewing, just some early comments.

internals/daemon/api_notices.go
benhoyt: Good call -- added a TODO comment in https://github.com/canonical/pebble/pull/434/commits/bc5d32d6acd44977d39c4f92b5290866707103e4

internals/daemon/api_notices.go
benhoyt: True. This is per the design of notices being very much UID-based (which was arguably short-sighted?). I've added a TODO comment in https://github.com/canonical/pebble/pull/434/commits/bc5d32d6acd44977d39c4f92b5290866707103e4

internals/daemon/api_notices.go
benhoyt: I've tweaded the comment, and added a TODO comment in https://github.com/canonical/pebble/pull/434/commits/bc5d32d6acd44977d39c4f92b5290866707103e4

hpidcock:
Unrelated flaky test in internals/daemon
```
----------------------------------------------------------------------
FAIL: api_exec_test.go:91: execSuite.TestStderr

api_exec_test.go:92:
    stdout, stderr, waitErr := s.exec(c, "", &client.ExecOptions{
        Command: []string{"/bin/sh", "-c", "echo some stderr! >&2"},
    })
api_exec_test.go:324:
    c.Assert(err, IsNil)
... value *fmt.wrapError = &fmt.wrapError{msg:"cannot connect to \"control\" websocket: read unix @->/tmp/check-2962214203/23.pebble.socket: i/o timeout", err:(*net.OpError)(0xc000326370)} ("cannot connect to \"control\" websocket: read unix @->/tmp/check-2962214203/23.pebble.socket: i/o timeout")

2024-07-11T05:20:52.589Z [test] Started daemon.
2024-07-11T05:20:52.602Z [test] POST /v1/exec 6.517775ms 202
2024-07-11T05:20:52.607Z [test] GET /v1/tasks/1/websocket/control 3.803267ms 200
2024-07-11T05:20:52.607Z [test] GET /v1/tasks/1/websocket/stdio 101.523µs 200
2024-07-11T05:20:52.608Z [test] GET /v1/tasks/1/websocket/stderr 137.962µs 200
2024-07-11T05:20:52.626Z [test] GET /v1/changes/1/wait 17.87651ms 200

----------------------------------------------------------------------
```

hpidcock (approved):
@benhoyt, your work on this is much appreciated.

I've performed the QA that I think was suitable for this (integrating with my Juju PR) + testing the login directly + a series of stress tests on the unit tests (with and without -race).

I am happy to approve this now. I only had one thought around possible security violations that this change could bring and that is to do with revocation of privileges.
If you have a long running request, with an identity that had access to do an action (such as exec, logs, push, pull, etc.), then that action will go uninterrupted, despite changes to the access level for that identity.

We can address that in a future PR.

internals/daemon/daemon.go
```diff
@@ -142,8 +142,22 @@ const (
...
+
+	st.Lock()
+	identity := st.IdentityFromInputs(&ucred.Uid)
+	st.Unlock()
+
+	if identity == nil {
+		// No identity that matches these inputs (for now, just UID).
+		return nil, nil
+	}
+	return &UserState{Access: identity.Access, UID: &ucred.Uid}, nil
```
thp-canonical, line 160: It looks like (I didn't look too closesly) that `IdentityFromInputs()` may return a pointer to an object inside the identities structure (which is part of `st`, and protected by the mutex?), and we later access `identity.Access` in line 160 without the lock held.

Better to keep the lock until it's no longer possible to access `identity` (which currently is at the end of the function). Checking for `identity == nil` and creating a `UserState` struct shouldn't take up too much time, so the lock shouldn't be kept held unnecessarily long this way.

```suggestion
	st.Lock()
	identity := st.IdentityFromInputs(&ucred.Uid)
	defer st.Unlock()

	if identity == nil {
		// No identity that matches these inputs (for now, just UID).
		return nil, nil
	}
	return &UserState{Access: identity.Access, UID: &ucred.Uid}, nil
```

internals/daemon/daemon.go
thp-canonical: This should be fine. In case there's another implementation in the future of `access` that is not of concrete type `OpenAccess`, but behaves the same, we will only "lose" the optimization (meaning we unnecessarily call `userFromRequest` and populate `user` properly), but it will still do the right thing.

internals/daemon/daemon.go
```diff
@@ -194,7 +202,31 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// If we don't have a named-identity user, use ucred UID to see if we have a default.
+	if user == nil && ucred != nil {
+		if ucred.Uid == 0 || ucred.Uid == uint32(os.Getuid()) {
+			// Admin if UID is 0 (root) or the UID the daemon is running as.
+			user = &UserState{Access: state.AdminAccess, UID: &ucred.Uid}
+		} else {
+			// Regular read access if any other local UID.
+			user = &UserState{Access: state.ReadAccess, UID: &ucred.Uid}
+		}
+	}
```
thp-canonical, line 227: Shouldn't this be in `userFromRequest()`? Seems like this is for backwards compatibility? Or I guess so that we can make the optimization above that `userFromRequest()` is called only conditionally?

internals/daemon/daemon_test.go
```diff
@@ -510,10 +531,55 @@ func (s *daemonSuite) TestSuperAccess(c *C) {
...
-		s.testAccessChecker(c, tests, remoteAddr)
+	s.testAccessChecker(c, tests, "pid=100;uid=0;socket=;")
+	s.testAccessChecker(c, tests, "pid=100;uid=1;socket=;") // adminuser
+	s.testAccessChecker(c, tests, fmt.Sprintf("pid=100;uid=%d;socket=;", os.Getuid()))
+}
+
+func (s *daemonSuite) TestDefaultUcredUsers(c *C) {
+	d := s.newDaemon(c)
+
+	var user *UserState
```
thp-canonical, line 542: Maybe call this `userSeenInCommand` (or something like that) to make it more obvious (if I got it right) that this is the user that was "seen" when the command was called?

internals/daemon/daemon_test.go
```diff
@@ -510,10 +531,55 @@ func (s *daemonSuite) TestSuperAccess(c *C) {
...
+	// Read access for UID not 0 and not daemon UID.
+	user = nil
+	req = &http.Request{Method: "GET", RemoteAddr: fmt.Sprintf("pid=100;uid=%d;socket=;", os.Getuid()+1)}
+	rec = httptest.NewRecorder()
+	cmd.ServeHTTP(rec, req)
+	c.Check(rec.Code, Equals, http.StatusOK)
+	c.Assert(user, NotNil)
+	c.Check(user.Access, Equals, state.ReadAccess)
+	c.Assert(user.UID, NotNil)
+	c.Check(*user.UID, Equals, uint32(os.Getuid()+1))
```
thp-canonical, line 582: Haven't checked if these 3 code blocks have the same structure, but if they do, maybe turn it into data + code (slice of 3 testcase data items, and then iterate over these and do the checks in 1 code block).

Maybe not worth it for 3 cases, though.

internals/overlord/state/identities.go
```diff
@@ -264,6 +264,22 @@ func (s *State) Identities() map[string]*Identity {
 	return result
 }
 
+// IdentityFromInputs returns an identity with the given inputs, or nil
+// if there is none.
+func (s *State) IdentityFromInputs(userID *uint32) *Identity {
```
thp-canonical, line 269: Maybe add to the comment that this requires the state lock to be held, also for as long as the returned `*Identity` is accessed? Or is `s.reading()` taking care of checking this?

internals/overlord/state/identities_test.go
```diff
@@ -538,3 +538,43 @@ func (s *identitiesSuite) TestIdentities(c *C) {
 	identities["changed"] = &state.Identity{}
 	c.Assert(identities2, DeepEquals, expected)
 }
+
+func (s *identitiesSuite) TestIdentityFromInputs(c *C) {
```
thp-canonical, line 542: Maybe also add a test case what happens when `st.IdentityFromInputs()` is called without `st.Lock()` held (if this is testable, not sure).

internals/daemon/access.go
```diff
@@ -16,40 +16,55 @@ package daemon
...
 
-func (ac AdminAccess) CheckAccess(d *Daemon, r *http.Request, ucred *Ucrednet, user *UserState) Response {
-	if ucred != nil && (ucred.Uid == 0 || ucred.Uid == uint32(os.Getuid())) {
+func (ac AdminAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	if user == nil {
+		return Unauthorized(accessDenied)
+	}
+	if user.Access == state.AdminAccess {
 		return nil
 	}
```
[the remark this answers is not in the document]
thp-canonical: ..and of course if we go with the "`user` can never be `nil`" as described below, this simplifies a bit, too.

internals/daemon/access.go
```diff
@@ -16,40 +16,55 @@ package daemon
...
-func (ac UserAccess) CheckAccess(d *Daemon, r *http.Request, ucred *Ucrednet, user *UserState) Response {
-	if ucred == nil {
-		return Unauthorized("access denied")
+func (ac UserAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	if user == nil {
+		return Unauthorized(accessDenied)
 	}
-	return nil
+	switch user.Access {
+	case state.ReadAccess, state.AdminAccess:
```
flotter, line 65: I see that only a ```state.ReadAccess``` exist. At the same time I can see that notices have:

```
    Path:        "/v1/notices",
    ReadAccess:  UserAccess{},
    WriteAccess: UserAccess{}, // any user is allowed to add a notice with their own uid
    GET:         v1GetNotices,
    POST:        v1PostNotices,
```

and ```UserAccess``` validation pass with only ```ReadAccess```. Should the state identify access type not simply say ```UserAccess```. It feels that the read/write control is already available at the API endpoint level ?

internals/daemon/daemon.go
flotter: I believe the trick is that identity objects never get mutated (they get replaced), so once you have a pointer to an instance, its safe to keep it.

flotter (approved):
Really nice!

benhoyt:
> If you have a long running request, with an identity that had access to do an action (such as exec, logs, push, pull, etc.), then that action will go uninterrupted, despite changes to the access level for that identity.

@hpidcock That doesn't really concern me. If you were able to start a long running request, you had access at the time.

internals/cli/cmd_notices.go
```diff
@@ -54,7 +54,7 @@ func init() {
 		Summary:     cmdNoticesSummary,
 		Description: cmdNoticesDescription,
 		ArgsHelp: merge(timeArgsHelp, map[string]string{
-			"--users":   "Show all notices with any user ID (admin only; cannot be used with --uid)",
+			"--users":   "'all' to list all notices with any user ID (admin only; cannot be used with --uid)",
```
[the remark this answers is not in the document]
benhoyt: Thanks -- I've improved the help text and the argument help for `--users` here.

internals/daemon/access.go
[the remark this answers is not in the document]
benhoyt: Yeah, I tried with the more complex conditionals but it seemed harder to get wrong with the simpler ones.

Regarding the "user can never be nil" thing, I've switched to that approach -- I think it makes sense!

internals/daemon/access_test.go
```diff
@@ -33,50 +31,43 @@ func (s *accessSuite) TestOpenAccess(c *C) {
...
+	user := &daemon.UserState{Access: state.AdminAccess}
+	c.Check(ac.CheckAccess(nil, nil, user), IsNil)
+	user = &daemon.UserState{Access: state.ReadAccess}
+	c.Check(ac.CheckAccess(nil, nil, user), IsNil)
 
-	// UserAccess allows access form normal user
-	ucred = &daemon.Ucrednet{Uid: 42, Pid: 100}
-	c.Check(ac.CheckAccess(nil, nil, ucred, nil), IsNil)
+	// But not UntrustedAccess
+	user = &daemon.UserState{Access: state.UntrustedAccess}
```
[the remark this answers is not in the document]
benhoyt: Yeah, this is a good point. I think it makes sense to always have `UserState` non-nil. I've made that change. (UntrustedAccess by default)

internals/daemon/access_test.go
```diff
@@ -33,50 +31,43 @@ func (s *accessSuite) TestOpenAccess(c *C) {
...
+	c.Check(ac.CheckAccess(nil, nil, nil), DeepEquals, errUnauthorized)
 
-	// Root is granted access
-	ucred = &daemon.Ucrednet{Uid: 0, Pid: 100}
-	c.Check(ac.CheckAccess(nil, nil, ucred, nil), IsNil)
+	// But not ReadAccess or UntrustedAccess
+	user := &daemon.UserState{Access: state.ReadAccess}
+	c.Check(ac.CheckAccess(nil, nil, user), DeepEquals, errUnauthorized)
+	user = &daemon.UserState{Access: state.UntrustedAccess}
+	c.Check(ac.CheckAccess(nil, nil, user), DeepEquals, errUnauthorized)
```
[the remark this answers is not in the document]
benhoyt: Good catch! Added.

internals/daemon/api_notices_test.go
```diff
@@ -107,9 +85,8 @@ func (s *apiSuite) testNoticesFilter(c *C, makeQuery func(after time.Time) url.V
 	query := makeQuery(after)
 	req, err := http.NewRequest("GET", "/v1/notices?"+query.Encode(), nil)
 	c.Assert(err, IsNil)
-	req.RemoteAddr = "pid=100;uid=0;socket=;"
 	noticesCmd := apiCmd("/v1/notices")
-	rsp, ok := noticesCmd.GET(noticesCmd, req, nil).(*resp)
+	rsp, ok := noticesCmd.GET(noticesCmd, req, userState(state.AdminAccess, 0)).(*resp)
```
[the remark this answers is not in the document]
benhoyt: I'm going to leave these smaller test nits as is for now. I could go either way -- I don't mind the literals as long as they're local to the test.

internals/daemon/api_notices_test.go
```diff
@@ -1007,3 +889,7 @@ func addNotice(c *C, st *state.State, userID *uint32, noticeType state.NoticeTyp
 	_, err := st.AddNotice(userID, noticeType, key, options)
 	c.Assert(err, IsNil)
 }
+
+func userState(access state.IdentityAccess, uid uint32) *UserState {
+	return &UserState{Access: access, UID: &uid}
+}
```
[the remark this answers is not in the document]
benhoyt: Yep, that's right -- just to work around the fact you can't do `&1000` in Go. Could have package-level globals, but I prefer the helper. Going to leave this as is.

internals/daemon/daemon.go
```diff
@@ -110,10 +110,10 @@ type Daemon struct {
...
 
 // UserState represents the state of an authenticated API user.
-//
-// The struct is currently empty as the behaviors haven't been implemented
-// yet.
-type UserState struct{}
+type UserState struct {
+	Access state.IdentityAccess
+	UID    *uint32
+}
```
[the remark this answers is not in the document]
benhoyt: Per above, I've changed this to always provide a non-nil UserState. Definitely a bit simpler!

internals/daemon/daemon.go
benhoyt: Yep, Fred's right here -- thanks.

internals/daemon/daemon_test.go
benhoyt: Fair enough. I've renamed to `userSeen`.

internals/daemon/daemon_test.go
benhoyt: Yeah, I almost did this, but then decided it wasn't worth it for 3x. Any more and I think it would be. I'll leave as is for now.

internals/daemon/daemon.go
benhoyt: Yep, I've moved this to `userFromRequest` -- good call.

internals/overlord/state/identities.go
benhoyt: Most of these state functions assume but don't explicitly document that you need to hold the state lock, so I think this is okay as is (and yeah, the `s.reading()` checks).

internals/overlord/state/identities_test.go
benhoyt: Going to leave this as is. Having `s.reading()` panic seems fine, and other State methods don't test for this.

internals/daemon/access.go
benhoyt: Yeah, it's a little weird that the AccessChecker is called UserAccess but the level is called "read". The latter was spec'd so I'm going to leave that one. I think I'll leave this as is for now, but in a future PR we could consider renaming UserAccess to ReadAccess.

benhoyt:
@thp-canonical Actually, after discussion with Harry, he was a bit concerned with that latest commit to always use Untrusted and a non-nil UserState. It may be good (in future) to distinguish between no auth and an untrusted identity. In addition, both of us had done manual testing on the previous iteration of the code, so any changes now are risky. We can always iterate further in future without user-facing changes.

PR #435, opened 2024-06-26
perf(daemon): avoid acquiring state lock (twice!) for every API request

### LINKED ISSUE #366: Avoid acquiring state lock in Command.ServeHTTP

benhoyt:
In daemon.Command.ServeHTTP, at the end of every request we acquire the state lock twice, once for including the maintenance (restart) info in the response, and once for including the warnings summary. We don't use either maintenance/restarts or warnings in Pebble at all, so let's just turn that off.

We could rip out those features completely, but the APIs need to stay for backwards compatibility, and it's possible we'll want them later, so let's just disable them and their tests for now.

Locking and unlocking the state for reading isn't very expensive, but if something else has locked the state for writing (for example, creating/updating a Change or Task), API requests won't be able to acquire the lock till the relatively slow operation of serialising and writing state is finished. Per #366, that can sometimes be over 1s when the system is under load.

Fixes #366

### DIFF
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -57,6 +57,16 @@ var (
 	sysGetuid       = sys.Getuid
 )
 
+const (
+	// Whether to include maintenance info (pending restarts) and warnings
+	// with every API response. These are turned off right now because
+	// restarts and warnings aren't used at all in Pebble now, and it's not
+	// worth locking state twice on every request for features we don't use.
+	// We could rip out the code for these features, but this will do for now.
+	includeMaintenance     = false
+	includeWarningsSummary = false
+)
+
 // Options holds the daemon setup required for the initialization of a new daemon.
 type Options struct {
 	// Dir is the pebble directory where all setup is found. Defaults to /var/lib/pebble/default.
@@ -203,18 +213,20 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if rsp, ok := rsp.(*resp); ok {
 		st := c.d.state
-		st.Lock()
-		_, rst := restart.Pending(st)
-		st.Unlock()
-		switch rst {
-		case restart.RestartSystem:
-			rsp.transmitMaintenance(errorKindSystemRestart, "system is restarting")
-		case restart.RestartDaemon:
-			rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is restarting")
-		case restart.RestartSocket:
-			rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is stopping to wait for socket activation")
+		if includeMaintenance {
+			st.Lock()
+			_, rst := restart.Pending(st)
+			st.Unlock()
+			switch rst {
+			case restart.RestartSystem:
+				rsp.transmitMaintenance(errorKindSystemRestart, "system is restarting")
+			case restart.RestartDaemon:
+				rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is restarting")
+			case restart.RestartSocket:
+				rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is stopping to wait for socket activation")
+			}
 		}
-		if rsp.Type != ResponseTypeError {
+		if includeWarningsSummary && rsp.Type != ResponseTypeError {
 			st.Lock()
 			count, stamp := st.WarningsSummary()
 			st.Unlock()
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -262,6 +262,9 @@ func (s *daemonSuite) TestCommandMethodDispatch(c *C) {
 }
 
 func (s *daemonSuite) TestCommandRestartingState(c *C) {
+	if !includeMaintenance {
+		c.Skip("includeMaintenance disabled")
+	}
 	d := s.newDaemon(c)
 
 	cmd := &Command{d: d, ReadAccess: OpenAccess{}}
@@ -312,6 +315,9 @@ func (s *daemonSuite) TestCommandRestartingState(c *C) {
 }
 
 func (s *daemonSuite) TestFillsWarnings(c *C) {
+	if !includeWarningsSummary {
+		c.Skip("includeWarningsSummary disabled")
+	}
 	d := s.newDaemon(c)
 
 	cmd := &Command{d: d, ReadAccess: OpenAccess{}}


hpidcock:
@thp-canonical or @flotter could you please check this out.

internals/daemon/daemon.go
```diff
@@ -57,6 +57,16 @@ var (
 	sysGetuid       = sys.Getuid
 )
 
+const (
+	// Whether to include maintenance info (pending restarts) and warnings
+	// with every API response. These are turned off right now because
+	// restarts and warnings aren't used at all in Pebble now, and it's not
+	// worth locking state twice on every request for features we don't use.
+	// We could rip out the code for these features, but this will do for now.
+	includeMaintenance     = false
```
hpidcock, line 66: My only concern here is that these are const. My affect @thp-canonical or @flotter's plans.

internals/daemon/daemon.go
```diff
@@ -203,18 +213,20 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		case restart.RestartSystem:
-			rsp.transmitMaintenance(errorKindSystemRestart, "system is restarting")
-		case restart.RestartDaemon:
-			rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is restarting")
-		case restart.RestartSocket:
-			rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is stopping to wait for socket activation")
+		if includeMaintenance {
+			st.Lock()
+			_, rst := restart.Pending(st)
+			st.Unlock()
```
thp-canonical, line 219: Since `restart.Pending()` already returns a `bool` as first return value (that's ignored here), what about tracking an `atomic.Bool` on `State`, so that checking whether `restart.Pending()` would return any of `RestartSystem`, `RestartDaemon` or `RestartSocket` would be lockless via the atomic bool (it will be false in > 99%-ish of cases?), and only when the atomic bool is actually true you take out the lock and request the reason with `restart.Pending(st)` with the lock held.

Having a quick look, it's the `restartStateKey{}` in `restart.go` that accesses the state, so tracking the access there should make it possible as a side-effect to track a boolean in `st`.

Instead of `if includeMaintenance {` above, this would become something like `if st.atomicBoolRestartPending {`.

Alternatively, because of `type RestartType int` you could also decide to store the full restart type as an `atomic.Int32` (32-bit ought to be enough to enumerate all possible restart types) and make this part here completely lock-free.

That would at least avoid adding `includeMaintenance` for the restart part.

internals/daemon/daemon.go
```diff
@@ -203,18 +213,20 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			case restart.RestartSystem:
+				rsp.transmitMaintenance(errorKindSystemRestart, "system is restarting")
+			case restart.RestartDaemon:
+				rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is restarting")
+			case restart.RestartSocket:
+				rsp.transmitMaintenance(errorKindDaemonRestart, "daemon is stopping to wait for socket activation")
+			}
 		}
-		if rsp.Type != ResponseTypeError {
+		if includeWarningsSummary && rsp.Type != ResponseTypeError {
```
thp-canonical, line 229: This one is a little bit more tricky, because `WarningsSummary()` seems to depend on the current time:

https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/overlord/state/warning.go#L281-L297

Looking at the `addWarningsToMeta()` function that gets called with the result from `WarningsSummary()`:

https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/daemon/response.go#L68-L77

We can see that if `count == 0`, `addWarningsToMeta()` does nothing (also if `r.WarningCount != 0` already).

One thing that could avoid locks here is that if we have the `rsp` object (`Response` interface, but sometimes(?) implemented as type `resp` from `response.go`?), and its `WarningCount != 0`, we don't need to take out the lock and call `WarningsSummary()` (but that only applies to cases where `rsp` already contains warnings, not sure if that will be the case often - I expect not).

The other thing could be to - like above - track an atomic Uint64 (say) of how many warnings there are in state, not taking into account expiry. This would only work nicely if warnings in state are eventually purged if they are expired (otherwise as soon as one warning is in there, it won't ever go away).

One other way to track it would be the following:

- Keep track of a count of "unexpired warnings" and a "dirty flag" for that count
- Whenever a new warning is added, set the dirty flag
- Whenever the dirty flag is set here, call `WarningsSummary()` here, which will update the unexpired warnings count and reset the dirty flag
- Whenever the dirty flag is not set here, check "unexpired warnings" and if it's zero, skip the check

This can probably be simplified into a single atomic boolean flag "st.needToSummarizeWarningsWarnings":

- Adding a warning or changing any warnings in any way sets the flag
- `WarningsSummary()` returning `count == 0` resets the flag (otherwise the flag is kept set)
- The code here checks the flag, and only takes out the lock and calls `WarningsSummary()` when the flag is set

This assumes that warnings will never be visible "only in the future" (meaning at time `t` they are not visible, but at time `t + 1` they are visible), otherwise this logic would break.

thp-canonical (commented):
Not sure if it's worth tracking the additional state as described in my comments, but for those two specific cases, it seems possible to add two atomic booleans and make sure they are correctly updated whenever the restart type or the list of warnings (or warning content) change, and then only take out the lock when it's likely that taking out the lock will be worth it (i.e. restart is pending or there are potentially warnings to be shown).

It's a possible source for bugs (forgetting to update the additional state would mask restart reasons and/or warnings), but might be less of an issue than just unconditionally not doing both things.

benhoyt:
Thanks for the comments, folks, especially your digging into details, @thp-canonical.

Regarding maintenance/restart, using a separate atomic bool or int is a good idea. I'll consider that further.

For warnings, note that they aren't used at all, and probably never should be. They are kind of broken, well, at least they assume only a single client (which isn't the case of Pebble, and probably isn't for KernOS either) -- this is because warnings are "okayed" on the server side. Whereas the new notices system does it via an "after" timestamp from the client, so you can have as many clients as you need. [Read more here.](https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/overlord/state/warning.go#L18-L33)

So for warnings, I'm going to have a go at removing them more completely. `State.Warnf` isn't called by Pebble at all (and KernOS definitely doesn't), so I think we should just remove it, and then warnings can never get added. I think we can even remove `warning-timestamp` and `warning-count` from the response struct, as they're marked `omitempty` and hence won't ever be present right now anyway. I'll need to keep the /v1/warnings API for backwards compatibility, however. I'm not 100% sure it's doable yet, but I think this is a better route for Pebble.

hpidcock:
> Thanks for the comments, folks, especially your digging into details, @thp-canonical.
> 
> Regarding maintenance/restart, using a separate atomic bool or int is a good idea. I'll consider that further.
> 
> For warnings, note that they aren't used at all, and probably never should be. They are kind of broken, well, at least they assume only a single client (which isn't the case of Pebble, and probably isn't for the other project either) -- this is because warnings are "okayed" on the server side. Whereas the new notices system does it via an "after" timestamp from the client, so you can have as many clients as you need. [Read more here.](https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/overlord/state/warning.go#L18-L33)
> 
> So for warnings, I'm going to have a go at removing them more completely. `State.Warnf` isn't called by Pebble at all (and the other project definitely doesn't), so I think we should just remove it, and then warnings can never get added. I think we can even remove `warning-timestamp` and `warning-count` from the response struct, as they're marked `omitempty` and hence won't ever be present right now anyway. I'll need to keep the /v1/warnings API for backwards compatibility, however. I'm not 100% sure it's doable yet, but I think this is a better route for Pebble.

I'm curious how hard it would be to re-implement on-top of the back of notices.

benhoyt:
> So for warnings, I'm going to have a go at removing them more completely

I'm having a go at this in https://github.com/canonical/pebble/pull/443

benhoyt:
Going to close this hacky approach in favour of https://github.com/canonical/pebble/pull/451 and https://github.com/canonical/pebble/pull/443

PR #436, opened 2024-06-26
fix(planstate): don't hold planLock while calling plan-changed listeners

benhoyt:
Some listeners do quite a bit of stuff (check manager and log manager), and they're defined in packages external to the planstate package, so it's best to avoid holding our plan lock for the duration.

See [spec OP040](https://docs.google.com/document/d/1cmZ5FWz0efglDAxGRciY5GRyD6KfWJCOe-kjWl6auig/edit#heading=h.po9vomna3y7d).

Also:

- PlanManager doesn't use state and runner, so remove them
- Rename planHandlers to changeListeners to match AddChangeListener

### DIFF
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/canonical/pebble/internals/overlord/state"
 	"github.com/canonical/pebble/internals/plan"
 )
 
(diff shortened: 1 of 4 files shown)


IronCore864 (approved):

internals/overlord/planstate/manager.go
```diff
@@ -33,23 +32,19 @@ func (e *LabelExists) Error() string {
 }
 
 type PlanManager struct {
-	state     *state.State
-	runner    *state.TaskRunner
```
benhoyt, line 37: Drive-by: PlanManager never used state and runner (maybe a copy 'n' paste issue), so remove them.

internals/overlord/planstate/manager.go
```diff
@@ -76,14 +74,16 @@ type PlanChangedFunc func(p *plan.Plan)
...
 
-func (m *PlanManager) planChanged(plan *plan.Plan) {
-	m.plan = plan
-	for _, f := range m.planHandlers {
+func (m *PlanManager) callChangeListeners(plan *plan.Plan) {
+	if plan == nil {
+		// Avoids if statement on every deferred call to this method (we
+		// shouldn't call listeners when the operation fails).
+		return
+	}
```
hpidcock, line 85: If we also had a `recover()` higher up, we'd want to recover here then re-panic rather than call the changeListeners. But I don't think we recover anywhere do we?

hpidcock (approved):
Thanks LGTM. Just the one question.

internals/overlord/planstate/manager.go
benhoyt: We do recover and re-panic in Run(). However, in case of a panic in the functions that call `callChangeListeners`, plan (newPlan) will be nil, so it won't call them due to this `if` statement.

Issue #231, opened 2023-06-02
Error starting one service causes others in task set to not start


troyanov:
Pebble gets _locked_ after starting the service which then exits immediately.

Layer specification used:
```yaml
services:
  a:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
  b:
    override: replace
    command: sh -c "exit 123"
    startup: enabled
  c:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
```

Pebble will not try to start service `c` as can be seen in debug output: 
```log
❯ PEBBLE_DEBUG=1 PEBBLE="/tmp/pebble-root" go run ./cmd/pebble run
2023-06-02T20:13:31.257Z [pebble] DEBUG socket "/tmp/pebble-root/.pebble.socket" was not activated; listening
2023-06-02T20:13:31.257Z [pebble] DEBUG Reaper started, waiting for SIGCHLD.
2023-06-02T20:13:31.257Z [pebble] DEBUG socket "/tmp/pebble-root/.pebble.socket.untrusted" was not activated; listening
2023-06-02T20:13:31.257Z [pebble] Started daemon.
2023-06-02T20:13:31.257Z [pebble] DEBUG activation done in 10ms
2023-06-02T20:13:31.266Z [pebble] DEBUG Configuring check manager (stopping 0, starting 0)
2023-06-02T20:13:31.280Z [pebble] DEBUG Running task 251 on Do: Start service "a"
2023-06-02T20:13:31.280Z [pebble] DEBUG pid=1871156;uid=1000;socket=/tmp/pebble-root/.pebble.socket; POST /v1/services 22.762402ms 202
2023-06-02T20:13:31.281Z [pebble] POST /v1/services 22.762402ms 202
2023-06-02T20:13:31.281Z [pebble] Started default services with change 79.
2023-06-02T20:13:31.295Z [pebble] Service "a" starting: sh -c "sleep 999"
2023-06-02T20:13:31.295Z [pebble] DEBUG Service "a" started with PID 1871169
2023-06-02T20:13:31.295Z [pebble] DEBUG Service "a" transitioning to state "starting"
2023-06-02T20:13:32.296Z [pebble] DEBUG Service "a" transitioning to state "running"
2023-06-02T20:13:32.310Z [pebble] DEBUG Running task 252 on Do: Start service "b"
2023-06-02T20:13:32.316Z [pebble] Service "b" starting: sh -c "exit 123"
[...]

PR #437, opened 2024-06-26
feat(plan): use lanes to start services with dependencies

IronCore864:
When starting services, put services that depend on each other (before/after/requires) in the same lane. If there is no dependency, put it in a stand-alone lane. Tasks only wait for tasks defined in the same lane.

Fixes https://github.com/canonical/pebble/issues/231.

## Logic

The logic is as follows:

- get services from plan so that before/after/requires can be fetched
- if a service doesn't depend on other services or isn't dependent by other services (it doesn't have before/after/requires, and it doesn't show up in other services' before/after/requires), put it in a single lane
- if a service has dependencies (before/after/requires), all these dependencies will be put in the same lane as the service itself
- tasks only wait for tasks of the same lane. Since the input is already sorted, the latter tasks of the same lane only need to wait for the previous task in the same lane.

## Tests

I mainly ran three types of test. The first is the one mentioned in https://github.com/canonical/pebble/issues/231: 

```yaml
services:
  a:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
  b:
    override: replace
    command: sh -c "exit 123"
    startup: enabled
  c:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
```

With this feature, service C won't be in "hold" status, but rather, started. Logs:

```bash
2024-07-23T17:24:56.172Z [pebble] Started daemon.
2024-07-23T17:24:56.182Z [pebble] POST /v1/services 2.781792ms 202
2024-07-23T17:24:56.185Z [pebble] Service "a" starting: sh -c "sleep 999"
2024-07-23T17:24:56.185Z [pebble] Service "b" starting: sh -c "exit 123"
2024-07-23T17:24:56.186Z [pebble] Service "c" starting: sh -c "sleep 999"
2024-07-23T17:24:56.189Z [pebble] Change 1 task (Start service "b") failed: cannot start service: exited quickly with code 123
2024-07-23T17:24:57.190Z [pebble] GET /v1/changes/1/wait 1.005492792s 200
2024-07-23T17:24:57.190Z [pebble] Started default services with change 1.
2024-07-23T17:25:01.404Z [pebble] GET /v1/services?names= 120.625µs 200
^C2024-07-23T17:25:21.617Z [pebble] Exiting on interrupt signal.
2024-07-23T17:25:21.621Z [pebble] Stopping all running services.
2024-07-23T17:25:21.628Z [pebble] Service "c" stopped
2024-07-23T17:25:21.628Z [pebble] Service "a" stopped
```

As a second test, if C depends on B, C is still in "hold" because they are in the same lane:

```yaml
services:
  a:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
  b:
    override: replace
    command: sh -c "exit 123"
    startup: enabled
  c:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
    requires:
      - b
```

Logs:

```bash
2024-07-23T17:25:37.799Z [pebble] Started daemon.
2024-07-23T17:25:37.810Z [pebble] POST /v1/services 2.859792ms 202
2024-07-23T17:25:37.813Z [pebble] Service "a" starting: sh -c "sleep 999"
2024-07-23T17:25:37.813Z [pebble] Service "b" starting: sh -c "exit 123"
2024-07-23T17:25:37.816Z [pebble] Change 1 task (Start service "b") failed: cannot start service: exited quickly with code 123
2024-07-23T17:25:38.819Z [pebble] GET /v1/changes/1/wait 1.005558292s 200
2024-07-23T17:25:38.820Z [pebble] Started default services with change 1.
2024-07-23T17:25:47.218Z [pebble] GET /v1/services?names= 103.208µs 200
^C2024-07-23T17:26:19.741Z [pebble] Exiting on interrupt signal.
2024-07-23T17:26:19.745Z [pebble] Stopping all running services.
2024-07-23T17:26:19.752Z [pebble] Service "a" stopped
```

As a third test, I tested the "requires + before" combination:

```yaml
services:
  a:
    override: replace
    command: sh -c "exit 123"
    startup: enabled
    requires:
      - b
    before:
      - b
  b:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
```

Logs:

```bash
2024-07-23T17:26:23.145Z [pebble] Started daemon.
2024-07-23T17:26:23.154Z [pebble] POST /v1/services 2.873792ms 202
2024-07-23T17:26:23.157Z [pebble] Service "a" starting: sh -c "exit 123"
2024-07-23T17:26:23.161Z [pebble] Change 1 task (Start service "a") failed: cannot start service: exited quickly with code 123
2024-07-23T17:26:23.167Z [pebble] GET /v1/changes/1/wait 12.322667ms 200
2024-07-23T17:26:23.167Z [pebble] Started default services with change 1.
2024-07-23T17:26:26.642Z [pebble] GET /v1/services?names= 583.833µs 200
^C2024-07-23T17:26:52.384Z [pebble] Exiting on interrupt signal.
```

### DIFF
--- a/internals/daemon/api_services.go
+++ b/internals/daemon/api_services.go
@@ -104,37 +104,38 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 	defer st.Unlock()
 
 	var taskSet *state.TaskSet
+	var lanes [][]string
 	var services []string
 	switch payload.Action {
 	case "start", "autostart":
-		services, err = servmgr.StartOrder(payload.Services)
+		lanes, err = servmgr.StartOrder(payload.Services)
 		if err != nil {
 			break
 		}
-		taskSet, err = servstate.Start(st, services, servmgr)
+		taskSet, err = servstate.Start(st, lanes)
 	case "stop":
-		services, err = servmgr.StopOrder(payload.Services)
+		lanes, err = servmgr.StopOrder(payload.Services)
 		if err != nil {
 			break
 		}
-		taskSet, err = servstate.Stop(st, services)
+		taskSet, err = servstate.Stop(st, lanes)
 	case "restart":
-		services, err = servmgr.StopOrder(payload.Services)
+		lanes, err = servmgr.StopOrder(payload.Services)
 		if err != nil {
 			break
 		}
-		services = intersectOrdered(payload.Services, services)
+		lanes = intersectOrdered(payload.Services, lanes)
 		var stopTasks *state.TaskSet
-		stopTasks, err = servstate.Stop(st, services)
+		stopTasks, err = servstate.Stop(st, lanes)
 		if err != nil {
 			break
 		}
-		services, err = servmgr.StartOrder(payload.Services)
+		lanes, err = servmgr.StartOrder(payload.Services)
 		if err != nil {
 			break
 		}
 		var startTasks *state.TaskSet
-		startTasks, err = servstate.Start(st, services, servmgr)
+		startTasks, err = servstate.Start(st, lanes)
 		if err != nil {
 			break
 		}
@@ -143,7 +144,7 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 		taskSet.AddAll(stopTasks)
 		taskSet.AddAll(startTasks)
 	case "replan":
-		var stopNames, startNames []string
+		var stopNames, startNames [][]string
 		stopNames, startNames, err = servmgr.Replan()
 		if err != nil {
 			break
@@ -154,7 +155,7 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 			break
 		}
 		var startTasks *state.TaskSet
-		startTasks, err = servstate.Start(st, startNames, servmgr)
+		startTasks, err = servstate.Start(st, startNames)
 		if err != nil {
 			break
 		}
@@ -165,11 +166,15 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 
 		// Populate a list of services affected by the replan for summary.
 		replanned := make(map[string]bool)
-		for _, v := range stopNames {
-			replanned[v] = true
+		for _, row := range stopNames {
+			for _, v := range row {
+				replanned[v] = true
+			}
 		}
-		for _, v := range startNames {
-			replanned[v] = true
+		for _, row := range startNames {
+			for _, v := range row {
+				replanned[v] = true
+			}
 		}
 		for k := range replanned {
 			services = append(services, k)
@@ -186,6 +191,9 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 	// Use the original requested service name for the summary, not the
 	// resolved one. But do use the resolved set for the count.
 	var summary string
+	for _, names := range lanes {
+		services = append(services, names...)
+	}
 	switch {
 	case len(taskSet.Tasks()) == 0:
 		// Can happen with a replan that has no services to stop/start. A
@@ -222,16 +230,21 @@ func v1PostService(c *Command, r *http.Request, _ *UserState) Response {
 
 // intersectOrdered returns the intersection of left and right where
 // the right's ordering is persisted in the resulting set.
-func intersectOrdered(left []string, orderedRight []string) []string {
+func intersectOrdered(left []string, orderedRight [][]string) [][]string {
 	m := map[string]bool{}
 	for _, v := range left {
 		m[v] = true
 	}
-	var out []string
-	for _, v := range orderedRight {
-		if m[v] {
-			out = append(out, v)
+
+	var out [][]string
+	for _, row := range orderedRight {
+		var intersectRow []string
+		for _, v := range row {
+			if m[v] {
+				intersectRow = append(intersectRow, v)
+			}
 		}
+		out = append(out, intersectRow)
 	}
 	return out
 }
--- a/internals/daemon/api_services.go
+++ b/internals/daemon/api_services.go
@@ -144,18 +144,18 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 		taskSet.AddAll(stopTasks)
 		taskSet.AddAll(startTasks)
 	case "replan":
-		var stopNames, startNames [][]string
-		stopNames, startNames, err = servmgr.Replan()
+		var stopLanes, startLanes [][]string
+		stopLanes, startLanes, err = servmgr.Replan()
 		if err != nil {
 			break
 		}
 		var stopTasks *state.TaskSet
-		stopTasks, err = servstate.Stop(st, stopNames)
+		stopTasks, err = servstate.Stop(st, stopLanes)
 		if err != nil {
 			break
 		}
 		var startTasks *state.TaskSet
-		startTasks, err = servstate.Start(st, startNames)
+		startTasks, err = servstate.Start(st, startLanes)
 		if err != nil {
 			break
 		}
@@ -166,13 +166,13 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 
 		// Populate a list of services affected by the replan for summary.
 		replanned := make(map[string]bool)
-		for _, row := range stopNames {
-			for _, v := range row {
+		for _, lane := range stopLanes {
+			for _, v := range lane {
 				replanned[v] = true
 			}
 		}
-		for _, row := range startNames {
-			for _, v := range row {
+		for _, lane := range startLanes {
+			for _, v := range lane {
 				replanned[v] = true
 			}
 		}
@@ -237,15 +237,15 @@ func intersectOrdered(left []string, orderedRight [][]string) [][]string {
 	}
 
 	var out [][]string
-	for _, row := range orderedRight {
-		var intersectRow []string
-		for _, v := range row {
+	for _, lane := range orderedRight {
+		var intersectLane []string
+		for _, v := range lane {
 			if m[v] {
-				intersectRow = append(intersectRow, v)
+				intersectLane = append(intersectLane, v)
 			}
 		}
-		if len(intersectRow) > 0 {
-			out = append(out, intersectRow)
+		if len(intersectLane) > 0 {
+			out = append(out, intersectLane)
 		}
 	}
 	return out
--- a/internals/daemon/api_services.go
+++ b/internals/daemon/api_services.go
@@ -111,7 +111,7 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 		if err != nil {
 			break
 		}
-		taskSet, err = servstate.Start(st, services)
+		taskSet, err = servstate.Start(st, services, servmgr)
 	case "stop":
 		services, err = servmgr.StopOrder(payload.Services)
 		if err != nil {
@@ -134,7 +134,7 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 			break
 		}
 		var startTasks *state.TaskSet
-		startTasks, err = servstate.Start(st, services)
+		startTasks, err = servstate.Start(st, services, servmgr)
 		if err != nil {
 			break
 		}
@@ -154,7 +154,7 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 			break
 		}
 		var startTasks *state.TaskSet
-		startTasks, err = servstate.Start(st, startNames)
+		startTasks, err = servstate.Start(st, startNames, servmgr)
 		if err != nil {
 			break
 		}
--- a/internals/daemon/api_services.go
+++ b/internals/daemon/api_services.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2020 Canonical Ltd
+// Copyright (c) 2014-2024 Canonical Ltd
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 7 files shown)


IronCore864:
Not in scope of this PR (or the issue https://github.com/canonical/pebble/issues/231 we are trying to fix), but I'd like to point out the painful experience of using requires/before/after when implementing and testing this feature.

This PR is actually my second one on this feature, because in the first draft, I made the mistake of assuming that "when handling task B, if B depends on task A, A should have already been processed, since the input is sorted already, so, simply put B in the lane where A already is." I ignored the possibility that you can actually configure A to require B but A before B, for example, the test case 3 mentioned above:

```yaml
services:
  a:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
    requires:
      - b
    before:
      - b
  b:
    override: replace
    command: sh -c "sleep 999"
    startup: enabled
```

The logic in my first draft won't work, because the sorted order is (A, B), so when handling A, B's lane isn't known yet. Hence my current implementation in this PR.

After giving it more thought, I reached the conclusion that the above use case doesn't seem to make much sense, because "A requires B, but A must be started before B" is illogical:

- If "A requires B", it means "if B doesn't start, A won't". Or to put it another way, "B must start before A".
- If "A must be started before B", it means "B requires A", not the other way around.

So, saying that "A requires B" and "A must start before B" is illogical, the two parts counter each other.

I tried but failed to think of any use case where "A requires B", but "A must be started before B". Some analogies:

- Let's say we want to create a CDN (say, with AWS CloudFront, as service A), and the origin server is hosted in an AWS S3 bucket (as service B). This whole "CDN" product can only work when both CloudFront and S3 are present, but you must create S3 before creating CloudFront because CloudFront needs to know the S3 URL to load content from. So, CloudFront depends on S3 (A requires B), and you must create S3 (B) before creating CloudFront (A).
- A web app comprises a backend (service A) and a database (B), and A requires B because the backend reads/writes to the DB. But you must start A before B? Starting A when B isn't started doesn't make sense because the whole product isn't ready for service anyway.
- Yet another analogy: I want to create an EC2 instance (A) but it requires an IAM role (B). The IAM role must be created first because when creating the EC2 instance, I need to assign the role to the instance. So, the EC2 requires the IAM role (A requires B) and the IAM role must be created before the EC2 (A must start AFTER, not BEFORE, B).

Using three keywords requires/before/after to handle dependency is an overkill, which seems to bring more problems than solve.

I think we can drop "before/after", and rename "requires" to something like "depend-on" (depend-on literally means requires + after).

I prefer renaming "requires" to "depends-on" because "requires" in many cases isn't right:
- In the CDN analogy, the CloudFront doesn't really "require" the S3 origin server. CloudFront can be created with an empty or invalid origin server I believe. It's the CDN (the whole product that requires S3).
- In the web app analogy, the backend also doesn't really "require" the DB; without the DB, when accessing the backend service, it would throw some error saying data can't be loaded/persisted. It's the whole web app product that requires the DB.

If we drop "before/after" and leave only "depends-on", the layer in test case 3 can be expressed as:

```yaml
services:
    a:
        override: replace
        command: /bin/sh -c "echo test1; sleep 10"
        startup: enabled

    b:
        override: replace
        command: /bin/sh -c "echo test2; sleep 10"
        depends-on:
          - a
```

And I think this is more than enough to solve any dependency requirements. Proof: Terraform does exactly the same thing, even the name "depends-on" is borrowed from Terraform.

If we can do this, after sorting, the order is (A, B) and when handling B, A's lane is already decided. In this case, the feature in this PR can be much more simpler:

```
# pseudo
for task in tasks:
  if task.depends_on:
    task.lane = task.depends_on.lane
    task.wait_for(task.depends_on)
```

internals/overlord/servstate/request.go
```diff
@@ -4,28 +4,85 @@ import (
...
 	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/plan"
 )
 
 // ServiceRequest holds the details required to perform service tasks.
 type ServiceRequest struct {
 	Name string
 }
 
+func get_or_create_lane(s *state.State, service *plan.Service, service_lane_mapping map[string]int) int {
```
benhoyt, line 15: Let's just make sure we use Go naming conventions throughout, for example `getOrCreateLane` -- similar below.

benhoyt (changes requested):
Let's discuss this when you're back. But per my comment, I think this would fit better in `plan.StartOrder` (and have that return a list of independent lanes).

benhoyt:
> Using three keywords requires/before/after to handle dependency is an overkill, which brings more problems than solves.

I think you're right -- it's overkill. And I and others have been confused by requires/before/after several times. [There's more in the README here.](https://github.com/canonical/pebble/?tab=readme-ov-file#service-dependencies)

"Requires" basically means depends-on, but doesn't specify any order of starting. "Before" and "after" deal with the ordering.

> I think we can drop "before/after", and rename "requires" to something like "depend-on" (depend-on means requires + after, in my book).

That said, I think that ship has sailed -- in other words, it's too late to change it now. We can probably refine before/after/requires, but not remove or change them significantly.

Let's chat about all this further when you're back, and then you can do a second iteration.

IronCore864:
> > Using three keywords requires/before/after to handle dependency is an overkill, which brings more problems than solves.
> 
> I think you're right -- it's overkill. And I and others have been confused by requires/before/after several times. [There's more in the README here.](https://github.com/canonical/pebble/?tab=readme-ov-file#service-dependencies)
> 
> "Requires" basically means depends-on, but doesn't specify any order of starting. "Before" and "after" deal with the ordering.
> 
> > I think we can drop "before/after", and rename "requires" to something like "depend-on" (depend-on means requires + after, in my book).
> 
> That said, I think that ship has sailed -- in other words, it's too late to change it now. We can probably refine before/after/requires, but not remove or change them significantly.
> 
> Let's chat about all this further when you're back, and then you can do a second iteration.

Refactored according to the above suggestion. Key changes:

- Move the logic with lanes from `servstate` to `plan`.
- Add UT.

Some explanation:

- In the `plan` package, when assembling `lanes [][]string`, it makes sure services with dependencies (`requires`) will be put in the same lane, but the lanes are not created here. Because to use `state.State.NewLane()`, a State must be passed to the Plan and using `State.NewLane()` requires lock. Different indices here mean services of the same index will be put in the same lane, but the indices are not lane IDs since lanes are not created here yet.
- The lanes are created in `servestate.Start()` (and `Stop()`) when creating tasks, where the lock is already aqquired.

internals/daemon/api_services.go
```diff
@@ -143,7 +144,7 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 		taskSet.AddAll(stopTasks)
 		taskSet.AddAll(startTasks)
 	case "replan":
-		var stopNames, startNames []string
+		var stopNames, startNames [][]string
```
benhoyt, line 147: Could these be named `stopLanes, startLanes` now?

internals/daemon/api_services.go
```diff
@@ -165,11 +166,15 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 
 		// Populate a list of services affected by the replan for summary.
 		replanned := make(map[string]bool)
-		for _, v := range stopNames {
-			replanned[v] = true
+		for _, row := range stopNames {
```
benhoyt, line 169:
```suggestion
		for _, lane := range stopLanes {
```

internals/daemon/api_services.go
```diff
@@ -222,15 +230,22 @@ func v1PostService(c *Command, r *http.Request, _ *UserState) Response {
...
 	for _, v := range left {
 		m[v] = true
 	}
-	var out []string
-	for _, v := range orderedRight {
-		if m[v] {
-			out = append(out, v)
+
+	var out [][]string
+	for _, row := range orderedRight {
```
benhoyt, line 240: Let's call this `lane` and `intersectLane`.

internals/overlord/servstate/manager.go
```diff
@@ -204,19 +204,28 @@ func (m *ServiceManager) DefaultServiceNames() ([]string, error) {
...
 	}
 
-	return currentPlan.StartOrder(names)
+	lanes, err := currentPlan.StartOrder(names)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []string
+	for _, row := range lanes {
```
benhoyt, line 213: Again, `lane`.

internals/overlord/servstate/manager_test.go
```diff
@@ -351,8 +351,17 @@ services:
 
 	stops, starts, err := s.manager.Replan()
 	c.Assert(err, IsNil)
-	c.Check(stops, DeepEquals, []string{"test2"})
-	c.Check(starts, DeepEquals, []string{"test1", "test2"})
+	for _, row := range stops {
```
benhoyt, line 354: Why can't we do a direct `DeepEquals` check, like `c.Check(stops, DeepEquals, [][]string{{"test2", "test1"}})`?

internals/overlord/servstate/request_test.go
```diff
@@ -18,13 +18,73 @@ import (
...
+	req, err := servstate.TaskServiceRequest(tasks[0])
+	c.Assert(err, IsNil)
+	c.Assert(req.Name, Equals, "one")
+
+	c.Assert(tasks[1].Kind(), Equals, "start")
+	req, err = servstate.TaskServiceRequest(tasks[1])
+	c.Assert(err, IsNil)
+	c.Assert(req.Name, Equals, "two")
+
+	c.Assert(tasks[0].Lanes()[0], IntNotEqual, tasks[1].Lanes()[0])
```
benhoyt, line 60: How is `IntNotEqual` different from `Not(Equals)`? I've never used that one.

internals/plan/plan.go
```diff
@@ -924,16 +924,78 @@ func (p *Plan) Validate() error {
...
+	for _, name := range names {
+		service, ok := services[name]
+		if !ok {
+			return nil, &FormatError{
+				Message: fmt.Sprintf("service %q does not exist", name),
+			}
+		}
+
+		lane = getOrCreateLane(lane, service, serviceLaneMapping)
+		mapServiceToLane(service, lane, serviceLaneMapping)
```
benhoyt, line 989: It seems odd to me that we don't do this inside `getOrCreateLane`, because as it is, `getOrCreateLane` doesn't ever actually create the lane, right? Can't we get rid of `mapServiceToLane` and just have `getOrCreateLane` fall through to the code in `mapServiceToLane` in the bottom case when it needs to create it?

benhoyt (changes requested):
This work looks good overall! Just requesting some minor changes.

internals/overlord/servstate/manager_test.go
IronCore864: Good spot! In my first draft, the first index of the lanes 2D array is the lane ID, and `lanes[0]` could be empty, so I ended up with this version. Later I changed it and it could not be empty any more. I will change it to a direct `DeepEquals`.

internals/overlord/servstate/request_test.go
IronCore864: I did some testings and it seems they are the same, at least in this case.

I also searched the repo, and although there are occurrences of `IntNotEqual`, there are more occurrences of `Not(Equals)`, so I will change it to the latter for consistency.

internals/plan/plan.go
IronCore864: You are right that `getOrCreateLane` doesn't really create the lane.

But if we move the logic of `mapServiceToLane` to the bottom case, there will be issues:

For example, if "svc1 depends on svc2, svc2 depends on svc3":

1. When handling svc1, it's not in the map, so both svc1 and svc2 are on the map.
2. When handling svc2, since it's already in the map, it will not reach the bottom case, hence svc3 will not be put in the map.
3. When handling svc3, since it's not in the map, a new lane will be created, while in fact it should be put in the same lane.

That's why I separated the logic of creating lanes and mapping service to lanes.

That said, We don't have to call `mapServiceToLane` in `createLanes`, but inside `getOrCreateLane`. Shall I change it?

benhoyt: Yes, I think that (calling `mapServiceToLane` inside `getOrCreateLane`) would make more sense.

IronCore864: I moved `mapServiceToLane` inside `getOrCreateLane`. I really don't like calling the function 3 times in 3 cases. I have two other versions:

The first version I created a bool like `laneAlreadyExist` and use conditions on it for the second/third case, but I find it slightly harder to read than simply return directly.

The second version I had a wrapper, something to this effect:

```go
func getOrCreateLane(currentLane int, service *Service, serviceLaneMapping map[string]int) int {
	// Check if the service has been mapped to a lane
	if lane, ok := getServiceLane(service, serviceLaneMapping); ok {
		return lane
	}

	// Check if any dependency has been mapped to a lane
	for _, dependency := range service.Requires {
		if lane, ok := getServiceLane(dependency, serviceLaneMapping); ok {
			return lane
		}
	}

	// Neither the service itself nor any of its dependencies is mapped to an existing lane
	lane := currentLane + 1
	mapServiceToLane(service, lane, serviceLaneMapping)
	return lane
}

func getServiceLane(name string, serviceLaneMapping map[string]int) (int, bool) {
	if lane, ok := serviceLaneMapping[name]; ok {
		mapServiceToLane(name, lane, serviceLaneMapping)
		return lane, true
	}
	return 0, false
}
```

But I think it's also harder to read.

After consideration, I kept the current version which is to call `mapServiceToLane` in each case.

benhoyt (approved):
Thanks for this! Looking good now, and I've done some basic manual testing too.

PR #438, opened 2024-06-27
docs: remove howtos and add references

IronCore864:
Remove most how-to docs since they are more like references. Create reference docs.

Changes:
- almost all how-tos are removed, except:
  - how to install (newly created)
  - how to manage service dependencies (will rework soon)
- how-tos are moved to references, all are updated using the template provided by Anne. Most reference is a single page, some are grouped together (like changes and tasks, pebble exec and file management commands, notices/notice/notify).
- refactor the tutorial after changing links for how-tos and references.

Preview: https://canonical-pebble--438.com.readthedocs.build/en/438/reference/


benhoyt:
I haven't reviewed this in depth. I have a couple of thoughts:

- I don't love how we're copying and pasting the help output of each command. Is there a nicer way to do that?
- I wonder if this structure would be:

```
Tutorial
How-to guides
Explanation
Reference
  Commands
    Start
    Stop
    Run
    ...
  Layers
  Notices
  ...
```

Let's chat about the structure of this with Anne when you're back, and we can go from there.

IronCore864:
The latest commit contains the two changes:
- Per our discussion elsewhere, all CLI command reference docs are moved under `Reference > CLI Commands`.
- According to Juju documentation, reference documents are ordered alphabetically.

Preview: https://canonical-pebble--438.com.readthedocs.build/en/438/reference/cli-commands/cli-commands/

docs/reference/notices/notices.md
```diff
@@ -35,7 +35,7 @@ See comment at the top of internals/overlord/state/warning.go for more info.
 :titlesonly:
 :maxdepth: 1
 
-Notify command <pebble-notify>
-Noticies command <pebble-notices>
 Notice command <pebble-notice>
+Noticies command <pebble-notices>
```
benhoyt, line 39:
```suggestion
Notices command <pebble-notices>
```

IronCore864:
I have removed the "index" title from the CLI command page and resolved previous comments. Anne's OK with merging this one before coming up with a solution for automated reference doc generation. Please take a final review and merge @benhoyt @AnneCYH.

benhoyt:
I'm happy to merge this in the name of incremental progress. However, the command documentation is a bit scattered right now, so I think there's a bunch of reorganisation that needs to be done for future work. In particular, only some of the commands are under "CLI Commands", and some are under "Changes and Tasks" (`changes`, `tasks`), some are under "Notices" (`notice`, `notices`, `notify`), and some under "Use Pebble in Containers" (`exec`, `ls`, `mkdir`, `rm`, `push`, `pull`).

I think the next step should be automating the command help generation, and that will put *all* the CLI commands under "CLI Commands".

Then we can remove the commands that are in the other subsections. But I think some content should also move to How-tos, for example: "How to use Pebble in containers" (exec, ls, etc), "How to use Pebble notices". Maybe "Changes and Tasks" should go to Explanation, as it's more of an implementation detail. Those pages can link to the CLI commands for full details, of course.

@IronCore864 Perhaps you could propose a draft structure (not a PR, just a table of contents / outline in a GitHub issue) and we can discuss with Anne before we go further?

benhoyt (approved):

Issue #439, opened 2024-06-27
Pebble alive and ready checks


amandahla:
Hi,

From the docs:

```
        # (Optional) Defines what happens when each of the named health checks
        # fail. Possible values are:
        #
        # - restart (default): restart the service once
        # - shutdown: shut down and exit the Pebble daemon (with exit code 11)
        # - success-shutdown: shut down and exit Pebble with exit code 0
        # - ignore: do nothing further
        on-check-failure:
            <check name>: restart | shutdown | success-shutdown | ignore
```

And

```
        # (Optional) Check level, which can be used for filtering checks when
        # calling the checks API or health endpoint.
        #
        # For the health endpoint, ready implies alive. In other words, if all
        # the "ready" checks are succeeding and there are no "alive" checks,
        # the /v1/health API will return success for level=alive.
        level: alive | ready
```

I assumed:
- If you don't set an alive check but set a ready check once it fails, /v1/health/hlevel=alive and /v1/health/level=ready returns `"result":{"healthy":false}`
- If a ready check fails, since on-check-failure default is restart, the service is restarted.

But actually:
- /v1/healt/level=alive returns true.
- When ready fails, nothing happens.

Is this the expected behavior and the documentation is unclear about it, or should the behavior be changed according to what one can assume from the docs?

benhoyt:
Hi @amandahla, thanks for the report. There is actually a small bug in the docs -- see [this PR](https://github.com/canonical/pebble/pull/442) for the fix. Basically, there's no such thing as a default for `on-check-failure`, as that's a map of service name to value pairs, and so doesn't have "defaults" like the `on-success` and `on-failure` fields (which are just strings). So I've just removed the word "default" there in the layer definition docs.

However, the other thing is working as expected. The [health endpoint](https://github.com/canonical/pebble/?tab=readme-ov-file#health-endpoint) section of the README clarifies this:

> Ready implies alive, and not-alive implies not-ready. If you've configured an "alive" check but no "ready" check, and the "alive" check is unhealthy, /v1/health?level=ready will report unhealthy as well, and the Kubernetes readiness probe will act on that.

So ready implies alive, but not-ready does not necessarily mean not-alive ("alive" roughly means the network is up, whereas "ready" means the service is ready to serve). Hence why /v1/health?level=alive or `pebble health --level=alive` returning healthy/true even when the ready check is failing. If you want finer-grained control over ready vs alive, you'll need to define both "ready" and "alive" level checks.

[...]

PR #442, opened 2024-07-01
docs: various improvements to "service dependencies" section

benhoyt:
`on-check-failure` is a map of check name to value, so this field doesn't have a default, you have to add keys and values to the map explicitly. I believe this incorrect text was just a copy 'n' paste error.

In addition, I've added a (hopefully-clarifying) paragraph and tweaked things a bit.

Fixes https://github.com/canonical/pebble/issues/439

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ When multiple services need to be started together, they're started in order acc
 
 Note that currently, `before` and `after` are of limited usefulness, because Pebble only waits 1 second before moving on to start the next service, with no additional checks that the previous service is operating correctly.
 
-If the configuration of `requires`, `before`, and `after` for a service results in a cycle or "loop", an error will be returned when attempting to start or stop the service.
+If the configuration of `before` and `after` for the services results in a cycle, an error will be returned when the Pebble daemon starts (and the plan is loaded) or when a layer that causes a cycle is added.
 
 ### Service auto-restart
 
--- a/README.md
+++ b/README.md
@@ -733,7 +733,7 @@ services:
         # (Optional) Defines what happens when each of the named health checks
         # fail. Possible values are:
         #
-        # - restart (default): restart the service once
+        # - restart: restart the service once
         # - shutdown: shut down and exit the Pebble daemon (with exit code 11)
         # - success-shutdown: shut down and exit Pebble with exit code 0
         # - ignore: do nothing further
--- a/README.md
+++ b/README.md
@@ -353,6 +353,8 @@ The tool running the Pebble server can make use of this, for example, under Kube
 
 Ready implies alive, and not-alive implies not-ready. If you've configured an "alive" check but no "ready" check, and the "alive" check is unhealthy, `/v1/health?level=ready` will report unhealthy as well, and the Kubernetes readiness probe will act on that.
 
+On the other hand, not-ready does not imply not-alive: if you've configured a "ready" check but no "alive" check, and the "ready" check is unhealthy, `/v1/health?level=alive` will still report healthy.
+
 If there are no checks configured, the `/v1/health` endpoint returns HTTP 200 so the liveness and readiness probes are successful by default. To use this feature, you must explicitly create checks with `level: alive` or `level: ready` in the layer configuration.
 
 ### Changes and tasks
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -251,11 +251,11 @@ If you want to force a service to restart even if its service configuration hasn
 
 ### Service dependencies
 
-Pebble takes service dependencies into account when starting and stopping services. When Pebble starts a service, it also starts the services which that service depends on (configured with `required`). Conversely, when stopping a service, Pebble also stops services which depend on that service.
+Pebble takes service dependencies into account when starting and stopping services. When Pebble starts a service, it also starts the services which that service depends on (configured with `requires`). Conversely, when stopping a service, Pebble also stops services which depend on that service.
```
dimaqq, line 254: Possibly off-topic:

I wonder if this would read a little better with a small doze or general computing jargon, for example:

"services which that service depends on" -> "dependencies"

Additionally, I'm left to guess what "also starts x" means, does it means that dependencies are started before or in parallel with the denepdent? Does pebble wait for dependencies to actually start?

README.md
```diff
@@ -251,11 +251,11 @@ If you want to force a service to restart even if its service configuration hasn
...
 
-When multiple services need to be started together, they're started in order according to the `before` and `after` configuration, waiting 1 second for each to ensure the command doesn't exit too quickly. The `before` option is a list of services that this service must start before (it may or may not `require` them). Or if it's easier to specify this ordering the other way around, `after` is a list of services that this service must start after.
+When multiple services need to be started together, they're started in order according to the `before` and `after` configuration, waiting 1 second for each to ensure the command doesn't exit too quickly. The `before` option is a list of services that this service must start before (it may or may not "require" them). Or if it's easier to specify this ordering the other way around, `after` is a list of services that this service must start after.
```
dimaqq, line 258: More off-topic:

I guess this paragraph somewhat disambiguates the above, I I just wonder if it's possible to somehow help a lazy/hurried reader and insert a small quip in the section above.

dimaqq (approved):
totally off-topic:

> If the configuration of `requires`, `before`, and `after` for a service results in a cycle or "loop", an error will be returned when attempting to start or stop the service.

So... it's OK to load config from disc with `--on-hold` or possibly replan, but an error would be raised later, when trying to start a service?

If some services are running, is it possible to end up in a situation where they no cannot be stopped?

Will pebble fail or succeed if some services are magically, conveniently started or stopped a priori?

Something is odd here.

README.md
```diff
@@ -353,6 +353,8 @@ The tool running the Pebble server can make use of this, for example, under Kube
 
 Ready implies alive, and not-alive implies not-ready. If you've configured an "alive" check but no "ready" check, and the "alive" check is unhealthy, `/v1/health?level=ready` will report unhealthy as well, and the Kubernetes readiness probe will act on that.
 
+On the other hand, not-ready does not imply not-alive: if you've configured a "ready" check but no "alive" check, and the "ready" check is unhealthy, `/v1/health?level=alive` will still report healthy.
```
dimaqq, line 356: Tangential: given that this needed to be clarified, how about this:

I'm assuming that it's possible to specify both ready and health checks on the same service.

What if their results are in conflict? That is raw results are `ready: true; alive: false`?

dimaqq: Or is it perhaps that "ready implies alive, if the latter is not specified"?

benhoyt:
> So... it's OK to load config from disc with --on-hold or possibly replan, but an error would be raised later, when trying to start a service?

That's a good question (you had me guessing), but no, it's actually when the plan is loaded (when `pebble run` starts):

```
$ cat /var/lib/pebble/default/layers/001-base.yaml 
services:
  nginx:
    override: replace
    command: sleep 3000
    after: [logger]
  logger:
    override: replace
    command: sleep 2000
    after: [nginx]
$ pebble run
cannot run daemon: cannot load plan: services in before/after loop: logger, nginx
# exit code 1
```

I've fixed this paragraph and clarified in https://github.com/canonical/pebble/pull/442/commits/4036e7aec6d2f4d45fa2e8b336aacaada24da5b5. (Also, `requires` isn't involved, as a loop there is okay; it's just before/after.)

README.md
benhoyt: 1) I intended to *define* the term "service dependencies" that I lead with, hence me spelling it out "which that service depends on". So I think I'll keep that.
2) Regarding "also starts", I do think that is clarified in the next paragraphs. `requires` doesn't actually do anything about ordering, just "starts with" or "also starts". `before` and `after` do the ordering part (kind of -- see the "of limited usefulness" note).

README.md
benhoyt: I've (hopefully) clarified slightly in https://github.com/canonical/pebble/pull/442/commits/c1fe1d346d2b302b403508241f909f3d3d7701a1

If they're both specified, they're not in conflict, it's just that there's no fallback: /v1/health?level=ready gives you the `ready` checks, and /v1/health?level=alive gives you the `alive` checks directly.

PR #443, opened 2024-07-04
fix!: remove warnings

benhoyt:
Pebble has never used the warnings system or recorded any warnings. It's getting in the way, so remove it (for one thing, before this PR it means a state lock/unlock on every request to get the warnings summary -- which is always 0).

Warnings (inherited from snapd) assume a single client, which isn't true and doesn't make a lot of sense for Pebble: they have a single lastShown timestamp (on the server) for when they were last shown to the (single) client, whereas notices aren't marked as "shown" on the server. Each server response included pending warnings (pending means not shown or having a lastShown earlier than repeatAfter ago).

This PR:

- removes the warnings from state
- keeps the /v1/warnings API (but it's hard-coded to return no warnings)
- keeps the Go client functions (though again, they're hard-coded to return nothing)
- removes the `warnings` CLI command (`okay` now *just* okays notices)

Separately (or in this PR if desired?) we could reinstate the `pebble warnings` command to use notices of type=warning and save the last warning timestamp locally on the client. But I'm not sure that gains much over the user just using `pebble notices --type=warning`.

Discussion points for spec review meeting:

* Do we need warnings in every request? Seems silly to try to keep this complexity for a feature that's not and has never been used. But we could reimplement this either with a timestamp sent to the server on every request, or just the CLI querying for warnings after every operation.
* Should we just get rid of the /v1/warnings API endpoints too (or change them to 410 Gone)? I'm pretty certain nobody will be using these, as warnings have never been a thing in Pebble.


internals/daemon/api_warnings.go
```diff
@@ -35,41 +36,20 @@ func v1AckWarnings(c *Command, r *http.Request, _ *UserState) Response {
...
-		ws, _ = statePendingWarnings(st)
-	}
-	if len(ws) == 0 {
-		// no need to confuse the issue
-		return SyncResponse([]state.Warning{})
-	}
-
-	return SyncResponse(ws)
+	// Do nothing; return empty JSON array (type doesn't matter).
+	return SyncResponse([]string{})
```
thp-canonical, line 54: Depending on whether we want to find out about clients using this or not (and clients finding out that they are doing an "unnecessary" API call that we might want them to stop doing eventually), a 410 status code could be worth considering here.

client/client.go
```diff
@@ -223,7 +221,9 @@ func (client *Client) Maintenance() error {
 // the user, and the timestamp of the most recently added warning (useful for
 // silencing the warning alerts, and OKing the returned warnings).
 func (client *Client) WarningsSummary() (count int, timestamp time.Time) {
-	return client.warningCount, client.warningTimestamp
+	// Pebble has never produced warnings (and now it can't), so we can just
```
thp-canonical, line 224: "and now it can't" could be changed to "support code for warnings was removed in July 2024" or "support code for warnings was removed in PR#443" or something, so that casual observers of the future won't have to wonder why it can't produce warnings "now".

This comment also applies to the other instances below where this comment is reproduced.

internals/cli/cmd_okay.go
```diff
@@ -23,20 +23,17 @@ import (
...
+'{{.ProgramName}} notices', so that they are omitted from future runs of the command.
+When a notice is repeated, it will show up again until the next '{{.ProgramName}} okay'.
 `
 
 type cmdOkay struct {
 	client *client.Client
 
 	socketPath string
-
-	Warnings bool `long:"warnings"`
```
thp-canonical, line 39: This might break CLI-using scripts that used `--warnings` before. Should we leave this in as a no-op, but undocumented flag (so that scripts that pass `--warnings` continue to work) to not break existing CLI users? (assuming `go-flags` fails for unknown command-line flags, and doesn't just ignore them by default, anyway)

thp-canonical (approved):

benhoyt:
Per spec review discussion today, we agreed it's better to update warnings to use notices properly, rather than remove them. And we still want to report active warnings in each response somehow.

To solve the "single client" issue (there's no `warning.timeShown` for notices), we'd have the server report in each response the timestamp of the latest warning (this was @dimaqq's suggestion). That way the client could compare against its stored timestamp of latest okayed warning, and it would know whether there were new warnings or not. It wouldn't know the number, but the client could easily query for that if there was 1 or more.

So closing this draft PR. I'll open an issue to track that work.

PR #444, opened 2024-07-04
feat: include check name in notice data for perform-check and recover-check

tonyandrewmeyer:
For notices of kind `change-updated`, the notice data will include the `kind` field that is standard for these notices, but also when `kind` is `perform-check` or `recover-check`, it will also include `check-name`, set to the name of the check that is being performed or recovered.

This is the Pebble portion of [OP046](https://docs.google.com/document/d/13ItH8l5ZSmmv9WqZXpqjV8EifZU5e3zxINiNLubnC68/edit). It is used by Juju ([PR](https://github.com/juju/juju/pull/17655)) and ops ([PR](https://github.com/canonical/operator/pull/1281)).

### DIFF
--- a/internals/overlord/state/change.go
+++ b/internals/overlord/state/change.go
@@ -17,6 +17,7 @@ package state
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -428,7 +429,16 @@ func (c *Change) addNotice() error {
 	opts := &AddNoticeOptions{
 		Data: map[string]string{"kind": c.Kind()},
 	}
-	_, err := c.state.AddNotice(nil, ChangeUpdateNotice, c.id, opts)
+	var extraData map[string]string
+	err := c.Get("notice-data", &extraData)
+	if err == nil {
+		for k, v := range extraData {
+			opts.Data[k] = v
+		}
+	} else if !errors.Is(err, &NoStateError{Key: "notice-data"}) {
+		return fmt.Errorf("could not get notice data from change %s: %v", c.ID(), err)
+	}
+	_, err = c.state.AddNotice(nil, ChangeUpdateNotice, c.id, opts)
 	return err
 }
 
--- a/internals/overlord/state/change.go
+++ b/internals/overlord/state/change.go
@@ -435,8 +435,8 @@ func (c *Change) addNotice() error {
 		for k, v := range extraData {
 			opts.Data[k] = v
 		}
-	} else if !errors.Is(err, &NoStateError{Key: "notice-data"}) {
-		return fmt.Errorf("could not get notice data from change %s: %v", c.ID(), err)
+	} else if !errors.Is(err, ErrNoState) {
+		return fmt.Errorf("cannot get notice data from change %s: %w", c.ID(), err)
 	}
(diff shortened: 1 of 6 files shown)


internals/overlord/state/state.go
```diff
@@ -333,11 +333,22 @@ func (s *State) Cache(key, value interface{}) {
 
 // NewChange adds a new change to the state.
 func (s *State) NewChange(kind, summary string) *Change {
+	return s.NewChangeWithNoticeData(kind, summary, nil)
+}
+
+// NewChangeWithNoticeData adds a new change to the state, adding in any provided extra data.
```
benhoyt, line 339: Let's be a bit more specific: the data is specifically data for the change-update notices, not data for the change itself. I'd say we should call this `noticeData` and tweak the doc comment.

internals/overlord/state/change.go
```diff
@@ -428,7 +429,16 @@ func (c *Change) addNotice() error {
...
-	_, err := c.state.AddNotice(nil, ChangeUpdateNotice, c.id, opts)
+	var extraData map[string]string
+	err := c.Get("notice-data", &extraData)
+	if err == nil {
+		for k, v := range extraData {
+			opts.Data[k] = v
+		}
+	} else if !errors.Is(err, &NoStateError{Key: "notice-data"}) {
+		return fmt.Errorf("could not get notice data from change %s: %v", c.ID(), err)
+	}
```
benhoyt, line 440: It's kind of odd, but `NoStateError` has an `Is(error) bool` method which checks against `*NoStateError` or `ErrNoState`. So it's simplest (and I think intended) to do `errors.Is(err, ErrNoState)`. [See this test.](https://github.com/canonical/pebble/blob/c330d5f0f37cd7c47b4ec3ce094b39e43fb3f6b4/internals/overlord/state/state_test.go#L1102-L1107)

Also, note we've settled on "cannot ..." for most Pebble errors. Also notice the `%w` verb, which wraps the error, instead of just including it as a string:

```suggestion
	if err == nil {
		for k, v := range extraData {
			opts.Data[k] = v
		}
	} else if !errors.Is(err, ErrNoState) {
		return fmt.Errorf("cannot get notice data from change %s: %w", c.ID(), err)
	}
```

benhoyt: Originally I was going to say that in Go we normally do early error returns, which is true, but I wrote it that way and it wasn't actually clearer. So I think the structure you have is fine.

benhoyt (changes requested):
Nice work! Some small changes requested for comments and error message.

I've also tested locally and see the check-name nicely in the state file:

```
    {
      "id": "1",
      "user-id": null,
      "type": "change-update",
      "key": "1",
      "first-occurred": "2024-07-05T03:50:52.556541656Z",
      "last-occurred": "2024-07-05T03:51:46.633442924Z",
      "last-repeated": "2024-07-05T03:51:46.633442924Z",
      "occurrences": 4,
      "last-data": {
        "check-name": "chk1",
        "kind": "perform-check"
      },
      "expire-after": "168h0m0s"
    }
```

tonyandrewmeyer:
> I've also tested locally and see the check-name nicely in the state file:

Ah, yes, thanks. I should have mentioned that I checked that and also locally tested that wait notices has it in the last data as well.

benhoyt (approved):

PR #448, opened 2024-07-12
fix(snap): build snap with Go 1.22 to fix CVE-2024-24790

cjdcordeiro:
Although not shown in the CI (because Trivy is being run with `trivy fs`, which [doesn't consider the Pebble binary](https://aquasecurity.github.io/trivy/v0.24.4/vulnerability/detection/language/)), there currently is a reported CVE in the Pebble snap (CVE-2024-24790).

This CVE seems to have been fixed on newer versions of Go, but we are still using Go 1.20 to build the Pebble snap.

This PR bumps this build dependency to Go 1.22.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -33,7 +33,7 @@ parts:
     plugin: go
     source: .
     build-snaps:
-      - go/1.20/stable
+      - go/1.22/stable
     override-build: |
       go generate ./cmd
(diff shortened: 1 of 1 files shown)


dimaqq (approved):

benhoyt:
Though I believe this CVE doesn't actually affect Pebble, I have no problems merging this -- thanks.

However, I also need to fix the Snap versioning before this goes out to the snap. Currently the (recent) snap versions are showing as hex commit hashes rather than version numbers (even for releases). I'll work on figuring this out today.

benhoyt (approved):

PR #449, opened 2024-07-16
chore: use Go 1.22 (supported), in go.mod and binary builds

benhoyt:
Per discussion with Harry, we'd like to bump up the minimum Go version to a supported version (1.21 and 1.22). Might as well go up to 1.22 (that's what Juju is on now as well).


hpidcock (approved):
LGTM, thank-you.

PR #450, opened 2024-07-16
fix(snap): set Pebble version before snap builds

cjdcordeiro:
# Problem

When doing a `snapcraft remote-build`, Snapcraft will create a copy of the project in Launchpad before running the snap build.
The **problem** is that Snapcraft seems to ignore tags while creating the copy of the project, which means the `go generate ./cmd` in the `snapcraft.yaml`'s `override-build` will never dynamically generate the right version.

# Solution

This PR enforces the execution of `go generate ./cmd` as a pre-requisite for running the snap build. I.e., for both local and remote builds, `go generate ./cmd` must be run first, such that the `VERSION` and `version_generated.go` files are staged in the repo, prior to the snap build, such that the `override-build` script can simply rely on that information and thus ignore the missing Git refs in the LP's project.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -30,6 +30,14 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: "go.mod"
+
+      # Generate the version for the local snap build.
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Looks good. Thanks for your help with this, @cjdcordeiro!

PR #451, opened 2024-07-17
fix: avoid need to lock state for restart.Pending()

benhoyt:
This change avoids the need to lock/unlock state when fetching `restart.Pending()`, avoiding the need for one state lock/unlock on every HTTP request. (We hope to get rid of the other lock/unlock by removing warnings, but that's a separate issue.)

Updates https://github.com/canonical/pebble/issues/366


internals/overlord/restart/restart.go
```diff
@@ -60,7 +61,7 @@ type restartManagerKey struct{}
 // RestartManager takes care of restart-related state.
 type RestartManager struct {
 	state      *state.State
-	restarting RestartType
+	restarting atomic.Int32 // really of type RestartType
```
hpidcock, line 64: In future we could use an `atomic.Pointer[RestartType]` instead, although unnecessary heap allocation for such a small type, keeps the type.

internals/overlord/restart/restart.go
```diff
@@ -149,24 +150,20 @@ func Request(st *state.State, t RestartType) {
...
 // Pending returns whether a restart was requested with Request and of which type.
-func Pending(st *state.State) (bool, RestartType) {
-	cached := st.Cached(restartManagerKey{})
-	if cached == nil {
-		return false, RestartUnset
-	}
-	rm := cached.(*RestartManager)
-	return rm.restarting != RestartUnset, rm.restarting
+// NOTE: the state does not need to be locked to fetch this information.
+func (rm *RestartManager) Pending() (bool, RestartType) {
```
hpidcock, line 159: an "ok" bool is usually the last return value (like error)

e.g.
`v, ok := myMap[k]`

Since this is existing functionality, I don't mind.

hpidcock (approved):

internals/overlord/restart/restart.go
benhoyt: Yeah, I was actually tempted to get rid of the bool entirely, because it's just `b != RestartUnset`, which is easy for the caller to do. However, given it has this signature in snapd, I want to keep it as is.

PR #452, opened 2024-07-17
fix: make Client.Pull properly handle API errors

### LINKED ISSUE #413: "pebble pull" gives wrong error in case of access denied
When pebble is running as root and you do a `pebble pull`, the error message is unexpected:

```
$ PEBBLE=/var/lib/pebble/default/ ./pebble pull /etc/hosts hosts
error: expected a multipart response, got "application/json"
```

Compare that to what you get with curl:

```
$ curl --unix-socket /var/lib/pebble/default/.pebble.socket 'http://x/v1/files?action=read&path=/etc/hosts'
```

Presumably this is happening because it's checking the media type before checking for an early error response.

benhoyt:
Currently the `Client.Pull` method is immediately checking the content type for multipart, but we should also check for a regular API error. This PR fixes that.

Before (in case of permission denied error):

```
$ pebble pull /etc/hosts hosts
error: expected a multipart response, got "application/json"
```

After:

```
$ pebble pull /etc/hosts hosts
error: access denied (try with sudo)
```

Fixes #413

### DIFF
--- a/client/files.go
+++ b/client/files.go
@@ -550,7 +550,16 @@ func (client *Client) Pull(opts *PullOptions) error {
 		return fmt.Errorf("cannot parse Content-Type: %w", err)
 	}
 	if mediaType != "multipart/form-data" {
-		// Not an error response after all.
+		// If it's not multipart, it's likely an API error, so try parsing it as such.
+		var serverResp response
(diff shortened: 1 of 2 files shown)


anpep (approved):
Nice catch, thanks!

PR #453, opened 2024-07-17
feat(daemon): add Err() to retrieve the tomb death reason

thp-canonical:
Daemon has a `.Dying()` method that mirrors the [`Tomb.Dying()`](https://pkg.go.dev/gopkg.in/tomb.v2#Tomb.Dying) API, but not `.Err()` for the [`Tomb.Err()`](https://pkg.go.dev/gopkg.in/tomb.v2#Tomb.Err) API. Add a `.Err()` pass-through function and print out the server exit reason for easier debugging should one of the daemon tomb goroutines die.

### DIFF
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -776,6 +776,12 @@ func (d *Daemon) Dying() <-chan struct{} {
 	return d.tomb.Dying()
 }
 
+// Err returns the death reason, or ErrStillAlive
+// if the tomb is not in a dying or dead state.
+func (d *Daemon) Err() error {
+	return d.tomb.Err()
+}
+
 func clearReboot(st *state.State) {
 	// FIXME See notes in the state package. This logic should be
 	// centralized in the overlord which is the orchestrator. Right
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -292,7 +292,7 @@ out:
 			break out
 		case <-d.Dying():
 			// something called Stop()
-			logger.Noticef("Server exiting!")
+			logger.Noticef("Server exiting! Reason: %v", d.Err())
 			break out
 		case <-checkTicker:
 			if err := sanityCheck(); err == nil {


internals/daemon/daemon.go
```diff
@@ -776,6 +776,12 @@ func (d *Daemon) Dying() <-chan struct{} {
 	return d.tomb.Dying()
 }
 
+// Err returns the death reason, or ErrStillAlive
+// if the tomb is not in a dying or dead state.
```
thp-canonical, line 780: Comment taken from [`d.tomb.Err()`](https://pkg.go.dev/gopkg.in/tomb.v2#Tomb.Err)'s documentation.

benhoyt:
Looks reasonable to me. However, I'm wondering how this can happen. Currently it looks like the daemon mostly calls `d.tomb.Kill(nil)` with a nil death reason. Do you have a repro?

thp-canonical:
> Looks reasonable to me. However, I'm wondering how this can happen. Currently it looks like the daemon mostly calls `d.tomb.Kill(nil)` with a nil death reason. Do you have a repro?

I think it also happens when one of the tomb goroutines exits with an error, right? From [the `Tomb.Go()` docs](https://pkg.go.dev/gopkg.in/tomb.v2#Tomb.Go):

> If f returns a non-nil error, t.Kill is called with that error as the death reason parameter.

So any errors returned from here would (I think) also be reported, not just `.Kill()` calls:

https://github.com/canonical/pebble/blob/02e5d2d72e53413d0ac7ed78c41240d5a6f1ed44/internals/daemon/daemon.go#L469-L474

Same for this related place:

https://github.com/canonical/pebble/blob/02e5d2d72e53413d0ac7ed78c41240d5a6f1ed44/internals/daemon/daemon.go#L479-L485

With the [documentation for `Tomb.Kill()`](https://pkg.go.dev/gopkg.in/tomb.v2#Tomb.Kill) saying:

> Althoguh (sic) Kill may be called multiple times, only the first non-nil error is recorded as the death reason. 

So the ` && d.tomb.Err() == tomb.ErrStillAlive` checks above are probably not needed (if their intention is only to avoid overwriting an existing error).

I haven't followed if there's a way to bubble up errors from some handlers all the way to the caller of the `.Serve()` function, but it seems like it at least any error return value from any of the `.Serve()` calls that's ` != http.ErrServerClosed` would be good to have reported instead of silently ignored.

benhoyt (approved):

PR #454, opened 2024-07-17
fix: actually turn off child subreaping in reaper.Stop

### LINKED ISSUE #412: Turn off child subreaper at OS level in reaper.Stop
Currently `reaper.Stop` kills the reaper tomb and does some bookkeeping, but doesn't actually make the OS call to turn off child subreaping. We should do that.

Also, while we're in there, we should clean up `setChildSubreaper`'s return values. It returns `(bool, error)` right now, and it's unnecessarily confusing.

benhoyt:
Previously we weren't actually turning off child subreaping at the OS level in `reaper.Stop`, so it didn't reverse `reaper.Start`. Fix this, and also simplify the error handling a little.

This almost certainly didn't cause problems, because as soon as the Pebble process exits it doesn't matter. But it wouldn't have turned it off in tests, and it's just the Right Thing To Do.

Fixes #412

### DIFF
--- a/internals/reaper/reaper.go
+++ b/internals/reaper/reaper.go
@@ -16,6 +16,7 @@ package reaper
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


internals/reaper/reaper.go
```diff
@@ -75,22 +73,28 @@ func Stop() error {
 	started = false
 	mutex.Unlock()
 
+	err := setChildSubreaper(0)
```
flotter, line 76: Would it be theoretically safer to disable receiving ```unix.SIGCHLD``` signals, before we terminate the processing loop that deals with them ?

flotter (approved):
Looks good to me. Just the one question :)

benhoyt:
Yep, fair point -- fixed.

PR #455, opened 2024-07-18
fix(timeutil): remove unreachable panic() call

thp-canonical:
This has been added in upstream snapd back in the day here:

* https://github.com/snapcore/snapd/pull/4775

Move conditionals to the correct `case` where they belong (e.g. `>= -cutoffDays` belongs to the "%d days ago", ...), and then just naturally move the non-special case of `then.Format("2006-01-02")` to the `default` case.

Looking at compiler explorer, it seems like Go (gc 1.21 on x86_64) can optimize out the old version's `panic()`, so it's just dead code in the source (as it should be): https://godbolt.org/z/fxqTzYfPv

### DIFF
--- a/internals/timeutil/human.go
+++ b/internals/timeutil/human.go
@@ -52,9 +52,7 @@ func delta(then, now time.Time) int {
 func humanTimeSince(then, now time.Time, cutoffDays int) string {
 	d := delta(then, now)
 	switch {
-	case d < -cutoffDays || d > cutoffDays:
-		return then.Format("2006-01-02")
-	case d < -1:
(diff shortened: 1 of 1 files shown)


benhoyt:
Does this actually cause any problems? I'm not opposed to this, though might be better to get it into snapd first? Otherwise it seems to me "if it ain't broke, don't fix it".

thp-canonical:
Proposed fix in snapd: https://github.com/canonical/snapd/pull/14427

benhoyt (approved):

benhoyt:
This was merged over in snapd -- merging here as well.

PR #456, opened 2024-07-23
feat(daemon): allow overriding layers directory in daemon.Options

thp-canonical:
This allows having the layers directory in a different directory than the "layers" sub-directory inside the pebble directory.

The defaults are unchanged, but one can pass an optional non-empty string in `daemon.Options{LayersDir: ...}` when calling `daemon.New()` to use a custom location for the layers.

This also removes the need to pass `pebbleDir` into `plan` and `planstate`, as it really only needs the `layersDir`.

### DIFF
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -62,6 +62,10 @@ type Options struct {
 	// Dir is the pebble directory where all setup is found. Defaults to /var/lib/pebble/default.
 	Dir string
 
+	// LayersDir is an optional path for the layers directory.
+	// Defaults to "layers" inside the pebble directory.
+	LayersDir string
+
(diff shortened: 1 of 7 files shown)


flotter:
This PR relates to KO064 for supporting a different storage arrangement for a derivative project, where R/W data persisted over reboot is located at a separate location.

benhoyt (approved):
This looks good to me, thanks. I like how it makes the ReadDir parameter more sensible, too.

thp-canonical:
@hpidcock What do you think about this change?

benhoyt:
Harry's busy (and working night) this week with the roadmap virtual sprint. I think this is a reasonable change and well done. Let's just go ahead and merge it.

PR #440, opened 2024-06-28
docs: add explanations for how to manage service dependencies

IronCore864:
A how-to guide on managing service dependencies.

Preview: https://canonical-pebble--440.com.readthedocs.build/en/440/how-to/service-dependencies/

Note that this PR include changes from https://github.com/canonical/pebble/pull/438; to only review the changes on the service dependencies how-to guide, see [here](https://github.com/IronCore864/pebble/pull/2).

PR #457, opened 2024-07-24
docs: how to manage service dependencies

AnneCYH:


IronCore864:
I have reviewed Anne's suggestions in detail and we discussed them on a 1-1 call.

FYI @benhoyt we decided to merge this PR first because the structural change is quite much and it would be more difficult to merge this one into my PR than to merge mine into the main.

After this PR is merged, in [my original PR](https://github.com/canonical/pebble/pull/440), I will rebase it on Anne's work, then add references for service dependencies and start orders, and add the links in the how-to.

docs/how-to/service-dependencies.md
```diff
@@ -1,11 +1,270 @@
...
+frontend  enabled  active    today at 10:20 CST
+```
+
+We can configure the layer so that Pebble only starts a given service if all
+other services it is dependent on are running successfully to avoid consuming
+resources unnecessarily.
+
+## Define service dependencies
+
+To create service dependencies in Pebble, use the ``requires`` key with
```
benhoyt, line 123: Why the double backticks? Markdown only requires a single backtick (here and below).

docs/how-to/service-dependencies.md
```diff
@@ -1,11 +1,270 @@
...
+``before`` / ``after`` in the
+[service definition](../reference/layer-specification.md).
+
+### Specify dependent services
+
+To specify one or more services that a given service requires to run
+successfully, use the ``require`` key.
```
benhoyt, line 130: Typo - `requires` needs to be plural:

```suggestion
successfully, use the `requires` key.
```

docs/how-to/service-dependencies.md
```diff
@@ -1,11 +1,270 @@
...
+      - database
+    after:
+      - database
+  database:
+    override: replace
+    command: python3 -m http.server 3306
+    startup: enabled
+```
+
+In the updated layer above, the ``backend`` service requires the ``database``
```
benhoyt, line 189: I think this paragraph should mention the services in the order they appear above, so start with frontend, then proceed to backend (rather than the other way around).

docs/how-to/service-dependencies.md
```diff
@@ -1,11 +1,270 @@
...
+```{terminal}
+   :input: pebble services
+   
+Service   Startup  Current  Since
+backend   enabled  active   today at 10:17 CST
+database  enabled  active   today at 10:17 CST
+frontend  enabled  active   today at 10:17 CST
+```
+
+To further verify the service dependencies, we can force fail a service that is
```
benhoyt, line 230: "Force fail" doesn't mean much to me. What about "if we force a service that is requires by another service to fail"? And how is it being forced to fail here -- that's unclear.

docs/how-to/service-dependencies.md
```diff
@@ -1,11 +1,270 @@
...
+required by another service.
+For example, if we force the ``database`` service to fail, the output for the
+``pebble run`` command should be similar to:
+
+```{terminal}
+   :input: pebble run 
+
+2024-06-28T02:28:06.569Z [pebble] Started daemon.
+[database backend frontend]
```
benhoyt, line 239: What is this line? I don't see it when I run this.

benhoyt (changes requested):
Looks good overall. Requesting a few minor changes.

docs/how-to/service-dependencies.md
AnneCYH: Good point @benhoyt ; I get confused switching between RST and MD. :D I'll update the whole doc.

docs/how-to/service-dependencies.md
AnneCYH: Will update it @benhoyt .

docs/how-to/service-dependencies.md
AnneCYH: Good point.

docs/how-to/service-dependencies.md
AnneCYH: Hi @IronCore864 , could you check this please? I copied the snippet from your branch.
Thanks!

docs/how-to/service-dependencies.md
AnneCYH: Hey @benhoyt , 

"How is it being forced to fail here." For the first part it is a general guide on how you may want to test this. So I added (back) more context for the web app example we are using here.

Hope the changes make sense.
---

To further verify the service dependencies, we force a service that is required
by another service to fail.

In this example, if we force the `database` service to fail (by using port 3306
for another process), the output for the `pebble run` command should be similar
to:

AnneCYH:
Hi @benhoyt , I've addressed all your comments but left one conversation open for your review.

Thanks.

benhoyt (approved):

PR #458, opened 2024-07-25
docs: automate cmd reference docs generation

IronCore864:
A script to help generate CLI CMD reference docs.

Usage: `tox -e commands`

Changes:
- Doc title command uses lowercase to match the exact command. E.g.: "Services command" -> "services command". The order is kept, not using alphabetic order.
- Add reference tags for myst usage. E.g.: "(reference_pebble_stop_command)=".
- A script that generate CLI CMD docs:
  - Get all commands using `pebble help --all`.
  - If `cmd.md` reference doc doesn't exist, create it from a simple template.
  - Get help command output and update the docs.
  - If the reference doc is created using template (in step 2 above), update the title and description as well.
  - Update toc tree in the index page.
  - Show warnings and exit with non-zero value if docs are updated.
- Run the above script to update existing docs and generate missing docs.

Preview: https://canonical-pebble--458.com.readthedocs.build/en/458/reference/cli-commands/cli-commands/

### DIFF
--- a/docs/generate_cmd_doc.py
+++ b/docs/generate_cmd_doc.py
@@ -1,0 +1,154 @@
+import logging
+import os
+import re
+import subprocess
+import typing
+
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+AUTOMATED_START_MARKER = "<!-- START AUTOMATED OUTPUT -->"
+AUTOMATED_STOP_MARKER = "<!-- END AUTOMATED OUTPUT -->"
+
+TEMPLATE = """\
+# CMD command
+
+Description
+
+## Usage
+
+<!-- START AUTOMATED OUTPUT -->
+```{terminal}
+   :input: command
+```
+<!-- END AUTOMATED OUTPUT -->
+"""
+
+
+def get_all_pebble_commands() -> typing.List[str]:
+    cmd = "pebble help --all"
+    p = subprocess.run(cmd.split(), text=True, capture_output=True)
+    if p.returncode != 0:
+        logging.error("Error running {}.".format(cmd))
+        exit(1)
+    return re.findall(r"\n\s{4}(\w+)", p.stdout)
+
+
+def get_command_help_output(cmd: str) -> str:
+    p = subprocess.run(cmd.split(), text=True, capture_output=True)
+    if p.returncode != 0:
+        logging.error("Error running {}.".format(cmd))
+        exit(1)
+    return p.stdout
+
+
+def get_description_from_output(text: str) -> str:
+    pattern = r"Usage:\n.*?\n\n(.*?)\n.*"
+    match = re.search(pattern, text, re.DOTALL)
+    if match:
+        desired_block = match.group(1).strip()
+        return desired_block
+    return ""
+
+
+def render_title(text: str, cmd: str) -> str:
+    return re.sub(
+        r"^# CMD command$",
+        f"# {cmd.capitalize()} command",
+        text,
+        flags=re.MULTILINE,
+    )
+
+
+def render_description(text: str, description: str) -> str:
+    return re.sub(r"^Description$", f"{description}", text, flags=re.MULTILINE)
+
+
+def render_code_block_cmd(text: str, cmd: str) -> str:
+    return re.sub(r"(:input: ).*$", rf"\1{cmd}", text, count=1, flags=re.MULTILINE)
+
+
+def render_code_block_output(text: str, output: str) -> str:
+    start_pos = text.find(AUTOMATED_START_MARKER)
+    end_pos = text.find(AUTOMATED_STOP_MARKER) + len(AUTOMATED_STOP_MARKER)
+    return text[:start_pos] + output + text[end_pos:]
+
+
+def update_toc(all_commands):
+    index_page = "reference/cli-commands/cli-commands.md"
+    with open(index_page, "r") as file:
+        text = file.read()
+
+    start_index = text.find("```{toctree}")
+    end_index = text.find("```", start_index + 1) + 3
+    cmd_list = "\n".join(["{cmd} <{cmd}>".format(cmd=cmd) for cmd in all_commands])
+    toc_tree = """\
+```{{toctree}}
+:titlesonly:
+:maxdepth: 1
+
+{cmd_list}
+```""".format(cmd_list=cmd_list)
+
+    text = text[:start_index] + toc_tree + text[end_index:]
+    with open(index_page, "w") as file:
+        file.write(text)
+
+
+def main():
+    all_commands = get_all_pebble_commands()
+    for command in all_commands:
+        logger.info("Processing doc for command {}".format(command))
+
+        file_path = "reference/cli-commands/{}.md".format(command)
+        file_existed = os.path.exists(file_path)
+        if not file_existed:
+            logger.info(
+                "The doc for command {} doesn't exist, "
+                "creating from the template.".format(command)
+            )
+            with open(file_path, "w") as file:
+                file.write(TEMPLATE)
+
+        with open(file_path, "r") as file:
+            text = file.read()
+
+        if AUTOMATED_START_MARKER not in text:
+            logger.info(
+                "The marker for automated doc generation is not found in the doc, ignore."
+            )
+            continue
+
+        help_cmd = "pebble {} --help".format(command)
+
+        help_cmd_output = """\
+<!-- START AUTOMATED OUTPUT -->
+```{{terminal}}
+:input: {help_cmd}
+{stdout}
+```
+<!-- END AUTOMATED OUTPUT -->""".format(
+            help_cmd=help_cmd,
+            stdout=get_command_help_output(help_cmd).strip(),
+        )
+
+        text = render_code_block_cmd(text, help_cmd)
+        text = render_code_block_output(text, help_cmd_output)
+
+        if not file_existed:
+            text = render_title(text, command)
+            text = render_description(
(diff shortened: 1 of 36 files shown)


docs/generate_cmd_doc.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def update_toc(all_cmds: typing.List[str]):
+    index_page = "reference/cli-commands/cli-commands.md"
+    with open(index_page, "r") as file:
+        text = file.read()
+
+    start_index = text.find("```{toctree}")
+    end_index = text.find("```", start_index + 1) + 3
+    cmd_list = "\n".join(["{cmd} <{cmd}>".format(cmd=cmd) for cmd in all_cmds])
```
benhoyt, line 92: Can we use f-strings? You've sometimes used f-strings in this PR, sometimes `str.format`. Let's be consistent. You can also drop the extra `[]` brackets and use a generator expression directly in the `join`:

```python
cmd_list = '\n'.join(f'{cmd} <{cmd}>' for cmd in all_cmds)
```

docs/generate_cmd_doc.py
```diff
@@ -0,0 +1,182 @@
...
+<!-- END AUTOMATED OUTPUT -->""".format(
+        help_cmd=help_cmd,
+        stdout=get_command_help_output(help_cmd).strip(),
+    )
+
+    return help_cmd, help_cmd_output
+
+
+def process_command(cmd: str):
+    logger.info("Processing doc for command {}".format(cmd))
```
benhoyt, line 144: It's standard [when logging to use `%`-formatting](https://docs.python.org/3/howto/logging.html#logging-variable-data), as it `logger.info()` and similar functions lazily-format the arguments. Or you can use f-strings, but `%` is probably most standard.

```
logger.info('Processing doc for command %s', cmd)
```

docs/reference/cli-commands/cli-commands.md
```diff
@@ -4,11 +4,31 @@
 :titlesonly:
 :maxdepth: 1
 
-help <help>
-logs <logs>
-replan <replan>
 run <run>
+help <help>
```
benhoyt, line 8: Not a comment on this line, but can we add a one-paragraph into to this page? How `pebble` uses subcommands as `pebble foo`, and that this is the list of all commands, something like that?

docs/reference/cli-commands/help.md
```diff
@@ -1,5 +1,5 @@
 (reference_pebble_help_command)=
-# Pebble help command
+# help command
```
benhoyt, line 2: Should `help` have the regular usage block too, just to be consistent?

docs/tox.ini
```diff
@@ -32,3 +32,10 @@ commands_pre =
 commands =
     sphinx-autobuild . _build/html
 
+[testenv:commands]
+description = Generate reference docs for CLI commands
+deps =
+    ruff==0.5.4
```
benhoyt, line 38: It seems odd to me that the `commands` item also runs the Ruff formatter. Can we have a separate tox command for that? (But maybe we can add that in a separate PR and have it format all the *.py files?)

benhoyt (changes requested):
I've added a bunch of code review comments. The overall approach is good, but I think there are a few things we could structure a bit better.

A few comments from when I run this script with the branch checked out locally:

* I get a bunch of diffs and new files created. It seems like the version you've committed doesn't have the latest changes on master (for example, the identities work?).
* I also get an error running it: "Error running pebble update --help.". Any ideas what the "update" command is? Probably a parsing error from the `update-identities` command?

docs/reference/cli-commands/add.md
```diff
@@ -0,0 +1,22 @@
+(reference_pebble_add_command)=
+# add command
+
+The add command reads the plan's layer YAML from the path specified and
```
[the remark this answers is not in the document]
IronCore864: Yes, the description is taken from the first paragraph below the usage in the help output.

Not all commands' short description fits in one line, I missed this and will fix it.

I did a comparison and found that the paragraph after the usage in the output is better than the short description in the `pebble help --all` command because for some commands it gives more useful information.

docs/generate_cmd_doc.py
```diff
@@ -0,0 +1,182 @@
...
+"""
+
+
+def get_all_pebble_commands() -> typing.List[str]:
+    cmd = "pebble help --all"
+    p = subprocess.run(cmd.split(), text=True, capture_output=True)
+    if p.returncode != 0:
+        logger.error("Error running {}.".format(cmd))
+        exit(1)
+    return re.findall(r"\n\s{4}(\w+)", p.stdout)
```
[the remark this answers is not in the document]
IronCore864: See the reply in the other comment.

docs/generate_cmd_doc.py
```diff
@@ -0,0 +1,182 @@
...
+
+def get_command_help_output(cmd: str) -> str:
+    p = subprocess.run(cmd.split(), text=True, capture_output=True)
+    if p.returncode != 0:
+        logger.error("Error running {}.".format(cmd))
+        exit(1)
+    return p.stdout
+
+
+def get_description_from_output(text: str) -> str:
```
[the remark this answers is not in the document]
IronCore864: See the reply in the other comment, will use the command output's first paragraph since it provides more information.

IronCore864:
Lastest preview:

https://canonical-pebble--458.com.readthedocs.build/en/458/reference/cli-commands/cli-commands/

Changes:

- Added a paragraph on the index page.
- Reordered commands alphabetically.
- Get descriptions from `pebble help --all` instead.
- Regenerate all docs.
- And some other minor tweaks according to the review.

docs/generate_command_docs.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def get_all_commands() -> typing.List[typing.Tuple[str, str]]:
+    process = subprocess.run(
+        ["go", "run", "../cmd/pebble", "help", "--all"],
+        text=True,
+        capture_output=True,
+        check=True,
+    )
+    pattern = r"^\s{4}(\w+(?:-\w+)?)(?:\s{2,}|$)([\w\s\'\-\.\n]+?)(?=\n\s{4}|$)"
```
benhoyt, line 39: To me this regex seems almost unreadable / unmaintainable. It seems like we're processing such a simple thing here: find the first word and the rest of the line for all lines that start with 4 spaces. Can we do this:

```python
return sorted(line.split(maxsplit=1) for line in process.stdout.splitlines()
              if line.startswith('    '))
```

Avoiding a fancy regex here is simpler and much more readable.

Also, for sorting lists of lists or lists of tuples on the first element, you don't need `key=lambda x: x[0]`, because when the items being sorted are lists or tuples, Python automatically compares the first element first, then the second, and so on.

docs/generate_command_docs.py
```diff
@@ -0,0 +1,148 @@
+"""This module generate reference docs for pebble CLI commands."""
```
benhoyt, line 1:
```suggestion
"""This module generates reference docs for pebble CLI commands."""
```

docs/reference/cli-commands/help.md
```diff
@@ -8,18 +8,47 @@ information about available `pebble` commands.
...
+
+         Run: run
+        Info: help, version
+        Plan: add, plan
+    Services: services, logs, start, restart, signal, stop, replan
+      Checks: checks, health
+       Files: push, pull, ls, mkdir, rm, exec
+     Changes: changes, tasks
+     Notices: warnings, okay, notices, notice, notify
+  Identities: identities --help
```
benhoyt, line 28: Can `help.md` be generated with automatic code blocks too? Otherwise this will have to be updated whenever we add a command, which we might forget. We might need a special case in the script, as you need to type `pebble help` or `pebble --help` to get it, rather than the usual pattern of `pebble help --help`.

docs/reference/cli-commands/run.md
```diff
@@ -26,64 +26,6 @@ pebble run --args myservice --port 8080 \; --hold
           --http=        Start HTTP API listening on this address (e.g., ":4000")
       -v, --verbose      Log all output from services to stdout
           --args=        Provide additional arguments to a service
+          --identities=  Seed identities from file (like update-identities --replace)
 ```
 <!-- END AUTOMATED OUTPUT -->
-
-## How it works
```
benhoyt, line 29: Was this content supposed to be deleted?

benhoyt (changes requested):

docs/reference/cli-commands/help.md
IronCore864: Yes, that's why I left it out. I'll add some logic in the script to unify it.

docs/reference/cli-commands/run.md
IronCore864: Might be a bug in the script, let me double check.

IronCore864:
Just did one more round of refactoring according to review, now it's ready for @AnneCYH's final review.

benhoyt (approved):

AnneCYH:
hey @IronCore864 , looks good.

Just some points that may need to be considered:
1. Does `tox -e command` need to be executed as part of the GitHub workflow action? If docs are updated when commands are updated / added then this isn't a concern.
2. Making sure entities that are mentioned in commands are also defined in a corresponding "parent" reference page. E.g. what is an identity in Pebble?

Other than that no major concerns, thanks!

- Anne -

IronCore864:
> * Does `tox -e command` need to be executed as part of the GitHub workflow action? If docs are updated when commands are updated / added then this isn't a concern.

We've resolved the concerns elsewhere in a meeting.

Merging now.

PR #460, opened 2024-07-29
feat(plan): do not allocate unused plan maps

flotter:
The Pebble plan library currently allocates a new layer in two places:

1. ```ParseLayer(order int, label string, data []byte) (*Layer, error)```
2. ```CombineLayers(layers ...*Layer) (*Layer, error)```

The current approach is to always allocate the plan section (```Services```, ```Checks``` or ```LogTargets```) maps of the layer first, irrespective if the section might be empty or not:

```go
func ParseLayer(order int, label string, data []byte) (*Layer, error) {
    layer := Layer{
        Services:   map[string]*Service{},
        Checks:     map[string]*Check{},
        LogTargets: map[string]*LogTarget{},
    }
    :
```

Since nil maps in Golang supports indexing and ```len()``` we can leave unused maps nil to save a little bit of memory in the process and remove the need for unit test code to instantiate empty maps, although this is not the main motivation.

**Proposed Change:**

1. When combining, only allocate sections once its clear the source layer(s) have a matching section type.
2. When Parsing a YAML, let the unmarshal process allocate when needed for non-empty sections.

**Why:**

While working on plan schema extension support, the current philosophy to always allocate empty sections creates an additional interface method requirement where the plan library has to ask the schema extension for an empty section (concrete backing type), beyond using the ```CombineSections``` and ```ParseSection``` interface methods of the extension.

In order to promote a consistent consumer experience, I would like to create a consistent understanding that your plan section will be ```nil``` if no configuration is present.

**Impact:**

Since none of the current plan consumers seem to make assumptions about a specific top level section map entry, the code already checks for the existence of an entry, or iterate over the entries. Both cases are supported with ```nil``` maps in Go.

### DIFF
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -549,17 +549,20 @@ func (e *FormatError) Error() string {
 	return e.Message
 }
 
+func makeMapIfNil[K comparable, V any](m map[K]V) map[K]V {
+	if m == nil {
+		m = make(map[K]V)
+	}
+	return m
+}
+
 // CombineLayers combines the given layers into a single layer, with the later
 // layers overriding earlier ones.
 // Neither the individual layers nor the combined layer are validated here - the
 // caller should have validated the individual layers prior to calling, and
 // validate the combined output if required.
 func CombineLayers(layers ...*Layer) (*Layer, error) {
-	combined := &Layer{
-		Services:   make(map[string]*Service),
-		Checks:     make(map[string]*Check),
-		LogTargets: make(map[string]*LogTarget),
-	}
+	combined := &Layer{}
 	if len(layers) == 0 {
 		return combined, nil
 	}
@@ -568,6 +571,7 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 	combined.Description = last.Description
 	for _, layer := range layers {
 		for name, service := range layer.Services {
+			combined.Services = makeMapIfNil(combined.Services)
 			switch service.Override {
 			case MergeOverride:
 				if old, ok := combined.Services[name]; ok {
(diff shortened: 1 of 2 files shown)


internals/plan/plan.go
```diff
@@ -549,17 +549,22 @@ func (e *FormatError) Error() string {
 	return e.Message
 }
 
+// sectionAllocateOnce only allocates a section if it is nil, allowing on
+// demand allocation of sections only if it is present in the configuration.
+func sectionAllocateOnce[K comparable, V any](section map[K]V) map[K]V {
```
benhoyt, line 554: The function name here is awkward, and more "how it may be used" rather than "what it does". Perhaps call it `makeMapIfNil`? Then the arg should be named something more generic too, maybe `m`, and the doc comment can probably just go away?

benhoyt (commented):
I think this idea and the motivation is fine, thanks. Just one minor quibble about naming.

internals/plan/plan.go
flotter: Made naming changes as proposed. Resolving.

flotter:
@benhoyt I will leave this PR in draft state until early next week when I will sync up with @niemeyer.

I would like to get his input on the plan extension side (and from you after that), to make sure the ```nil``` map approach (if config is not present) is the preferred model on the extension side. If not, then this PR will fall away, and we would continue the empty map strategy.

flotter:
Following discussion with @niemeyer I will close this for now. This strategy is limited to maps and slices, and if the top level item is a struct, handling becomes awkward. I am currently reworking the plan extensions to revert back to the empty instance approach.

PR #466, opened 2024-08-02
fix(overlord): send EOF to master pty in interactive mode

### LINKED ISSUE #306: Pebble exec, when in enforced interactive mode, is prone to losing some of the command output
If one enforces the pebble exec interactive mode when executing a command and stdin returns EOF sooner than reading from the commands stdout ends, it may result in some output lost. Consider:

```
> pebble exec -i -t -- tee < sometextfile
> pebble exec -i -t -- ls </dev/null
```

In both cases, the command's output will either be nonexistent or partial. This is likely due to the client closing the websocket connection with the server (`/v1/tasks/<task-id>/websocket/stdio`) before the command's stdout is fully read. In the interactive mode, `stdio` websocket is used to redirect both stdin and stdout, hence, if the client's stdin go routine sends the end message to the websocket server it affects the routine reading stdout (see the line 68): https://github.com/canonical/pebble/blob/613429877dfafaa891f8d57fd4b7aee6e5a3d494/internals/wsutil/wsutil.go#L54-L70

If sending the end message is delayed until the stdout routine reads all the produced output, it resolves the issue.

dmitry-lyfar:
If the interactive mode is enforced in `pebble exec`, there are scenarios like:

```go run ./cmd/pebble exec -i -- tee < ./HACKING.md```

that would prompt the server to shutdown the execution earlier than the whole output from 'tee' will be read and sent to the client. Hence, instead of closing the master pty descriptor when there is no more input from the client, send
Ctrl-D (EOF) to indicate to the process that there will be no more input.

Fixes #306.

### DIFF
--- a/internals/overlord/cmdstate/handlers.go
+++ b/internals/overlord/cmdstate/handlers.go
@@ -252,7 +252,12 @@ func (e *execution) do(ctx context.Context, task *state.Task) error {
 			// websocket and write to the PTY.
 			go func() {
 				<-wsutil.WebsocketRecvStream(master, ioConn)
-				master.Close()
+				// If the interactive is enforced, it is possible to finish
+				// reading earlier than the mirroring go routine sends all the
+				// output to the client. Thus, closing the master descriptor
+				// here will terminate mirroring prematurely. Instead, we
+				// should send Ctrl-D to the fd to indicate the end of input.
+				master.Write([]byte{byte(unix.VEOF)})
 			}()
 		} else {
 			// Non-interactive: start goroutine to receive stdin from "stdio"
--- a/internals/wsutil/wsutil_linux.go
+++ b/internals/wsutil/wsutil_linux.go
@@ -32,7 +32,7 @@ func MirrorToWebsocket(conn MessageWriter, r io.ReadCloser, exited chan struct{}
 	for {
 		buf, ok := <-in
 		if !ok {
-			r.Close()
+			_ = r.Close()
 			logger.Debugf("Sending write barrier")
 			err := conn.WriteMessage(websocket.TextMessage, endCommandJSON)
 			if err != nil {

### CODE REVIEW

**Review by IronCore864 — APPROVED** [review]

File: internals/overlord/cmdstate/handlers.go [hunk 1]
```diff
@@ -252,7 +252,12 @@ func (e *execution) do(ctx context.Context, task *state.Task) error {
 			// websocket and write to the PTY.
 			go func() {
 				<-wsutil.WebsocketRecvStream(master, ioConn)
-				master.Close()
+				// If the interactive is enforced, it is possible to finish
+				// reading earlier than the mirroring go routine sends all the
+				// output to the client. Thus, closing the master descriptor
+				// here will terminate mirroring prematurely. Instead, we
+				// should send Ctrl-D to the fd to indicate the end of input.
+				master.Write([]byte{byte(unix.VEOF)})
```
dimaqq, line 260: The fix makes sense, see e.g. https://stackoverflow.com/a/18366397/705086

I do wonder about some corner cases, e.g.:

- `pebble -i yada yada < somefile` where input is not from a terminal, but the output is
- all of stdin/stdout/stderr are redirected, and yet `-i` is supplied

Please consider how to test these -- unit tests? functional? one-off manual tests? I'm not sure what would be more appropriate.

dimaqq (approved):

internals/overlord/cmdstate/handlers.go
benhoyt: Yeah, it's not ideal, but we currently don't have any automated tests for interactive mode. We have on our roadmap for @IronCore864 to write end-to-end tests for `pebble run`; maybe some `pebble exec -i` tests can be included in that work?

In any case, I've done some pretty extensive manual testing of all the various exec modes, and this fix seems to work nicely. I've also tested that the number of file descriptors isn't increasing (leaking) by watching `/bin/sh -c 'ls /proc/{pebble_pid}/fd/ | wc -l'`, and it's stable (stays at 10).

Tiexin has also done manual testing on this PR.

benhoyt:
This does seem like the right fix -- thanks very much, @dmitry-lyfar -- really appreciated! As mentioned above, both I and Tiexin have done fairly extensive manual testing on this PR, and we think it's good to go. So I'm going to merge this now.

dmitry-lyfar:
@benhoyt @IronCore864 @dimaqq amazing, thanks for looking into and testing this!

Issue #464, opened 2024-08-01
Improve the tutorial doc


IronCore864:
Update the title to "Get started with Pebble". Currently, the title is "Getting started", which is a bit more 'vague'; is the title talking about WHETHER you should get started with Pebble? See the advice on [naming how-to guides](https://diataxis.fr/how-to-guides/#pay-attention-to-naming).

A checklist for minor improvements:

- It starts with a clear description of what the learner can expect to achieve from the tutorial
- The language of the tutorial assures the user that the instructor is on the same journey (“we”, not just “you”)
- The tutorial is reproducible with a mock setup which can easily be duplicated for the target user (a mock HTTP server instead of unknown services)
- It focuses on a single installation method and does not provide alternatives to users
- Tutorial is broken down into multiple smaller, logical steps and (where necessary) guides the user to observe the expected outcome at each step
- The amount of explanation is minimal and generally limited to necessary topics for the learner; links are used to connect the user to further reading on related topics
- Maybe having Python installed (though seemingly basic) might be a good Prerequisite to spell out early on so that users don't fail halfway through the tutorial because they can't run http.serve

PR #471, opened 2024-08-05
docs: update tutorial according to review

IronCore864:
Minor changes to the tutorial according to Anne's suggestion.

Closes https://github.com/canonical/pebble/issues/464.

This PR makes sure:

- [x] It starts with a clear description of what the learner can expect to achieve from the tutorial
- [x] The language of the tutorial assures the user that the instructor is on the same journey (“we”, not just “you”)
- [x] The tutorial is reproducible with a mock setup which can easily be duplicated for the target user (a mock HTTP server instead of unknown services)
- [x] It focuses on a single installation method and does not provide alternatives to users
- [x] Tutorial is broken down into multiple smaller, logical steps and (where necessary) guides the user to observe the expected outcome at each step
- [x] The amount of explanation is minimal and generally limited to necessary topics for the learner; links are used to connect the user to further reading on related topics
- [x] Listing Python as a prerequisite to spell out early on so that users don't fail halfway through the tutorial because they can't run `http.serve`

### DIFF
--- a/docs/tutorial/getting-started.md
+++ b/docs/tutorial/getting-started.md
@@ -7,7 +7,7 @@ After this tutorial, you will have a basic understanding of what Pebble is and h
 ## Prerequisites
 
 - A Linux machine.
-- Python3 (used to run basic HTTP servers as sample services managed by Pebble).
+- Python 3.x (used to run basic HTTP servers as sample services managed by Pebble).
 
 ## Download and install Pebble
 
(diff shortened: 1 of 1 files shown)


docs/tutorial/getting-started.md
```diff
@@ -7,6 +7,7 @@ After this tutorial, you will have a basic understanding of what Pebble is and h
 ## Prerequisites
 
 - A Linux machine.
+- Python3 (used to run basic HTTP servers as sample services managed by Pebble).
```
benhoyt, line 10:
```suggestion
- Python 3.x (used to run basic HTTP servers as sample services managed by Pebble).
```

benhoyt (approved):
Looks good. One minor comment.

PR #474, opened 2024-08-05
feat(planstate): make plan propagation safer

flotter:
The original Pebble plan was part of the service manager, only concerned with services. The plan was lazy loaded, so the first service manager public method called (internally or over the HTTP API) would result in plan layers getting loaded, combined and propagated. This scheme was later extended to support additional managers, which also added the ```PlanChanged``` subscription mechanism. Subscribed managers would receive updates of the plan on the initial load and subsequent runtime changes, due to layers ```add```ed using the HTTP API.

[The plan manager](https://github.com/canonical/pebble/pull/387) decoupled the Pebble Plan from the service manager. The first version of the plan manager simply loaded the plan from disk as early as possible, before the HTTP API endpoints were activated. However, since the load also triggered a ```PlanChanged``` update to all subscribed managers, these managers received an update of the plan before the State Engine was running. This resulted in [extra code](https://github.com/canonical/pebble/blob/1f50b018ce2a34fb60a442c2a1cdfca70c16f9cd/internals/overlord/checkstate/manager.go#L167) required in ```checkstate``` to defer calling ```Ensure()``` inside the ```PlanChanged``` callback.

The following changes are proposed:

1. Adapt the planstate load and plan changed notifications to only use standard State Engine events ```StartUp()``` and ```Ensure()``` during startup. Note that because we now guarantee a call to Ensure before the HTTP API is enabled, it is safe to do the initial load and propagate lock-less (care has to be taken so external callers do to call us before the state engine is running).

2. Make a change to the overlord to only enable the HTTP API endpoints once both ```StartUp``` and at least one ```Ensure()``` pass was completed. Note that this does not impact startup performance of managers, but would ensure external HTTP API calls, and self-calls (i.e. autostart) would only happen after ```StartUp``` and at least one ```Ensure()``` completed, giving managers a guarantee on startup behaviour. 

Performing a simple startup duration test from the application entrypoint until HTTP API is enabled resulted in no measurable difference.

### DIFF
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -199,16 +199,6 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 		return err
 	}
 
-	if rcmd.Args != nil {
-		mappedArgs, err := convertArgs(rcmd.Args)
-		if err != nil {
-			return err
-		}
-		if err := d.SetServiceArgs(mappedArgs); err != nil {
-			return err
-		}
-	}
-
 	// Run sanity check now, if anything goes wrong with the
 	// check we go into "degraded" mode where we always report
 	// the given error to any client.
@@ -225,6 +215,20 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 	d.Version = cmd.Version
 	if err := d.Start(); err != nil {
 		return err
+	}
+
+	// When we reach this point, we know every overlord manager has received a
+	// state engine StartUp() and at least one Ensure() call. This also guarantees
+	// the plan has been loaded and propagated to every change listener.
+
+	if rcmd.Args != nil {
+		mappedArgs, err := convertArgs(rcmd.Args)
+		if err != nil {
+			return err
+		}
+		if err := d.SetServiceArgs(mappedArgs); err != nil {
+			return err
+		}
 	}
 
 	watchdog, err := runWatchdog(d)
--- a/internals/overlord/checkstate/manager.go
+++ b/internals/overlord/checkstate/manager.go
@@ -20,7 +20,6 @@ import (
 	"reflect"
 	"sort"
 	"sync"
-	"sync/atomic"
 
 	"gopkg.in/tomb.v2"
 
@@ -38,8 +37,7 @@ const (
 
 // CheckManager starts and manages the health checks.
 type CheckManager struct {
-	state      *state.State
-	ensureDone atomic.Bool
+	state *state.State
 
 	failureHandlers []FailureFunc
 
@@ -87,7 +85,6 @@ func NewManager(s *state.State, runner *state.TaskRunner) *CheckManager {
 }
 
 func (m *CheckManager) Ensure() error {
-	m.ensureDone.Store(true)
 	return nil
 }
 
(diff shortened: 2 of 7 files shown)


internals/cli/cmd_run.go
```diff
@@ -227,6 +217,20 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
...
+	// When we reach this point, we know every overlord manager has received a
+	// state engine StartUp() and at least one Ensure() call. This also guarantees
+	// the plan has been loaded and propagated to every change listener.
+
+	if rcmd.Args != nil {
+		mappedArgs, err := convertArgs(rcmd.Args)
+		if err != nil {
+			return err
+		}
+		if err := d.SetServiceArgs(mappedArgs); err != nil {
```
flotter, line 229: ```SetServiceArgs``` requires the service manager plan so we have to move it down slightly.

internals/overlord/overlord.go
```diff
@@ -379,15 +371,34 @@ func (o *Overlord) Loop() {
 			}
 		}
 	})
+	o.ensureWaitRun()
```
flotter, line 374: I am also discussing this proposal with @pedronis in a separate thread, to make sure this is compatible with snapd.

internals/overlord/overlord.go
```diff
@@ -89,7 +88,7 @@ type Overlord struct {
 	ensureLock  sync.Mutex
 	ensureTimer *time.Timer
 	ensureNext  time.Time
-	ensureRun   int32
+	ensureRun   chan struct{}
```
flotter, line 91: I am extending an existing feature here to also allow for waiting, not only checking if ensure ran already.

flotter:
@benhoyt and @hpidcock please could you give me some feedback on this proposal when you next have some time, thank you!

internals/overlord/checkstate/manager.go
```diff
@@ -38,8 +37,7 @@ const (
 
 // CheckManager starts and manages the health checks.
 type CheckManager struct {
-	state      *state.State
-	ensureDone atomic.Bool
```
flotter, line 42: This logic, which would also have been needed in derived project managers, can now safely be removed. The race is resolved in the overlord.

internals/overlord/overlord.go
```diff
@@ -379,15 +371,34 @@ func (o *Overlord) Loop() {
...
+	default:
+		close(o.ensureRun)
+	}
 }
 
-func (o *Overlord) CanStandby() bool {
-	run := atomic.LoadInt32(&o.ensureRun)
-	return run != 0
+// ensureWaitRun waits until StateEngine.Ensure() was called at least once.
+func (o *Overlord) ensureWaitRun() {
```
benhoyt, line 387: This function waits, so it would make more sense if it were called `waitEnsureRun`, that is, "wait till ensure has run". The name `ensureWaitRun` sounds more like "ensure that the wait has run".

internals/overlord/overlord.go
```diff
@@ -379,15 +371,34 @@ func (o *Overlord) Loop() {
...
+}
+
+func (o *Overlord) CanStandby() (ensured bool) {
+	select {
+	case <-o.ensureRun:
+		// Already closed. Ensure already ran at least once.
+		ensured = true
+	default:
+	}
+	return ensured
```
benhoyt, line 401: The named return seems confusing here. What about:
```suggestion
func (o *Overlord) CanStandby() bool {
	select {
	case <-o.ensureRun:
		// Already closed. Ensure already ran at least once.
		return true
	default:
		return false
	}
```

benhoyt (commented):
This looks good to me. Just a couple of style comments.

internals/overlord/overlord.go
flotter: Improved as proposed. Resolving.

internals/overlord/overlord.go
flotter: Improved as per proposal, thank you. Resolving.

flotter:
Following super helpful conversations with @pedronis and @niemeyer I realised I am solving this the wrong way. Closing this PR - will create a new PR shortly.

PR #477, opened 2024-08-08
chore: minor improvement to docs generation

IronCore864:
Minor fix to the automated doc generation.

Changes:

- When getting command help output, use `go run ./cmd/pebble`.
- GitHub Actions job to check if the automated CLI reference docs have been changed.
- Regenerate CLI command reference doc to fix `stty cols 80` issue.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -87,3 +87,5 @@ jobs:
       run: |
         python3 generate_command_docs.py
         git diff --exit-code
+      working-directory: ./docs
+
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,7 +74,7 @@ jobs:
   docs:
     runs-on: ubuntu-latest
 
-    name: Format check
+    name: Automated docs check
     steps:
     - uses: actions/checkout@v3
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -81,11 +81,13 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.10' 
+        python-version: '3.12' 
+
+    - name: Install tox
+      run: pip install tox~=4.17
 
     - name: Ensure no CLI reference doc changes
       run: |
-        python3 generate_command_docs.py
+        tox -e commands
         git diff --exit-code
       working-directory: ./docs
-
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,3 +70,20 @@ jobs:
       run: |
         go fmt ./...
         git diff --exit-code
+
+  docs:
+    runs-on: ubuntu-latest
+
+    name: Format check
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -70,3 +70,22 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    name: Automated docs check
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
```
benhoyt, line 84: Why not the latest, 3.12?

.github/workflows/tests.yml
```diff
@@ -70,3 +70,22 @@ jobs:
...
+    - uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10' 
+
+    - name: Ensure no CLI reference doc changes
+      run: |
+        python3 generate_command_docs.py
```
benhoyt, line 88: I guess we should probably make this `tox -e commands` (and install tox like we do in Ops framework_tests) in case the script starts to grow dependencies.

docs/generate_command_docs.py
```diff
@@ -102,7 +102,8 @@ def create_file_if_not_exist(filepath: str, cmd: str) -> bool:
 def generate_help_command_and_output(cmd: str) -> typing.Tuple[str, str]:
     help_cmd = ["pebble", "help"] if cmd == "help" else ["pebble", cmd, "--help"]
     help_cmd_str = " ".join(help_cmd)
-    help_cmd_output = get_command_help_output(help_cmd_str).strip()
+    go_run_cmd = " ".join(["go", "run", "../cmd/pebble"] + help_cmd[1:])
+    help_cmd_output = get_command_help_output(go_run_cmd).strip()
```
benhoyt, line 106: It's weird to create `help_cmd` with "pebble" at the start, only to remove it and replace it here. Can we do something like:

```python
args = ["help"] if cmd == "help" else [cmd, "--help"]
help_cmd_str = " ".join(["pebble"] + args)
go_run_cmd = " ".join(["go", "run", "../cmd/pebble"] + args)
```

benhoyt (changes requested):
Thanks for this. Couple of comments. I also don't fully understand why the columns thing is now fixed.

IronCore864:
Add some notes for the issue of the column in case we need to reinvestigate it in the future:

- My env is multipass.
- After abundant testing, `stty cols` works steadily if no terminal-resizing is done during the period the script is running. The behaviour is consistent regardless of the terminal width.
- Broken output can be reproduced steadily if the terminal width is resized while the script runs.

.github/workflows/tests.yml
IronCore864: Upgraded to python 3.12 and the latest version of tox (4.17). Triggering CI test now.

IronCore864:
Upgraded to python 3.12 and the latest version of tox (4.17). Triggered the CI test and it passed: https://github.com/canonical/pebble/actions/runs/10313091679/job/28549325132?pr=477.

benhoyt (approved):

PR #478, opened 2024-08-09
fix(cmdstate): don't store execSetup and environment in state

### LINKED ISSUE #411: Every exec stores several KB of JSON in state (due to environment)
I just ran a `pebble exec` of a simple command (`echo foo`) locally, and because it stores all the command arguments in the task data, it expanded to over 4KB of JSON data in state -- mostly due to the environment which we now (since #234) copy `os.Environ` as a base. It would be good to find a way to avoid storing this much in state, for example, can we expand `os.Environ` lazily, later in the game?

benhoyt:
Per #411, the environment (and other task data) can get quite large for exec objects. Given that exec commands are one-shot and only relevant for the current run of Pebble, there's no need to persist them.

Note that none of this data is in "api-data" so it's not accessible via the API at all right now, and so this is a non-breaking change.

Later if we want to make exec tasks a bit more introspectable we can add the command line or other fields to "api-data" (but we shouldn't add the entire environment).

Fixes #411.

### DIFF
--- a/internals/overlord/cmdstate/handlers.go
+++ b/internals/overlord/cmdstate/handlers.go
@@ -70,13 +70,13 @@ type execution struct {
 }
 
 func (m *CommandManager) doExec(task *state.Task, tomb *tomb.Tomb) error {
-	var setup execSetup
 	st := task.State()
 	st.Lock()
-	err := task.Get("exec-setup", &setup)
+	setupObj := st.Cached(execSetupKey{task.ID()})
 	st.Unlock()
-	if err != nil {
-		return fmt.Errorf("cannot get exec setup object for task %q: %v", task.ID(), err)
+	setup, ok := setupObj.(*execSetup)
+	if !ok || setup == nil {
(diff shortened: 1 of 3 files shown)


dmitry-lyfar (approved):

dmitry-lyfar:
A very naive comparison does see the difference, it seems. Every run was for a clean, empty state.

```
#!/usr/bin/bash

for i in {1..50}; do
  pebble exec -- echo "perftest" > /dev/null &
done

wait
echo "Done."
```

Before:
`bash ./naive.sh > /dev/null  0.13s user 0.21s system 43% cpu 0.780 total`
After:
`bash ./naive.sh > /dev/null  0.13s user 0.19s system 105% cpu 0.311 total`


After growing the number of simultaneous execs to 500, the difference becomes quite dramatic as the old version that keeps the exec setup in the state file gives up and starts to fail requests with a timeout. Given that the state is clean for every test as with repetitive runs a growing state file degrades the performance seriously.

benhoyt:
Thanks for the review and the performance numbers, @dmitry-lyfar!

PR #479, opened 2024-08-12
docs: add Identities documentation

benhoyt:
This adds docs for Identities:

* Access control information under "API and clients" - [preview](https://canonical-pebble--479.com.readthedocs.build/en/479/explanation/api-and-clients/)
* A "how to manage identities" guide - [preview](https://canonical-pebble--479.com.readthedocs.build/en/479/how-to/manage-identities/)
* And an "Identities" reference page to define/describe the concept - [preview](https://canonical-pebble--479.com.readthedocs.build/en/479/reference/identities/)

### DIFF
--- a/docs/explanation/api-and-clients.md
+++ b/docs/explanation/api-and-clients.md
@@ -1,6 +1,8 @@
 # API and clients
 
-The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers the plan, and so on.
+The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers to the plan, and so on.
+
+If `pebble run` is started with the `--http <address>` option, Pebble also allows access to "untrusted" HTTP endpoints using the given TCP address. See "access control and identities" below.
 
 There is currently no official documentation for the API at the HTTP level (apart from the [code itself](https://github.com/canonical/pebble/blob/master/internals/daemon/api.go)!); most users will interact with it via the Pebble command line interface or by using the Go or Python clients.
 
@@ -9,3 +11,28 @@ The Go client is used primarily by the CLI, but is importable and can be used by
 We try to never change the underlying HTTP API in a backwards-incompatible way, however, in rare cases we may change the Go client in a backwards-incompatible way.
 
 In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the [`ops` library](https://github.com/canonical/operator) used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
+
+
+## Access control and identities
+
+API endpoints fall into one of three access levels, from least restricted to most restricted:
+
+* `untrusted`: these are allowed from any user, even unauthenticated users using the HTTP-over-TCP listener. The only untrusted endpoints are `/v1/system-info` and `/v1/health`.
+* `read`: these are allowed from any authenticated user, regardless of access levey. They are usually read operations that use the HTTP `GET` method, such as listing services or viewing notices.
+* `admin`: these are only allowed from admin users. They are usually write or modify operations that use the HTTP `POST` method, such as adding a layer or starting a service.
+
+Pebble authenticates clients that connect to the socket API using peer credentials ([`SO_PEERCRED`](https://man7.org/linux/man-pages/man7/socket.7.html)) to determine the user ID (UID) of the connecting process. If this UID is 0 (root) or the UID of the Pebble daemon, the user's access level is `admin`, otherwise the access level is `read`.
+
+If Pebble can't authenticate the user at all, or if it's an unauthenticated user connecting over TCP, the user is considered `untrusted`.
+
+In addition to the default access control, Pebble admins can also set up named [identities](../reference/identities.md) with specific access levels using the identities CLI commands (see [how to manage identities](../how-to/manage-identities.md)).
+
+Currently the only identity type is `local`, that is, a local user ID determined using peer credentials. An example admin identity named "bob" is shown below:
+
+```yaml
+identities:
+    bob:
+        access: admin
+        local:
+            user-id: 42
+```
--- a/docs/explanation/api-and-clients.md
+++ b/docs/explanation/api-and-clients.md
@@ -2,7 +2,7 @@
 
 The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers to the plan, and so on.
 
-If `pebble run` is started with the `--http <address>` option, Pebble also allows access to "untrusted" HTTP endpoints using the given TCP address. See "access control and identities" below.
+If `pebble run` is started with the `--http <address>` option, Pebble also allows access to "untrusted" HTTP endpoints using the given TCP address (see {ref}`api-access-levels` below).
 
 There is currently no official documentation for the API at the HTTP level (apart from the [code itself](https://github.com/canonical/pebble/blob/master/internals/daemon/api.go)!); most users will interact with it via the Pebble command line interface or by using the Go or Python clients.
 
@@ -13,21 +13,27 @@ We try to never change the underlying HTTP API in a backwards-incompatible way, 
 In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the [`ops` library](https://github.com/canonical/operator) used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
 
 
-## Access control and identities
+(api-access-levels)=
+## API access levels
 
 API endpoints fall into one of three access levels, from least restricted to most restricted:
 
 * `untrusted`: these are allowed from any user, even unauthenticated users using the HTTP-over-TCP listener. The only untrusted endpoints are `/v1/system-info` and `/v1/health`.
-* `read`: these are allowed from any authenticated user, regardless of access levey. They are usually read operations that use the HTTP `GET` method, such as listing services or viewing notices.
+* `read`: these are allowed from any authenticated user, regardless of access level. They are usually read operations that use the HTTP `GET` method, such as listing services or viewing notices.
 * `admin`: these are only allowed from admin users. They are usually write or modify operations that use the HTTP `POST` method, such as adding a layer or starting a service.
 
 Pebble authenticates clients that connect to the socket API using peer credentials ([`SO_PEERCRED`](https://man7.org/linux/man-pages/man7/socket.7.html)) to determine the user ID (UID) of the connecting process. If this UID is 0 (root) or the UID of the Pebble daemon, the user's access level is `admin`, otherwise the access level is `read`.
 
 If Pebble can't authenticate the user at all, or if it's an unauthenticated user connecting over TCP, the user is considered `untrusted`.
 
-In addition to the default access control, Pebble admins can also set up named [identities](../reference/identities.md) with specific access levels using the identities CLI commands (see [how to manage identities](../how-to/manage-identities.md)).
 
-Currently the only identity type is `local`, that is, a local user ID determined using peer credentials. An example admin identity named "bob" is shown below:
+## Controlling API access using identities
+
+In addition to the default access control, Pebble admins can also set up named [identities](../reference/identities.md) with specific access levels using the identities CLI commands.
+
+Each identity has a name, an access level, and an authentication type with type-specific configuration.
+
+Currently the only authentication type is `local`, that is, a local user ID determined using peer credentials. An example admin identity named "bob" is shown below:
 
 ```yaml
 identities:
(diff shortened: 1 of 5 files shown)


docs/explanation/api-and-clients.md
```diff
@@ -9,3 +11,28 @@ The Go client is used primarily by the CLI, but is importable and can be used by
...
 
 In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the [`ops` library](https://github.com/canonical/operator) used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
+
+
+## Access control and identities
+
+API endpoints fall into one of three access levels, from least restricted to most restricted:
+
+* `untrusted`: these are allowed from any user, even unauthenticated users using the HTTP-over-TCP listener. The only untrusted endpoints are `/v1/system-info` and `/v1/health`.
+* `read`: these are allowed from any authenticated user, regardless of access levey. They are usually read operations that use the HTTP `GET` method, such as listing services or viewing notices.
```
IronCore864, line 21:
```suggestion
* `read`: these are allowed from any authenticated user, regardless of access level. They are usually read operations that use the HTTP `GET` method, such as listing services or viewing notices.
```

docs/explanation/api-and-clients.md
```diff
@@ -9,3 +11,28 @@ The Go client is used primarily by the CLI, but is importable and can be used by
...
+* `admin`: these are only allowed from admin users. They are usually write or modify operations that use the HTTP `POST` method, such as adding a layer or starting a service.
+
+Pebble authenticates clients that connect to the socket API using peer credentials ([`SO_PEERCRED`](https://man7.org/linux/man-pages/man7/socket.7.html)) to determine the user ID (UID) of the connecting process. If this UID is 0 (root) or the UID of the Pebble daemon, the user's access level is `admin`, otherwise the access level is `read`.
+
+If Pebble can't authenticate the user at all, or if it's an unauthenticated user connecting over TCP, the user is considered `untrusted`.
+
+In addition to the default access control, Pebble admins can also set up named [identities](../reference/identities.md) with specific access levels using the identities CLI commands (see [how to manage identities](../how-to/manage-identities.md)).
+
+Currently the only identity type is `local`, that is, a local user ID determined using peer credentials. An example admin identity named "bob" is shown below:
```
IronCore864, line 30: Here I find the introduction of identity "type" a little abrupt, it was not introduced before, and then there is a sudden mention of identity, followed by an example.

I don't know how to make this better, but maybe move the sentence "Each identity has a name, an access level, and an authentication type with type-specific configuration." from the how-to doc here? Like:


```yaml
Each identity has a name, an access level, and an authentication type with type-specific configuration. An example of an identity named "bob" with admin access level and local type is shown below:

...

Currently the only identity type is `local`, that is, a local user ID determined using peer credentials.
```

Or, maybe we can have two subsections under `## Access control and identities`, one is `### Access level` and the other is `### Identity type`. Although this might be an overkill since the only identity type is local...

docs/how-to/manage-identities.md
```diff
@@ -0,0 +1,127 @@
+
+# How to manage identities
+
+Pebble uses named [identities](../reference/identities.md) to extend access to its [API](../explanation/api-and-clients.md).
+
+By default any UID (user ID) connected to the API socket is a `read` user; UID 0 (root) or the UID of the Pebble server process gives `admin` access.
+
+To extend this to additional users, add named identities using the [`add-identities`](../reference/cli-commands/add-identities.md) command with a YAML file configuring the details. For example, to add a new admin "bob" with UID 42 and a new read user "alice" with UID 2000, prepare this file:
```
IronCore864, line 8: I think it's better to paint a picture here, of what problem we are trying to solve here. Why the users might want to extend this to additional users, why not run the Pebble daemon and API with the same user, why not run API with the root user, etc?

A picture in my mind could be:

Imagine you have a cross-functioning team working on the same product on the same range of servers, Ops might need to deploy stuff so they need admin access, and developers might need read access to monitor and debug services. You could allow everybody to SSH using the same user with which the daemon is running, but sharing accounts is not a good practice. You could also let everybody run API with root to bypass the permission issue, but this violates the principle of least privilege in security. To solve this problem, you can use Pebble's identity and access management feature.

docs/how-to/manage-identities.md
```diff
@@ -0,0 +1,127 @@
...
+To show a single identity in YAML format, use the `identity` command:
+
+```{terminal}
+   :input: pebble identity alice
+access: read
+local:
+    user-id: 2000
+```
+
+## Seeding initial identities
```
IronCore864, line 119: This is the only subsection in this how-to doc, it feels not so good. Maybe we can create more sections, for example:

```yaml
# How to manage identities

painting the picture

## Create identities
### add-identities
### Seeding initial identities

## View identities
### list all
### list single
### output format

## Update identities
### update
### remove
### idempotency

## Read more

some more links
```

docs/explanation/api-and-clients.md
```diff
@@ -9,3 +11,28 @@ The Go client is used primarily by the CLI, but is importable and can be used by
 We try to never change the underlying HTTP API in a backwards-incompatible way, however, in rare cases we may change the Go client in a backwards-incompatible way.
 
 In addition to the Go client, there's also a [Python client](https://github.com/canonical/operator/blob/master/ops/pebble.py) for the Pebble API that's part of the [`ops` library](https://github.com/canonical/operator) used by Juju charms ([documentation here](https://juju.is/docs/sdk/interact-with-pebble)).
+
+
```
AnneCYH, line 15:
```suggestion
(explanation-api-access-control)=
```

Suggest adding an anchor here.

docs/explanation/api-and-clients.md
```diff
@@ -1,6 +1,8 @@
 # API and clients
 
-The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers the plan, and so on.
+The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers to the plan, and so on.
+
+If `pebble run` is started with the `--http <address>` option, Pebble also allows access to "untrusted" HTTP endpoints using the given TCP address. See "access control and identities" below.
```
AnneCYH, line 5:
```suggestion
If `pebble run` is started with the `--http <address>` option, Pebble also allows access to "untrusted" HTTP endpoints using the given TCP address. See {ref}`explanation-api-access-control` below.
```

docs/explanation/api-and-clients.md
AnneCYH: I agree the flow could be improved here.

I suggest something like this (this isn't perfect, open to suggestions).

## Access control and identities -> ## API access levels

API endpoints fall into...

If Pebble can't authenticate the user...

### Control API access by identity

In addition to...

docs/how-to/index.md
```diff
@@ -13,7 +14,8 @@ Installation follows a broadly similar pattern on all architectures, and you can
 Install Pebble <install-pebble>
 ```
 
-## Service Orchestration
+
+## Service orchestration
```
AnneCYH, line 18: Thanks for making these changes!

docs/reference/identities.md
```diff
@@ -0,0 +1,37 @@
+# Identities
+
+Pebble has the concept of named "identities", which lets administrators allow additional users to access the [API](../explanation/api-and-clients.md).
```
AnneCYH, line 3: Hi @benhoyt , @IronCore864 ,

From the current implementation it seems to make sense to use identities to give specific users elevated access to the APIs. But "allow additional users" doesn't seem so accurate since you can (theoretically) force an identity's access to "untrusted" right?

So maybe reword "allow additional users to access the API" to "manage users' access to the API"?

Thanks.

docs/reference/identities.md
```diff
@@ -0,0 +1,37 @@
...
+        #
+        # Currently the only suported authentication type is "local". Other
+        # types may be added in future, at which point you may configure an
+        # identity with one or more authentication types.
+        local:
+            # (Required) Peer credential UID.
+            user-id: <uid>
+```
+
+For example, a local identity named "bob" with UID 42 would be defined as follows:
```
AnneCYH, line 29: I would suggest being a bit more explicit here (since it's a reference and it's where you can describe the machinery / tools in Pebble.

```suggestion
For example, a local identity named "bob" with UID 42 that is granted `admin` level access would be defined as follows:
```

docs/explanation/api-and-clients.md
AnneCYH: And I would move the link to the "how to manage identities" towards the end of the section instead of early on.

docs/how-to/manage-identities.md
AnneCYH: Thanks @IronCore864 ! I agree it would be helpful to first provide a bit of context as to when you would want to use identities in Pebble, and how that would help your real-world scenario.

As to the current format, this is something to which I have a slightly different opinion compared to existing doc sets.

I think all the different options for `pebble *identities*` don't actually need to be in how-tos but can be kept in the Reference doc for Identities, with links to the relevant CLI commands.
Why? Because all those options describe the machinery / tool, what Pebble can do.
See this part "Refer to the x reference guide for a full list of options.".
https://diataxis.fr/how-to-guides/#the-language-of-how-to-guides

If we go back to the very basic analogy, it's like a recipe.
https://diataxis.fr/how-to-guides/#applied-to-food-and-cooking
So what is the user trying to accomplish?

But I know it's a difficult question to address :) 

As a start I would suggest the following changes (the rest can be improved later on as needed)
1. Provide a bit more context as @IronCore864 suggested so that when a user lands on this page, they have an idea how / when they would want to manage / use Identities in Pebble.
2. Structure each section with relevant headings so it's easier to navigate. It might seem repetitive but it makes it a lot easier to get to where you want. E.g. "How to manage identities", "Add a new identity", "Remove an identity" etc. 
3. I would suggest moving the "list identities" command before you update / remove / delete since you would (presumably) need to verify what exists before you try to change it.

Hope this makes sense, thanks!

docs/how-to/manage-identities.md
AnneCYH: I agree a bit more structure would be helpful to navigate the available content here.

docs/how-to/manage-identities.md
AnneCYH: Please ignore my suggestion about (potentially) moving this to Reference. :)

docs/explanation/api-and-clients.md
benhoyt: Fixed, thanks!

docs/explanation/api-and-clients.md
benhoyt: I've made these changes, thanks.

docs/how-to/manage-identities.md
benhoyt: Thanks. I've added a paragraph with a scenario. It's not quite as "big" as Tiexin's suggestion, but hopefully it servers as a context-setting introduction.

I've also added headings.

@AnneCYH Regarding moving "list identities" to the top, the problem with that is that at that stage there are no identities to list, so the output isn't useful. So I'd prefer to keep this one till the bottom so that we have some useful stuff to show (especially with the YAML/JSON modes).

docs/explanation/api-and-clients.md
benhoyt: Done.

docs/explanation/api-and-clients.md
benhoyt: Done.

docs/reference/identities.md
benhoyt: Good suggestion -- thanks.

docs/reference/identities.md
benhoyt: Good suggestion, thanks.

docs/how-to/manage-identities.md
benhoyt: Added

IronCore864 (approved):

docs/how-to/manage-identities.md
AnneCYH: Thanks @benhoyt , i think the brief intro is sufficient. 
And understood about the List identities sequence. I thought it might be good to List right after you Add before you Update or Remove.
But it's a minor thing.

Thanks!

docs/how-to/manage-identities.md
```diff
@@ -0,0 +1,145 @@
...
+Pebble uses named [identities](../reference/identities.md) to extend access to its [API](../explanation/api-and-clients.md).
+
+For example, let's say you're building a container-based system with one container running Pebble and the workload, and another container in the same user namespace talking to Pebble over the API. You may want to give API access to a non-root user -- that's where identities come in.
+
+By default any UID (user ID) connected to the API socket is a `read` user; UID 0 (root) or the UID of the Pebble server process gives `admin` access.
+
+
+## Add new identities
+
+To extend this to additional users, add named identities using the [`add-identities`](../reference/cli-commands/add-identities.md) command with a YAML file configuring the details. For example, to add a new admin "bob" with UID 42 and a new read user "alice" with UID 2000, prepare this file:
```
AnneCYH, line 13: Suggest a slight change since the flow is broken a little after sectioning out the content:
To extend access to Pebble's API to additional users, add...

Thanks.

Issue #462, opened 2024-08-01
Pebble silently fails on a `readOnlyRootFilesystem`


berkayoz:
Hello,

Kubernetes has a `securityContext` option called `readOnlyRootFilesystem` that makes a container's filesystem read-only. In this scenario, Pebble tries to create some state like `/var/lib/pebble/default/.pebble.state.2NLvJGDxF8jn~` and silently fails/hangs at startup. The container reports as running, and no logs indicating an error are communicated back(since I assume pebble couldn't start properly). 

A good number of projects on Kubernetes make use of this option by default in their manifests/helm charts. This is a pain point in the drop-in docker image replacement story of rocks. We are currently trying to work around this by disabling this option, which might not be accepted/appreciated by users.

Your help is much appreciated, thank you!

cc: @cjdcordeiro

benhoyt:
I am concerned that it fails/hangs silently at startup. That's definitely problematic, and we can look into that (probably next week). It should fail loudly, or at the very least log an error.

In terms of the readonly issue, Pebble does need a non-readonly directory to write the state file into. Can you set the `PEBBLE` environment variable to point to a writable directory?

benhoyt:
I've confirmed that we're hanging silently on startup when writing the state file fails (one can reproduce this easily by modifying `overlordStateBackend.Checkpoint` to always return an error).

berkayoz:
Thank you for triaging the issue @benhoyt 

One possible workaround would be to mount an `emptyDir` volume from Kubernetes and point pebble there, this is still not ideal for drop-in replacement scenarios however it might be a more acceptable workaround to keep the security benefits. We'll try and see if this works. I'll keep updating here on the progress, thanks again!

IronCore864:
I've provided [a quick fix](https://github.com/canonical/pebble/pull/481) to check if the state dir is writable before creating the backend.

Here are some of my other thoughts coming from debugging this issue:

[...]

PR #481, opened 2024-08-12
fix(overlord): pebble hangs with a read-only file system when starting the daemon

IronCore864:
Fix an issue where the Pebble daemon fails to start with a read-only file system.

Closes https://github.com/canonical/pebble/issues/462.

Changes:
- Check if Pebble dir is writable before creating the overlord state backend.
- Add some logging when backend writes fail to enhance user experience.
- Listen for interrupt signals when state unlock retries because of backend writes failure.

### DIFF
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -125,7 +125,7 @@ func New(opts *Options) (*Overlord, error) {
 		return nil, fmt.Errorf("directory %q does not exist", o.pebbleDir)
 	}
 	if !osutil.IsWritable(o.pebbleDir) {
-		return nil, fmt.Errorf("no write permission in directory %q", o.pebbleDir)
+		return nil, fmt.Errorf("directory %q not writable", o.pebbleDir)
 	}
 
 	statePath := filepath.Join(o.pebbleDir, cmd.StateFile)
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -124,6 +124,10 @@ func New(opts *Options) (*Overlord, error) {
 	if !osutil.IsDir(o.pebbleDir) {
 		return nil, fmt.Errorf("directory %q does not exist", o.pebbleDir)
 	}
+	if !osutil.IsWritable(o.pebbleDir) {
+		return nil, fmt.Errorf("no write permission in directory %q", o.pebbleDir)
+	}
+
 	statePath := filepath.Join(o.pebbleDir, cmd.StateFile)
 
 	backend := &overlordStateBackend{
@@ -256,7 +260,6 @@ func loadState(statePath string, restartHandler restart.Handler, backend state.B
 		patch.Init(s)
 		return s, restartMgr, nil
 	}
-
 	r, err := os.Open(statePath)
 	if err != nil {
 		return nil, nil, fmt.Errorf("cannot read the state file: %s", err)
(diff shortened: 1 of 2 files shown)


internals/overlord/overlord.go
```diff
@@ -124,6 +124,10 @@ func New(opts *Options) (*Overlord, error) {
 	if !osutil.IsDir(o.pebbleDir) {
 		return nil, fmt.Errorf("directory %q does not exist", o.pebbleDir)
 	}
+	if !osutil.IsWritable(o.pebbleDir) {
+		return nil, fmt.Errorf("no write permission in directory %q", o.pebbleDir)
```
benhoyt, line 128: Can we make this error message consistent with the other message above? `directory %q not writeable` or similar.

internals/overlord/state/state.go
```diff
@@ -293,6 +295,22 @@ func (s *State) Unlock() {
 			s.modified = false
 			return
 		}
+
+		// Let the user know what's going on if the backend writes failed.
+		if _, ok := err.(*os.PathError); ok {
```
benhoyt, line 300: Why only log on PathError? Shouldn't we log in all cases? Also, let's make the log message consistent with most other log messages: "Cannot" and starting with a capital letter. Maybe `Cannot write state file, retrying: %v`.

internals/overlord/state/state.go
```diff
@@ -293,6 +295,22 @@ func (s *State) Unlock() {
...
+			logger.Noticef("backend checkpoint failed: %v", err)
+		}
+
+		// Channel to listen for interrupt signals.
+		interrupt := make(chan os.Signal, 1)
+		signal.Notify(interrupt, syscall.SIGINT, syscall.SIGTERM)
```
benhoyt, line 306: This has a couple of related issues:

- You're calling `signal.Notify` once every loop. Only needs to be done once.
- In any case, we don't want this here (`signal.Notify` is global state polluting low-level code). It'd be much cleaner if we gave state a channel that was closed by a new `st.Stop()` method, and this was a `select` that waited on the stop channel or the timer. We already listen for signals in `cmd_run.go` at the top level, and that's probably where that should stay.

Still, it's a bit tricky to know when to call `st.Stop()`, because `state.Unlock` is called in the setup phase, when `cmd_run.go` is not yet waiting for interrupt signals. I wonder if we have a `state.InitDone()` that is called when we're ready; before that it doesn't do retries, after that it does (and waits on the close channel).

Or maybe we don't need to retry at all here? I'm not sure why we retry the writing when it's local disk. Surely if it would fail once it'll fail again? I'd like to get @hpidcock's input on this.

benhoyt (changes requested):

internals/overlord/state/state.go
benhoyt: Yeah, after discussion with Harry, he pointed out this isn't what we want to do anyway -- we want the "atomic commit" to try as hard as possible and *not* be interruptible. If they really want to kill the process they can `SIGKILL` it hard.

So I think we should leave this as is, but keep the `IsWriteable` check (which will catch most problems much earlier) and keep the additional log.

benhoyt (approved):
Great, thanks!

benhoyt:
Looks good, let's go ahead and merge @IronCore864

PR #483, opened 2024-08-13
docs: reduce readme

IronCore864:
Reducing the readme to a minimum to get ready for launching our doc site.

Preview: https://github.com/canonical/pebble/blob/9ab785b226fdb74a3723b79ff6b82eb4e9c52843/README.md


README.md
```diff
@@ -7,916 +7,62 @@
 
 _Take control of your internal daemons!_
 
-**Pebble** helps you to orchestrate a set of local service processes as an organized set.
-It resembles well known tools such as _supervisord_, _runit_, or _s6_, in that it can
-easily manage non-system processes independently from the system services, but it was
-designed with unique features that help with more specific use cases.
+**Pebble** is a lightweight Linux service manager that helps you orchestrate a set of local service processes as an organized set. It resembles well known tools such as _supervisord_, _runit_, or _s6_, in that it can easily manage non-system processes independently from the system services, but it was designed with unique features that help with more specific use cases.
```
tonyandrewmeyer, line 10: I am not 100% on this, and it's in the original text, but I believe that generally it's "well-known" when you're modifying an object and "well known" when you're using it standalone. So here it should be hyphenated.

README.md
```diff
@@ -7,916 +7,62 @@
...
-  - [Using Pebble](#using-pebble)
-  - [Container usage](#container-usage)
-  - [Layer specification](#layer-specification)
-  - [API and clients](#api-and-clients)
-  - [Hacking / Development](#hacking--development)
-  - [Contributing](#contributing)
+Pebble's key features:
 
-## General model
+- Service management with [layers](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layers/)
```
tonyandrewmeyer, line 14: Does the publishing system provide any way for these to be local links here and transformed into full (or rtd-local) ones when it's published? That would be nicer than having full ones here, and also avoid the need to adjust them when there's a nicer domain than `canonical-pebble.readthedocs-hosted.com`.

README.md
```diff
@@ -7,916 +7,62 @@
...
+- Service [dependencies](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/service-dependencies/)
+- Service [logs](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/cli-commands/logs/) and [log forwarding](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/log-forwarding/)
+- [Health checks](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/health-checks/)
+- [Notices](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/notices/)
+- Identities
+- Can be used in Machines and containers
```
tonyandrewmeyer, line 20:
```suggestion
- Can be used in machines and containers
```

More generally: "machines and containers" feels very Juju-y language. For someone unfamiliar with our systems, maybe it's more something like "used on bare metal, VMs, and containers"?

README.md
```diff
@@ -7,916 +7,62 @@
...
-The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of
-configuration files with names similar to `001-base-layer.yaml`, where the digits define
-the order of the layer and the following label uniquely identifies it. Each
-layer in the stack sits above the former one, and has the chance to improve or
-redefine the service configuration as desired.
+Prerequisites:
 
-## Layer configuration examples
+- A Linux machine.
+- Python 3.x (used to run a basic HTTP server as a sample service managed by Pebble).
```
tonyandrewmeyer, line 28: Is there some other service that's installed in a typical Ubuntu install that could be used? It feels a bit odd to be saying that you need Python, even though there's the "just for the sample service" explanation right afterwards.

README.md
```diff
@@ -7,916 +7,62 @@
...
 
-Below is an example of the current configuration format.
-For full details of all fields, see the [complete layer specification](#layer-specification).
-
-```yaml
+```bash
+git clone https://github.com/canonical/pebble.git
+cd pebble
+mkdir -p ~/PEBBLE/layers
+export PEBBLE=$HOME/PEBBLE
```
tonyandrewmeyer, line 34: I haven't been following the other docs closely, so maybe this has already been decided on, but I don't really like having a suggestion of an all-caps folder in my home directory. Couldn't it be `$HOME/pebble`? Or better, `$HOME/.pebble`? Or, since this is an example, `/tmp/pebble`?

README.md
```diff
@@ -7,916 +7,62 @@
...
-    A better description for a simple layer.
-
+    A simple layer.
 services:
-    srv1:
+    http-server:
         override: replace
-        summary: Service summary
-        command: cmd arg1 "arg2a arg2b"
+        summary: demo http server
```
tonyandrewmeyer, line 42: We could probably leave this line out to help make this more minimal.

README.md
```diff
@@ -7,916 +7,62 @@
...
+- A Linux machine.
+- Python 3.x (used to run a basic HTTP server as a sample service managed by Pebble).
 
-Below is an example of the current configuration format.
-For full details of all fields, see the [complete layer specification](#layer-specification).
-
-```yaml
+```bash
+git clone https://github.com/canonical/pebble.git
+cd pebble
```
tonyandrewmeyer, line 32: For a quick start that's for users not Pebble devs, probably it would be cleaner to `go install github.com/canonical/pebble/cmd/pebble@latest` or `sudo snap install pebble` and also adjust the command for running pebble, right? Probably the snap would be the best way, so that you don't have to knowingly be doing anything with Go.

README.md
```diff
@@ -7,916 +7,62 @@
...
-## API and clients
-
-The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers the plan, and so on.
+You can also follow the [Getting started with Pebble tutorial](https://canonical-pebble.readthedocs-hosted.com/en/latest/tutorial/getting-started/).
 
-There is currently no official documentation for the API at the HTTP level (apart from the [code itself](https://github.com/canonical/pebble/blob/master/internals/daemon/api.go)!); most users will interact with it via the Pebble command line interface or by using the Go or Python clients.
+## Getting help
 
-The Go client is used primarily by the CLI, but is importable and can be used by other tools too. See the [reference documentation and examples](https://pkg.go.dev/github.com/canonical/pebble/client) at pkg.go.dev.
+If you need support, start with the [documentation](https://canonical-pebble.readthedocs-hosted.com/en/latest/).
```
tonyandrewmeyer, line 53: I think it'd be worth having a link to the MM channel here as well (or maybe a Matrix channel can be added and be the new place to go?).

tonyandrewmeyer (commented):
Looking nice!

This is in the original, so I can't do a PR comment, but the final "and enjoy the rest of the year": do we need that? It sounds like some sort of inside joke or reference and I find it vaguely offputting as such. I don't have strong feelings on it, but if no-one actually knows why it's there, maybe it's time to retire it?

README.md
benhoyt: Agreed. This is already fixed in the docs (https://canonical-pebble.readthedocs-hosted.com/en/latest/), but let's fix it here too.

Let's also copy the other changes we've done in the new docs paragraph, eg "unique features such as layered configuration and an HTTP API that help with more specific use cases".

README.md
benhoyt: What do you mean by "the publishing system"? Isn't the main place this lives the github.com repo page?

tonyandrewmeyer: > What do you mean by "the publishing system"? Isn't the main place this lives the github.com repo page?

Ah, I was thinking that this also ended up somewhere on the docs site, probably mixing it up in my mind with the way that PyPI has the ops README.

So irrelevant, sorry.

benhoyt: I'm also asking on `~Documentation` and `~IS` whether we can get something like `pebble.canonical.com`.

README.md
benhoyt: Agreed. I think we should change this here and in the main docs.

benhoyt: Let's finalise `$HOME/pebble` vs `$HOME/.pebble` in daily sync.

README.md
benhoyt: Agreed the either of these is better for normal users than cloning. We also present a different way again in the tutorial (download the pre-compiled binaries and unpack). The snap approach looks tantalisingly simple ... but maybe for the README the `go install ...@latest` is best.

README.md
benhoyt: How about we point people to open GitHub Issues for now? The Mattermost channel is private to Canonical, and I'm not sure about a Matrix channel just yet.

README.md
benhoyt: It will actually show up on https://pkg.go.dev/github.com/canonical/pebble, but I don't think many people will be reading it there (you'd have to be looking for the [Go client](https://pkg.go.dev/github.com/canonical/pebble@v1.15.0/client)).

README.md
```diff
@@ -7,916 +7,62 @@
...
 
-## General model
+- Service management with [layers](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layers/)
+- Service [dependencies](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/service-dependencies/)
+- Service [logs](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/cli-commands/logs/) and [log forwarding](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/log-forwarding/)
+- [Health checks](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/health-checks/)
+- [Notices](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/notices/)
+- Identities
+- Can be used in Machines and containers
+- [CLI](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/cli-commands/cli-commands/), [HTTP API and Go client](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/api-and-clients/)
```
benhoyt, line 21: What about two bullet points. Maybe:

```
* CLI commands
* [HTTP API] with a [Go client] and a [Python client]
```

README.md
benhoyt: Yeah, I wonder if we should just make the service command a "sleep 1000" here.

README.md
benhoyt: Yeah, let's remove the top-level summary and description as well.

README.md
```diff
@@ -7,916 +7,62 @@
...
-    # on the outgoing logs. The label values may contain $ENV_VARS, which will
-    # be substituted using the environment for the corresponding service.
-    labels:
-      <label name>: <label value>
-```
-
-## API and clients
-
-The Pebble daemon exposes an API (HTTP over a unix socket) to allow remote clients to interact with the daemon. It can start and stop services, add configuration layers the plan, and so on.
+You can also follow the [Getting started with Pebble tutorial](https://canonical-pebble.readthedocs-hosted.com/en/latest/tutorial/getting-started/).
```
benhoyt, line 49: What about "For a slightly longer introduction, you can also go through the [official tutorial](#)."

README.md
```diff
@@ -7,916 +7,62 @@
...
 
-There is currently no official documentation for the API at the HTTP level (apart from the [code itself](https://github.com/canonical/pebble/blob/master/internals/daemon/api.go)!); most users will interact with it via the Pebble command line interface or by using the Go or Python clients.
+## Getting help
 
-The Go client is used primarily by the CLI, but is importable and can be used by other tools too. See the [reference documentation and examples](https://pkg.go.dev/github.com/canonical/pebble/client) at pkg.go.dev.
+If you need support, start with the [documentation](https://canonical-pebble.readthedocs-hosted.com/en/latest/).
 
-We try to never change the underlying HTTP API in a backwards-incompatible way, however, in rare cases we may change the Go client in a backwards-incompatible way.
+To learn more about Pebble, read the following sections:
```
benhoyt, line 55: Some of these seems like duplicates of the "key features" list above. Can we combine the features into that list?

And the Hacking/contributing links could just go in the next section.

README.md
```diff
@@ -7,916 +7,62 @@
...
+- [General model](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/general-model/)
+- [Layer configuration examples](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layers/)
+- [Container usage](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/pebble-in-containers/)
+- [Layer specification](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layer-specification/)
+- [API and clients](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/api-and-clients/)
+- [Hacking / Development](#hacking--development)
+- [Contributing](#contributing)
 
-## Hacking / Development
+## Hacking / development
```
benhoyt, line 65: Maybe `Hacking and development`

benhoyt: I also reckon we should remove the "Have fun! and enjoy the rest of the year!" section ... just seems a bit weird to me.

README.md
```diff
@@ -7,916 +7,62 @@
...
-
-Logs are viewable via the logs API or using `pebble logs`, for example:
-
-```
-$ pebble logs
-2022-11-14T01:35:06.979Z [srv1] Log 0 from srv1
-2022-11-14T01:35:08.041Z [srv2] Log 0 from srv2
-2022-11-14T01:35:09.982Z [srv1] Log 1 from srv1
+""" > $PEBBLE/layers/001-http-server.yaml
+go run ./cmd/pebble run
```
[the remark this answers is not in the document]
IronCore864: I agree it will be easier for the eyes. But I'm not sure if it's necessary, because the point here is that users can copy the whole block and run it. Thoughts? @benhoyt @tonyandrewmeyer

README.md
IronCore864: OK, I will remove this list. I think one item here adds value - the general model link. I will add a sentence about it in the quick start part maybe.

README.md
benhoyt: Decision was to use `$HOME/.config/pebble` (and update other docs to be consistent).

README.md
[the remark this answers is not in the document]
IronCore864: Per our daily sync, we will add empty lines here.

README.md
IronCore864: Per our discussion in the daily, we decide to use `~/.config/pebble` as pebble dir. The tutorial will be changed to the same as well.

README.md
```diff
@@ -7,916 +7,53 @@
...
-
-When multiple services need to be started together, they're started in order according to the `before` and `after` configuration, waiting 1 second for each to ensure the command doesn't exit too quickly. The `before` option is a list of services that this service must start before (it may or may not "require" them). Or if it's easier to specify this ordering the other way around, `after` is a list of services that this service must start after.
+Read more about Pebble's general model [here](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/general-model/).
 
-Note that currently, `before` and `after` are of limited usefulness, because Pebble only waits 1 second before moving on to start the next service, with no additional checks that the previous service is operating correctly.
+For a slightly longer introduction, you can also go through the [official tutorial](https://canonical-pebble.readthedocs-hosted.com/en/latest/tutorial/getting-started/)
```
AnneCYH, line 48:
```suggestion
For a hands-on introduction to Pebble, we recommend going through the [tutorial](https://canonical-pebble.readthedocs-hosted.com/en/latest/tutorial/getting-started/).
```

AnneCYH: Because "longer" doesn't necessarily mean "better" so hands-on might be more appealing since users know they'll be getting their feet wet.

Thanks.

README.md
```diff
@@ -7,916 +7,53 @@
...
 
-If the configuration of `before` and `after` for the services results in a cycle, an error will be returned when the Pebble daemon starts (and the plan is loaded) or when a layer that causes a cycle is added.
+## Getting help
 
-### Service auto-restart
+If you need support, start with the [documentation](https://canonical-pebble.readthedocs-hosted.com/en/latest/).
```
AnneCYH, line 52: To get the most out of Pebble, we recommend starting with the [documentation](https://canonical-pebble.readthedocs-hosted.com/en/latest/).

README.md
```diff
@@ -7,62 +7,53 @@
...
 Pebble's key features:
 
-- Service management with [layers](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layers/)
+- [Layer](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/layers/)-based configuration
 - Service [dependencies](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/service-dependencies/)
 - Service [logs](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/cli-commands/logs/) and [log forwarding](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/log-forwarding/)
 - [Health checks](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/health-checks/)
-- [Notices](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/notices/)
+- [Notices](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/notices/) (aggregated events)
 - Identities
```
benhoyt, line 19: We can add a link for Identities now. Maybe the how-to guide?

README.md
```diff
@@ -7,62 +7,53 @@
...
-- Can be used in Machines and containers
-- [CLI](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/cli-commands/cli-commands/), [HTTP API and Go client](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/api-and-clients/)
+- Can be used in machines and [containers](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/pebble-in-containers/)
```
benhoyt, line 20: Per Tony's earlier suggestion, "VMs" or "virtual machines"

```suggestion
- Can be used in virtual machines and [containers](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/pebble-in-containers/)
```

README.md
```diff
@@ -7,916 +7,53 @@
...
 
-The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of
-configuration files with names similar to `001-base-layer.yaml`, where the digits define
-the order of the layer and the following label uniquely identifies it. Each
-layer in the stack sits above the former one, and has the chance to improve or
-redefine the service configuration as desired.
+Prerequisites:
 
-## Layer configuration examples
+- A Linux machine.
```
benhoyt, line 28: This kinda reads funny to have only a single item rerequisite bulleted list. Perhaps remove "Prerequisites:" and "- A linux machine." and just introduce it with something like this:

```
At any Linux shell:
```

README.md
```diff
@@ -7,916 +7,53 @@
...
-Below is an example of the current configuration format.
-For full details of all fields, see the [complete layer specification](#layer-specification).
-
-```yaml
-summary: Simple layer
-
-description: |
-    A better description for a simple layer.
+```bash
+go install github.com/canonical/pebble/cmd/pebble@latest
```
benhoyt, line 31: We should figure out which installation approach we want to recommend, and use the same one as in the tutorial. On balance it's probably best to use standard Linux tools and not require Go to be installed, so IMO let's go with what's in the tutorial.

Perhaps we can just link to the installation doc to avoid repeating a bunch of stuff. Maybe:

> At any Linux shell, [install Pebble](https://canonical-pebble.readthedocs-hosted.com/en/latest/how-to/install-pebble/#how-to-install-pebble) and then run the following:
>
>     mkdir -p ~/.config/pebble/layers
>     ...

benhoyt (changes requested):

AnneCYH (approved):

README.md
IronCore864: Technically it can be used in physical machines as well, that is why I used "machines" in the first place. But physical machines already come with systemd, Pebble probably won't be used there, although it can be. Changing to "virtual machines" then.

README.md
IronCore864: Although using a link reduces duplicated content, it increases the barrier for people who want to give it a "quick start". The user journey would become:

Found the Pebble repo -> read the README -> click a link and jump to another tab -> copy-paste-run three commands -> switch back to the repo tab -> copy-paste the following command.

I think we want to reduce the friction to a minimum so that users can easily start without any hassle.

Even if we copy the three commands from the [how to install pebble](https://canonical-pebble.readthedocs-hosted.com/en/latest/how-to/install-pebble/#how-to-install-pebble) doc, users still need to replace the version tag.

I think `go install` is the simplest way, just one line of command, no need to replace anything, and it's OK to be used in the README because people who check out the repo are most likely developers.

It's also OK to use a different approach in the official doc because first of all, it's a more "official" way of installing a specific version; and second, the doc's audience includes normal users who are not developers.

So, I'd like to keep the `go install` here. Any objections? 

```bash

benhoyt: Good reasoning -- you've convinced me, thanks. Let's keep `go install`.

benhoyt (approved):
Thanks!

Issue #480, opened 2024-08-12
Improve pebble doc site landing page "in this documentation" page


IronCore864:
Instead of using words from the template, rewrite it to highlight anything we want to.

PR #485, opened 2024-08-15
docs: update in this documentation section

IronCore864:
Update the "In this documentation" section on the main page.

Closes https://github.com/canonical/pebble/issues/480.

Changes:

- Add links to highlight important content for how-to, explanations and references.

Preview:

Screenshot 2024-08-15 at 08 24 29

Some explanations of my experiments and how I arrived at this version:

- Tutorial: nothing has changed. I wanted to list the tutorial there like the LXD doc does, but since we have only one, it would look weird and the link would be the same as the section title's link, so I left this part untouched.
- How-to: all 3 existing how-to guides there. Since we only had 3, it doesn't make much sense to pick just 2 of the more important ones, so I listed them all.
- Explanations: only "General models" and "API and clients"; "Service dependencies" skipped because of duplication as the how-to.
- Reference, only 3 the most reference-y items. In my first draft, I added 4 more items like "Changes and tasks", "Health check", "Notices", and "Identities", which are all important features. But in that way, the list would be 7 items long and extremely imbalanced visually compared to the left explanation box which has only 2 items. Making it 4-5 items would be less bad, but it's difficult to pick just 1 or 2 from the 4 items I left out to highlight here, so I just left them all out.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,7 +11,7 @@ Pebble is useful for developers who are building [Juju charms on Kubernetes](htt
 ## In this documentation
 
 ````{grid} 1 1 2 2
-```{grid-item-card} [Tutorials](tutorial/getting-started)
+```{grid-item-card} [Tutorial](tutorial/getting-started)
 **Start here**: a hands-on getting started introduction to Pebble for new users.
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -12,11 +12,11 @@ Pebble is useful for developers who are building [Juju charms on Kubernetes](htt
 
 ````{grid} 1 1 2 2
 ```{grid-item-card} [Tutorial](tutorial/getting-started)
-**Start here**: a hands-on getting started introduction to Pebble for new users.
+**Start here**: a hands-on introduction to Pebble, guiding you through your first steps using the CLI
 ```
 
 ```{grid-item-card} [How-to guides](how-to/index)
-**Step-by-step guides** covering key operations and common tasks.
+**Step-by-step guides** covering key operations and common tasks
 - [Install Pebble](how-to/install-pebble)
 - [Manage service dependencies](how-to/service-dependencies)
 - [Manage identities](how-to/manage-identities)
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -17,16 +17,24 @@ Pebble is useful for developers who are building [Juju charms on Kubernetes](htt
...
 **Step-by-step guides** covering key operations and common tasks.
+- [Install Pebble](how-to/install-pebble)
+- [Manage service dependencies](how-to/service-dependencies)
+- [Manage identities](how-to/manage-identities)
 ```
 ````
 
 ````{grid} 1 1 2 2
 ```{grid-item-card} [Explanation](explanation/index)
 **Discussion and clarification** of key topics.
```
benhoyt, line 28: This probably shouldn't have a `.` at the end, as it's not a full sentence. Same with the other subtitles.

docs/index.md
```diff
@@ -17,16 +17,24 @@ Pebble is useful for developers who are building [Juju charms on Kubernetes](htt
 
 ```{grid-item-card} [How-to guides](how-to/index)
 **Step-by-step guides** covering key operations and common tasks.
+- [Install Pebble](how-to/install-pebble)
+- [Manage service dependencies](how-to/service-dependencies)
+- [Manage identities](how-to/manage-identities)
```
benhoyt, line 22: Side note: do you know how we configure the ordering of these items in the left-hand sidebar menu? I would have thought they'd be ordered alphabetically. (Here I think ordering by importance makes sense.)

benhoyt (commented):
Looks like a nice improvement!

Can we change "Start here ..." to the following? An introduction is by definition for new users, and "getting started" ~= "introduction". Plus we can add a bit more info. This is similar to https://documentation.ubuntu.com/lxd/en/latest/

**Start here:** a hands-on introduction to Pebble, guiding you through your first steps using the CLI

docs/index.md
IronCore864: It depends on the order of the `toctree` defined in `docs/how-to/index.md`.

It makes more sense to order by importance. Which order shall we use? I think:

```yaml
# How-to guides

## Service orchestration
## Identities
## Installation
```

Is this OK? However, it seems a little odd on the how-to index page because normally you wouldn't finish an article with how to install something...

benhoyt: Per discussion, I think the current order is fine, actually -- I was just wondering why it wasn't alphabetical. But the manual ordering here is fine as is. I like "Install Pebble" first.

benhoyt (approved):
Thanks!

PR #486, opened 2024-08-16
refactor: implement warnings as notices

### LINKED ISSUE #475: Fully update warnings system to use notices
Per https://github.com/canonical/pebble/pull/443 and subsequent spec review discussion, we want to update warnings to use notices properly and report whether there are active warnings in each response somehow.

To solve the "single client" issue (there's no `warning.timeShown` for notices), we'd have the server report in each response the timestamp of the latest warning (this was @dimaqq's suggestion). This is similar to how we currently report the warning count in each response today (that would go away). That way the client could compare against its stored timestamp of latest okayed warning, and it would know whether there were new warnings or not. It wouldn't know the number, but the client could easily query for that if there was 1 or more.

Some details still to be worked out, but that's the basic idea.

benhoyt:
This deletes the legacy warnings code and re-implements `pebble warnings` and `pebble okay` using notices as warnings (as we originally intended). The `/v1/warnings` endpoints have been removed, and `pebble warnings` now hits `/v1/notices?types=warning`.

On every request, instead of returning the count of warnings and the latest timestamp, we only return the timestamp of the latest (most recent last-repeated time). This way the server doesn't have to track which notices have been okayed -- the client can compare against its stored timestamp and see if the server-reported timestamp is later.

So there are a few breaking changes in this PR:

* `Client.WarningsSummary() (count int, timestamp time.Time)` is now `Client.LatestWarningTime() time.Time`, as there's no count anymore
* `GET /v1/warnings` (list warnings) is gone: `pebble warnings` now uses `GET /v1/notices?types=warning`
* `POST /v1/warnings` (okay warnings) are gone: `pebble okay` now just updates the local state file, and doesn't use the API/client at all
* Similarly, `Client.Warnings()` and `Client.Okay()` are gone; use `Client.Notices()` and maintain local client-side state, respectively
* `State.Warnf` is still around, but it's now just a helper that calls `State.AddNotice` with the right type (warning) and key (message)

Currently Pebble itself doesn't record any warnings, so you can't test easily without making a code change. To perform manual tests, I hacked the API endpoint used by `pebble notify` to record `warning` notices instead of `custom` ones:

```diff
diff --git a/internals/daemon/api_notices.go b/internals/daemon/api_notices.go
index 8cd2745..7b62f19 100644
--- a/internals/daemon/api_notices.go
+++ b/internals/daemon/api_notices.go
@@ -201,9 +201,9 @@ func v1PostNotices(c *Command, r *http.Request, user *UserState) Response {
        if payload.Type != "custom" {
                return BadRequest(`invalid type %q (can only add "custom" notices)`, payload.Type)
        }
-       if !customKeyRegexp.MatchString(payload.Key) {
-               return BadRequest(`invalid key %q (must be in "example.com/path" format)`, payload.Key)
-       }
+       //if !customKeyRegexp.MatchString(payload.Key) {
+       //      return BadRequest(`invalid key %q (must be in "example.com/path" format)`, payload.Key)
+       //}
        if len(payload.Key) > maxNoticeKeyLength {
                return BadRequest("key must be %d bytes or less", maxNoticeKeyLength)
        }
@@ -228,7 +228,7 @@ func v1PostNotices(c *Command, r *http.Request, user *UserState) Response {
        st.Lock()
        defer st.Unlock()
 
-       noticeId, err := st.AddNotice(user.UID, state.CustomNotice, payload.Key, &state.AddNoticeOptions{
+       noticeId, err := st.AddNotice(nil, state.WarningNotice, payload.Key, &state.AddNoticeOptions{
                Data:        data,
                RepeatAfter: repeatAfter,
        })
```

Fixes #475.


internals/cli/cmd_okay.go
```diff
@@ -61,33 +61,28 @@ func (cmd *cmdOkay) Execute(args []string) error {
...
 			state.NoticesLastOkayed = state.NoticesLastListed
-			err = saveCLIState(cmd.socketPath, state)
-			if err != nil {
-				return fmt.Errorf("cannot save CLI state: %w", err)
-			}
 		}
 	}
 
-	last, err := lastWarningTimestamp()
+	err = saveCLIState(cmd.socketPath, state)
```
hpidcock, line 83: I think, at some point, we need to improve `saveCLIState` to atomically write the state file (and to include conflict resolution).

internals/cli/cmd_warnings.go
```diff
@@ -71,56 +72,57 @@ func (cmd *cmdWarnings) Execute(args []string) error {
...
 			fmt.Fprintf(w, "repeats-after:\t%s\n", quantity.FormatDuration(warning.RepeatAfter.Seconds()))
 			fmt.Fprintf(w, "expires-after:\t%s\n", quantity.FormatDuration(warning.ExpireAfter.Seconds()))
 		}
 		fmt.Fprintln(w, "warning: |")
-		writeWarning(w, warning.Message, termWidth)
+		writeWarning(w, warning.Key, termWidth)
 		w.Flush()
 	}
 
+	state.WarningsLastListed = warnings[len(warnings)-1].LastRepeated
```
hpidcock, line 120: Nice detail.

hpidcock (approved):
LGTM. A few comments. Looking forward to the follow-up review that removes the mutex lock on the daemon response code for warnings.

internals/overlord/state/notices.go
```diff
@@ -352,6 +369,22 @@ func (s *State) Notices(filter *NoticeFilter) []*Notice {
 	return notices
 }
 
+// LatestWarningTime returns the most recent time a warning notice was
+// repeated, or the zero value if there are no warnings.
+func (s *State) LatestWarningTime() time.Time {
+	s.reading()
+
+	// TODO(benhoyt): optimise with an in-memory atomic cache when warnings are added (or pruned),
```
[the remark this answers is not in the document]
benhoyt: Yes, it's my plan to do a follow-up PR shortly after this one.

client/warnings.go
```diff
@@ -1,82 +0,0 @@
-// Copyright (c) 2014-2020 Canonical Ltd
```
IronCore864, line 1: Why is `internals/cli/cmd_warnings` kept, but `client/warnings` removed? I would assume that CLI and client have the same set of commands/APIs.

internals/overlord/state/notices.go
```diff
@@ -278,6 +277,24 @@ func (s *State) AddNotice(userID *uint32, noticeType NoticeType, key string, opt
 	return notice.id, nil
 }
 
+// Warnf records a warning: if it's the first Warning with this message it'll
+// be added (with its firstOccurred and lastOccurred set to the current time),
+// otherwise the existing one will have its lastOccurred updated.
+func (s *State) Warnf(template string, args ...interface{}) {
```
IronCore864, line 283: It seems this function is not called anywhere. I have two questions:

1. How is a warning created? 

- Is it that in the future, if something goes wrong, the Pebble code may call this function to create a warning?
- Or is it that a charmer can create a notice with the type set to `WarningNotice`? If this is the case, currently, [only custom notice is allowed](https://github.com/canonical/pebble/blob/master/internals/daemon/api_notices.go#L201). This feature is not easy to test, I have to comment this place out and use `state.NoticeType(payload.Type)` when calling `AddNotice`, or call `Warnf` in `api_notices.go`. After testing, though, the feature works as expected, and `last-occurrence` can be updated if the same warning is issued again later.

2. Why it doesn't return a notice ID?

Asking because it seems `state.AddNotice` returns an ID. And if we want to support the feature to okay just a single warning (see my other comment), the ID will be needed I think.

internals/cli/cmd_okay.go
```diff
@@ -61,33 +61,28 @@ func (cmd *cmdOkay) Execute(args []string) error {
...
 
+	state, err := loadCLIState(cmd.socketPath)
+	if err != nil {
+		return fmt.Errorf("cannot load CLI state: %w", err)
+	}
+
+	okayedWarnings := false
+	if !state.WarningsLastListed.IsZero() {
+		okayedWarnings = true
+		state.WarningsLastOkayed = state.WarningsLastListed
```
IronCore864, line 72: If there are multiple warnings, okay will acknowledge them all, and when users run `pebble warnings` again, all warnings are gone.

What if I only want to acknowledge a single warning and leave the others there as a memo? For example, I have three warnings about three failed services, after checking the warnings, I fix the first failed service, but I don't know what to do with the others yet, so I only want to make the first warning disappear.

client/client.go
```diff
@@ -147,9 +147,8 @@ type Config struct {
 type Client struct {
 	requester Requester
 
-	maintenance      error
-	warningCount     int
```
IronCore864, line 151: Maybe the count is important because, for example, the warning is the workload restarted, you want to know how many times it restarted in the last hour or minute because if it restarted like 100 times in the past minute, something is seriously wrong. I got this idea from Kubernetes where you can see the counts.

IronCore864 (commented):
Some questions/comments on this feature in general, not on detailed code.

IronCore864 (approved):

client/notices.go
```diff
@@ -130,6 +130,10 @@ const (
 	// The key and data fields are provided by the user. The key must be in
 	// the format "example.com/path" to ensure well-namespaced notice keys.
 	CustomNotice NoticeType = "custom"
+
+	// Warnings are a subset of notices where the key is a human-readable
+	// warning message.
+	WarningNotice NoticeType = "warning"
```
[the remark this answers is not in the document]
benhoyt: Ah, good catch -- updated.

client/warnings.go
benhoyt: The `pebble warnings` command lists warnings, but because warnings are now implemented in terms of notices, there's no warnings API or state anymore -- it's all done via notices.

internals/overlord/state/notices.go
benhoyt: Yes, currently Pebble itself doesn't create warnings or call `Warnf` anywhere. We might add that in the future, or other users of Pebble as a library might call it. It's not the intention to expose this to charmers (as you notices, the API prevents that now) -- charms would use custom notices.

It doesn't return a notice ID as there's no need for now. And you can actually look up a notice of type warning by key (which is the textual message of the warning).

benhoyt: So to test this -- you're right, how have to add `Warnf` calls to the code manually. I'll update the PR description once I've finished doing my own manual testing.

internals/cli/cmd_okay.go
benhoyt: Yes, this is by design. There's no way to acknowledge individual warnings, just all the ones you've listed with `pebble warnings`. If we need something finer-grained in future, we can consider it then.

client/client.go
benhoyt: The server can't return the count directly any more because the server side doesn't know which warnings are acknowledged anymore. However, if the client sees there's 1 or more warnings it could query `GET /v1/notices?types=warning` to get the number -- though I decided not to do that for now. If we want to add that later as warnings actually start to get used, it won't be hard.

Issue #145, opened 2022-08-26
Add support for a delay before starting health checks


arturo-seijas:
It would be useful to have ability to specify a delay before start executing them analogous to kubernetes' [initialDelaySeconds](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes):

```
checks:
    up:
        override: replace
        level: alive
        period: 10s
        timeout: 3s
        threshold: 3
        delay: 0s # this would be the new parameter
        exec:
            command: service nginx status
```

benhoyt:
This is a reasonable request, thanks. We did consider it when initially implementing checks, but decided it generally wouldn't be needed -- the period (10s in this case) would almost always cover startup time. Just to get more info: what's your use case, or how long does the container take to start up (i.e., what would your delay value be here)? Could you use a slightly longer `period` instead?

arturo-seijas:
We're deploying a web application. On startup it might run some database migrations and spin off an application server. This can take up to a couple minutes, which seems like a lot of time to be detected by adjusting the existing checks.

If we did, in the worst case scenario, we'd risk having the container down for two minutes before triggering a restart, which seems like sacrificing a lot of availability.

We did increase the period to mitigate our issues, but for the reasons above, the way it's configured still triggers some restarts while spinning up.

benhoyt:
Just copying my quick-n-dirty proof of concept code for the record, before I delete it locally:

```go
	delay := 100 * time.Millisecond // TODO: from plan
	logger.Debugf("Check %q starting with period %v, delay %v", c.config.Name, c.config.Period.Value, delay)

	if delay != 0 {
		// Before starting check, wait for initial delay time.
		select {
		case <-time.After(delay):
		case <-c.ctx.Done():
			logger.Debugf("Check %q stopped: %v", c.config.Name, c.ctx.Err())
			return
		}
	}
```

mthaddon:
Btw, just to note that we already see this defined in the liveness checks for containers:

```
http-get http://:38816/v1/health%3Flevel=alive delay=30s timeout=1s period=5s #success=1 #failure=1
```

[...]

PR #487, opened 2024-08-19
feat(delay): Add initial delay

merkata:
Adds an option to `checks` to configure an initial delay before starting checks - this is helpful for slow services that might take minutes to start up initially (think Synapse that is a python app that connects to a DB and the more used and federated the home server is, the slower the initial start up). This relates to #145 requested for a similar use case (Indico being a python app with a DB that causes a slow start up time).

I've tested this locally with a toy app as seen in the screenshot, though I haven't written a test yet, I would need some guidance where to put it and how to structure it (provided this PR fits the purpose and moves forward).

pebble

Thanks!

### DIFF
--- a/internals/overlord/checkstate/handlers.go
+++ b/internals/overlord/checkstate/handlers.go
@@ -38,6 +38,16 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 		return fmt.Errorf("cannot get check details for perform-check task %q: %v", task.ID(), err)
 	}
 
+	if !details.HasInitialDelay && config.Delay.Value > 0 {
+		details.HasInitialDelay = true
+		logger.Debugf("Initial delay for check %q: %v", config.Name, config.Delay.Value)
+		select {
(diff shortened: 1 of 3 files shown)


benhoyt:
Just cross-posting Tiexin's comment on #145 here for visibility:

> Hi @merkata, thanks very much for the PR! However, we are still in the process of spec reviewing this feature. Please wait for a while. I'll share the spec doc later and keep you posted :)

We have bumped up the priority of this significantly, but want to spec it properly, so we'll aim to get it in the September release of Pebble.

IronCore864:
Hi @merkata, thanks very much for the PR! However, we are still in the process of spec review.
Here's the spec to be reviewed, hopefully, next week: https://docs.google.com/document/d/1_U92ssI-DGAojkMJUULyur23NkjCRu6gSM0dkr7PuGY/edit.

benhoyt:
We've decided to not to initial-delay and take a different approach here, see: https://github.com/canonical/pebble/issues/145#issuecomment-2314206029

PR #488, opened 2024-08-19
feat(plan): add plan section support

flotter:
The plan library was originally written as a configuration schema for the services manager (servstate). Over time the need arose to support configurations for other managers such as checks (checkstate) and log-targets (logstate). The configuration schema for these managers, closely related to the services manager, has since also been built in to the plan library.

The services manager and its related checks and log-targets functionality will always be part of the Pebble core. However, as Pebble is getting more functionality (additional managers) and also used as a core in derivative projects, a more modular and dynamic approach to extending the schema is needed.

Add an the ```SectionExtension``` interface for use by managers who wishes to register a schema extension during Pebble startup.

Inside each layer, top level entries are now referred to as ```sections``` (built-in sections includes ```summary```, ```description```, ```services```, ```log-targets``` and ```checks```).

Each section has an associated ```field``` that is the top level key, and if supplied by an extension, an opaque backing type ```Section```.

**SectionExtension interface:**

```
// SectionExtension allows the plan layer schema to be extended without
// adding centralised schema knowledge to the plan library.
type SectionExtension interface {
	// ParseSection returns a newly allocated concrete type containing the
	// unmarshalled section content.
	ParseSection(data yaml.Node) (LayerSection, error)

	// CombineSections returns a newly allocated concrete type containing the
	// result of combining the supplied sections in order.
	CombineSections(sections ...LayerSection) (LayerSection, error)

	// ValidatePlan takes the complete plan as input, and allows the
	// extension to validate the plan. This can be used for cross section
	// dependency validation.
	ValidatePlan(plan *Plan) error
}

type Section interface {
	// Validate checks whether the section is valid, returning an error if not.
	Validate() error
	
        // IsZero reports whether the section is empty.
	IsZero() bool
}
```

**Example usage:**

```
// New SectionExtension type
type fooExtension struct{}
func (f *fooExtension) ParseSection(data yaml.Node) (LayerSection, error) {...}
func (f *fooExtension) CombineSections(sections ...LayerSection) (LayerSection, error) {...}
func (f *fooExtension) ValidatePlan(plan *Plan) error {...}

// New Section type
type FooSection struct {
    Entries map[string]Bar `yaml:",inline,omitempty"`
}
type Bar struct {
    Name string `yaml:"name,omitempty"`
}
func (s *FooSection) Validate() error {...}
func (s *FooSection) IsZero() bool {...}
```
```
// Early startup
plan.RegisterExtension("foo", &fooExtension{})
:
// Load layers containing new section
newPlan := plan.ReadDir(layersDir)
:
// Show plan
yaml.Marshal(newPlan)
```
Example YAML output:
```
foo:
    bar1:
           name: test1
    bar2:
           name: test2          
```

### DIFF
--- a/internals/plan/extensions_test.go
+++ b/internals/plan/extensions_test.go
@@ -42,18 +42,24 @@ type planResult struct {
 }
 
 var extensionTests = []struct {
+	summary    string
 	extensions map[string]plan.LayerSectionExtension
 	layers     []*inputLayer
 	error      string
 	result     *planResult
 	resultYaml string
 }{
-	// Index 0: No Sections
 	{
+		summary:    "No Sections",
 		resultYaml: "{}\n",
-	},
-	// Index 1: Sections with empty YAML
-	{
+	}, {
+		summary: "Section using built-in name",
+		extensions: map[string]plan.LayerSectionExtension{
+			"summary": &xExtension{},
+		},
+		error: ".*already used as built-in field.*",
+	}, {
+		summary: "Sections with empty YAML",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -63,9 +69,8 @@ var extensionTests = []struct {
 			y: &ySection{},
 		},
 		resultYaml: "{}\n",
-	},
-	// Index 2: Load file layers invalid section
-	{
+	}, {
+		summary: "Load file layers invalid section",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -82,9 +87,8 @@ var extensionTests = []struct {
 			},
 		},
 		error: "cannot parse layer .*: unknown section .*",
-	},
-	// Index 3: Load file layers not unique order
-	{
+	}, {
+		summary: "Load file layers not unique order",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -108,9 +112,8 @@ var extensionTests = []struct {
 			},
 		},
 		error: "invalid layer filename: .* not unique .*",
-	},
-	// Index 4: Load file layers not unique label
-	{
+	}, {
+		summary: "Load file layers not unique label",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -134,9 +137,8 @@ var extensionTests = []struct {
 			},
 		},
 		error: "invalid layer filename: .* not unique .*",
-	},
-	// Index 5: Load file layers with empty section
-	{
+	}, {
+		summary: "Load file layers with empty section",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -164,9 +166,8 @@ var extensionTests = []struct {
 			y: &ySection{},
 		},
 		resultYaml: "{}\n",
-	},
-	// Index 6: Load file layers with section validation failure #1
-	{
+	}, {
+		summary: "Load file layers with section validation failure #1",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -187,9 +188,8 @@ var extensionTests = []struct {
 			},
 		},
 		error: "cannot validate layer section .* cannot accept entry not starting .*",
-	},
-	// Index 7: Load file layers with section validation failure #2
-	{
+	}, {
+		summary: "Load file layers with section validation failure #2",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -207,9 +207,8 @@ var extensionTests = []struct {
 			},
 		},
 		error: "cannot validate layer section .* cannot have nil entry .*",
-	},
-	// Index 8: Load file layers failed plan validation
-	{
+	}, {
+		summary: "Load file layers failed plan validation",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -245,9 +244,8 @@ var extensionTests = []struct {
 			},
 		},
 		error: "cannot validate plan section .* cannot find .* as required by .*",
-	},
-	// Index 9: Check empty section omits entry
-	{
+	}, {
+		summary: "Check empty section omits entry",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -277,9 +275,8 @@ var extensionTests = []struct {
 			y: &ySection{},
 		},
 		resultYaml: "{}\n",
-	},
-	// Index 10: Load file layers
-	{
+	}, {
+		summary: "Load file layers",
 		extensions: map[string]plan.LayerSectionExtension{
 			"x-field": &xExtension{},
 			"y-field": &yExtension{},
@@ -356,49 +353,62 @@ var extensionTests = []struct {
 }
 
 func (s *S) TestPlanExtensions(c *C) {
-	for testIndex, planTest := range extensionTests {
-		c.Logf("Running TestPlanExtensions with test data index %v", testIndex)
+nexttest:
+	for testIndex, testData := range extensionTests {
+		c.Logf("TestPlanExtensions :: %s (data index %v)", testData.summary, testIndex)
 
 		// Write layers to test directory.
 		layersDir := filepath.Join(c.MkDir(), "layers")
-		s.writeLayerFiles(c, layersDir, planTest.layers)
+		s.writeLayerFiles(c, layersDir, testData.layers)
 		var p *plan.Plan
 
 		// Register test extensions.
-		for field, extension := range planTest.extensions {
-			plan.RegisterExtension(field, extension)
+		for field, extension := range testData.extensions {
+			err := func() (err error) {
+				defer func() {
+					if r := recover(); r != nil {
+						err = fmt.Errorf("%v", r)
+					}
+				}()
+				plan.RegisterExtension(field, extension)
+				return nil
+			}()
+			if err != nil {
+				c.Assert(err, ErrorMatches, testData.error)
+				continue nexttest
+			}
 		}
 
 		// Load the plan layer from disk (parse, combine and validate).
 		p, err := plan.ReadDir(layersDir)
-		if planTest.error != "" || err != nil {
+		if testData.error != "" || err != nil {
 			// Expected error.
-			c.Assert(err, ErrorMatches, planTest.error)
+			c.Assert(err, ErrorMatches, testData.error)
 		} else {
-			if _, ok := planTest.extensions[xField]; ok {
+			if _, ok := testData.extensions[xField]; ok {
 				// Verify "x-field" data.
 				var x *xSection
 				x = p.Section(xField).(*xSection)
 				c.Assert(err, IsNil)
-				c.Assert(x.Entries, DeepEquals, planTest.result.x.Entries)
+				c.Assert(x.Entries, DeepEquals, testData.result.x.Entries)
 			}
 
-			if _, ok := planTest.extensions[yField]; ok {
+			if _, ok := testData.extensions[yField]; ok {
 				// Verify "y-field" data.
 				var y *ySection
 				y = p.Section(yField).(*ySection)
 				c.Assert(err, IsNil)
-				c.Assert(y.Entries, DeepEquals, planTest.result.y.Entries)
+				c.Assert(y.Entries, DeepEquals, testData.result.y.Entries)
 			}
 
 			// Verify combined plan YAML.
 			planYAML, err := yaml.Marshal(p)
 			c.Assert(err, IsNil)
-			c.Assert(string(planYAML), Equals, planTest.resultYaml)
+			c.Assert(string(planYAML), Equals, testData.resultYaml)
 		}
 
 		// Unregister test extensions.
-		for field, _ := range planTest.extensions {
+		for field, _ := range testData.extensions {
 			plan.UnregisterExtension(field)
 		}
 	}
--- a/internals/plan/extensions_test.go
+++ b/internals/plan/extensions_test.go
@@ -353,7 +353,7 @@ func (s *S) TestPlanExtensions(c *C) {
 	defer func() {
 		// Remove remaining registered extensions.
 		for _, field := range registeredExtensions {
-			plan.UnregisterExtension(field)
+			plan.UnregisterSectionExtension(field)
 		}
 	}()
 
@@ -363,7 +363,7 @@ nexttest:
 
 		// Unregister extensions from previous test iteraton.
 		for _, field := range registeredExtensions {
-			plan.UnregisterExtension(field)
+			plan.UnregisterSectionExtension(field)
 		}
 		registeredExtensions = []string{}
 
@@ -380,7 +380,7 @@ nexttest:
 						err = fmt.Errorf("%v", r)
 					}
 				}()
-				plan.RegisterExtension(e.field, e.ext)
+				plan.RegisterSectionExtension(e.field, e.ext)
 				registeredExtensions = append(registeredExtensions, e.field)
 				return nil
 			}()
@@ -430,11 +430,11 @@ nexttest:
 // registration and follows the built-in sections which are ordered
 // the same way they are defined in the Plan struct.
 func (s *S) TestSectionOrderExt(c *C) {
-	plan.RegisterExtension("x-field", &xExtension{})
-	plan.RegisterExtension("y-field", &yExtension{})
+	plan.RegisterSectionExtension("x-field", &xExtension{})
+	plan.RegisterSectionExtension("y-field", &yExtension{})
 	defer func() {
-		plan.UnregisterExtension("x-field")
-		plan.UnregisterExtension("y-field")
+		plan.UnregisterSectionExtension("x-field")
+		plan.UnregisterSectionExtension("y-field")
 	}()
 
 	layer, err := plan.ParseLayer(1, "label", reindent(`
--- a/internals/plan/extensions_test.go
+++ b/internals/plan/extensions_test.go
@@ -44,7 +44,7 @@ type planResult struct {
 
 type extension struct {
 	field string
-	ext   plan.LayerSectionExtension
+	ext   plan.SectionExtension
 }
 
 var extensionTests = []struct {
@@ -349,16 +349,30 @@ var extensionTests = []struct {
 }}
 
 func (s *S) TestPlanExtensions(c *C) {
+	registeredExtensions := []string{}
+	defer func() {
+		// Remove remaining registered extensions.
+		for _, field := range registeredExtensions {
+			plan.UnregisterExtension(field)
+		}
+	}()
+
 nexttest:
 	for testIndex, testData := range extensionTests {
 		c.Logf("TestPlanExtensions :: %s (data index %v)", testData.summary, testIndex)
+
+		// Unregister extensions from previous test iteraton.
+		for _, field := range registeredExtensions {
+			plan.UnregisterExtension(field)
+		}
+		registeredExtensions = []string{}
 
 		// Write layers to test directory.
 		layersDir := filepath.Join(c.MkDir(), "layers")
 		s.writeLayerFiles(c, layersDir, testData.layers)
 		var p *plan.Plan
 
-		// Register test extensions.
+		// Register extensions for this test iteration.
 		for _, e := range testData.extensions {
 			err := func() (err error) {
 				defer func() {
@@ -367,6 +381,7 @@ nexttest:
 					}
 				}()
 				plan.RegisterExtension(e.field, e.ext)
+				registeredExtensions = append(registeredExtensions, e.field)
 				return nil
 			}()
 			if err != nil {
@@ -380,37 +395,33 @@ nexttest:
 		if testData.error != "" || err != nil {
 			// Expected error.
 			c.Assert(err, ErrorMatches, testData.error)
-		} else {
-			if slices.ContainsFunc(testData.extensions, func(n extension) bool {
-				return n.field == xField
-			}) {
-				// Verify "x-field" data.
-				var x *xSection
-				x = p.Section(xField).(*xSection)
-				c.Assert(err, IsNil)
-				c.Assert(x.Entries, DeepEquals, testData.result.x.Entries)
-			}
+			continue nexttest
+		}
 
-			if slices.ContainsFunc(testData.extensions, func(n extension) bool {
-				return n.field == yField
-			}) {
-				// Verify "y-field" data.
-				var y *ySection
-				y = p.Section(yField).(*ySection)
-				c.Assert(err, IsNil)
-				c.Assert(y.Entries, DeepEquals, testData.result.y.Entries)
-			}
-
-			// Verify combined plan YAML.
-			planYAML, err := yaml.Marshal(p)
+		if slices.ContainsFunc(testData.extensions, func(n extension) bool {
+			return n.field == xField
+		}) {
+			// Verify "x-field" data.
+			var x *xSection
+			x = p.Sections[xField].(*xSection)
 			c.Assert(err, IsNil)
-			c.Assert(string(planYAML), Equals, testData.resultYaml)
+			c.Assert(x.Entries, DeepEquals, testData.result.x.Entries)
 		}
 
-		// Unregister test extensions.
-		for _, e := range testData.extensions {
-			plan.UnregisterExtension(e.field)
+		if slices.ContainsFunc(testData.extensions, func(n extension) bool {
+			return n.field == yField
+		}) {
+			// Verify "y-field" data.
+			var y *ySection
+			y = p.Sections[yField].(*ySection)
+			c.Assert(err, IsNil)
+			c.Assert(y.Entries, DeepEquals, testData.result.y.Entries)
 		}
+
+		// Verify combined plan YAML.
+		planYAML, err := yaml.Marshal(p)
+		c.Assert(err, IsNil)
+		c.Assert(string(planYAML), Equals, testData.resultYaml)
 	}
 }
 
@@ -506,10 +517,10 @@ func (s *S) writeLayerFiles(c *C, layersDir string, inputs []*inputLayer) {
 
 const xField string = "x-field"
 
-// xExtension implements the LayerSectionExtension interface.
+// xExtension implements the SectionExtension interface.
 type xExtension struct{}
 
-func (x xExtension) ParseSection(data yaml.Node) (plan.LayerSection, error) {
+func (x xExtension) ParseSection(data yaml.Node) (plan.Section, error) {
 	xs := &xSection{}
 	err := data.Decode(xs)
 	if err != nil {
@@ -524,7 +535,7 @@ func (x xExtension) ParseSection(data yaml.Node) (plan.LayerSection, error) {
 	return xs, nil
 }
 
-func (x xExtension) CombineSections(sections ...plan.LayerSection) (plan.LayerSection, error) {
+func (x xExtension) CombineSections(sections ...plan.Section) (plan.Section, error) {
 	xs := &xSection{}
 	for _, section := range sections {
 		err := xs.Combine(section)
@@ -537,10 +548,10 @@ func (x xExtension) CombineSections(sections ...plan.LayerSection) (plan.LayerSe
 
 func (x xExtension) ValidatePlan(p *plan.Plan) error {
 	var xs *xSection
-	xs = p.Section(xField).(*xSection)
+	xs = p.Sections[xField].(*xSection)
 	if xs != nil {
 		var ys *ySection
-		ys = p.Section(yField).(*ySection)
+		ys = p.Sections[yField].(*ySection)
 
 		// Test dependency: Make sure every Y field in X refer to an existing Y entry.
 		for xEntryField, xEntryValue := range xs.Entries {
@@ -583,7 +594,7 @@ func (xs *xSection) IsZero() bool {
 	return xs.Entries == nil
 }
 
-func (xs *xSection) Combine(other plan.LayerSection) error {
+func (xs *xSection) Combine(other plan.Section) error {
 	otherxSection, ok := other.(*xSection)
 	if !ok {
 		return fmt.Errorf("cannot combine incompatible section type")
@@ -645,10 +656,10 @@ func (x *X) Merge(other *X) {
 
 const yField string = "y-field"
 
-// yExtension implements the LayerSectionExtension interface.
+// yExtension implements the SectionExtension interface.
 type yExtension struct{}
 
-func (y yExtension) ParseSection(data yaml.Node) (plan.LayerSection, error) {
+func (y yExtension) ParseSection(data yaml.Node) (plan.Section, error) {
 	ys := &ySection{}
 	err := data.Decode(ys)
 	if err != nil {
@@ -663,7 +674,7 @@ func (y yExtension) ParseSection(data yaml.Node) (plan.LayerSection, error) {
 	return ys, nil
 }
 
-func (y yExtension) CombineSections(sections ...plan.LayerSection) (plan.LayerSection, error) {
+func (y yExtension) CombineSections(sections ...plan.Section) (plan.Section, error) {
 	ys := &ySection{}
 	for _, section := range sections {
 		err := ys.Combine(section)
@@ -701,7 +712,7 @@ func (ys *ySection) IsZero() bool {
 	return ys.Entries == nil
 }
 
-func (ys *ySection) Combine(other plan.LayerSection) error {
+func (ys *ySection) Combine(other plan.Section) error {
 	otherySection, ok := other.(*ySection)
 	if !ok {
 		return fmt.Errorf("cannot combine incompatible section type")
--- a/internals/plan/extensions_test.go
+++ b/internals/plan/extensions_test.go
@@ -190,7 +190,7 @@ var extensionTests = []struct {
 					a: a
 					b: b`,
 	}},
-	error: "cannot validate layer section .* cannot accept entry not starting .*",
+	error: ".*cannot accept entry not starting.*",
 }, {
 	summary: "Load file layers with section validation failure #2",
 	extensions: []extension{{
@@ -209,7 +209,7 @@ var extensionTests = []struct {
 			x-field:
 				x1:`,
 	}},
-	error: "cannot validate layer section .* cannot have nil entry .*",
+	error: ".*cannot have nil entry.*",
 }, {
 	summary: "Load file layers failed plan validation",
 	extensions: []extension{{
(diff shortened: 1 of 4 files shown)


internals/plan/plan.go
```diff
@@ -32,6 +32,31 @@ import (
 	"github.com/canonical/pebble/internals/osutil"
 )
 
+// LayerSectionExtension allows the plan layer schema to be extended without
+// adding centralised schema knowledge to the plan library.
+type LayerSectionExtension interface {
```
flotter, line 37: Question from Ben: Could this simply become ```plan.Extension``` ?

internals/plan/plan.go
```diff
@@ -32,6 +32,31 @@ import (
...
+	// result of combining the supplied sections in order.
+	CombineSections(sections ...LayerSection) (LayerSection, error)
+
+	// ValidatePlan takes the complete plan as input, and allows the
+	// extension to validate the plan. This can be used for cross section
+	// dependency validation.
+	ValidatePlan(plan *Plan) error
+}
+
+type LayerSection interface {
```
flotter, line 52: Question from Ben: Could this simply become ```plan.Section``` ?

internals/plan/plan.go
```diff
@@ -42,11 +67,80 @@ const (
...
+
+	sectionVal := reflect.ValueOf(section)
+	sectionType := sectionVal.Type()
+	outValPtrType := outVal.Elem().Type()
+	if !sectionType.AssignableTo(outValPtrType) {
+		return fmt.Errorf("cannot assign value of type %s to out argument of type %s", sectionType, outValPtrType)
+	}
+	outVal.Elem().Set(sectionVal)
+	return nil
+}
```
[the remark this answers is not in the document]
flotter: Done. Resolving.

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,680 @@
...
+	x *xSection
+	y *ySection
+}
+
+var extensionTests = []struct {
+	extensions             map[string]plan.LayerSectionExtension
+	layers                 []*inputLayer
+	errorString            string
+	combinedPlanResult     *planResult
+	combinedPlanResultYaml string
```
benhoyt, line 49: These field names are quite verbose. What about:
```suggestion
	error      string
	result     *planResult
	resultYAML string
```

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,680 @@
...
+			{}`)),
+	},
+	// Index 2: Load file layers invalid section
+	{
+		extensions: map[string]plan.LayerSectionExtension{
+			"x-field": &xExtension{},
+			"y-field": &yExtension{},
+		},
+		layers: []*inputLayer{
+			&inputLayer{
```
benhoyt, line 76: I *think* you can drop the `&inputLayer` on these lines, as the type above already specifies it.

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,680 @@
...
+					summary: y
+					description: desc-y
+				`,
+			},
+		},
+		combinedPlanResult: &planResult{
+			x: &xSection{},
+			y: &ySection{},
+		},
+		combinedPlanResultYaml: string("{}\n"),
```
benhoyt, line 168:
```suggestion
		combinedPlanResultYaml: "{}\n",
```

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,680 @@
...
+}
+
+var extensionTests = []struct {
+	extensions             map[string]plan.LayerSectionExtension
+	layers                 []*inputLayer
+	errorString            string
+	combinedPlanResult     *planResult
+	combinedPlanResultYaml string
+}{
+	// Index 0: No Sections
```
benhoyt, line 51: Instead of using indexes here, what about a test summary field like `plan_test` does? Like `summary: "No Sections"`. Seems a bit more human-friendly.

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,680 @@
...
+
+func (xs *xSection) Combine(other plan.LayerSection) error {
+	otherxSection, ok := other.(*xSection)
+	if !ok {
+		return fmt.Errorf("cannot combine incompatible section type")
+	}
+
+	for field, entry := range otherxSection.Entries {
+		xs.Entries = makeMapIfNil(xs.Entries)
+		switch entry.Override {
```
benhoyt, line 508: It would be nice to be able to unify the merge/override logic somehow, so each extension didn't have to reimplement it. Any ideas? But maybe that's a pipe dream.

benhoyt (commented):
I really like this direction -- thanks. Left a few comments.

benhoyt:
One other thing: it'd be quite interesting to me if instead of having builtin sections we can just implement "services" etc using `RegisterExtension` (and register those sections in `init()` by default). That would really prove the concept and avoid the special-casing. Would probably be best in a follow-up PR though to avoid too much churn here.

internals/plan/plan.go
```diff
@@ -32,6 +32,31 @@ import (
...
+	// extension to validate the plan. This can be used for cross section
+	// dependency validation.
+	ValidatePlan(plan *Plan) error
+}
+
+type LayerSection interface {
+	// Ask the section to validate itself.
+	Validate() error
+
+	// Returns true if the section is empty.
```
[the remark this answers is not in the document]
flotter: Improved as per proposal. Resolving.

internals/plan/plan.go
```diff
@@ -32,6 +32,31 @@ import (
...
+	CombineSections(sections ...LayerSection) (LayerSection, error)
+
+	// ValidatePlan takes the complete plan as input, and allows the
+	// extension to validate the plan. This can be used for cross section
+	// dependency validation.
+	ValidatePlan(plan *Plan) error
+}
+
+type LayerSection interface {
+	// Ask the section to validate itself.
```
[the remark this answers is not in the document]
flotter: Improved as per proposal. Resolving.

internals/plan/plan.go
```diff
@@ -42,11 +67,61 @@ const (
...
 )
 
+// layerExtensions keeps a map of registered extensions.
+var layerExtensions = map[string]LayerSectionExtension{}
+
+// RegisterExtension adds a plan schema extension. All registrations must be
+// done before the plan library is used.
+func RegisterExtension(field string, ext LayerSectionExtension) error {
+	if _, ok := layerExtensions[field]; ok {
+		return fmt.Errorf("internal error: extension %q already registered", field)
```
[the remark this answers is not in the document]
flotter: Applied. Thanks. Resolving.

internals/plan/plan.go
```diff
@@ -42,11 +67,61 @@ const (
...
+// will return nil.
+func (p *Plan) Section(field string) LayerSection {
+	return p.Sections[field]
+}
+
+// MarshalYAML implements an override for top level omitempty tags handling.
+// This is required since Sections are based on an inlined map, for which
+// omitempty and inline together is not currently supported.
+func (p *Plan) MarshalYAML() (interface{}, error) {
+	marshalData := make(map[string]interface{})
```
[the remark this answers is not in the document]
flotter: Improved. Thanks. Resolving.

internals/plan/plan.go
```diff
@@ -559,7 +636,29 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
...
+		for _, layer := range layers {
+			if section := layer.Sections[field]; section != nil {
+				sections = append(sections, section)
+			}
+		}
+		var err error
+		combined.Sections[field], err = extension.CombineSections(sections...)
+		if err != nil {
+			return nil, &FormatError{
+				Message: fmt.Sprintf(`cannot combine section %q: %v`, field, err),
```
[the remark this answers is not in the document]
flotter: Thanks. Fixed. Resolving.

internals/plan/plan.go
```diff
@@ -1085,19 +1200,84 @@ func (p *Plan) checkCycles() error {
...
+			// with KnownFields = true behaviour. Once one of the proposals get
+			// merged, we can remove the intermediate Marshal step.
+			// https://github.com/go-yaml/yaml/issues/460
+			data, err := yaml.Marshal(&section)
+			if err != nil {
+				return nil, fmt.Errorf("internal error: cannot marshal %v section: %w", field, err)
+			}
+			dec := yaml.NewDecoder(bytes.NewReader(data))
+			dec.KnownFields(true)
+			if err = dec.Decode(builtinSections[field]); err != nil {
```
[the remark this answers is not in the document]
flotter: We cannot update the "copy" ```builtin```, so I have to index the map again here to update the map itself. (tests break to moment I change this).

internals/plan/plan.go
```diff
@@ -1085,19 +1200,84 @@ func (p *Plan) checkCycles() error {
...
+				return nil, &FormatError{
+					Message: fmt.Sprintf("cannot parse layer %q section %q: %v", label, field, err),
+				}
+			}
+		} else {
+			if extension, ok := layerExtensions[field]; ok {
```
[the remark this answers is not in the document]
flotter: Much nicer, thanks. Applied. Resolving.

internals/plan/extensions_test.go
flotter: Done. Thanks. Resolving.

internals/plan/extensions_test.go
flotter: Nice. Applied. Resolving.

internals/plan/extensions_test.go
flotter: Applied in multiple places. Resolving.

internals/plan/extensions_test.go
flotter: This is just a personal preference, but I am happy to change it. This is how I use it:

If I see an unit test failure, the last message contains an index, e.g:

```
Running TestPlanExtensions with test data index 3"
```

All I have to do now is append the index to the unit test data set:

```
var extensionTests = []struct {
    extensions map[string]plan.LayerSectionExtension
    layers     []*inputLayer
    error      string
    result     *planResult
    resultYaml string
}{
  : // lots and lots of test cases
}[3:]  // <=======
```

and when I delve debug the issue, I get to immediately debug the failed test.

```
dlv exec plan.test -- -check.f TestPlanExtensions
> break TestPlanExtensions
```

So both the test name, and data index logged helps me to immediately debug the issue.

Without an index, the offset within the test data set is not immediately available.

Let me know if this is a good enough reason to try something different. Happy to revert back to a description only apporach.

internals/plan/extensions_test.go
flotter: You are right that this block repeats itself somewhat with each section implementation.

What I have found, however, while implementing a new extension based on a mature and complex existing schema, is that current Pebble examples where a section is simply a top level ```map[string]*Entry``` structure is not always the case. For example, you could easily have a extension that looks something like this at the top level:

```
type foo struct {
    bar string
    baz map[string]EntryX
}
```

where at the Pebble plan root level, it is positioned as this:

```
sections map[string]section `yaml:",inline"`
```

so that the yaml will look like this:

```
services:
   srv1:
     override:
      :
foo:
   bar: something
   baz:
      zzz1:
        override:
        :
```

So the override levels are not necessarily at the same depth, and the example above have a top level scalar.

flotter:
> One other thing: it'd be quite interesting to me if instead of having builtin sections we can just implement "services" etc using `RegisterExtension` (and register those sections in `init()` by default). That would really prove the concept and avoid the special-casing. Would probably be best in a follow-up PR though to avoid too much churn here.

This is definitely the next step, and should not be difficult to do. I would be super happy to make this happen in followup PRs. I do feel confident (after implementing a quite complex extension), that the existing structures will be trivial to map to extensions.

internals/plan/extensions_test.go
benhoyt: Interesting approach. I'd probably a conditional breakpoint for this. I use GoLand, though it looks with Delve you can set a [condition](https://github.com/go-delve/delve/tree/master/Documentation/cli#condition) on your breakpoints too: `condition <breakpoint name or id> test.summary == "Foo bar"`.

Or if you're changing your code anyway, you could add an `if test.summary != "Foo Bar" { continue }` line in the test loop.

If the above helps, great, let's use the non-index approach. But if you'd prefer to leave it as is, I don't have strong feelings.

internals/plan/plan.go
```diff
@@ -42,11 +67,60 @@ const (
...
+// layerExtensions keeps a map of registered extensions.
+var layerExtensions = map[string]LayerSectionExtension{}
+
+// RegisterExtension adds a plan schema extension. All registrations must be
+// done before the plan library is used.
+func RegisterExtension(field string, ext LayerSectionExtension) {
+	if _, ok := layerExtensions[field]; ok {
+		panic(fmt.Sprintf("internal error: extension %q already registered", field))
+	}
+	layerExtensions[field] = ext
```
hpidcock, line 79: We need to panic if the name clashes with a pebble field name.

hpidcock (approved):
Thanks @flotter for putting this together. I'm not totally a fan of this method of extending pebble. I understand that this is probably the cleanest way to achieve this currently. My fear is that we need to have additional extensions in other parts of pebble for different super projects and their needs. We can probably cross that bridge when we get to it, hence why I am approving this now.

My initial thoughts on an alternative path to achieve the same thing would be through the use of a stub extensions module. Pebble would import "github.com/canonical/pebble/extensions", this would be a separate go module.

Then through composition, we can achieve something similar.
```
type Layer struct {
	extensions.Layer
	Order       int                   `yaml:"-"`
	Label       string                `yaml:"-"`
	Summary     string                `yaml:"summary,omitempty"`
	Description string                `yaml:"description,omitempty"`
	Services    map[string]*Service   `yaml:"services,omitempty"`
	Checks      map[string]*Check     `yaml:"checks,omitempty"`
	LogTargets  map[string]*LogTarget `yaml:"log-targets,omitempty"`
}
```

The "github.com/canonical/pebble/extensions" module on the pebble project would be a series of empty structs, interfaces and noop functions. Pebble would embed these structs and interfaces as well as call the stub functions in key places.

Then in the super project, you can use go work/go mod replace directives to achieve redirecting "github.com/canonical/pebble/extensions" to the super project's implementation.

There are obviously a host of issues with this approach (mostly around testing), but that's the best I could come up with to challenge what this PR is proposing.

internals/plan/extensions_test.go
flotter: I've changed to the summary approach, but the message printed by each test (visible on failure) shows:

```c.Logf("TestPlanExtensions :: %s (data index %v)", testData.summary, testIndex)```

So now its the same as plan_test, but I do still add a hint on the data index. This is more consistent, thanks.

Comment (flotter) [reply internals/plan/extensions_test.go:51]: Resolving.

File: internals/plan/extensions_test.go (hunk 8 above)
Comment (flotter) [reply internals/plan/extensions_test.go:505]: Resolving.

File: internals/plan/plan.go (hunk 16 above)
Comment (flotter) [reply internals/plan/plan.go:79]: Yes! Fixed, thank you! Added a unit test as well. Resolving.

File: internals/plan/plan.go (hunk 14 above)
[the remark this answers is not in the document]
Comment (flotter) [reply internals/plan/plan.go:1259]: I am closing this for now as per my previous comment. Please let me know if I did something stupid here. Resolving.

File: internals/plan/plan.go [hunk 17]
```diff
@@ -1085,19 +1199,85 @@ func (p *Plan) checkCycles() error {
...
+	//
+	// 1. Extended sections are YAML inlined, and also do not have a
+	// concrete type at this level, we cannot simply unmarshal the layer
+	// in one step.
+	//
+	// 2. We honor KnownFields = true behaviour for non extended schema
+	// sections, and at the top field level, which includes Section field
+	// names.
+	builtinSections := map[string]interface{}{
+		"summary":     &layer.Summary,
```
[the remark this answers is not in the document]
flotter: Could you take a look what I did in the last commit to solve this ? Please let me know if this could work, or if you have a different proposal. I chose a solution that also uses reflection in the unit test.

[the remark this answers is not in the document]
flotter: https://github.com/canonical/pebble/pull/488/commits/d3ae02cf61a972f2ea377ab2b8e8af6f55b85d90

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,687 @@
...
+
+var extensionTests = []struct {
+	summary    string
+	extensions map[string]plan.LayerSectionExtension
+	layers     []*inputLayer
+	error      string
+	result     *planResult
+	resultYaml string
+}{
+	{
```
niemeyer, line 52: There's an unnecessary indendation on every line below:
```
}{{
        summary: ...
}, {
    ...
}}

File: internals/plan/extensions_test.go [hunk 19]
```diff
@@ -0,0 +1,687 @@
...
+		},
+		resultYaml: "{}\n",
+	}, {
+		summary: "Load file layers invalid section",
+		extensions: map[string]plan.LayerSectionExtension{
+			"x-field": &xExtension{},
+			"y-field": &yExtension{},
+		},
+		layers: []*inputLayer{
+			{
```
niemeyer, line 79: Same here: in cases like that the two lines should be merged so the outcome is cleaner:
```
layers: []*inputLayer{{
        order:1,
        ...
}},
```

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,687 @@
...
+				order: 1,
+				label: "layer-1",
+				yaml: `
+					summary: xy
+					description: desc
+				`,
+			},
+			{
```
niemeyer, line 105: Same here:
```
}. {
```

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,687 @@
...
+
+			// Verify combined plan YAML.
+			planYAML, err := yaml.Marshal(p)
+			c.Assert(err, IsNil)
+			c.Assert(string(planYAML), Equals, testData.resultYaml)
+		}
+
+		// Unregister test extensions.
+		for field, _ := range testData.extensions {
+			plan.UnregisterExtension(field)
```
niemeyer, line 412: This cannot be here, I believe, as any Assert above will leave the runtime in a bad state for further tests. Such cases need to be on a defer, or as part of the suite tear down functions.

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,687 @@
...
+				c.Assert(err, ErrorMatches, testData.error)
+				continue nexttest
+			}
+		}
+
+		// Load the plan layer from disk (parse, combine and validate).
+		p, err := plan.ReadDir(layersDir)
+		if testData.error != "" || err != nil {
+			// Expected error.
+			c.Assert(err, ErrorMatches, testData.error)
```
niemeyer, line 386: May use continue nexttest here too and deindent.  The assumption that termination code will happen anyway due to the else isn't sound because any Assert in the branch below will break the flow too.

internals/plan/plan.go
niemeyer: `SectionExtension` seems good. This is hooked into the system for a particular field, which we call a section, so it makes sense.  The function name should be renamed too, as RegisterSectionExtension, so we have the same terminology throughout the surface.

internals/plan/plan.go
```diff
@@ -42,11 +68,68 @@ const (
...
 	LogTargets map[string]*LogTarget `yaml:"log-targets,omitempty"`
+
+	Sections map[string]LayerSection `yaml:",inline"`
+}
+
+// Section retrieves a section from the plan. If Section is called
+// before the plan is loaded, or with an unregistered field, this method
+// will return nil.
+func (p *Plan) Section(field string) LayerSection {
+	return p.Sections[field]
```
niemeyer, line 110: What's the purpose of this method when its implementation is public and that trivial?

internals/plan/plan.go
```diff
@@ -42,11 +68,68 @@ const (
...
+	delete(layerExtensions, field)
+}
+
 type Plan struct {
 	Layers     []*Layer              `yaml:"-"`
 	Services   map[string]*Service   `yaml:"services,omitempty"`
 	Checks     map[string]*Check     `yaml:"checks,omitempty"`
 	LogTargets map[string]*LogTarget `yaml:"log-targets,omitempty"`
+
+	Sections map[string]LayerSection `yaml:",inline"`
```
niemeyer, line 103: This seems a bit awkward to read. I realize it's mimicking what was there before, with Services, Checks, etc, but now it seems even more weird as we have Layers, Sections, and individual Sections in the same level. This is okay if it's temporary, but is it?  And if so, we need a comment in here properly spelling out the plan so the next person can make sense of the current state. Can we do better so it's more sensible meanwhile?

internals/plan/plan.go
```diff
@@ -825,11 +932,18 @@ func (layer *Layer) Validate() error {
 		}
 	}
 
+	for field, section := range layer.Sections {
+		err := section.Validate()
+		if err != nil {
+			return fmt.Errorf("cannot validate layer section %q: %w", field, err)
```
niemeyer, line 938: That feels like unnecessary verbosity given the context. Consider the errors above, for example:
```
fmt.Sprintf("plan check %q timeout must not be zero", name),
```
This is the sort of error that a section extension should emit, so there's no need to prefix these and get:

> cannot validate layer section "checks": plan check %q timeout must not be zero

for example.

IOW, this is "our turf", so let's make sure the extensions are emitting pleasant messages.

internals/plan/plan.go
```diff
@@ -917,6 +1031,15 @@ func (p *Plan) Validate() error {
 	if err != nil {
 		return err
 	}
+
+	// Each section extension must validate the combined plan.
+	for field, extension := range layerExtensions {
+		err = extension.ValidatePlan(p)
+		if err != nil {
+			return fmt.Errorf("cannot validate plan section %q: %w", field, err)
```
niemeyer, line 1039: Same here.

internals/plan/plan.go
```diff
@@ -1085,19 +1208,90 @@ func (p *Plan) checkCycles() error {
...
+	builtinSections := map[string]interface{}{
+		"summary":     &layer.Summary,
+		"description": &layer.Description,
+		"services":    &layer.Services,
+		"checks":      &layer.Checks,
+		"log-targets": &layer.LogTargets,
+	}
+	// Make sure builtinSections contains the exact same fields as expected
+	// in the Layer type.
+	if !mapHasKeys(builtinSections, layerBuiltins) {
```
niemeyer, line 1236: mapHasKeys doesn't sound like a bijective assertion - that is, typically hasKey is true if the key is present, not if it is the only key. That's how I read this function and was confused about how it satisfied the comment above.

internals/plan/plan.go
```diff
@@ -1125,7 +1319,21 @@ func ParseLayer(order int, label string, data []byte) (*Layer, error) {
...
+}
+
+// mapHasKeys returns true if the key list supplied is an exact match of the
+// keys in the map (ordering is ignored).
+func mapHasKeys[M ~map[K]V, K comparable, V any](inMap M, keyList []K) bool {
+	if len(inMap) != len(keyList) {
+		return false
+	}
+	for _, key := range keyList {
+		if _, ok := inMap[key]; !ok {
```
niemeyer, line 1332: FWIW, this implementation is not fool-proof: if you have N times the exact same key in keyList, where N matches len(inMap), this function still returns true.

niemeyer (approved):
Thanks for this, Fred.  Here is a final pass from my side with details only.  The overall aspect looks good, and I'm happy once you and the team are happy.

benhoyt:
Thanks Harry and Gustavo for your reviews. @flotter I spoke a bit further with Harry about his concerns, as I'd misunderstood. He's basically concerned that we'll start to accumulate more and more dynamic hooks and ways to extend Pebble, which don't actually help Pebble itself, and will make the codebase less simple/static and hence harder to understood. There are other ways to skin the cat (that still require complexity, but they might push it from dynamic/runtime complexity to version control and workspace complexity) that we may want to think about in future.

I do share this concern, but we both agreed that this is reasonable for now, and that we'll keep an eye on it over time.

internals/plan/plan.go
```diff
@@ -42,11 +68,68 @@ const (
...
+// before the plan is loaded, or with an unregistered field, this method
+// will return nil.
+func (p *Plan) Section(field string) LayerSection {
+	return p.Sections[field]
+}
+
+// MarshalYAML implements an override for top level omitempty tags handling.
+// This is required since Sections are based on an inlined map, for which
+// omitempty and inline together is not currently supported.
+func (p *Plan) MarshalYAML() (interface{}, error) {
```
flotter, line 116: TODO Fred: Introducing a map here makes the ordering random.

flotter: If we create a structure with reflection, this can be done in a more defined way - https://github.com/ompluscator/dynamic-struct

flotter: https://pkg.go.dev/reflect#StructOf

internals/plan/extensions_test.go
flotter: Fixed, thank you. Resolving.

internals/plan/extensions_test.go
flotter: Fixed, thank you. Resolving.

internals/plan/extensions_test.go
flotter: Fixed as part of previous comment. Resolving.

internals/plan/extensions_test.go
flotter: Done. Much better, thank you. Resolving.

internals/plan/extensions_test.go
flotter: I fixed this so that extension removal now happens irrespective of the test iteration result. Resolving.

internals/plan/plan.go
flotter: Applied ```SectionExtension``` name. Resolving.

internals/plan/plan.go
flotter: Applied change to ```plan.Section```. It also feels a bit more accurate as sections to appear directly in the Plan, not only a layer.

flotter: Resolving.

internals/plan/plan.go
flotter: I removed this as rightfully it no longer really serves a purpose. Thanks you. Resolving.

internals/plan/plan.go
flotter: I previously played with having:

```
type Plan struct {
	Layers     []*Layer              `yaml:"-"`
	combined *Layer               `yaml:",inline"
}
```

The main issue I observed was that you lose the ability to simply access content from the plan directly, unless you add helper methods, and this propagated far into the codebase:

```
func (p *Plan) Services() map[string]*Service {
    return p.combined.Services
}
```

If we do that, then perhaps waiting until we migrated the 3 built-in sections out, so that it looks like this:

```
type Plan struct {
	Layers     []*Layer              `yaml:"-"`
	Sections map[string]LayerSection `yaml:",inline"`
}
```

@benhoyt Perhaps we can brainstorm what could be nice here?

internals/plan/plan.go
flotter: This issue is now fixed, with two tests added to ensure we do not break ordering in the future. Resolving.

internals/plan/plan.go
flotter: I changed this to ```mapMatchKeys```.

@benhoyt Do you have a better proposal ?

internals/plan/plan.go
flotter: Fixed, thank you! Resolving.

internals/plan/plan.go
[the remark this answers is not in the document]
flotter: Resolving.

internals/plan/plan.go
flotter: Removed wrapping of the error. Thank you. Resolved.

internals/plan/plan.go
flotter: Removed wrapping of the error. Thank you. Resolved.

internals/plan/plan.go
```diff
@@ -42,11 +68,90 @@ const (
 	defaultCheckThreshold = 3
 )
 
+var (
+	// sectionExtensions keeps a map of registered extensions.
+	sectionExtensions = map[string]SectionExtension{}
+
+	// sectionExtensionsOrder records the order in which the extensions were registered.
+	sectionExtensionsOrder = []string{}
```
flotter, line 76: @benhoyt I experimented with a slice of structs to make ```sectionExtensions``` ordered, but reverted to just adding a separate slice for key ordering. I left it left the overall code in a simpler state, as ordering is only applicable at Marshal time, while map access simplifies the remainder of the code. Let me know what you think.

internals/plan/plan.go
```diff
@@ -42,11 +68,90 @@ const (
...
 	Checks     map[string]*Check     `yaml:"checks,omitempty"`
 	LogTargets map[string]*LogTarget `yaml:"log-targets,omitempty"`
+
+	Sections map[string]Section `yaml:",inline"`
+}
+
+// MarshalYAML implements an override for top level omitempty tags handling.
+// This is required since Sections are based on an inlined map, for which
+// omitempty and inline together is not currently supported.
+func (p *Plan) MarshalYAML() (interface{}, error) {
```
flotter, line 120: @benhoyt This new implementation of Marshal now insures the ordering of top level sections according to the discussed proposal. Let me know what you think.

benhoyt: This is definitely an interesting implementation, but I'm guessing there's no way other than `reflect` to tell the YAML package to order the fields (a map doesn't have order or maybe uses alpha order, and a regular struct is static). I didn't even know you could do this (build a brand new struct type dynamically).

internals/plan/plan.go
```diff
@@ -1085,19 +1228,88 @@ func (p *Plan) checkCycles() error {
...
+	builtins := map[string]interface{}{
+		"summary":     &layer.Summary,
+		"description": &layer.Description,
+		"services":    &layer.Services,
+		"checks":      &layer.Checks,
+		"log-targets": &layer.LogTargets,
+	}
+	// Make sure builtins contains the exact same fields as expected
+	// in the Layer type.
+	if !mapMatchKeys(builtins, builtinSections) {
```
benhoyt, line 1256: Is this something we even need to do at runtime, given it's a coding bug if we get it wrong? Or could we just have a test for it?

internals/plan/plan_test.go
```diff
@@ -2045,3 +2071,87 @@ func (s *S) TestStartStopOrderMultipleLanes(c *C) {
 	c.Assert(lanes[1], DeepEquals, []string{"srv2"})
 	c.Assert(lanes[2], DeepEquals, []string{"srv3"})
 }
+
+// TestLayerBuiltinCompatible ensures layerBuiltins used in the plan package
+// reflects the same YAML fields as exposed in the Layer type.
+func (s *S) TestLayerBuiltinCompatible(c *C) {
+	fields := structYamlFields(plan.Layer{})
```
benhoyt, line 2078: It seems to me it'd make a simpler and better test if we just hard-coded this as `[]string{"summary", "description", "services", "checks", "log-targets"}`. If we get out of sync the test will tell us. Here we have a lot of machinery to do something we can easily do by hand (and who's testing the machinery :-).

Also, you might find `testutil.DeepUnsortedMatches` useful.

internals/plan/plan_test.go
```diff
@@ -2045,3 +2071,87 @@ func (s *S) TestStartStopOrderMultipleLanes(c *C) {
...
+		return s
+	}
+	r := []rune(s)
+	r[0] = unicode.ToLower(r[0])
+	return string(r)
+}
+
+// TestSectionOrder ensures built-in section order is maintained
+// during Plan marshal operations.
+func (s *S) TestSectionOrder(c *C) {
```
benhoyt, line 2115: Thanks for adding this.

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,778 @@
...
+)
+
+type inputLayer struct {
+	order int
+	label string
+	yaml  string
+}
+
+// PlanResult represents the final content of a combined plan. Since this
+// package exclusively focuses extensions, all built-in sections are empty
```
benhoyt, line 38:
```suggestion
// test file exclusively focuses on extensions, all built-in sections are empty
```

internals/plan/extensions_test.go
```diff
@@ -0,0 +1,778 @@
...
+}
+
+var extensionTests = []struct {
+	summary    string
+	extensions []extension
+	layers     []*inputLayer
+	error      string
+	result     *planResult
+	resultYaml string
+}{{
```
benhoyt, line 57: Ah yes, the cuddled braces make this look much cleaner, thanks.

benhoyt (approved):
This looks good, thanks. A few minor comments, but I'm happy with the approach/naming/implementation overall.

internals/plan/plan.go
flotter: OK from Ben. Resolving.

internals/plan/plan.go
flotter: If we only had the known built-in keys to work with (and then this was not required in the first place), it would have been possible to simply do something like:

```
func (p *Plan) MarshalYAML() (interface{}, error) {
    // Inline type to redefine ordering
    layout := &struct {
        A interface{} `yaml:"check,omitempty"`
        B interface{} `yaml:"logtargets,omitempty"`
        C interface{} `yaml:"services,omitempty"`
    } {
        A: p.Checks,
        B: p.LogTargets,
        C: p.Services,
    }
    return layout
}
```

A/B/C above can be defined with the proper name and type as well, but since YAML processing is reflection magic in any case, you can go crazy as above, and it will still work perfectly.

However, the issue is simply we do not know at runtime which extensions were added, so we need dynamic creation of some type for which order is still maintained, i.e struct.

flotter: I believe this is the only way, but since we can improve this at any point in time, and Marshalling is only something we currently do on request from the API (and given YAML processing is reflection based in any case), I believe this is fine.

Resolving for now.

internals/plan/plan.go
flotter: Further feedback from Ben and further thinking, I've moved section field change detection purely into the test domain. Remove this runtime check, as at best it only covers some areas in the code, not all. Resolving.

internals/plan/plan.go
flotter: I've removed the runtime check. Also, since it does not cover all the hard-coded field cases, and settled with an updated test which fails on changes to the exported YAML field in both Layer and Plan types.

internals/plan/extensions_test.go
flotter: Fixed, Resolving.

internals/plan/plan_test.go
flotter: I've changed the comment on the test function to indicate this is simply a change detector for ```plan.Layer``` and ```plan.Plan```. The ```builtinSections``` list of fields is just one source of hard-coded fields, so only testing that could be a too limiting check to really be of use. Overall, this test may not be that useful in practise, is test coverage is pretty good, and failing to update internal fields will almost certainly result in other tests failing.

I've updated the test to use ```testutil.DeepUnsortedMatches```. Happy to further simplify or remove this in a followup PR. Resolving for now.

benhoyt:
Thanks! Let's go ahead and merge this -- we can have further discussion about planstate in #489.

benhoyt:
Actually one more thing from Gustavo above. In the renaming thread, he said:

> The function name should be renamed too, as RegisterSectionExtension, so we have the same terminology throughout the surface.

I think that's reasonable to keep things consistent -- so please make that tweak and then merge this.

PR #489, opened 2024-08-19
feat(planstate): add plan section support

flotter:
Built on top of https://github.com/canonical/pebble/pull/488.

Add section support to the Plan Manager (planstate).

### DIFF
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -48,6 +48,12 @@ func NewManager(layersDir string) (*PlanManager, error) {
 	return manager, nil
 }
 
+// RegisterExtension adds a plan extension which extends the plan schema. All
+// extensions must be registered before Load() is called.
+func (m *PlanManager) RegisterExtension(field string, ext plan.LayerSectionExtension) {
+	plan.RegisterExtension(field, ext)
+}
+
 // Load reads plan layers from the pebble directory, combines and validates the
 // final plan, and finally notifies registered managers of the plan update. In
 // the case of a non-existent layers directory, or no layers in the layers
@@ -181,6 +187,7 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 		Services:   combined.Services,
 		Checks:     combined.Checks,
 		LogTargets: combined.LogTargets,
+		Sections:   combined.Sections,
 	}
 	err = p.Validate()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internals/overlord/planstate/manager.go
```diff
@@ -48,6 +48,12 @@ func NewManager(layersDir string) (*PlanManager, error) {
 	return manager, nil
 }
 
+// RegisterExtension adds a plan extension which extends the plan schema. All
+// extensions must be registered before Load() is called.
+func (m *PlanManager) RegisterExtension(field string, ext plan.LayerSectionExtension) error {
```
benhoyt, line 53: This makes it look like registering an extension is tied to the plan manager, but it's really global. I think it'd be better for callers just to use `plan.RegisterExtension` directly so that that's clear.

flotter: Makes sense, thanks. Applied. Resolving.

internals/overlord/planstate/manager.go
```diff
@@ -181,6 +181,7 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 		Services:   combined.Services,
 		Checks:     combined.Checks,
 		LogTargets: combined.LogTargets,
+		Sections:   combined.Sections,
```
benhoyt, line 184: I don't quite understand this PR. Are we really adding hundreds of lines of tests just to test this 1 line of code, that `Sections` is wired up correctly? If so, I'd argue it's not worth it.

flotter: Hi Ben, just to make sure I understand your concern: The ```planstate``` manager builds on top of the ```plan``` package to implement the ```CombineLayers``` and ```AppendLayers``` functions (simple, but there are some additional logic added, which feels not ideal to not cover in a test). This is the reason the current planstate testsuite covers them, but only for built-in sections.

This change you commented on here was actually a result of the added tests failing (I discovered that only afterwards).

Instead of having separate ```Load```, ```CombineLayers``` and ```AppendLayers``` tests, I have now merged it with the built-in tests, which makes the delta much smaller. There is still however the need for a test extension to be define, which I kept as it.

Alternatively, I could simply include this line in the ```plan``` PR, and drop adding extension coverage at this level ?

benhoyt: Ah, I see now, thanks.

benhoyt (approved):

hpidcock (approved):
Thanks @flotter, I'm happy with the added testing too.

PR #490, opened 2024-08-20
Add test coverage to osutil.Mkdir()

holmanb:
I came across [this comment](https://github.com/canonical/pebble/blob/0dd6abc626f6cabd840a577bc1d1ecbb7f5a5af0/internals/osutil/mkdir.go#L26), while gaining familiarity with Pebble. I noticed that the mutex that prevents a race in `osutil.Mkdir()` doesn't appear to be tested. This adds a test which fails without the mutex.

Related: https://github.com/canonical/snapd/commit/390243e9fe809ecd2322feca9e232b8ea6f3a906

### DIFF
--- a/internals/osutil/mkdir_test.go
+++ b/internals/osutil/mkdir_test.go
@@ -15,6 +15,7 @@
 package osutil_test
 
 import (
+	"fmt"
 	"os"
 	"os/user"
 	"strconv"
@@ -162,6 +163,32 @@ func (mkdirSuite) TestMakeParentsAndChmod(c *check.C) {
 	info, err = os.Stat(tmpDir + "/foo/bar")
 	c.Assert(err, check.IsNil)
 	c.Assert(info.Mode().Perm(), check.Equals, os.FileMode(0o777))
+}
+
+// Concurrent calls to osutil.Mkdir must not race
+func (mkdirSuite) TestMakeParentsCollideAndChmod(c *check.C) {
+	tmpDir := c.MkDir()
+	results := make(chan error)
+	spread := 2
+	nFiles := 0
+	dirs := make([]string, spread*spread)
+	for pDir := range spread {
+		for cDir := range spread {
+			dir := fmt.Sprintf("%s/foo/%d/%d", tmpDir, pDir, cDir)
+			go func(path string) {
+				results <- osutil.Mkdir(path, 0o777, &osutil.MkdirOptions{MakeParents: true, Chmod: true})
+			}(dir)
+			dirs[nFiles] = dir
+			nFiles++
+		}
+	}
+	for range nFiles {
+		c.Assert(<-results, check.IsNil)
+	}
+	for _, dir := range dirs {
+		fmt.Printf(dir)
+		c.Assert(osutil.IsDir(dir), check.Equals, true)
+	}
 }
 
 func (mkdirSuite) TestMakeParentsAndNoChmod(c *check.C) {


benhoyt (commented):
Thanks for the contribution. I guess I'd prefer if this were submitted to snapd first, seeing they introduced the mutex code, then Pebble can merge it from there. (Though I know we've deviated from snapd quite a bit at this point.)

Also, how does the test work? Won't it sometimes succeed, sometimes fail (if osutil.Mkdir doesn't acquire the mutex), depending on the timing of the goroutine scheduling?

holmanb:
> Thanks for the contribution. I guess I'd prefer if this were submitted to snapd first, seeing they introduced the mutex code, then Pebble can merge it from there. (Though I know we've deviated from snapd quite a bit at this point.)

I'm having trouble wrapping my mind around the development model. I thought that since pebble had more recent changes to this code that Pebble's code would eventually get synced back to snapd - which now sounds like an invalid assumption. Regardless, as a single test I don't expect much trouble from the deviation - I'm happy to submit the change to snapd first if that's preferred. 

> Also, how does the test work?

I think you've got the right idea. It starts several goroutines which call Mkdir() with overlapping paths to attempt to force the race condition.

> Won't it sometimes succeed, sometimes fail (if osutil.Mkdir doesn't acquire the mutex), depending on the timing of the goroutine scheduling?

This will always succeed due to the existence of the mutex (as the code is written now). If the mutex is removed it would be possible for the test to still succeed if, for example, each goroutine were scheduled serially. Testing this locally[1], this test fails 100% of the time on my local machine with the mutex Lock/Unlock removed. I originally had the test running 100 goroutines (2 directories deep - 10 iterations per loop) but backed it off because that seems unnecessary to reliably produce test failures. I suspect that this reliably produces test failures due to syscall runtime being relative large compared to goroutine scheduler time.

A test like this that sometimes fails and sometimes passes has limited value. Racy behavior like this is usually difficult to test - and especially when the effects of the race occur external to the Go code. I've written tests before for race conditions that instrument / force a race, however that approach typically has the downside of being so tightly coupled with the implementation that is being tested that such a test isn't useful for future refactors since the test code cannot stand alone.

@benhoyt please let me know if you have any suggestions for a better test approach or if you think that the signal from a test of this nature aren't useful enough to be worth maintaining.

Aside: Regarding getting rid of the mutex altogether, I also think that we could make it more fine-grained (I have a few ideas on how to accomplish that), but the expectation that the function can allow concurrent callsites to modify the same paths appears to carry an ownership problem that requires some form of serialization - locking or otherwise.

[1] I recognize that goroutine scheduling is highly platform-dependent

benhoyt:
> I'm having trouble wrapping my mind around the development model. I thought that since pebble had more recent changes to this code that Pebble's code would eventually get synced back to snapd - which now sounds like an invalid assumption. 

Actually, you're probably right there. The `Mkdir` refactoring should probably get synced back to snapd. I'll open an issue for our team to put up a PR there to port those change back.

Okay, so it's a best-effort "fails most of the time" (without the mutex) kind of test. I've certainly used that kind of test before, and I don't have a problem with it per se.

> if you think that the signal from a test of this nature aren't useful enough to be worth maintaining

I guess that's where I'm at right now. I think it's a tricky test for a very simple thing, `mu.Lock(); defer mu.Unlock()`. I don't think the maintenance trade-off is worth it.

I'd probably prefer to expend our effort on whether we need the lock at all, or ways to get rid of it. Though even that I'd probably only do if there was a concrete need.

holmanb:
> Actually, you're probably right there. The `Mkdir` refactoring should probably get synced back to snapd. I'll open an issue for our team to put up a PR there to port those change back.

Thanks!

> Okay, so it's a best-effort "fails most of the time" (without the mutex) kind of test. I've certainly used that kind of test before, and I don't have a problem with it per se.
> 
> > if you think that the signal from a test of this nature aren't useful enough to be worth maintaining
> 
> I guess that's where I'm at right now. I think it's a tricky test for a very simple thing, `mu.Lock(); defer mu.Unlock()`. I don't think the maintenance trade-off is worth it.

That's reasonable. I will close this request.

> I'd probably prefer to expend our effort on whether we need the lock at all, or ways to get rid of it. Though even that I'd probably only do if there was a concrete need.

That's fine by me. Thanks for considering this contribution!

PR #493, opened 2024-08-27
refactor(state): avoid need for state lock in LatestWarningTime

### LINKED ISSUE #366: Avoid acquiring state lock in Command.ServeHTTP

benhoyt:
This means that not every request/response needs to acquire the state lock, as shown by the removal of the custom response type in the /v1/health endpoint.

Note that there's a test, TestHealthStateLockNotHeldSuccess, that specifically tests that the state lock is not held during /v1/health requests. If we accidentally introduce locking again, that will fail.

Fixes #366.

### DIFF
--- a/internals/overlord/state/notices.go
+++ b/internals/overlord/state/notices.go
@@ -418,7 +418,7 @@ func (s *State) flattenNotices(filter *NoticeFilter) []*Notice {
 func (s *State) unflattenNotices(flat []*Notice) {
 	now := time.Now()
 	s.notices = make(map[noticeKey]*Notice)
-	latestWarningTime := time.Time{}
+	var latestWarningTime time.Time
 	for _, n := range flat {
 		if n.expired(now) {
 			continue
--- a/internals/overlord/state/notices.go
+++ b/internals/overlord/state/notices.go
@@ -270,6 +270,13 @@ func (s *State) AddNotice(userID *uint32, noticeType NoticeType, key string, opt
 	notice.lastData = options.Data
 	notice.repeatAfter = options.RepeatAfter
 
+	// Update the latest warning time cache if needed. There's no need to
+	// actually update atomically here, because the state lock is held.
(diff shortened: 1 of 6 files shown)


IronCore864 (approved):

internals/overlord/state/notices.go
```diff
@@ -270,6 +270,13 @@ func (s *State) AddNotice(userID *uint32, noticeType NoticeType, key string, opt
 	notice.lastData = options.Data
 	notice.repeatAfter = options.RepeatAfter
 
+	// Update the latest warning time cache if needed. There's no need to
+	// actually update atomically here, because the state lock is held.
```
hpidcock, line 274: Yep, I was about to comment on that. Looks like you've already considered that. If we ever update this without the state mutex, then we'll need to reconsider this.

hpidcock (approved):
LGTM, thank-you for this follow up PR.

PR #494, opened 2024-08-27
ci: add a monthly (or on demand) TIOBE reporting workflow

tonyandrewmeyer:
Once a month, send the latest code quality (including coverage) data to TIOBE for central storage.

### DIFF
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -13,7 +13,7 @@ jobs:
 
       - name: Set up Go
         uses: actions/setup-go@v4
-  
+
       - name: Install dependencies
         run: |
           go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
@@ -23,7 +23,7 @@ jobs:
       # We could store a report from the regular run, but this is cheap to do and keeps this isolated.
       - name: Test and generate coverage report
         run: |
-          go test -race -coverprofile=coverage.out ./...
+          go test -coverprofile=coverage.out ./...
           gocov convert coverage.out > coverage.json
           # Annoyingly, the coverage.xml file needs to be in a .coverage folder.
           mkdir .coverage
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -17,8 +17,8 @@ jobs:
       - name: Install dependencies
         run: |
           go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
-          go get github.com/axw/gocov/gocov
-          go get github.com/AlekSi/gocov-xml
+          go install github.com/axw/gocov/gocov@v1.1.0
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
@benhoyt the TIOBE piece ought to be correct (although you'll need to add the secret here, as you did for `operator`) - I'm not entirely sure whether this is the way you'd best get an XML coverage report for Pebble or not, though, so review around that part would be particularly useful.

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install dependencies
+        run: |
+          go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
+          go get github.com/axw/gocov/gocov
+          go get github.com/AlekSi/gocov-xml
+
+      # We could store a report from the regular run, but this is cheap to do and keeps this isolated.
+      - name: Test and generate coverage report
+        run: |
+          go test -race -coverprofile=coverage.out ./...
```
benhoyt, line 26: Do we need the race detector on here? It may not matter, but it makes the tests run somewhat slower, so I'd probably take it out here.

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+  
+      - name: Install dependencies
+        run: |
+          go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
+          go get github.com/axw/gocov/gocov
+          go get github.com/AlekSi/gocov-xml
```
benhoyt, line 21: Should we `go install` these and pin them at a specific version too, like the staticcheck one?

benhoyt, line 21:
```suggestion
          go install github.com/axw/gocov/gocov@v1.1.0
          go install github.com/AlekSi/gocov-xml@v1.1.0
```

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+      # We could store a report from the regular run, but this is cheap to do and keeps this isolated.
+      - name: Test and generate coverage report
+        run: |
+          go test -race -coverprofile=coverage.out ./...
+          gocov convert coverage.out > coverage.json
+          # Annoyingly, the coverage.xml file needs to be in a .coverage folder.
+          mkdir .coverage
+          gocov-xml < coverage.json > .coverage/coverage.xml
```
benhoyt, line 30: I don't know these tools, but I also don't know a better way to do it, so this seems like a reasonable start to me.

benhoyt (commented):
A few minor comments. I've also added the token to the GitHub Actions secrets.

benhoyt (approved):
Thanks!

PR #497, opened 2024-09-04
test: add integration tests for "pebble run"

IronCore864:
A PoC to solve the Pebble integration test, see [issue here](https://github.com/canonical/pebble/issues/351).

Two issues I want to call out:

1. It seems the test suite can't be used with build flags, it can't collect any tests: `testing: warning: no tests to run`.

Maybe this is possible but I haven't made it work yet, and I think this could be the reason: as per the [Golang Doc](https://golang.org/pkg/go/build/), the build tag lists the conditions under which a file should be included in the package, and maybe this is the reason why it doesn't work with suite.

It means we can't do something like `func (s *IntegrationSuite) TestXXX(c *C)`, but can only do `func TestXXX(t *testing.T)`; which is OK I think, just something worth mentioning.

2. In the PoC I built the binary in setup, then created some functions to help create layers, run pebble and return logs. The logs part is tricky:

The daemon is a long-running process and we can't wait for it to "finish". Here I used something like "if there are no new logs in the past second, kill the process and return". It doesn't feel ideal to me.

I had another draft where I simply passed a value for timeout into it, sleep, then kill.

Both worked, but I'm not sure if this is the best we can do here.


internals/testintegration/pebble_run_test.go
```diff
@@ -0,0 +1,51 @@
...
+services:
+    demo-service:
+        override: replace
+        command: sleep 1000
+        startup: enabled
+    demo-service2:
+        override: replace
+        command: sleep 1000
+        startup: enabled
+`[1:]
```
dimaqq, line 36: Yaml is actually pretty chill wrt. initial indent.
This should work just as well:

```suggestion
	layerYAML := `---
        services:
            demo-service:
                override: replace
                command: sleep 1000
                startup: enabled
            demo-service2:
                override: replace
                command: sleep 1000
                startup: enabled`

```

internals/testintegration/utils.go
```diff
@@ -0,0 +1,105 @@
...
+				keywordFound = true
+				break
+			}
+		}
+		if !keywordFound {
+			notFound = append(notFound, keyword)
+		}
+	}
+
+	return len(notFound) == 0, notFound
```
dimaqq, line 51: My go is a bit rusty 🦀 but would it make more sense to return a slice?

```suggestion
	return len(notFound) == 0, []notFound
```

internals/testintegration/utils.go
```diff
@@ -0,0 +1,105 @@
...
+	stderrPipe, err := cmd.StderrPipe()
+	if err != nil {
+		t.Fatalf("Error creating stderr pipe: %v", err)
+	}
+
+	err = cmd.Start()
+	defer cmd.Process.Kill()
+	if err != nil {
+		t.Fatalf("Error starting 'pebble run': %v", err)
+	}
```
dimaqq, line 82: Should these be swapped, like so:

```suggestion
	if err != nil {
		t.Fatalf("Error starting 'pebble run': %v", err)
	}
	defer cmd.Process.Kill()
```

Or is it possible for cmd.Start to both start something and return an error?
Or is it always safe to .Process.Kill() even if the command fails to start?

dimaqq (commented):
The approach overall makes sense, though I wonder: do build flags allow you to fudge the code under test somehow?

I can see how this allows running integration tests overall, but I'm a bit unclear about how this helps testing code that needs root or what not.

P.S. maybe get someone from Juju to review the go bits?

internals/testintegration/utils.go
```diff
@@ -0,0 +1,105 @@
...
+	var logs []string
+
+	lastOutputTime := time.Now()
+
+	go func() {
+		scanner := bufio.NewScanner(stderrPipe)
+		for scanner.Scan() {
+			lastOutputTime = time.Now()
+			line := scanner.Text()
+			logs = append(logs, line)
```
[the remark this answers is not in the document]
benhoyt: Messy notes from brainstorming:

```
# Maybe something more like this:
func pebbleRun(t *testing.T, args []string) (cmd *cmd.Command, logs <-chan string) {
   t.Helper()
   cmd := start command
   // handle logs channel
   t.Cleanup(func() { send SIGTERM to pebble })
}

func waitForLogRegex(t *testing.T, logs chan string, regex string) { ... }

func waitFor(condition func() bool) {
start := time.Now()
for {
    if f() {
        break
    }
    if time.Since(start) > 10 * time.Second {
        t.Fatal("timed out waiting for X")
     }
     time.Sleep(100 * time.Millisecond)
}
}

wait(t, func() { return haveGotLogs() })
```

internals/testintegration/pebble_run_test.go
```diff
@@ -0,0 +1,51 @@
...
+	CreateLayer(t, pebbleDir, "001-simple-layer.yaml", layerYAML)
+
+	logs := PebbleRun(t, pebbleDir)
+
+	expected := []string{
+		"Started daemon",
+		"Service \"demo-service\" starting",
+		"Service \"demo-service2\" starting",
+		"Started default services with change",
+	}
```
benhoyt, line 46:
```
// start pebble
// pebble will start demo-service and demo-service 2
// check "demo-service" is actually running: side effect performed (written to file?)
```

Tests we want:

```
[run command options]
          --create-dirs  Create Pebble directory on startup if it doesn't exist
          --hold         Do not start default services automatically
          --http=        Start HTTP API listening on this address (e.g., ":4000")
      -v, --verbose      Log all output from services to stdout
          --args=        Provide additional arguments to a service
          --identities=  Seed identities from file (like update-identities --replace)
```

benhoyt (commented):
Leaving comments per our discussion.

tests/utils.go
```diff
@@ -0,0 +1,135 @@
...
+	logsCh := make(chan string)
+
+	cmd := exec.Command("../pebble", "run")
+	cmd.Env = append(os.Environ(), fmt.Sprintf("PEBBLE=%s", pebbleDir))
+
+	t.Cleanup(func() {
+		err := cmd.Process.Signal(os.Interrupt)
+		if err != nil {
+			t.Errorf("Error sending SIGINT/Ctrl+C to pebble: %v", err)
+		}
```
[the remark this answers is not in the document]
IronCore864: Good catch, I never met this issue when testing, but after you pointed it out, I did a for loop and it did occur intermittently.

IronCore864:
Resolving all the above comments, and here is a list of test cases for `pebble run`:

## Pebble Run Tests (`run_test.go`)

### 1 `TestNormal`

```yaml
services:
    svc1:
        override: replace
        command: /bin/sh -c "touch svc1; sleep 1000"
        startup: enabled
    svc2:
        override: replace
        command: /bin/sh -c "touch svc2; sleep 1000"
        startup: enabled
```

1. Start pebble.
1. Pebble will start svc1 and svc2
1. Check svc1 and svc2 are running: check files svc1 and svc2.

Need a helper function like `waitForServices` with a timeout to check files.

### 2 `TestCreateDirs`

```
tmpDir := t.TempDir()
pebbleDir := filepath.Join(tmpDir, "PEBBLE_HOME")
```

1. Start pebble with `--create-dirs`.
1. Pebble will create dir `tmpDir/PEBBLE_HOME`
1. Check dir `tmpDir/PEBBLE_HOME` exists.

No need for a helper function, use `os.path.exists`.

### 3 `TestHold`

```yaml
services:
    svc1:
        override: replace
        command: /bin/sh -c "touch /home/ubuntu/PEBBLE_HOME/svc1; sleep 1000"
        startup: enabled
```

1. Start pebble with `--hold`.
1. Pebble daemon starts.
1. Wait for log "Started daemon." before issuing any check (the `waitForLogs` helper function is needed).
1. Sleep a second before checking services (immediate check can't guarantee that svc1 is started shortly after the log "Started daemon.").
1. Check that svc1 is not running: check file svc1 does not exist. No need for a helper, use `os.path.exists`.

### 4 `TestHttpPort`

1. Start pebble with `--http=:4000`.
1. Pebble will start on port 4000
1. Check port 4000 is being listened by Pebble.

Need a helper function like `func isPortInUseByProcess(port string, processName string) bool {}`.

### 5 `TestVerbose`

```yaml
services:
    svc1:
        override: replace
        command: /bin/sh -c "cat /home/ubuntu/PEBBLE_HOME/layers/001-layer.yaml; sleep 1000"
        startup: enabled
```

1. Start pebble with `--verbose`
1. Check "services:", "svc1:", "override: replace", "startup: enabled" are in the logs, need the `waitForLogs` helper func.

### 6 `TestArgs`

```yaml
services:
    svc1:
        override: replace
        command: /bin/sh
        startup: enabled
```

1. Start pebble with `--verbose --args svc1 -c "cat /home/ubuntu/PEBBLE_HOME/layers/001-layer.yaml; sleep 1000"` (verbose is used so that I can test args, and I use the layers file for testing, have to be a bit innovative here to test args...)
2. Same checks as the previous test case, `waitForLogs`.

### 7 `TestIdentities`

1. Create a file named `idents-add.yaml`, maybe in pebbleDir:

```yaml
identities:
    bob:
        access: admin
        local:
            user-id: 42
    alice:
        access: read
        local:
            user-id: 2000
```

1. Start pebble with `--identities`.
1. Run `pebble identity bob` command and check `access: admin`, `user-id: 42` are in the output.
1. Run `pebble identity alice` command and check `access: read`, `user-id: 2000` are in the output.

Need a helper like `runPebbleCmdAndCheckOutput`.

## Summary

- `waitForLogs` is needed
- a few more helpers are needed, see the above test cases.

Since these test cases are relatively simple to implement, I will go ahead and do them now. We can review and refactor later.

internals/testintegration/utils.go
IronCore864: This comment is rendered invalid by the recent discussion with Ben because we are removing this. Closing now.

internals/testintegration/utils.go
IronCore864: After discussion with Ben, we decided to use SIGINT and wait for the process to finish.

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+	pebbleDir := t.TempDir()
+
+	port := "4000"
+	logsCh := pebbleRun(t, pebbleDir, "--http=:"+port)
+	expectedLogs := []string{"Started daemon"}
+	if err := waitForLogs(logsCh, expectedLogs, time.Second*3); err != nil {
+		t.Errorf("Error waiting for logs: %v", err)
+	}
+
+	if !isPortUsedByProcess(t, port, "pebble") {
```
benhoyt, line 116: Once again, I think this helper obscures the test (I had no idea it would use `lsof`, for example). Also, I think we should just make an `http.Get` call to one of the HTTP endpoints (probably `http://localhost:<port>/v1/health`).

```go
resp, err := http.Get("http://localhost:4000/v1/health")
if err != nil { /* error */ }
defer resp.Body.Close()
if resp.StatusCode != 200 { /* error */ }
```

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+	layerYAML := `
+services:
+    svc1:
+        override: replace
+        command: /bin/sh -c "{{.svc1Cmd}}"
+        startup: enabled
+`
+	layersFileName := "001-simple-layer.yaml"
+	svc1Cmd := fmt.Sprintf("cat %s; sleep 1000", filepath.Join(pebbleDir, "layers", layersFileName))
+	layerYAML = strings.Replace(layerYAML, "{{.svc1Cmd}}", svc1Cmd, -1)
```
benhoyt, line 133: Same comment about using one `fmt.Sprintf`.

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+	identitiesPath := filepath.Join(pebbleDir, identitiesFileName)
+	if err := os.WriteFile(identitiesPath, []byte(identitiesYAML), 0o755); err != nil {
+		t.Fatalf("Cannot create layers file: %v", err)
+	}
+}
+
+func pebbleRun(t *testing.T, pebbleDir string, args ...string) <-chan servicelog.Entry {
+	t.Helper()
+
+	logsCh := make(chan servicelog.Entry)
```
benhoyt, line 73: I wonder if we should return `(stdoutCh, stderrCh <-chan servicelog.Entry)`, so the channels are separated, and the `TestVerbose` can ensure that the service logs show up on stdout.

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+	cmd := exec.Command("../pebble", append([]string{"run"}, args...)...)
+	cmd.Env = append(os.Environ(), "PEBBLE="+pebbleDir)
+
+	t.Cleanup(func() {
```
benhoyt, line 78: I think this `Cleanup` call should be moved to after the `cmd.Start` call, in case the test fails before then, we don't have anything to clean up as the command hasn't started yet.

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+		}
+
+		// Both stderr and stdout are needed, because pebble logs to stderr
+		// while with "--verbose", services otuput to stdout.
+		stderrParser := servicelog.NewParser(stderrPipe, 4*1024)
+		stdoutParser := servicelog.NewParser(stdoutPipe, 4*1024)
+
+		// Channel to signal completion and close logsCh
+		done := make(chan struct{})
```
benhoyt, line 123: This is confusing: there's a `done` channel defined above, and again here. What are we trying to achieve, and why are there "nested" goroutines here? I'm guessing we want the `readLogs` goroutines to stop when the test finishes? In which case should a `close(done)` go into a `t.Cleanup` func? Perhaps you can clarify or simplify this a bit.

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+	createLayer(t, pebbleDir, layersFileName, layerYAML)
+
+	logsCh := pebbleRun(t, pebbleDir, "--verbose")
+	expectedLogs := []string{
+		"Started daemon",
+		"services:",
```
benhoyt, line 140: It seems a bit confusing to have the service output the layers file itself. How about we just have it `echo` a single fixed string and wait for that?

Assuming we change `pebbleRun` to return both stdout and stderr log channels, we'd do something like:

```go
waitForLogs(t, stdoutLogs, "Some fixed string", 3*time.Second)
```

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+	logsCh := pebbleRun(t, pebbleDir, "--verbose",
+		"--args",
+		"svc1",
+		"-c",
+		fmt.Sprintf("cat %s; sleep 1000", filepath.Join(pebbleDir, "layers", layersFileName)),
+	)
+	expectedLogs := []string{
+		"Started daemon",
+		"services:",
```
benhoyt, line 175: Similar comments here as above: can we just have it output a fixed string, rather than printing its own layers file?

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+        local:
+            user-id: 2000
+`
+	identitiesFileName := "idents-add.yaml"
+	createIdentitiesFile(t, pebbleDir, identitiesFileName, identitiesYAML)
+
+	logsCh := pebbleRun(t, pebbleDir, "--identities="+filepath.Join(pebbleDir, identitiesFileName))
+	expectedLogs := []string{
+		"Started daemon",
+		"POST /v1/services",
```
benhoyt, line 206: I don't think we should bother checking the logs in the tests (I don't think stable log formatting should even be part of the contract). So let's just check the expected behaviour -- the output of `pebble identity` below.

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	if err := waitForLogs(logsCh, expectedLogs, time.Second*3); err != nil {
+		t.Errorf("Error waiting for logs: %v", err)
+	}
+
+	expectedOutput := []string{"access: admin", "local:", "user-id: 42"}
+	runPebbleCmdAndCheckOutput(t, pebbleDir, expectedOutput, "identity", "bob")
```
benhoyt, line 213: This is another case where it's quite unclear what the helper is doing. I think we should refactor or maybe not even use a helper. If we do stick with a helper, I think it's much more natural for the helper to run the command and return the output (just as a string), then have the test check that:

```go
output := runPebbleCommand(t, pebbleDir, "identity", "bob")
expected := `
access: admin
local:
  user-id: 42
`[1:]
if output != expected { /* error */ }
```

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+		// Wait for both parsers to finish
+		<-done
+		<-done
+	}()
+
+	return logsCh
+}
+
+func waitForLogs(logsCh <-chan servicelog.Entry, expectedLogs []string, timeout time.Duration) error {
```
benhoyt, line 137: After applying my comments in the tests, we may not need this to wait for more than one log at a time, so perhaps we can simplify and just have it wait for a single log? Then we can also check the service name is "pebble" or "svc1" or whatever it should be. I also think that like the other test helpers, this should take `t` and fail, rather than return an error.

```go
func waitForLog(t, logsCh <-chan servicelog.Entry, serviceName, expectedLog string, timeout time.Duration) { ... }

// Example usage:
waitForLog(t, stderrLogs, "pebble", "Started daemon", 3*time.Second)
```

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+				if _, ok := receivedLogs[log]; !ok {
+					missingLogs = append(missingLogs, log)
+				}
+			}
+			return errors.New("timed out waiting for log: " + strings.Join(missingLogs, ", "))
+		}
+	}
+}
+
+func containsSubstring(s, substr string) bool {
```
benhoyt, line 179: Why is this needed? `containsSubstring` is actually longer to type than `strings.Contains`!

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+	return strings.Contains(s, substr)
+}
+
+func waitForServices(t *testing.T, pebbleDir string, expectedServices []string, timeout time.Duration) {
```
benhoyt, line 183: Per comment elsewhere, I think we should remove this helper and just call `waitForService` multiple times (though that should actually become `waitForFile`).

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+			stat, err := os.Stat(serviceFilePath)
+			if err == nil && stat.Mode().IsRegular() {
+				os.Remove(serviceFilePath)
+				return
+			}
+		}
+	}
+}
+
+func isPortUsedByProcess(t *testing.T, port string, processName string) bool {
```
benhoyt, line 213: This is another one that's very specific for a helper function. Per comments in the test, I think we should remove this helper and just use `http.Get` to do an actual HTTP request in the test.

tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+	outputStr := string(output)
+	if strings.Contains(outputStr, processName) {
+		return true
+	}
+
+	return false
+}
+
+func runPebbleCmdAndCheckOutput(t *testing.T, pebbleDir string, expectedOutput []string, args ...string) {
```
benhoyt, line 240: Per comments in the test, I think we should simplify this helper to just run the command and return the output, and the test can check it. 

```go
func runPebbleCommand(t *testing.T, pebbleDir string, args ...string) string { ... }

output := runPebbleCommand(t, pebbleDir, "identity", "bob")
if output != expected { ... }
```

benhoyt (changes requested):
Thanks for this -- I think it's heading in the right direction. Lots of comments, but they're basically about how we structure the helpers and make the tests more obvious. I think in general it's better to have fewer, more generic helpers. One tell-tale sign is that same of the names start to get long or specific, like `isPortUsedByProcess` (very specific to a particular test), `writeIdentitiesFile` (doesn't actually do anything specific to identities), or `runPebbleCmdAndCheckOutput` (often an "and" in a function name means you should split it). Happy to discuss any of these further on video if you want.

benhoyt:
> The reason is that if some code weren't put in a separate function, the test functions would become much longer and harder to read. So, I kept them as they were. Instead of thinking of them as helper functions, think of them as a way to refactor the tests to improve readability.

I realise this is somewhat subjective, but I think mere length is okay. I definitely disagree about "harder to read" -- I think several of the helpers obscure the logic of the test and make it unclear what it's actually testing. I think there are some minor tweaks we can make to the structure and names to help with this, and I've argued my case in the comments above.

benhoyt (commented):
FWIW, I just ran the integration tests with the pebble daemon running in another window, and got this error:

```
$ go test -count=1 -tags=integration ./tests/
--- FAIL: TestHttpPort (3.05s)
    run_test.go:113: Error waiting for logs: timed out waiting for log: Started daemon
```

It seems to me the two instances should be completely independent, as the integration tests point to a temporary `PEBBLE` directory. Any ideas why this would fail?

tests/README.md
```diff
@@ -0,0 +1,23 @@
+# Pebble Integration Tests
+
+## Run Tests
+
+```bash
+go test -count=1 -tags=integration ./tests/
```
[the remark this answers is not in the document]
IronCore864: Yeah, I have taken note of this in a previous comment.

Split or in `tests.yml`: I'm OK with both, but I lean slightly towards splitting because logically, it's a little different from unit tests (it requires building a binary); and functionally, we can utilize multiple runners to run it in parallel. It also makes debugging easier because we don't have to go through a whole lot unit test successful logs to get to the failed integration ones.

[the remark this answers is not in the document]
benhoyt: Sounds good, let's split it.

tests/run_test.go
```diff
@@ -0,0 +1,217 @@
...
+        access: admin
+        local:
+            user-id: 42
+    alice:
+        access: read
+        local:
+            user-id: 2000
+`
+	identitiesFileName := "idents-add.yaml"
+	createIdentitiesFile(t, pebbleDir, identitiesFileName, identitiesYAML)
```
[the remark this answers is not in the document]
IronCore864: I've watched the talk, I think this part makes sense, I'll restructure it to the first option above.

tests/run_test.go
IronCore864: OK, then I will change it to the same as `TestHold` where after `pebble run` we sleep 100 millisecond just to be sure the daemon is up and running before issuing `pebble identity` commands.

tests/main_test.go
IronCore864: After applying all comments above, the only use case of `waitForLog` would be to wait for the stdout that comes from `--verbose`.

Another use case would be to wait for "Started daemon", to make sure the daemon is already started before checking other things (like running `pebble identity` commands), but this could be solved by waiting for 100 milliseconds, just as in `TestHold`.

So, shall we change it to something like `waitForStdout`, and make `pebbleRun` only return one channel for stdout, not stderr (we are not checking pebble logs anyway if we change waiting for `Started daemon" to sleep)?

IronCore864:
> FWIW, I just ran the integration tests with the pebble daemon running in another window, and got this error:
> 
> ```
> $ go test -count=1 -tags=integration ./tests/
> --- FAIL: TestHttpPort (3.05s)
>     run_test.go:113: Error waiting for logs: timed out waiting for log: Started daemon
> ```
> 
> It seems to me the two instances should be completely independent, as the integration tests point to a temporary `PEBBLE` directory. Any ideas why this would fail?

I saw this comment at last after fixing all comments above, and after the fixes, I could not reproduce. Could you confirm?

benhoyt:
Regarding the `TestHttpPort` failure when the daemon is running -- no, I can't reproduce now either. We'll call it fixed.

tests/main_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	timeoutCh := time.After(timeout)
+	for {
+		select {
+		case log, ok := <-logsCh:
+			if !ok {
+				t.Error("channel closed before all expected logs were received")
+			}
+
+			if strings.Contains(log.Message, expectedLog) {
```
benhoyt, line 135: Let's add an `expectedService string` argument before `expectedLog` and test equality of that too. It'll be "pebble" for pebble logs on stderr, or the service name for service logs on stderr. Makes the test a bit more complete.
```suggestion
			if log.Service == expectedService && strings.Contains(log.Message, expectedLog) {
```

tests/run_test.go
```diff
@@ -0,0 +1,210 @@
...
+        local:
+            user-id: 2000
+`
+	identitiesFileName := "idents-add.yaml"
+	if err := os.WriteFile(filepath.Join(pebbleDir, identitiesFileName), []byte(identitiesYAML), 0o755); err != nil {
+		t.Fatalf("Cannot write identities file: %v", err)
+	}
+
+	_, _ = pebbleRun(t, pebbleDir, "--identities="+filepath.Join(pebbleDir, identitiesFileName))
+	time.Sleep(100 * time.Millisecond)
```
benhoyt, line 189: I think we should wait for the "Starting daemon" log here, rather than a time.Sleep. The time.Sleep is okay in the "directories *not* created" case, but here we're waiting for something positive (not lack of something).

tests/run_test.go
```diff
@@ -0,0 +1,210 @@
...
+func TestIdentities(t *testing.T) {
+	pebbleDir := t.TempDir()
+
+	identitiesYAML := `
+identities:
+    bob:
+        access: admin
+        local:
+            user-id: 42
+    alice:
```
benhoyt, line 178: We probably only need one user in this test -- we're just testing that `--identities` works -- the full functionality of "add identity" is tested elsewhere.

benhoyt (changes requested):
Nice, this is much cleaner now. A few minor comments, and one comment about the nested goroutines in `pebbleRun`.

benhoyt (approved):
Looking good, thanks!

One remaining thing I noticed after running the integration tests -- they don't seem to clean up after themselves very well. After running them, I get:

```
$ ps aux | grep sleep
ben       137333  0.0  0.0   2800  1536 pts/2    S    08:56   0:00 /bin/sh -c touch /tmp/TestStartupEnabledServices3374203583/001/svc1; sleep 1000
ben       137334  0.0  0.0   2800  1536 pts/2    S    08:56   0:00 /bin/sh -c touch /tmp/TestStartupEnabledServices3374203583/001/svc2; sleep 1000
ben       137337  0.0  0.0   8288  1920 pts/2    S    08:56   0:00 sleep 1000
ben       137338  0.0  0.0   8288  1920 pts/2    S    08:56   0:00 sleep 1000
ben       137373  0.0  0.0   2800  1536 pts/2    S    08:56   0:00 /bin/sh -c echo 'hello world'; sleep 1000
ben       137374  0.0  0.0   8288  1920 pts/2    S    08:56   0:00 sleep 1000
ben       137384  0.0  0.0   2800  1536 pts/2    S    08:56   0:00 /bin/sh -c echo 'hello world'; sleep 1000
ben       137385  0.0  0.0   8288  1920 pts/2    S    08:56   0:00 sleep 1000
```

Any idea why? Maybe we can do some logging to see why. I would have thought sending SIGINT should cause Pebble to stop the running services before exiting, but it's clearly not doing that (properly, at any rate).

One thing we could do to mitigate it (but not fix it, so we should still look into it) is changing the `sleep 1000` to `sleep 10` -- that's still plenty long enough for these tests, but at least they'll exit sooner themselves.

I'll also ask Harry to review this, as he originally created the issue.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/canonical/pebble/internals/servicelog"
+)
+
+// TestMain builds the pebble binary before running the integration tests.
+func TestMain(m *testing.M) {
+	goBuild := exec.Command("go", "build", "-o", "../pebble", "../cmd/pebble")
```
hpidcock, line 33: Not a fan of invoking go build in a test, but if that's the path @benhoyt is happy with, then I am ok.

But we should be able to run the integration tests against any pre-compiled pebble binary. This way the juju team can add these tests into our gating test pipeline.

So I suggest we add a flag to the test executable like `--no-build`, where instead it uses `exec.LookPath`.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+	stopStdout := make(chan struct{}, 1)
+	stopStderr := make(chan struct{}, 1)
+
+	t.Cleanup(func() {
+		err := cmd.Process.Signal(os.Interrupt)
+		if err != nil {
+			t.Errorf("Error sending SIGINT/Ctrl+C to pebble: %v", err)
+		}
+		cmd.Wait()
+		stopStdout <- struct{}{}
```
hpidcock, line 91: Cancellation channels should be unbuffered channels that are closed.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+	if err := goBuild.Run(); err != nil {
+		fmt.Println("Cannot build pebble binary:", err)
+		os.Exit(1)
+	}
+
+	exitCode := m.Run()
+	os.Exit(exitCode)
+}
+
+func createLayer(t *testing.T, pebbleDir, layerFileName, layerYAML string) {
```
hpidcock, line 43: These utility functions deserve comments to describe their use.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	layerPath := filepath.Join(layersDir, layerFileName)
+	err = os.WriteFile(layerPath, []byte(layerYAML), 0o755)
+	if err != nil {
+		t.Fatalf("Cannot create layers file: %v", err)
+	}
+}
+
+func pebbleRun(t *testing.T, pebbleDir string, args ...string) (stdoutCh chan servicelog.Entry, stderrCh chan servicelog.Entry) {
```
hpidcock, line 59: Comment me.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+	stderrParser := servicelog.NewParser(stderrPipe, 4*1024)
+	stdoutParser := servicelog.NewParser(stdoutPipe, 4*1024)
+
+	go readLogs(stdoutParser, stdoutCh, stopStdout)
+	go readLogs(stderrParser, stderrCh, stopStderr)
+
+	return stdoutCh, stderrCh
+}
+
+func waitForLog(t *testing.T, logsCh <-chan servicelog.Entry, expectedService, expectedLog string, timeout time.Duration) {
```
hpidcock, line 119: Comment me.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+				return
+			}
+
+		case <-timeoutCh:
+			t.Fatalf("timed out after %v waiting for log %s", 3*time.Second, expectedLog)
+		}
+	}
+}
+
+func waitForFile(t *testing.T, file string, timeout time.Duration) {
```
hpidcock, line 140: Comment me.

tests/main_test.go
```diff
@@ -0,0 +1,171 @@
...
+		case <-ticker.C:
+			stat, err := os.Stat(file)
+			if err == nil && stat.Mode().IsRegular() {
+				return
+			}
+		}
+	}
+}
+
+func runPebbleCommand(t *testing.T, pebbleDir string, args ...string) string {
```
hpidcock, line 159: Comment me.

tests/run_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"fmt"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestStartupEnabledServices(t *testing.T) {
```
hpidcock, line 30: I think all these tests should have a verbose comment describing what they are testing.

hpidcock (commented):
Looking good so far. Just a few nitpicks.

tests/main_test.go
benhoyt: Yeah, I really like that it makes them just work by default. However, it's a good idea to allow a more explicit override. I'd prefer a slightly more explicit approach: a flag `-pebblebin` that, if set, uses that as the Pebble binary rather than building one.

benhoyt:
Agreed with all of Harry's comments (but a slightly different explicit approach than `--no-build`) -- thanks for the review.

tests/main_test.go
benhoyt: That way you can point it at a binary you or the Juju CI or whatever has just built, rather than necessarily having to install it or put it in PATH.

IronCore864:
Reply to the above comments:

- I have added some comments for all the tests and most of the helper functions.
- A new flag `-pebbleBin` is added as Ben suggested, so that if set, it will be used for the integration tests rather than building one.
- Services not stopped after SIGINT is sent to Pebble is mitigated by changing `sleep 1000` to `sleep 10`, but _it's not resolved_.

benhoyt (approved):
Great, thanks!

benhoyt:
> Services not stopped after SIGINT is sent to Pebble is mitigated by changing sleep 1000 to sleep 10, but it's not resolved.

Can you please spend a bit of time looking into why is? I just don't want to paper over something if it's an actual bug.

benhoyt:
I looked into this further, and debugged by printing out the Pebble/stderr logs (and running the tests with `-v` so they showed up). Then I enabled `PEBBLE_DEBUG=1` in the subprocess, and this helped me see that when SIGTERM was sent, it was before the "start" had completed (which takes 1s). So [`stopRunningServices` thought there were "No services to stop."](https://github.com/canonical/pebble/blob/183eadfbbb342daf4ea30186e76bfd3e18f07fa8/internals/daemon/daemon.go#L629) as they were in "starting" state, and didn't try to send SIGTERM to the children. Arguably `stopRunningServices` should include the services in "starting" state. I might open a separate issue on that.

In the meantime, it's probably best if we wait till the "startup: enabled" services are started when the daemon starts up, and the simplest way to do this now is to wait for the "Started default services ..." log. Note that the full log looks like this.

```
2024-09-22T22:10:40.147Z [pebble] Started default services with change 1.
```

I recommend the following diff, which cleans up nicely in my tests:

```diff
diff --git a/tests/run_test.go b/tests/run_test.go
index c73c365..c98c693 100644
--- a/tests/run_test.go
+++ b/tests/run_test.go
@@ -49,7 +49,8 @@ services:
 
        createLayer(t, pebbleDir, "001-simple-layer.yaml", layerYAML)
 
-       _, _ = pebbleRun(t, pebbleDir)
+       _, stderrCh := pebbleRun(t, pebbleDir)
+       waitForLog(t, stderrCh, "pebble", "Started default services", 3*time.Second)
 
        waitForFile(t, filepath.Join(pebbleDir, "svc1"), 3*time.Second)
        waitForFile(t, filepath.Join(pebbleDir, "svc2"), 3*time.Second)
@@ -141,6 +142,7 @@ services:
        stdoutCh, stderrCh := pebbleRun(t, pebbleDir, "--verbose")
        waitForLog(t, stderrCh, "pebble", "Started daemon", 3*time.Second)
        waitForLog(t, stdoutCh, "svc1", "hello world", 3*time.Second)
+       waitForLog(t, stderrCh, "pebble", "Started default services", 3*time.Second)
 }
 
 // TestArgs tests that Pebble provides additional arguments to a service
@@ -166,6 +168,7 @@ services:
        )
        waitForLog(t, stderrCh, "pebble", "Started daemon", 3*time.Second)
        waitForLog(t, stdoutCh, "svc1", "hello world", 3*time.Second)
+       waitForLog(t, stderrCh, "pebble", "Started default services", 3*time.Second)
 }
 
 // TestIdentities tests that Pebble seeds identities from a file
```

benhoyt:
I've opened https://github.com/canonical/pebble/issues/502 to track having Pebble terminate services in "starting" state as well.

IronCore864:
While testing the leaked "sleep" issue, I found [another place](https://github.com/canonical/pebble/blob/master/internals/cli/cmd_enter_test.go#L174) that leaks a sleep. Since it's not related to this feature, I will merge this PR and continue debugging it in another branch.

PR #501, opened 2024-09-22
test: make root tests easier to run

IronCore864:
Background: Previously, when adding a new root test, we had to modify the GitHub Actions workflow to include the newly added test, which is some extra operational overhead.

Per discussion [here](https://github.com/canonical/pebble/pull/500#issuecomment-2362290802), we agreed to run all tests with root and skip those tests that would fail with root (checking permissions and stuff), and this PR skips tests that would fail when running with root and made the GitHub Actions workflow for root tests much easier.

### DIFF
--- a/internals/cli/cmd_add_test.go
+++ b/internals/cli/cmd_add_test.go
@@ -100,10 +100,9 @@ services:
 			} else if path == triggerLayerPath {
 				c.Assert(err, check.ErrorMatches, "triggered")
 			} else if path == unreadableLayerPath {
-				if os.Getuid() == 0 {
-					c.Skip("requires running as non-root users")
+				if os.Getuid() != 0 {
+					c.Assert(os.IsPermission(err), check.Equals, true)
 				}
-				c.Assert(os.IsPermission(err), check.Equals, true)
 			}
 		}
 
--- a/internals/cli/cmd_add_test.go
+++ b/internals/cli/cmd_add_test.go
@@ -109,5 +109,6 @@ services:
 		args = append(args, "extra", "arguments", "invalid")
 		_, err = cli.ParserForTest().ParseArgs(args)
 		c.Assert(err, check.Equals, cli.ErrExtraArgs)
+		s.ResetStdStreams()
 	}
 }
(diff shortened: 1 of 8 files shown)


internals/cli/cmd_add_test.go
```diff
@@ -100,6 +100,9 @@ services:
 			} else if path == triggerLayerPath {
 				c.Assert(err, check.ErrorMatches, "triggered")
 			} else if path == unreadableLayerPath {
+				if os.Getuid() == 0 {
+					c.Skip("requires running as non-root users")
```
benhoyt, line 104: It's a bit weird to skip the test only once you get to this code path. We could skip the whole test, but I'd probably just only do this `IsPermission` assert if it's non-root. For example:

```go
if os.Getuid() != 0 {
    c.Assert(os.IsPermission(err), check.Equals, true)
}
```

internals/osutil/io_test.go
```diff
@@ -84,6 +84,9 @@ func (ts *AtomicWriteTestSuite) TestAtomicWriteFileSymlinkNoFollow(c *C) {
 	c.Assert(os.Chmod(rodir, 0500), IsNil)
 	defer os.Chmod(rodir, 0700)
 
+	if os.Getuid() == 0 {
+		c.Skip("requires running as non-root users")
```
benhoyt, line 88: Grammar nit (here and elsewhere):
```suggestion
		c.Skip("requires running as non-root user")
```

benhoyt (changes requested):
Thanks for this, looks very simple. Just two minor tweaks requested.

internals/cli/cmd_add_test.go
```diff
@@ -100,12 +100,15 @@ services:
...
+				if os.Getuid() != 0 {
+					c.Assert(os.IsPermission(err), check.Equals, true)
+				}
 			}
 		}
 
 		args = append(args, "extra", "arguments", "invalid")
 		_, err = cli.ParserForTest().ParseArgs(args)
 		c.Assert(err, check.Equals, cli.ErrExtraArgs)
+		s.ResetStdStreams()
```
IronCore864, line 112: This test case uses a for loop to test `combine=false` and `combine=true`.

When running as root, adding the `unreadableLayer` would succeed and the stdout interferes with the next loop. So, here I added a reset on the stdout so that the two cases don't interfere with each other.

IronCore864:
[Root tests passed](https://github.com/canonical/pebble/actions/runs/10986680141/job/30500379396?pr=501).

benhoyt (approved):
Great, thanks -- looks good to me!

PR #503, opened 2024-09-24
fix: allow stopping services in "starting" state

### LINKED ISSUE #410: Allow service to be stopped while in 1s "starting" state
Per https://github.com/canonical/pebble/issues/408, if you try to stop a service while it's in the 1-second "starting" status, it will fail with the error "cannot start service while starting". We should probably make the service manager a bit cleverer in this case and allow the stop to succeed -- the fact that it's just been started shouldn't mean you can't stop it.

### LINKED ISSUE #502: stopRunningServices should probably also stop "starting" services
If the "startup: enabled" services are still starting (because it's been less than 1s) and you send SIGTERM to the Pebble daemon, it exits without trying to terminate or clean up those services, because [`servicesToStop`](https://github.com/canonical/pebble/blob/183eadfbbb342daf4ea30186e76bfd3e18f07fa8/internals/overlord/servstate/manager.go#L356) only includes services in "running" and "backoff" state right now.

We should probably also include "starting" services. Let's just make sure there are no other bad implications of doing this.

See also https://github.com/canonical/pebble/pull/497#issuecomment-2366990079

IronCore864:
Pebble has a 1-second okayWait time period for starting services, after which, if the service is still not exited, it's considered as "running". Previously, when services were still in the starting state within this 1-second okayWait period, they couldn't be stopped. This causes some issues in tests. And, just because services have just started doesn't mean they shouldn't be allowed to stop. This PR allows services in the starting state (within the 1s okayWait period) to be stopped, no matter if it's the daemon exits (Ctrl+C in the console) or the `pebble stop` command is issued.

Fixes #410
Fixes #502

Also adding two test cases:

1. servstate: use a mock service which sleeps a while then create a side effect, stop the service immediately after it's started, and the side effect won't be created.
2. daemon: start a service, stop the daemon before the okay delay, and a stop change should be created. Whether the service is actually terminated or not isn't tested because it's tested above in the servstate.


internals/overlord/servstate/state-diagram.dot
```diff
@@ -7,6 +7,7 @@ digraph service_state_machine {
     node [penwidth=1]
     initial -> starting [label="start"]
     starting -> running [label="okay wait\nelapsed"]
+    starting -> terminating [label="stopped before\nokay wait elapses"]
```
benhoyt, line 10:
```suggestion
    starting -> terminating [label="stop (before\nokay wait elapses)"]
```

internals/overlord/servstate/handlers.go
```diff
@@ -682,6 +682,10 @@ func (s *serviceData) stop() error {
 	defer s.manager.servicesLock.Unlock()
 
 	switch s.state {
+	case stateStarting:
+		s.started <- fmt.Errorf("stopped before the 1 second okay delay")
```
benhoyt, line 686:
```suggestion
		s.started <- fmt.Errorf("stopped before the %s okay delay", okayDelay)
```

internals/overlord/servstate/manager_test.go
```diff
@@ -297,6 +297,53 @@ services:
...
+	s.newServiceManager(c)
+	// The service sleeps for fakeOkayDelay second then creates a side effect (a file at donePath).
+	layer := `
+services:
+    %s:
+        override: replace
+        command: /bin/sh -c "sleep %g; {{.NotifyDoneCheck}}"
+`
+	serviceName := "test-stop-within-okaywait"
+	s.planAddLayer(c, fmt.Sprintf(layer, serviceName, fakeOkayDelay))
```
benhoyt, line 315: I think this is buggy, as `fakeOkayDelay` is an integer `time.Duration` and can't be formatted with `%g`. I think we want `fakeOkayDelay.Seconds()`. I think the test was erroneously passing as a result.

Nit: I think it's easier to read to have the `fmt.Sprintf(...)` call above on the `layer :=` line, so the sprintf values are more closely associated with the format string.

internals/overlord/servstate/manager_test.go
```diff
@@ -297,6 +297,53 @@ services:
...
+	chg := s.stopServices(c, [][]string{{serviceName}})
+	s.st.Lock()
+	c.Assert(chg.Err(), IsNil)
+	s.st.Unlock()
+
+	waitChangeReady(c, s.runner, chgStart, "Start test")
+
+	s.st.Lock()
+	c.Check(chgStart.Status(), Equals, state.ErrorStatus)
+	c.Check(chgStart.Err(), ErrorMatches, `(?s).*cannot start service: exited quickly with code.*`)
```
benhoyt, line 337: Shouldn't we be looking for `stopped before the %s okay delay` message here?

benhoyt:
Also seems like the new daemon test is failing (sometimes). Let's ensure it's not flakey with some stress testing of the new tests.

internals/overlord/servstate/manager_test.go
IronCore864: `fakeOkayDelay.Seconds()` (together with the next comment) is fixed.

On the readability issue, I do agree if we have something like the following, it will be more readable:

```go
layer := `some template`
layer := fmt.Sprintf(layer, foo, bar)
s.planAddLayer(c, layer)
```

However, in the current test file, there are many test cases with the following code:

```go
layer := `some template`
s.planAddLayer(c, fmt.Sprintf(
    layer,
    foo, 
    bar
))
```

So, for consistency, I used the same code (I will move `serviceName` above `layer`, so that this part is exactly the same code organization as other test cases).

benhoyt (approved):
Thanks!

benhoyt (approved):
Test fix looks reasonable to me. Feel free to merge this PR (and we can discuss the TestDeadlock issue you found separately).

PR #504, opened 2024-09-25
docs: add a security policy document

tonyandrewmeyer:
Document the security policy for the Pebble project.

See also [SEC0026](https://docs.google.com/document/d/1RolsiWCWhM-sVmHFjLt-Mq5yKcgQJ8GLXssGJREjSL0/edit).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,38 @@
+# Security policy
+
+## Supported versions
+
+Security updates will be released for all major versions that have had releases in the last year,
+and for all versions of Pebble that are bundled with [Juju](https://github.com/juju/juju)
+releases that [receive security updates](https://juju.is/docs/juju/roadmap).
+
+## Reporting a vulnerability
+
+Please provide a description of the issue, the steps you took to
+create the issue, affected versions, and, if known, mitigations for
+the issue.
+
+The easiest way to report a security issue is through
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer:
I'll rebase after #505.

SECURITY.md
```diff
@@ -0,0 +1,38 @@
+# Security policy
+
+## Supported versions
+
+Security updates will be released for all major versions that have had releases in the last year,
```
benhoyt, line 5: Just to clarify "all major versions" -- given we only have 1.x.y, I guess this means just one security update?

benhoyt (approved):
Thank you for this! It looks like a good start to me.

SECURITY.md
tonyandrewmeyer: Yes, plus the ones that Juju needs. The "only for majors" is where we currently landed for ops, so I re-used that here.

We could have a table here listing versions like some projects do, but that seems a pain to maintain. [LXD](https://github.com/canonical/lxd/blob/main/SECURITY.md) has rules for monthly and LTS releases - we could copy that but have Juju instead of LTS?

(The "what qualifies" section in the LXD one I think would be good here too, but probably after the current modelling work is done).

benhoyt: I think what you have now is fine -- I just wanted to clarify the intent. Could add a little parenthetical to clarify for others like "(1.x versions)" if you want.

Issue #508, opened 2024-10-10
TestDeadlock fails intermittently in local environment with `go test -race`s


IronCore864:
We have a test case `TestDeadlock` in [`internals/daemon/api_services_test.go`](https://github.com/canonical/pebble/blob/master/internals/daemon/api_services_test.go#L367) where we randomly issue a bunch of pebble start/stop commands (c.a. 50 in total or so) within a second and expect all the changes to be done. This test case fails intermittently with `go test -race` because not all changes can be ready in certain cases. In my local Ubuntu 22.04 machine it's relatively easy to reproduce, failing once every five times or so, but weirdly it never occurs in GitHub Action runs.

Error message:

```bash
----------------------------------------------------------------------
FAIL: api_services_test.go:367: apiSuite.TestDeadlock

api_services_test.go:450:
    c.Fatal("timed out waiting for final request -- deadlock!")
... Error: timed out waiting for final request -- deadlock!

==================
WARNING: DATA RACE
Write at 0x0000012d2198 by goroutine 774:
  github.com/canonical/pebble/internals/daemon.(*Daemon).addRoutes()
      /home/ubuntu/pebble/internals/daemon/daemon.go:390 +0x1bb
  github.com/canonical/pebble/internals/daemon.(*apiSuite).daemon()
      /home/ubuntu/pebble/internals/daemon/api_test.go:77 +0x17c
  github.com/canonical/pebble/internals/daemon.(*apiSuite).TestGetPlan()
      /home/ubuntu/pebble/internals/daemon/api_plan_test.go:63 +0xaa
  runtime.call16()
      /snap/go/10730/src/runtime/asm_amd64.s:775 +0x42
  reflect.Value.Call()
      /snap/go/10730/src/reflect/value.go:365 +0xb5
  gopkg.in/check%2ev1.(*suiteRunner).forkTest.func1()
      /home/ubuntu/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:775 +0x965
  gopkg.in/check%2ev1.(*suiteRunner).forkCall.func1()
      /home/ubuntu/go/pkg/mod/gopkg.in/check.v1@v1.0.0-20201130134442-10cb98267c6c/check.go:669 +0xe9

Previous read at 0x0000012d2198 by goroutine 761:
  github.com/canonical/pebble/internals/daemon.v1PostServices()
      /home/ubuntu/pebble/internals/daemon/api_services.go:74 +0x313
[...]

PR #510, opened 2024-10-14
chore: revert changes introduced by allowing stopping services in starting state

IronCore864:
In [this PR](https://github.com/canonical/pebble/pull/503), we introduced a feature to allow stopping services that are in the starting state / within the okayDelay. However, this makes [a deadlock issue in ringbuffer a real issue](https://github.com/canonical/pebble/issues/508). So we are reverting this change now and will redo it after the deadlock is resolved.


benhoyt (approved):
Thanks! Yes, better be safe than sorry and accidentally ship this. :-)

PR #511, opened 2024-10-15
fix: ring buffer deadlock

IronCore864:
Fix an issue where the ring buffer might deadlock because rwlock and iterator mutex are acquired in a different order.

Fixes: https://github.com/canonical/pebble/issues/508

Notes:

1. As discussed with Harry and Ben, `rb.Write` isn't adjusted because the rwlock and iterator mutex are acquired at different times.
2. Besides an internal `close` function, an internal `positions` function is created for the same reason.
3. The [test case here](https://github.com/canonical/pebble/issues/508#issuecomment-2409839516) is not added because it was supposed to be an easy way to reproduce the deadlock. Adding an additional benchmark test is not useful, and the deadlock issue can be caught by the `TestDeadlock` test case anyway.
4. All three functions `signalIterators`, `releaseIterators` and `removeIterator` are adjusted (although technical only `releaseIterators` needed to be ) because they share a common naming pattern so it's nicer to share a common implementation pattern as well: all internal functions don't operate locks.

Benchmark test, manual test to reproduce the 3-way deadlock issue, and race test are all done and passed.

### DIFF
--- a/internals/servicelog/ringbuffer.go
+++ b/internals/servicelog/ringbuffer.go
@@ -75,10 +75,6 @@ func (rb *RingBuffer) Close() error {
 func (rb *RingBuffer) Closed() bool {
 	rb.rwlock.RLock()
 	defer rb.rwlock.RUnlock()
-	return rb.closed()
-}
-
-func (rb *RingBuffer) closed() bool {
 	return rb.writeClosed
 }
 
@@ -166,10 +162,6 @@ func (rb *RingBuffer) Size() int {
 func (rb *RingBuffer) Positions() (start RingPos, end RingPos) {
 	rb.rwlock.RLock()
 	defer rb.rwlock.RUnlock()
-	return rb.positions()
-}
-
-func (rb *RingBuffer) positions() (start RingPos, end RingPos) {
 	return rb.readIndex, rb.writeIndex
 }
 
@@ -253,14 +245,14 @@ func (rb *RingBuffer) TailIterator() Iterator {
 	defer rb.rwlock.RUnlock()
 	rb.iteratorMutex.Lock()
 	defer rb.iteratorMutex.Unlock()
-	start, _ := rb.positions()
+	start := rb.readIndex
 	iter := &iterator{
 		rb:        rb,
 		index:     start,
 		nextChan:  make(chan bool, 1),
 		closeChan: make(chan struct{}),
 	}
-	if rb.closed() {
+	if rb.writeClosed {
 		close(iter.closeChan)
 	}
 	rb.iteratorList = append(rb.iteratorList, iter)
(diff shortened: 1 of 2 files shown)


internals/servicelog/ringbuffer.go
```diff
@@ -158,6 +166,10 @@ func (rb *RingBuffer) Size() int {
 func (rb *RingBuffer) Positions() (start RingPos, end RingPos) {
 	rb.rwlock.RLock()
 	defer rb.rwlock.RUnlock()
+	return rb.positions()
+}
+
+func (rb *RingBuffer) positions() (start RingPos, end RingPos) {
```
benhoyt, line 172: We could just get rid of `positions` entirely, as the only other place it's used is in `TailIterator`, where we use the read index and throw away the write index. Could just use `rb.readIndex` directly there.

Similar with `closed` -- could just get rid of that and use `rb.writeClosed` directly in the couple of places it's needed.

internals/servicelog/ringbuffer.go
```diff
@@ -258,6 +272,8 @@ func (rb *RingBuffer) TailIterator() Iterator {
 // backwards from the head.
 func (rb *RingBuffer) HeadIterator(lines int) Iterator {
 	firstLine := rb.reverseLinePosition(lines)
+	rb.rwlock.RLock()
+	defer rb.rwlock.RUnlock()
```
benhoyt, line 276: This is now a bit messy, as `rwlock` is locked and unlocked twice now, once inside `reverseLinePosition` and once here. Could we get rid of the locking in `reverseLinePosition` and do it once in this function?

benhoyt (commented):
Looking good, thanks. Just a couple of nit comments to see if we can tidy up a couple of things.

internals/servicelog/ringbuffer.go
```diff
@@ -73,6 +75,10 @@ func (rb *RingBuffer) Close() error {
 func (rb *RingBuffer) Closed() bool {
 	rb.rwlock.RLock()
 	defer rb.rwlock.RUnlock()
+	return rb.closed()
+}
+
+func (rb *RingBuffer) closed() bool {
```
dimaqq, line 81: Given that it's a simple attribute access, there's arguably no need to protect it with a lock

benhoyt: I don't think that's true in general. It might work here, as it's a single-machine-word `bool` type, but it doesn't seem like good practice to rely on this. For example, a string, slice, or interface attribute would be a "simple attribute", but 2 or 3 machine words.

IronCore864:
I have refactored according to the code review from Ben and Dima, and then the final tests with `-race` and manual deadlock testing passed successfully. @hpidcock please take a look.

benhoyt (approved):
Looks good to me, and passes my deadlock tests locally.

benhoyt:
I'm going to merge this optimistically so we can move ahead with https://github.com/canonical/pebble/commit/c3cb594050df6822c4a71da50848bb63854ee845, but @hpidcock -- we'd still appreciate your review-after-the-fact when you get a chance.

hpidcock:
> I'm going to merge this optimistically so we can move ahead with [c3cb594](https://github.com/canonical/pebble/commit/c3cb594050df6822c4a71da50848bb63854ee845), but @hpidcock -- we'd still appreciate your review-after-the-fact when you get a chance.

LGTM. Looks like all the previous locking behaviour is preserved with the consistent ordering to avoid deadlocks.

PR #512, opened 2024-10-18
fix: allow stopping services in the starting state

IronCore864:
Previously, we introduced [a fix to allow stopping services in the starting state](https://github.com/canonical/pebble/pull/503).

Because of this fix, we discovered [another deadlock issue](https://github.com/canonical/pebble/issues/508), so we rolled it back. Now that the deadlock issue is [fixed by this PR](https://github.com/canonical/pebble/pull/511/files), we are reintroducing the fix about "allowing stopping services in the starting state".

Benchmark test, manual test to reproduce the 3-way deadlock issue, and race test are all done and passed.


benhoyt (approved):
Thanks! Change is identical to the already-reviewed #503, approving.

PR #516, opened 2024-11-08
docs: remove reusable content from search result

IronCore864:
Remove reusable content pages by excluding the `reuse` directory.

Currently, if we search certain keywords in the reusable content, the reusable content pages themselves also show up in the search result. This is because they are rendered as real pages. This PR fixes that by ignoring all pages under the `reuse` directory.

We have considered two other options:

1. We could add `nosearch: true` keyword to the reusable content pages so they are not searchable, but they are still rendered as doc pages and if you know the URL you can still access them.
2. Change the reusable content pages from markdown to text files, and txt files won't be rendered as separate pages. In the starter pack there is already a `link.txt` reusable page, but it's actually "lying" since the file isn't a txt but rst. And, if we accidentally added some other files into the reuse folder in the future, like a new markdown or rst file, it's still not ignored, so we preferred the [custom_excludes](https://github.com/canonical/pebble/blob/5f5973d2a1734baa856493236dcf5d82eec3decc/docs/custom_conf.py#L182-L185) method in this PR.

[Preview here.](https://canonical-pebble--516.com.readthedocs.build/en/516/)


benhoyt:
This looks like it prevents it surfacing in search, but it's still accessible as a page: https://canonical-pebble--516.com.readthedocs.build/en/516/reuse/service-start-order/

Can we prevent that?

IronCore864:
> This looks like it prevents it surfacing in search, but it's still accessible as a page: https://canonical-pebble--516.com.readthedocs.build/en/516/reuse/service-start-order/
> 
> Can we prevent that?

Resolved it once and for all :) 404 for `reuse/service-start-order` now.

benhoyt:
That looks better, thanks.

I don't mind the renaming if that's the only way to do that. But is there a way to say "don't add pages for anything under the 'reuse' directory?" Because these files are actually Markdown, right, so we're now kind of lying about the file extension. Would it work to add `reuse` to [`custom_excludes`](https://github.com/canonical/pebble/blob/5f5973d2a1734baa856493236dcf5d82eec3decc/docs/custom_conf.py#L182-L185)?

benhoyt (approved):
Yep, this looks like the right approach and much simpler to me, thanks.

IronCore864:
> reuse/service-start-order

Technically they are not markdown but rst, and in the starter pack there is already a `link.txt` reusable file (although we didn't use it) which is in fact rst.

With the previous change on reusable content and terminal output format for CLI command references, many "*.md" (like `verify.md` and other markdown files not in the reuse directory) are not fully markdown anymore and can't be fully rendered properly by a local markdown editor.

But I do agree that excluding the reuse directory in the [custom_excludes](https://github.com/canonical/pebble/blob/5f5973d2a1734baa856493236dcf5d82eec3decc/docs/custom_conf.py#L182-L185) is much better, because no matter what file extensions we might have in the reuse directory in the future, it's guaranteed to be ignored. See the latest commit.

PR #517, opened 2024-11-11
PoC: auto-restart services failed within okaydelay when startup is enabled

IronCore864:
Per [spec here](https://docs.google.com/document/d/1J4XeP-e_UH5fSIiJ0urO2PWNKAmcNFC6o95igOHsH-c/edit?tab=t.0) and [previous discussion here](https://github.com/canonical/pebble/pull/513), we transition the service straight into the running state, by passing okayDelay and the starting state, so that Pebble will restart it if it fails, if the service is configured with `startup: enabled`.

### DIFF
--- a/internals/overlord/servstate/handlers.go
+++ b/internals/overlord/servstate/handlers.go
@@ -331,8 +331,16 @@ func (s *serviceData) start() error {
 		if err != nil {
 			return err
 		}
-		s.transition(stateStarting)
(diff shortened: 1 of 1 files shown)


IronCore864:
After reviewing the spec, we are in favour of the other option, see the [PoC PR here](https://github.com/canonical/pebble/pull/514). Closing this one, and will create a new PR for this feature.

PR #519, opened 2024-11-13
feat: metrics endpoint (for services and checks)

IronCore864:
Add simple metrics for services and health checks in OpenTelemetry exposition format.

Metrics include:

- `pebble_service_active{service="foo"}`
- `pebble_service_starts_total{service="foo"}`
- `pebble_check_up{check="bar"}`
- `pebble_check_success_count{check="bar"}`
- `pebble_check_failure_count{check="bar"}`

More details:

- Previously, we thought about a self-implemented metrics library, but after spec review and discussion, it is not easy for memory management (removing metrics when services are removed, for example). So we implemented these metrics on existing structs (`serviceData` and `CheckInfo`).
- Further, metrics should be stored on an unexported struct, so a new `checkData` is created.
- To make updating metrics easier, `CheckManager.checks` is updated from `map[string]CheckInfo` to pointers `map[string]*checkData`.
- An interface is created so that if we need to export the metrics in another format, we can extend it easier.
- There is a discussion after the spec review where we agreed to update `perform_check_count` and `recover_check_count` to `check_success_count` and `check_failure_count` since they provide more insight.
- Note that this PR contains a basic type identity from https://github.com/canonical/pebble/pull/563, which needs to be merged first.


IronCore864:
Screenshot-counter
Screenshot-gauge
Screenshot-gauge

IronCore864:
Some investigation into the default metrics that come with the Prometheus Go client:

## 1 List of Metrics from Go `runtime/metrics`

c.a. 80 items.

- /cgo/go-to-c-calls:calls
- /cpu/*
- /gc/cycles/*
- /gc/gogc:percent
- /gc/gomemlimit:bytes
- /gc/heap/*
- /gc/limiter/last-enabled:gc-cycle
- /gc/pauses:seconds
- /gc/scan/*
- /gc/stack/starting-size:bytes
- /godebug/*
- /memory/*
- /sched/gomaxprocs:threads
- /sched/goroutines:goroutines
- /sched/latencies:seconds
- /sched/pauses/*
- /sync/mutex/wait/total:seconds

---

## 2 List of Default Metrics from `prometheus/client_golang`

c.a. 40 items.

- go_gc_duration_seconds{quantile="*"}
- go_gc_duration_seconds_sum
- go_gc_duration_seconds_count
- go_gc_gogc_percent
- go_gc_gomemlimit_bytes
- go_goroutines
- go_info{version="go1.23.1"}
- go_memstats_*
- go_sched_gomaxprocs_threads
- go_threads
- promhttp_metric_handler_requests_in_flight
- promhttp_metric_handler_requests_total{code="200"}
- promhttp_metric_handler_requests_total{code="500"}
- promhttp_metric_handler_requests_total{code="503"}

---

## 3 How Are the Default Metrics from `prometheus/client_golang` Fetched/Calculated

### 3.1 From `runtime` and `runtime/debug`

- go_goroutines
- go_threads
- go_gc_duration_seconds
- go_memstats_last_gc_time_seconds
- go_info

### 3.2 From `runtime.MemStats` with Some Calculations

- go_memstats_alloc_bytes
- go_memstats_alloc_bytes_total
- go_memstats_buck_hash_sys_bytes
- go_memstats_frees_total
- go_memstats_gc_sys_bytes
- go_memstats_heap_alloc_bytes
- go_memstats_heap_idle_bytes
- go_memstats_heap_inuse_bytes
- go_memstats_heap_objects
- go_memstats_heap_released_bytes
- go_memstats_heap_sys_bytes
- go_memstats_last_gc_time_seconds
- go_memstats_mallocs_total
- go_memstats_mcache_inuse_bytes
- go_memstats_mcache_sys_bytes
- go_memstats_mspan_inuse_bytes
- go_memstats_mspan_sys_bytes
- go_memstats_next_gc_bytes
- go_memstats_other_sys_bytes
- go_memstats_stack_inuse_bytes
- go_memstats_stack_sys_bytes
- go_memstats_sys_bytes

### 3.3 Directly from runtime/metrics but Go-Version-Dependant

- go_gc_gogc_percent
- go_gc_gomemlimit_bytes
- go_sched_gomaxprocs_threads

### 3.4 Prometheus-Related

- promhttp_metric_handler_requests_in_flight
- promhttp_metric_handler_requests_total{code="200"}
- promhttp_metric_handler_requests_total{code="500"}
- promhttp_metric_handler_requests_total{code="503"}

Self-processed.

IronCore864:
According to the last spec review, the following changes have been made:

1. The identity type is renamed from "basicauth" to just "basic".
2. The username field in the basic identity is removed -- we can just use the name of the identity instead.
3. Adding a new access type "metrics".
4. Passwords are hashed using OpenSSL (_TODO_).

After the first round of refactoring, here are some results:

### 1 Baisc Identity Name with Special Characters

```bash
$ cat identity.yaml
identities:
    "bob:asdf":
        access: read
        basic:
            password: bar
$ ./pebble add-identities --from ./identity.yaml
error: identity "bob:asdf" invalid: identity name "bob:asdf" contains invalid characters (only
       alphanumeric, underscore, and hyphen allowed)
```

### 2 Baisc Identity without Username

```bash
$ cat identity.yaml
identities:
    bob:
        access: read
        basic:
            password: bar
ubuntu@primary:~/work/pebble2$ ./pebble add-identities --from ./identity.yaml
Added 1 new identity.
```

### 3 Basic Identity Type "metrics"

```bash
$ # access type: read
$ cat identity.yaml
identities:
    bob:
        access: read
        basic:
            password: bar
$ ./pebble add-identities --from ./identity.yaml
Added 1 new identity.
$ # open access is fine
$ curl -u bob:bar localhost:4000/v1/health
{"type":"sync","status-code":200,"status":"OK","result":{"healthy":true}}
$ # no access on the metrics endpoint
$ curl -u bob:bar localhost:4000/metrics
```

```bash
$ # access type: metrics
$ cat identity.yaml
identities:
    bob:
        access: metrics
        basic:
            password: bar
$ ./pebble update-identities --from ./identity.yaml
Updated 1 identity.
$ # open access is fine
$ curl -u bob:bar localhost:4000/v1/health
{"type":"sync","status-code":200,"status":"OK","result":{"healthy":true}}
$ # accessing metrics
$ curl -u bob:bar localhost:4000/metrics
# HELP my_counter Total number of something processed.
# TYPE my_counter counter
my_counter{operation=read,status=success} 11
my_counter{operation=write,status=success} 22
my_counter{operation=read,status=failed} 11
# HELP my_gauge Current value of something.
# TYPE my_gauge gauge
my_gauge{sensor=temperature} 28.12
$ # no access on other endpoints
$ curl -u bob:bar localhost:4000/v1/changes
```

```bash
$ # access type: admin
$ cat identity.yaml
identities:
    bob:
        access: admin
        basic:
            password: bar
$ ./pebble update-identities --from ./identity.yaml
Updated 1 identity.
$ # admin can read metrics
$ curl -u bob:bar localhost:4000/v1/metrics
# HELP my_counter Total number of something processed.
# TYPE my_counter counter
my_counter{operation=read,status=success} 176
my_counter{operation=write,status=success} 352
my_counter{operation=read,status=failed} 176
# HELP my_gauge Current value of something.
# TYPE my_gauge gauge
my_gauge{sensor=temperature} 24.48
```

_TODO_: hashing password.

internals/overlord/state/state.go
```diff
@@ -172,17 +172,16 @@ type marshalledState struct {
 // marshalledIdentity is used specifically for marshalling to the state
 // database file. Unlike apiIdentity, it should include secrets.
 type marshalledIdentity struct {
-	Access    string                       `json:"access"`
-	Local     *marshalledLocalIdentity     `json:"local,omitempty"`
-	BasicAuth *marshalledBasicAuthIdentity `json:"basicauth,omitempty"`
+	Access string                   `json:"access"`
+	Local  *marshalledLocalIdentity `json:"local,omitempty"`
+	Basic  *marshalledBasicIdentity `json:"basi,omitempty"`
```
benhoyt, line 177: Just skimming through this PR (not reviewing it yet), but noticed this, so commenting before I forget: basi -> basic

IronCore864:
Notes:

We need to handle the memory usage issue in the future, to make this easier, after discussion, we decided not to use a self-implemented Prometheus-like module to store the metrics centrally, but rather, store the metrics on existing structs like `serviceData` and `CheckInfo`. In the last commit, the service-related metrics are implemented on `serviceData`. Check-related metrics are to be implemented.

IronCore864:
Taking some note before the holiday season:

Currently, the check metrics is only working when the check is successful. When it fails, both counters will be reset to 0 and never increase. Need more debugging.

internals/daemon/api_metrics.go
```diff
@@ -0,0 +1,53 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	err = r.chkMgr.Metrics(&buffer)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	w.Write([]byte(buffer.String()))
```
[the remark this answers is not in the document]
benhoyt: Do we need to handle the error here, and at least log it (we probably can't return an HTTP error at this point)?

benhoyt (changes requested):
Several comments, most importantly the design of the Metric type and the WriteMetrics interface.

internals/daemon/api_metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	err = r.chkMgr.Metrics(&buffer)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
IronCore864: I noticed that `api_files` has it explicitly. To be consistent with existing code, I'd like to keep this line.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/canonical/pebble
 go 1.22
 
 require (
+	github.com/GehirnInc/crypt v0.0.0-20230320061759-8cc1b52080c5
```
gruyaume, line 6: Is this package maintained? Are we confident about introducing it in every charm / pebble deployment out there?

gruyaume (commented):
I was curious about this change listening to the mid-cycle roadmap presentation. Nobody asked for this review, please feel free to ignore it.

go.mod
benhoyt: It hasn't been recently changed, but the part of it we're using is small a straight-forward, and the algorithm being implemented is stable and well-documented ([SHA-crypt](https://www.akkadia.org/drepper/SHA-crypt.txt)). So I would say it's "stable" rather than "unmaintained". I'm actually contemplating vendoring this (the core part of it is only ~100 LoC). But open to other ideas/concerns too.

benhoyt: We've done some diligence on it, see https://github.com/canonical/pebble/pull/563#issuecomment-2635646862 for example.

internals/overlord/checkstate/manager_test.go
```diff
@@ -869,3 +868,54 @@ func (s *ManagerSuite) TestReplan(c *C) {
...
+				Period:    plan.OptionalDuration{Value: 10 * time.Millisecond},
+				Timeout:   plan.OptionalDuration{Value: time.Second},
+				Threshold: 3,
+				Exec:      &plan.ExecCheck{Command: "echo chk1"},
+				Startup:   plan.CheckStartupEnabled,
+			},
+		},
+	})
+	waitChecks(c, s.manager, []*checkstate.CheckInfo{
+		{Name: "chk1", Startup: "enabled", Status: "up", Threshold: 3, PerformCheckCount: 2, RecoverCheckCount: 0},
```
benhoyt, line 904: Hmm, yes, we definitely shouldn't be getting these fields coming through here. In fact, we should either have an internal `checkInfo` (with the metrics fields) and an exported one `CheckInfo` without them, or we should explicitly set them to zero after we've made the copy in `Checks()` -- probably the latter is simplest.

For testing, we should call the `WriteMetrics` method with a fake `MetricsWriter` that just appends to a slice or something, and compare that. Or just use an OpenTelemetryWriter and compare the output -- I see you're already doing that for the services metrics.

internals/overlord/servstate/handlers.go
```diff
@@ -287,7 +290,7 @@ func (s *serviceData) transition(state serviceState) {
 
 // transitionRestarting changes the service's state and also sets the restarting flag.
 func (s *serviceData) transitionRestarting(state serviceState, restarting bool) {
-	// Update current-since time if derived status is changing.
+	// Update current-s_e time if derived status is changing.
```
benhoyt, line 293: Search/replace issue?

internals/overlord/servstate/handlers.go
```diff
@@ -825,6 +828,37 @@ func (s *serviceData) checkFailed(action plan.ServiceAction) {
...
+
+	active := 0
+	if stateToStatus(d.state) == StatusActive {
+		active = 1
+	}
+	err = writer.Write(metrics.Metric{
+		Name:       "pebble_service_active",
+		Type:       metrics.TypeGaugeInt,
+		ValueInt64: int64(active),
+		Comment:    "Indicates if the service is currently active (1) or not (0)",
```
benhoyt, line 852: Let's use the same phrasing here as for the checks one, so "Whether the service is..."

internals/overlord/servstate/manager_test.go
```diff
@@ -2093,3 +2094,49 @@ func waitChecks(c *C, checkMgr *checkstate.CheckManager, f func(checks []*checks
...
+	s.newServiceManager(c)
+	s.planAddLayer(c, testPlanLayer)
+	s.planChanged(c)
+
+	s.startTestServices(c, true)
+	if c.Failed() {
+		return
+	}
+	buf := new(bytes.Buffer)
+	openTelemetryWriter := metrics.NewOpenTelemetryWriter(buf)
```
benhoyt, line 2108: Nit: given it's clear from the context, how about just `writer := ...`?

internals/overlord/servstate/manager_test.go
```diff
@@ -2093,3 +2094,49 @@ func waitChecks(c *C, checkMgr *checkstate.CheckManager, f func(checks []*checks
...
+	s.manager.WriteMetrics(openTelemetryWriter)
+	metrics := buf.String()
+	c.Assert(metrics, testutil.Contains, "pebble_service_start_count{service=test1} 1")
```
benhoyt, line 2111: Should we just assert on the entire output, for simplicity, and so we're testing the HELP and TYPE parts?

internals/overlord/servstate/manager.go
```diff
@@ -334,6 +335,20 @@ func (m *ServiceManager) CheckFailed(name string) {
 	}
 }
 
+// WriteMetrics collects and writes metrics for all services to the provided writer.
+func (m *ServiceManager) WriteMetrics(writer metrics.Writer) error {
+	m.servicesLock.Lock()
+	defer m.servicesLock.Unlock()
+
+	for _, service := range m.services {
```
benhoyt, line 343: It's probably best to make the order stable, so let's grab the service names into a slice and `sort.Strings` them before looping. Otherwise the order of `/v1/metrics` will change each time you refresh it, which is not ideal.

Similar for the checks `WriteMetrics`.

internals/overlord/checkstate/manager.go
```diff
@@ -387,6 +427,62 @@ type checker interface {
...
+
+	return nil
+}
+
+// WriteMetrics collects and writes metrics for all checks to the provided writer.
+func (m *CheckManager) WriteMetrics(writer metrics.Writer) error {
+	m.checksLock.Lock()
+	defer m.checksLock.Unlock()
+
+	for _, info := range m.checks {
```
benhoyt, line 476: Per similar comment on `ServiceManager.WriteMetrics`, let's sort by check name to ensure the output is stable.

benhoyt (changes requested):
Mostly minor comments, though a few structural things. Main thing that needs discussion is whether the perform_check and recover_check counts are actually going to provide the monitoring we want -- let's discuss.

benhoyt:
@IronCore864 Can you please update the PR description to match our new approach?

Also, it probably goes without saying, but let's be sure **not to merge** this before the underlying identities PR that this builds on (#563) is reviewed for security and merged.

internals/metrics/metrics_test.go
```diff
@@ -0,0 +1,106 @@
...
+			metric: metrics.Metric{
+				Name:       "my_gauge",
+				Type:       metrics.TypeGaugeInt,
+				ValueInt64: 1,
+				Comment:    "A simple gauge",
+				Labels:     []metrics.Label{}, // Test with no labels
+			},
+			expected: `# HELP my_gauge A simple gauge
+# TYPE my_gauge gauge
+my_gauge{} 1
```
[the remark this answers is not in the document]
IronCore864: I did more research: If there is no label, there should not be brackets. If a metric must have a label like the ones we have, we can't set it without a label anyway. I will update the code and test accordingly.

internals/metrics/metrics_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		{
+			name: "NoComment", // Test without comment
+			metric: metrics.Metric{
+				Name:       "no_comment_metric",
+				Type:       metrics.TypeCounterInt,
+				ValueInt64: 42,
+				Labels:     []metrics.Label{metrics.NewLabel("env", "prod")},
+			},
+			expected: `# HELP no_comment_metric
```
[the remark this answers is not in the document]
IronCore864: It's possible to have no `# HELP` and no `# TYPE` lines for a metric. I will remove the `# HELP` line if there is no comment. However, since we will definitely have a type, I will make the `# TYPE` mandatory.

internals/overlord/checkstate/handlers.go
```diff
@@ -91,6 +91,7 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 				task.Set(checkDetailsAttr, &details)
 				m.state.Unlock()
 			}
+			m.incPerformCheckCount(config)
```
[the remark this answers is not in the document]
IronCore864: If we put it in the beginning right after `runCheck`, it's possible that there is no CheckInfo yet (`updateCheckInfo` is called later).

I agree `check_failure_count` (and maybe `check_success_count` as well, which was what we had in the first place in the spec) is more useful. In the spec review, we only decided to use perform/recover check count, because they coincided with the type of the change, which does not seem to be a strong argument now that I look at it again. It does not matter where the `runCheck` is called (`doPerformCheck` or `doRecoverCheck`), it should depend on the result of `runCheck`. I will update them to `check_failure_count` and `check_success_count`

Whether the check failed or succeeded: I accidentally forgot to add this metric. I will add it (`pebble_check_up{check=bar}`).

internals/metrics/metrics.go
```diff
@@ -27,6 +26,17 @@ const (
...
 )
 
+func (mt MetricType) String() string {
+	switch mt {
+	case TypeCounterInt:
+		return "counter"
+	case TypeGaugeInt:
+		return "gauge"
+	default:
+		panic("invalid metric type")
```
benhoyt, line 36: Nit: best to include the type number in the message:
```suggestion
		panic(fmt.Sprintf("invalid metric type %d", mt))
```

internals/metrics/metrics.go
```diff
@@ -63,28 +73,29 @@ func NewOpenTelemetryWriter(w io.Writer) *OpenTelemetryWriter {
...
-	_, err = fmt.Fprintf(otw.w, "# TYPE %s %s\n", m.Name, metricType)
+	_, err := fmt.Fprintf(otw.w, "# TYPE %s %s\n", m.Name, m.Type.String())
 	if err != nil {
 		return err
 	}
 
-	labels := make([]string, len(m.Labels))
-	for i, label := range m.Labels {
-		labels[i] = fmt.Sprintf("%s=%s", label.key, label.value)
+	io.WriteString(otw.w, m.Name)
```
benhoyt, line 88: Note that we're not checking errors here (but we are above). In this case that's actually okay given that we know the writer is a `bytes.Buffer` (which never fails), but for future-proofing and reliability, let's check each error here.

internals/metrics/metrics.go
```diff
@@ -63,28 +73,29 @@ func NewOpenTelemetryWriter(w io.Writer) *OpenTelemetryWriter {
...
+	if len(m.Labels) > 0 {
+		io.WriteString(otw.w, "{")
+		for i, label := range m.Labels {
+			if i > 0 {
+				io.WriteString(otw.w, ",")
+			}
+			fmt.Fprintf(otw.w, "%s=%q", label.key, label.value) // Use %q to quote values.
```
benhoyt, line 95: Ah, good catch that these are quoted.

internals/overlord/checkstate/manager.go
```diff
@@ -359,40 +348,37 @@ func (m *CheckManager) updateCheckInfo(config *plan.Check, changeID string, fail
...
+	check.Threshold = config.Threshold
+	check.ChangeID = changeID
 }
 
 func (m *CheckManager) incPerformCheckCount(config *plan.Check) {
 	m.checksLock.Lock()
 	defer m.checksLock.Unlock()
 
-	m.checks[config.Name].PerformCheckCount += 1
+	check, ok := m.checks[config.Name]
```
benhoyt, line 368: We should do this check in `incRecoverCheckCount` as well. Maybe we can have a little helper method `ensureCheck(name string) *CheckInfo` that does this dance, to avoid repeating it three times.

internals/metrics/metrics.go
IronCore864: Got it, I will check all errors for all `io.WriteString` and `fmt.Fprintf` to the writer.

internals/overlord/checkstate/handlers.go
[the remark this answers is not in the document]
IronCore864: After discussing it with Ben, Tony, and Harry, we decided to use `pebble_check_success_count` and `pebble_check_failure_count` because they provide more insight than the previously decided `pebble_perform_check_count` and `pebble_recover_check_count`.

IronCore864:
Check metrics updated, tests also updated.

internals/daemon/api_metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+
+	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/metrics"
+	"github.com/canonical/pebble/internals/overlord/checkstate"
+	"github.com/canonical/pebble/internals/overlord/servstate"
+)
+
+func v1GetMetrics(c *Command, r *http.Request, _ *UserState) Response {
+	return metricsResponse{
```
benhoyt, line 28: I realise I've done this elsewhere (like the files API), but you can actually simplify and do a handler like inline with:

```go
return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
    func v1GetMetrics(c *Command, r *http.Request, _ *UserState) Response {
        ...
    }
}
```

internals/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+
+		_, err = io.WriteString(otw.w, "}")
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err = fmt.Fprintf(otw.w, " %d\n", m.ValueInt64)
```
benhoyt, line 118: For a human debugging the output, it would be nice to separate each comment+metric block with a blank line:
```suggestion
	_, err = fmt.Fprintf(otw.w, " %d\n\n", m.ValueInt64)
```

internals/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+type MetricType int
+
+const (
+	TypeCounterInt MetricType = iota
```
benhoyt, line 25: It's probably worth doing `iota+1` so these start at one, so we don't accidentally forget to set `Type` and get the zero value (`TypeCounterInt`) by accident.

internals/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (otw *OpenTelemetryWriter) Write(m Metric) error {
+	if m.Comment != "" {
+		_, err := fmt.Fprintf(otw.w, "# HELP %s %s\n", m.Name, m.Comment)
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err := fmt.Fprintf(otw.w, "# TYPE %s %s\n", m.Name, m.Type.String())
```
benhoyt, line 83: `.String()` is done for you by the `fmt` package when you use the `%s` verb:
```suggestion
	_, err := fmt.Fprintf(otw.w, "# TYPE %s %s\n", m.Name, m.Type)
```

internals/overlord/checkstate/manager.go
```diff
@@ -375,6 +408,19 @@ type CheckInfo struct {
 	ChangeID  string
 }
 
+// checkData holds the metrics and other data for a a single check.
+type checkData struct {
```
benhoyt, line 412: Yeah, I think this separation makes things a lot cleaner now, thanks.

internals/overlord/checkstate/manager.go
```diff
@@ -387,6 +433,69 @@ type checker interface {
 	check(ctx context.Context) error
 }
 
+func (c *checkData) writeMetrics(writer metrics.Writer) error {
```
benhoyt, line 436: This should be singular:
```suggestion
func (c *checkData) writeMetric(writer metrics.Writer) error {
```

internals/overlord/checkstate/manager.go
```diff
@@ -387,6 +433,69 @@ type checker interface {
...
+	checkUp := int64(0)
+	if c.status == CheckStatusUp {
```
benhoyt, line 438: @tonyandrewmeyer I forget what we decided here, do you remember: should CheckStatusInactive mean the "up" metric report up (1) or down (0)? I was thinking it would be up.

internals/overlord/checkstate/manager.go
```diff
@@ -387,6 +433,69 @@ type checker interface {
...
+		checkUp = 1
+	}
+	err := writer.Write(metrics.Metric{
+		Name:       "pebble_check_up",
+		Type:       metrics.TypeGaugeInt,
+		ValueInt64: checkUp,
```
benhoyt, line 444: I don't really care which way we do this (int64 here or above), but let's at least be consistent between checks and services (we're currently doing it int64-on-variable here and int64-on-field in services). Now that I see them I like the services way better, so let's change this to int64-on-field (this line).

internals/overlord/servstate/handlers.go
```diff
@@ -825,6 +828,37 @@ func (s *serviceData) checkFailed(action plan.ServiceAction) {
 	}
 }
 
+// writeMetrics writes the service's metrics.
+func (d *serviceData) writeMetrics(writer metrics.Writer) error {
```
benhoyt, line 832:
```suggestion
func (d *serviceData) writeMetric(writer metrics.Writer) error {
```

internals/overlord/checkstate/manager.go
tonyandrewmeyer: I don't think we decided anything here - the discussion on this predated the start/stop work being completed, if I recall correctly.

When we combine checks together (e.g. for health determination) stopped checks are skipped - treated as if they don't exist. When we list checks or get information about a single check elsewhere, we don't have to have a binary choice, so we have up/down/inactive.

Here:
 * 0 (down) definitely seems wrong because it implies that the result of the check is that it's failing.
 * 1 (up) seems pretty wrong because it implies that the result of the check is success, even though we don't actually know - but it's a little better because it's the same effect for things like checking health
 * 2 or some other number meaning "inactive" doesn't seem great for a gauge type - I feel like gauge values ought to be actual numbers rather than enums.

I think the most correct behaviour would be to not list any inactive checks, because they don't have an up or down status. This is like health: they are the same as if they didn't exist.

For the other check metrics, check_success_count, check_failure_count, having inactive checks makes sense (and we did adjust the start/stop code so that the check-info failure count is preserved for inactive checks, so having the metric available is consistent).

benhoyt: > I think the most correct behaviour would be to not list any inactive checks, because they don't have an up or down status. This is like health: they are the same as if they didn't exist.

Ah yes, that's definitely what we should do -- exclude them from /v1/metrics altogether.

@IronCore864 Let's do that. It would have to be checked in the `WriteMetrics` loop.

IronCore864: I agree using gauge like it's an enum doesn't feel right. I will choose to not report pebble_check_up for inactive checks.

docs/specs/openapi.yaml
```diff
@@ -801,6 +801,45 @@ paths:
                   "service": "svc1",
                   "message": " * Serving Flask app 'main'"
                 }
+  /v1/metrics:
+    get:
+      summary: Get Pebble metrics
+      description: |
+        Get Pebble services and health checks metrics in [OpenMetrics](https://github.com/prometheus/OpenMetrics) format.
+
+        This endpoint requires HTTP basic authentication using an identity of type "basic". See [Identities](../identities) for more information.
```
benhoyt, line 810:
```suggestion
        When used over TCP, this endpoint requires HTTP basic authentication using an identity of type "basic". See [Identities](../identities) for more information.
```

internals/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+func (mt MetricType) String() string {
+	switch mt {
+	case TypeCounterInt:
+		return "counter"
+	case TypeGaugeInt:
+		return "gauge"
+	default:
+		panic(fmt.Sprintf("invalid metric type %d", mt))
```
benhoyt, line 36:
```suggestion
		panic(fmt.Sprintf("internal error: invalid metric type %d", mt))
```

internals/metrics/metrics_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+func Test(t *testing.T) {
+	TestingT(t)
+}
+
+type OpenTelemetryWriterSuite struct{}
+
+var _ = Suite(&OpenTelemetryWriterSuite{})
+
+func (s *OpenTelemetryWriterSuite) TestOpenTelemetryWriter(c *C) {
```
benhoyt, line 34: Nit: because this is in `OpenTelemetryWriterSuite`, the test name can just be `TestWriter`:
```suggestion
func (s *OpenTelemetryWriterSuite) TestWriter(c *C) {
```

internals/overlord/checkstate/manager.go
```diff
@@ -375,6 +408,19 @@ type CheckInfo struct {
 	ChangeID  string
 }
 
+// checkData holds the metrics and other data for a a single check.
```
benhoyt, line 411:
```suggestion
// checkData holds the metrics and other data for a single check.
```

internals/overlord/checkstate/manager.go
```diff
@@ -387,6 +433,72 @@ type checker interface {
 	check(ctx context.Context) error
 }
 
+func (c *checkData) writeMetric(writer metrics.Writer) error {
+	// Not to list any inactive checks because they don't have an up or down status.
```
benhoyt, line 437:
```suggestion
	// Don't list any inactive checks because they don't have an up or down status.
```

internals/overlord/checkstate/manager_test.go
```diff
@@ -869,3 +870,133 @@ func (s *ManagerSuite) TestReplan(c *C) {
...
+	writer := metrics.NewOpenTelemetryWriter(buf)
+	s.manager.WriteMetrics(writer)
+	expected := `
+# HELP pebble_check_up Whether the health check is up (1) or not (0)
+# TYPE pebble_check_up gauge
+pebble_check_up{check="chk1"} 1
+
+# HELP pebble_check_success_count Number of times the check has succeeded
+# TYPE pebble_check_success_count counter
+pebble_check_success_count{check="chk1"} 2
```
benhoyt, line 914: Hmm, this is racy. If CI is slow and the check (which has a short period) runs 3 or 4 times, this will fail. The simplest way I can think of is using a regex to match , with `\d+` in this position. (I think any number, even 1 here, would be okay. I don't think there's a specific reason we need to wait for it to run twice.)

internals/overlord/checkstate/manager_test.go
```diff
@@ -869,3 +870,133 @@ func (s *ManagerSuite) TestReplan(c *C) {
...
+		},
+	})
+
+	// Wait for check to run at least twice
+	for i := 0; ; i++ {
+		if i >= 1000 {
+			c.Fatalf("failed waiting for check to run")
+		}
+		b, _ := os.ReadFile(tempFile)
+		if len(b) >= 2 {
```
benhoyt, line 898: See below -- could just wait for >= 1

internals/overlord/checkstate/manager_test.go
```diff
@@ -869,3 +870,133 @@ func (s *ManagerSuite) TestReplan(c *C) {
...
+# TYPE pebble_check_up gauge
+pebble_check_up{check="chk1"} 1
+
+# HELP pebble_check_success_count Number of times the check has succeeded
+# TYPE pebble_check_success_count counter
+pebble_check_success_count{check="chk1"} 0
+
+# HELP pebble_check_failure_count Number of times the check has failed
+# TYPE pebble_check_failure_count counter
+pebble_check_failure_count{check="chk1"} 2
```
benhoyt, line 963: I believe this is racy for the same reason as above.

internals/overlord/servstate/handlers.go
```diff
@@ -825,6 +828,37 @@ func (s *serviceData) checkFailed(action plan.ServiceAction) {
...
+		Type:       metrics.TypeCounterInt,
+		ValueInt64: d.startCount.Load(),
+		Comment:    "Number of times the service has started",
+		Labels:     []metrics.Label{metrics.NewLabel("service", d.config.Name)},
+	})
+	if err != nil {
+		return err
+	}
+
+	active := 0
```
benhoyt, line 844: It won't affect the metrics, but it would be nice to move the `pebble_service_active` metric to the top of `writeMetric` so that it's in the same position as the `pebble_check_up` gauge (top). It feels odd that this status gauge is last but the checks one is first. I think putting them both first is best.

benhoyt (changes requested):
This looks really good now, thanks. I've left some (minor) comments.

In addition, I think this probably needs to be highlighted a bit more in our docs. Maybe a new section with example output at https://documentation.ubuntu.com/pebble/how-to/run-services-reliably/? What do you think, @dwilding?

IronCore864:
> In addition, I think this probably needs to be highlighted a bit more in our docs. Maybe a new section with example output at https://documentation.ubuntu.com/pebble/how-to/run-services-reliably/? What do you think, @dwilding?

I discussed this with David and we agreed on the following outline:

```
Intro

## Use HTTP health checks

## Restart a service when the health check fails

## Access health metrics in OpenMetrics format

... This would explain how to create an identity and show a curl example of how to access metrics. Then link to somewhere about configuring Prometheus.

## Limitations of health checks

## See more
```

I will work on this in the same PR.

IronCore864:
"How to run services reliably" updated, preview: https://canonical-ubuntu-documentation-library--519.com.readthedocs.build/pebble/how-to/run-services-reliably/#access-health-metrics-in-openmetrics-format.

docs/how-to/run-services-reliably.md
```diff
@@ -57,6 +57,52 @@ services:
       svc1-up: restart
 ```
 
+## Access health metrics in OpenMetrics format
+
+If we run Pebble with the `--http` option, Pebble exposes the `/v1/metrics` endpoint over HTTP, providing metrics data in [OpenMetrics](https://github.com/prometheus/OpenMetrics) format. This endpoint is not open-access and requires HTTP basic authentication.
+
+To allow accessing the metrics endpoint with HTTP basic authentication, create a "basic" type identity and give it "metrics" access. Prepare this file:
```
dwilding, line 64:
```suggestion
To access the metrics endpoint with HTTP basic authentication, first create a "basic" type identity and give it "metrics" access. Prepare this file:
```

docs/how-to/run-services-reliably.md
```diff
@@ -57,6 +57,52 @@ services:
       svc1-up: restart
 ```
 
+## Access health metrics in OpenMetrics format
+
+If we run Pebble with the `--http` option, Pebble exposes the `/v1/metrics` endpoint over HTTP, providing metrics data in [OpenMetrics](https://github.com/prometheus/OpenMetrics) format. This endpoint is not open-access and requires HTTP basic authentication.
```
dwilding, line 62: Could drop the mention of open-access if you like:
```suggestion
If we run Pebble with the `--http` option, Pebble exposes the `/v1/metrics` endpoint over HTTP, providing metrics data in [OpenMetrics](https://github.com/prometheus/OpenMetrics) format. This endpoint and requires HTTP basic authentication.
```

docs/how-to/run-services-reliably.md
```diff
@@ -57,6 +57,52 @@ services:
...
+  alice:
+    access: metrics
+      basic:
+      # The password is hashed using sha512-crypt, as generated by "openssl passwd -6".
+      password: <password hash>
+```
+
+Then run `pebble add-identities --from idents-add.yaml`. See [Identities](../reference/identities) and [How to manage identities](../how-to/manage-identities) for more information.
+
+To access the metrics endpoint, make sure we run Pebble with the `--http` option: `pebble run --http=:4000`, then we can get the metrics data with HTTP basic authentication:
```
dwilding, line 78: I think we could shorten this to make the instructions more obvious:
```suggestion
To access the metrics endpoint, run `pebble run --http=:4000`, then use `curl` and specify the identity that we created:
```

docs/how-to/run-services-reliably.md
```diff
@@ -57,6 +57,52 @@ services:
...
+# TYPE pebble_check_success_count counter
+pebble_check_success_count{check="check1"} 2
+
+# HELP pebble_check_failure_count Number of times the check has failed
+# TYPE pebble_check_failure_count counter
+pebble_check_failure_count{check="check1"} 0
+
+```
+
+To configure Prometheus to scrape a target protected by HTTP basic authentication, add an `http_config` section in the `scrape_config`. Read more [here in the official documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
```
dwilding, line 104: Recommend adjusting the link:
```suggestion
To configure Prometheus to scrape a target protected by HTTP basic authentication, add an `http_config` section in the `scrape_config`. See the [Prometheus configuration documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
```

benhoyt:
Looks good to me, thanks! I've tested this a bit locally too. Merging now so I can include this in today's release.

benhoyt (approved):

Issue #183, opened 2023-01-23
HOME not set when only "user" is specified


axinojolais:
Using pebble on a charm deployed on a 2.9.32 juju controller

My app was misbehaving because HOME wasn't set to /home/foo even though I specified `user: "foo"` in the pebble plan. I found https://github.com/canonical/pebble/pull/129 which sets HOME if both `user` and `group` are set, but I would expect HOME to be set as soon as just `user` is set.

I don't have a reason to not set `group`, I just don't really care about the group (and I would also assume that just setting the user would also set the group to the user's GID for me).

Anyway, I can work around the issue by setting HOME explicitly (my juju version is too old to have #129), but it would be nice if pebble could do that for me.

Thanks !

benhoyt:
Thanks for this. We can dig in deeper, but at first glance it makes sense to set HOME when only user is specified.

For the record, yes, Juju 2.9.32 is on Pebble commit 32ead042e947, which doesn't include the change you referenced (or things like https://github.com/canonical/pebble/pull/114). The latest 2.9.x version (.39 I think) does. But as you point out, you'd still have to set user and group.

benhoyt:
Per discussion on #521, we can't actually see how this can happen looking at the code now, because of how `osutil.NormalizeUidGid` looks up the gid if group is not set (and, at the bottom of the function, returns an error if the plan didn't specify both user and group).

If this happens again or you can find a reproducer, feel free to reopen (or open a new issue).

PR #521, opened 2024-11-19
feat: set home when user is set

IronCore864:
Set the `HOME` environment variable when only uid is set.

Closes https://github.com/canonical/pebble/issues/183.

The above issue could not be reproduced, because even if only "user" is set without "group", [Pebble will find and use the user's primary group](https://github.com/canonical/pebble/blob/master/internals/osutil/user.go#L117), and [when setting home, both uid and gid are not empty](https://github.com/canonical/pebble/blob/master/internals/overlord/servstate/handlers.go#L374).

The code change still feels right after a discussion with Tony and Dima in the standup, though, because setting HOME should have nothing to do with gid, empty or not.

On a probably unrelated topic: There is an unrelated issue (or feature) where [if user-id is used (instead of user), group-id must be specified at the same time](https://github.com/canonical/pebble/blob/master/internals/osutil/user.go#L141); otherwise, the service start will fail. But I believe this isn't what the original issue is about.

After discussing the issue with the owner @axinojolais who didn't have time to try to reproduce, we decided it's OK to close this issue. If it happens again, they will reopen this.

### DIFF
--- a/internals/overlord/servstate/handlers.go
+++ b/internals/overlord/servstate/handlers.go
@@ -382,19 +382,19 @@ func (s *serviceData) startInternal() error {
 				Gid: uint32(*gid),
 			})
 		}
+	}
 
-		// Also set HOME and USER if not explicitly specified in config.
-		if environment["HOME"] == "" || environment["USER"] == "" {
-			u, err := user.LookupId(strconv.Itoa(*uid))
-			if err != nil {
-				logger.Noticef("Cannot look up user %d: %v", *uid, err)
-			} else {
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Yeah, I also can't see how this can happen looking at the code now, because of how `osutil.NormalizeUidGid` looks up the gid if group is not set (and, at the bottom, returns an error if they didn't specify both user and group).

However, this change makes the code "more correct" as we only need `uid` in this code path, so approving.

PR #524, opened 2024-11-26
docs: make landing pages more informative

dwilding:
The purpose of this PR is to use a consistent style for the three landing pages [How-to guides](https://canonical-pebble.readthedocs-hosted.com/en/latest/how-to/), [Explanation](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/), and [Reference](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/). How-to guides already has the style we want, so I'm using the same style for Explanation and Reference.

Summary of changes:
- **How-to guides** - Adjusted some wording for readability and consistency of language
- **Explanation** - Moved [General model](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/general-model/) to the top of the TOC; split the landing page into sections with brief summaries
- **Reference** - Applied a logical order to the TOC; split the landing page into sections with brief summaries

### DIFF
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -1,11 +1,11 @@
 # How-to guides
 
-These guides walk you through the operations lifecycle of Pebble.
+These guides walk you through key operations you can perform with Pebble.
 
 
 ## Installation
 
-Installation follows a broadly similar pattern on all architectures. You can choose to install the pre-built binary or build it from the source by yourself.
+Installation follows a similar pattern on all architectures. You can choose to install the pre-built binary or build it from the source by yourself.
 
 ```{toctree}
 :titlesonly:
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -29,7 +29,7 @@ Manage service dependencies <service-dependencies>
 
 ## Identities
 
-Use named identities to allow additional users to access the API.
+Use named "identities" to allow additional users to access the API.
 
 ```{toctree}
 :titlesonly:
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -1,11 +1,11 @@
 # How-to guides
 
-These guides will walk you through every step of using Pebble through its complete operations lifecycle.
+These guides walk you through the operations lifecycle of Pebble.
 
 
 ## Installation
 
-Installation follows a broadly similar pattern on all architectures, and you can choose to install the pre-built binary or build it from the source by yourself.
+Installation follows a broadly similar pattern on all architectures. You can choose to install the pre-built binary or build it from the source by yourself.
 
 ```{toctree}
 :titlesonly:
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -30,7 +30,7 @@ CLI Commands <cli-commands/cli-commands>
 
 ## Pebble in containers
 
-When Pebble is configured as a client connected to a remote system (for example, a separate container), you can use subcommands on the client to manage the remote system.
+When the Pebble daemon is running inside a remote system (for example, a separate container), you can manage the remote system using subcommands on the Pebble client.
 
 ```{toctree}
 :titlesonly:
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -1,17 +1,101 @@
 # Reference
 
+These guides provide technical information about Pebble.
+
+
+## Layers
+
+Your service configuration is defined as a stack of "layers".
+
 ```{toctree}
 :titlesonly:
 :maxdepth: 1
 
 Layers <layers>
 Layer specification <layer-specification>
+```
+
+
+## Pebble commands
+
+The `pebble` command has several subcommands.
+
+```{toctree}
+:titlesonly:
+:maxdepth: 1
+
 CLI Commands <cli-commands/cli-commands>
+```
+
+
+## Pebble in containers
+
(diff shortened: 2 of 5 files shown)


docs/how-to/index.md
```diff
@@ -1,11 +1,11 @@
 # How-to guides
 
-These guides will walk you through every step of using Pebble through its complete operations lifecycle.
+These guides walk you through the operations lifecycle of Pebble.
```
benhoyt, line 3: Thanks -- I agree what we had wasn't great (it certainly isn't yet "complete" or "every step"). However, I don't really like the "operations lifecycle" jargon or really know what it means. What about something like:
```suggestion
These guides walk you through key operations you can perform with Pebble.
```

docs/how-to/index.md
```diff
@@ -1,11 +1,11 @@
...
 
 
 ## Installation
 
-Installation follows a broadly similar pattern on all architectures, and you can choose to install the pre-built binary or build it from the source by yourself.
+Installation follows a broadly similar pattern on all architectures. You can choose to install the pre-built binary or build it from the source by yourself.
```
benhoyt, line 8: While we're here, what about:
```suggestion
Installation follows a similar pattern on all architectures. You can choose to install the pre-built binary or build it from the source by yourself.
```

docs/reference/index.md
```diff
@@ -1,17 +1,101 @@
 # Reference
 
+These guides provide technical information about Pebble.
+
+
+## Layers
+
+Your service configuration is defined as a stack of "layers".
```
benhoyt, line 8: It's not just "service" configuration but also health checks and log targets. Also, "Your" feels a bit funny here. So how about:
```suggestion
Pebble configuration is defined as a stack of "layers".
```

docs/reference/index.md
```diff
@@ -1,17 +1,101 @@
...
+:titlesonly:
+:maxdepth: 1
+
+CLI Commands <cli-commands/cli-commands>
+```
+
+
+## Pebble in containers
+
+When Pebble is configured as a client connected to a remote system (e.g., a separate container), you can use subcommands on the client to manage the remote system.
```
benhoyt, line 33: We try to avoid (especially Latin) abbreviations like "e.g." and "i.e.":
```suggestion
When Pebble is configured as a client connected to a remote system (for example, a separate container), you can use subcommands on the client to manage the remote system.
```

benhoyt (changes requested):
Thanks, this is very good -- just requesting a few nit comments/changes.

In addition, I think we should discuss the ToC order. I think the "natural" or "logical" order is good in the main document, but (especially as things get longer) alphabetical order would be better for scanning/finding things in the ToC on the left. But let's discuss at/after daily.

docs/reference/index.md
benhoyt: See https://github.com/canonical/operator/blob/main/STYLE.md#spell-out-abbreviations

dwilding:
Thanks @benhoyt! I've included your suggestions in https://github.com/dwilding/canonical-pebble/commit/e90e04e

Let's discuss the TOC approach today

docs/reference/index.md
IronCore864: I find this sentence slightly inaccurate because Pebble can't connect to a container.

How about being more precise here, something like "When the Pebble daemon is running inside a remote system (for example, a separate container), the Pebble client subcommands can be used to manage the remote system."? Maybe you can come up with something better than this :)

docs/reference/index.md
```diff
@@ -1,17 +1,101 @@
...
+:titlesonly:
+:maxdepth: 1
+
+CLI Commands <cli-commands/cli-commands>
+```
+
+
+## Pebble in containers
+
+When Pebble is configured as a client connected to a remote system (for example, a separate container), you can use subcommands on the client to manage the remote system.
```
IronCore864, line 33: I find this sentence slightly inaccurate because Pebble can't connect to a container.

How about being more precise here, something like "When the Pebble daemon is running inside a remote system (for example, a separate container), the Pebble client subcommands can be used to manage the remote system."? Maybe you can come up with something better than this :)

IronCore864 (approved):

docs/reference/index.md
dwilding: Good point, thanks, I think this is a better approach. I've added https://github.com/canonical/pebble/pull/524/commits/7ab69e1e37c015c9f1c4018a7713b1b47bb17c0a with new wording

benhoyt (approved):
Great, thanks. Feel free to merge as is and figure out the ToC ordering in a separate PR, or include that in this one -- doesn't matter to me either way.

dwilding:
@benhoyt, @IronCore864 I've reorganized the links after our discussion. Changes in https://github.com/canonical/pebble/pull/524/commits/261e6da6b06671271c3321e3ab4fb231f70b5cdc:

- Links under Reference are now listed alphabetically in the nav sidebar
- On the Reference page, I've bumped a few sections to the top, for a more logical reading order. And I feel that grouping auto-restart & health checks together makes for a better explanation
- Minor adjustments to a couple of link titles:
    - Use Pebble in containers --> Pebble in containers
    - CLI Commands --> CLI commands
- Added comments in the source to explain how Reference is set up

Let me know if you think anything should be changed!

PR #526, opened 2024-11-27
docs: add snap as an installation method

dwilding:
The purpose of this PR is to list the `pebble` snap as an option in [How to install Pebble](https://canonical-pebble.readthedocs-hosted.com/en/latest/how-to/install-pebble/).

Changes:

- Added a new section called "Install the snap" after "Install the binary". I put the new section immediately after "Install the binary" because these two sections are more user focused, whereas "Install from source" is more developer focused
- Adjusted some language within the page for readability and consistency
- Changed the link titles in "Install from source" to make it more obvious what we're linking to

### DIFF
--- a/docs/how-to/install-pebble.md
+++ b/docs/how-to/install-pebble.md
@@ -7,6 +7,7 @@ To install the latest version of Pebble, choose any of the following methods:
 - {ref}`install_pebble_snap`
 - {ref}`install_pebble_from_source`
 
+
 (install_pebble_binary)=
 ## Install the binary
 
@@ -16,6 +17,7 @@ To install the binary for the latest version of Pebble:
    :start-after: Start: Install Pebble binary
    :end-before: End: Install Pebble binary
 ```
+
 
 (install_pebble_snap)=
 ## Install the snap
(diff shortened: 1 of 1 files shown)


IronCore864 (approved):

docs/how-to/install-pebble.md
```diff
@@ -16,15 +18,29 @@ To install the binary for the latest version of Pebble:
...
 ```
 
+
+(install_pebble_snap)=
+## Install the snap
+
+To install the latest version of Pebble from the Snap Store:
+
+```
+sudo snap install pebble --classic
```
tonyandrewmeyer, line 28: Do we recommend installing snaps with `sudo`? Without `sudo`, you'll get Ubuntu prompting for the privileges (assuming you're using Ubuntu), which feels nicer to me (both in terms of use and in terms of what we recommend people do). On the other hand, if you're using Pebble, you're probably very familiar with the command line so maybe you want to stay confined there.

You can also install the snap by searching for pebble in the Ubuntu software app :)

docs/how-to/install-pebble.md
```diff
@@ -16,15 +18,29 @@ To install the binary for the latest version of Pebble:
...
+
 (install_pebble_from_source)=
 ## Install from source
 
-Alternatively, you can install the latest version of Pebble from source:
+To install the latest version of Pebble from source:
 
-1. Follow the official Go documentation [here](https://go.dev/doc/install) to download and install Go.
-2. After installing, you will want to add the `$GOBIN` directory to your `$PATH` so you can use the installed tools. For more information, refer to the [official documentation](https://go.dev/doc/install/source#environment).
+1. Follow the [Go installation documentation](https://go.dev/doc/install) to download and install Go.
```
tonyandrewmeyer, line 39: :heart_eyes: for getting rid of a "here" link!

docs/how-to/install-pebble.md
benhoyt: At first I was going to say I agree that it would be better to drop the `sudo`. However, the official snap [tutorial](https://snapcraft.io/docs/get-started#install-the-snap) uses it -- I guess the thinking is they're already in the terminal, so why not keep them there? So maybe we should leave it with `sudo` for consistency.

PR #529, opened 2024-11-28
feat(daemon): add `ExternalMode` reboot

anpep:
This patch adds a new `daemon.ExternalMode` reboot mode that does nothing, so that code in charge of starting and stopping the Pebble daemon can run custom logic after the daemon has stopped.

When `ExternalMode` is set and the Pebble daemon stops after a system restart has been requested, `daemon.Stop()` returns the special `daemon.ErrRestartExternal` error that can be detected by the caller in order to execute custom reboot logic.

Furthermore, the fallback reboot handler is set to `syscallModeReboot()`, so that a system reboot is forced after `rebootWaitTimeout` is reached. In other words, the code responsible for handling the reboot with custom logic has a time span given by `rebootWaitTimeout` to perform a system reboot, or it will be forced.

```go
func exampleUsage() error {
    /* ... */
    daemon.SetRebootMode(daemon.ExternalMode)
    d, err := daemon.New(...)
    if err != nil {
        return err
    }
    /* ... */

    sigs := make(chan os.Signal, 1)
    signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
    select {
    case <-sigs:
        // Exiting due to signal
    case <-d.Dying():
        // Exiting due to daemon error/requested reboot
    }

    err := d.Stop()
    if errors.Is(err, daemon.ErrRestartExternal) {
        // Custom reboot logic (~10min window before reboot(2) is called)
        /* ... */
        return nil
    }
    return err
}
```

### DIFF
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -52,6 +52,7 @@ var (
 	ErrRestartSocket         = fmt.Errorf("daemon stop requested to wait for socket activation")
 	ErrRestartServiceFailure = fmt.Errorf("daemon stop requested due to service failure")
 	ErrRestartCheckFailure   = fmt.Errorf("daemon stop requested due to check failure")
+	ErrRestartExternal       = fmt.Errorf("daemon stop requested due to externally-handled reboot")
 
 	systemdSdNotify = systemd.SdNotify
 	sysGetuid       = sys.Getuid
@@ -508,7 +509,7 @@ func (d *Daemon) HandleRestart(t restart.RestartType) {
 	case restart.RestartSystem:
 		// try to schedule a fallback slow reboot already here,
 		// in case we get stuck shutting down
-		if err := rebootHandler(rebootWaitTimeout); err != nil {
+		if err := fallbackRebootHandler(rebootWaitTimeout); err != nil {
 			logger.Noticef("%s", err)
 		}
 		d.mu.Lock()
@@ -702,7 +703,10 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
 
 const rebootMsg = "reboot scheduled to update the system"
 
-var rebootHandler = systemdModeReboot
+var (
+	rebootHandler         = systemdModeReboot
+	fallbackRebootHandler = systemdModeReboot
+)
 
 type RebootMode int
 
@@ -711,6 +715,8 @@ const (
 	SystemdMode RebootMode = iota + 1
 	// Reboot uses direct kernel syscalls
 	SyscallMode
+	// Reboot is handled externally after the daemon stops
+	ExternalMode
 )
 
 // SetRebootMode configures how the system issues a reboot. The default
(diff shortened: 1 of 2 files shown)


internals/daemon/daemon.go
```diff
@@ -702,7 +703,10 @@ func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) err
 
 const rebootMsg = "reboot scheduled to update the system"
 
-var rebootHandler = systemdModeReboot
+var (
+	rebootHandler         = systemdModeReboot
+	fallbackRebootHandler = systemdModeReboot
```
benhoyt, line 708: I'm a bit unclear on when `rebootHandler` is used and when `fallbackRebootHandler` is used -- can you clarify?

benhoyt (commented):
This looks reasonable to me, thanks. (Though I had one clarifying question.)

The only other thing I'd suggest is to use `errors.Is` instead of `err == daemon.ErrRestartExternal` in your PR description (though I see you've used it in the test code).

internals/daemon/daemon.go
anpep: I hope my understanding on this is correct: up until now, Pebble was using `rebootHandler` both for requesting systemd for a system reboot (via `shutdown`) *and* scheduling a *fallback* reboot to happen after `rebootWaitTimeout` (~10min, also via `shutdown`). I think the rationale behind this is, if Pebble fails to immediately reboot, the system wouldn't be stuck running forever, and a reboot is forced after that time window.

When introducing an external reboot mode that essentially does nothing actively to reboot the machine, `rebootHandler` is no longer useful for scheduling this fallback reboot, so what I'm doing is having two reboot handlers now, which for systemd/syscall reboot are one and the same. For external reboot mode though, the `rebootHandler` is set to this "return a special error and do nothing" operation, while the `fallbackRebootHandler` is set to `syscallModeReboot`. If the caller fails to detect `ErrRestartExternal` and reboot the machine, Pebble will do it after `rebootWaitTimeout` has elapsed.

internals/daemon/daemon.go
```diff
@@ -720,8 +726,13 @@ func SetRebootMode(mode RebootMode) {
 	switch mode {
 	case SystemdMode:
 		rebootHandler = systemdModeReboot
+		fallbackRebootHandler = systemdModeReboot
```
flotter, line 729: @anpep I think this proposal is solid. Just one other option which may result in slightly less code changes:

Instead of added ```fallbackRebootHandler```, you could record the ```rebootMode```. In this model, you only ever have one ```rebootHandler```, and then for both ```SyscallMode``` and ```ExternalMode``` the handler is set to the syscall handler. 

Then the only other thing you do is in ```doReboot```, just add an early return with your special error code:

```go
func (d *Daemon) doReboot(sigCh chan<- os.Signal, waitTimeout time.Duration) error {
    if rebootMode == ExternalMode {
        return ErrRestartExternal
    }
:
}
```

flotter (approved):
Looks good for me now, thanks @anpep

benhoyt (approved):
Thanks!

PR #532, opened 2024-12-09
docs: generate API reference doc from an OpenAPI spec

IronCore864:
This PR demonstrates the usage of Swagger javascript/CSS to render OpenAPI 3.1.0 specs directly into RTD HTML.

See more investigation on this [here](https://gist.github.com/IronCore864/f2b957033b6bf6ec7ca2b4e308a0f1e9). However, after trying them out, we figured out there are some issues:

- With the sphinx OpenAPI extension + Wwagger 2.0 spec, examples can't be rendered.
- With the sphinx OpenAPI extension + OpenAPI 3.0 spec, the response body can't be rendered.

So, we decided to give up the extension but use Swagger javascript/CSS to render the result directly, which is similar to what the LXD and Anbox cloud docs do.

Preview: https://canonical-pebble--532.com.readthedocs.build/en/532/reference/api/.

Included APIs in the PoC:

- `/v1/services`, both GET and POST methods.
- `/v1/changes/id/wait`, GET.


docs/reference/index.md
```diff
@@ -21,6 +21,7 @@ Log forwarding <log-forwarding>
 Notices <notices>
 Pebble in containers <pebble-in-containers>
 Service auto-restart <service-auto-restart>
+API <api>
```
benhoyt, line 24: Let's add this to the top of the list to keep things in alphabetical order.

docs/reference/index.md
```diff
@@ -86,3 +87,7 @@ Pebble can send service logs to a Loki server.
 Pebble records events as "notices". In addition to the built-in notices, clients can report custom notices.
 
 * [Notices](notices)
+
+## API
```
benhoyt, line 91: Maybe this heading can be "Accessing the API", and a sentence about the API (like the other sections have).

docs/tox.ini
```diff
@@ -18,19 +18,22 @@ commands =
...
 description = Build the Sphinx docs
+allowlist_externals = rm
 deps = pip-tools
 commands_pre =
     pip-sync {toxinidir}/requirements.txt
 commands =
-    sphinx-build -W --keep-going . _build/html
+    rm -rf __pycache__ _build/
+    sphinx-build -W --keep-going -b dirhtml . _build/html
+    python -m http.server -b 127.0.0.1 -d _build/html/
```
benhoyt, line 28: I think this block should just stay the "build" part, and the serving part should be in the "serve" Tox command. Or is this just for debugging?

docs/reference/api.md
```diff
@@ -0,0 +1,29 @@
+# API
+
+<link rel="stylesheet" type="text/css" href="https://unpkg.com/swagger-ui-dist@5.11.0/swagger-ui.css" ></link>
```
benhoyt, line 3: This is probably fine for now. But just for reference, the LXD docs import this from the repo into `_static` at build time: https://github.com/canonical/lxd/blob/510c0117cb20591f92126e0b8d3bb583118cd93e/doc/custom_conf.py#L291

benhoyt (commented):
This is a great start, thanks. I haven't done a detailed review of the text (can do that later, but want to get the structure and overall things right first). A few things I think could be improved in the rendered output:

- The heading looks ugly to me, with the "v1" and "OAS 3.1" (they don't need to know what we built it in) and the ugly "/en/.../openapi.yaml" link. Can we remove those? In fact, the "Pebble API" is kind of unnecessary too, given the page header.
- Is this going to be able to make requests from the browser? I think it's not, given that the Pebble API is (mostly) Unix sockets. So should we disable the "Servers" list and the input text fields? (Though it looks like the input text boxes are in disabled state anyway.)
- Where are you getting the text from, or is that new?
  - The short text looks fine: "Wait for a change to complete". 
  - Then I think we need some longer text once you expand that. Could be copied from the Go client API doc comments, like the doc comment on `Client.WaitChange`? At least as a good start.
  - The descriptions of the fields. Could potentially copy these from Go client doc comments as a good start too? (Though Go-specific terms like "zero value" will have to be changed). Note that we don't need to say "optional" as required fields are clearly marked.
  - For things like durations, these might be used in various places. Can we document the duration format once and link to a type to describe the "300ms" type format once?
- What's the Links / "No links" column for? Can we remove it?
- What's the "Media type" and "Examples" dropdown for? Can we remove them?
- Would be nice to have `curl` examples, at least one (for GET and one for POST) at the top.
- We don't need 500 errors specified for each endpoint, as they're always "internal server error". But I think the other errors are good to list (though probably don't need a separate example for each. Maybe we can have error examples in the top section once, and just the error codes/meanings under each endpoint.
- It's not clear what the request body should look like for the POST (until you click Examples > Schema?). Ideally this can be clearer.

Let's discuss these points and any feedback David has in a meeting with him, either this afternoon or next week (Tuesday or after).

docs/.sphinx/_static/swagger-override.css
```diff
@@ -0,0 +1,98 @@
...
+.swagger-ui .errors-wrapper hgroup h4 {
+    font-family: "Ubuntu", -apple-system, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+}
+
+.swagger-ui .response-col_description,
+.swagger-ui .parameters-col_description {
+    width: 85%
+}
+
+.swagger-ui .information-container .url {
```
benhoyt, line 72: Let's put a comment on what each of these blocks is doing, for example `/* Hide the unnecessary foobar link */`.

docs/tox.ini
benhoyt: When I check out this code locally, run `tox -e docs`, then make changes to `openapi.yaml`, refreshing the web page doesn't update. I have to Ctrl-C to quit and then restart and then do a hard refresh in the browser. Is this expected? Can we make this a bit smoother for local development?

docs/tox.ini
```diff
@@ -18,19 +18,22 @@ commands =
...
 
 [testenv:serve]
 description = Live reload and serve the Sphinx docs
 deps = pip-tools
 commands_pre =
     pip-sync {toxinidir}/requirements.txt
 commands =
-    sphinx-autobuild . _build/html
+    sphinx-autobuild -b dirhtml . _build/html
```
benhoyt, line 36: If we're incorporating "serve" into `tox -e docs`, can we get rid of `tox -e serve`?

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
+# API
+
+Pebble exposes an API to allow remote clients to interact with the daemon. The API can start and stop services, add configuration layers to the plan, and so on.
+
+## Accessing the API
+
+The API uses HTTP over a Unix socket, with access to the API controlled by user ID. If `pebble run` is started with the `--http <address>` option, Pebble exposes a limited set of open-access API endpoints using the given TCP address. See {ref}`api-access-levels`.
+
+See below for some examples of how to use the API. For more examples, see "How to use the Pebble API". <!-- [David] Link to the how-to guide -->
```
benhoyt, line 9: I think we should drop this paragraph (below is also the API reference), and just add a link to the how-to guide under examples when that's ready.

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
...
+
+There's a Go client for the API. See the [Go client documentation](https://pkg.go.dev/github.com/canonical/pebble/client) and the examples below.
+
+We try to never change the underlying HTTP API in a backwards-incompatible way. However, in rare cases we may change the Go client in a backwards-incompatible way.
```
benhoyt, line 13: This paragraph is kind of stuck in the middle of mentioning the Go client and the Python client. How about we move it up.

In fact, instead of having these paragraphs here, why don't we make the structure like this:

```
# Accessing the API

## Using curl 

Introductory paragraph and a couple of examples (one GET, one POST).

## Using the Go client

Introductory paragraph and one example (one is probably enough here, just the one from https://pkg.go.dev/github.com/canonical/pebble/client).

## Using the Python client

Paragraph mentioning and linking to the docs (can we link to https://ops.readthedocs.io/en/latest/pebble.html to be more specific?). Maybe a 3-line example like:

     import ops
     client = ops.pebble.Client('/path/to/.pebble.socket')
     client.start_services(['svc1', 'svc2'])
```

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Go client examples
+
+TODO: go client code snippets.
+
+### cURL examples
+
+TODO: curl examples.
+
+## API access levels
```
benhoyt, line 25: For now, can we just drop this section (including "Identities") and have this content at `/explanation/api-and-clients`. I think we should merge the two, but not needed in this PR.

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
...
+
+API endpoints fall into one of three access levels, from least restricted to most restricted. See {ref}`api-access-levels`.
+
+### Identities
+
+Pebble has the concept of named "identities", which lets administrators manage users' access to the API. See {ref}`controlling-api-access-using-identities`.
+
+## Common parameters
```
benhoyt, line 33: How about we call this section "Data formats" or similar? Then have an introductory paragraph like:

> Some API parameters use specific formats as defined below.

We'll also want to describe the time format used in some fields (like `current-since`). RFC3339 with an example.

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
...
+
+### timeout
```
benhoyt, line 35: The parameter field name is "timeout", but what I'd like to document here is the "duration" format. So let's change this to "duration", and then in the API endpoint we say "`timeout` is a [duration](link)".

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
...
+
+The format of the timeout is a decimal number with an optional unit suffix (e.g., "300ms", "1.5s", "2h45m").
```
benhoyt, line 37: Can we basically copy this from the Go docs (to mention that it can be a sequence of number+unit.

> A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
...
+
+Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+
+## Common errors
```
benhoyt, line 41: Looking over the rendered docs again, I think we should actually move all error codes to this section, and move them all out of each endpoint. They don't add value there.

So I think this section should be just "Errors" and have an introductory paragraph like so:

> API endpoints may return errors as 4xx status codes (client errors) and 5xx status codes (server errors). Errors are always returned in JSON format with the following structure:
>
>     { ... }

And then just a bulleted list or table of error codes:

* 400: bad request, for example, a badly-formatted parameter
* 404: not found, for example, a change with the specified ID does not exist
* ...
* 500: internal server error, for example, the Pebble database is corrupt; if these errors keep happening, please [open an issue](link)

docs/reference/api.md
```diff
@@ -0,0 +1,77 @@
+# API
+
+Pebble exposes an API to allow remote clients to interact with the daemon. The API can start and stop services, add configuration layers to the plan, and so on.
+
+## Accessing the API
```
benhoyt, line 5: On reflection, maybe we can just drop this section heading -- this whole doc is a reference for "accessing the API".

However, we do need a section something like "structure of the API" where we mention the different parts of the API in general terms: URL with path or query params, request, response, JSON format, request types (sync, async).

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+  title: Pebble API
+  version: v1
+paths:
+  /v1/changes/{id}/wait:
+    get:
+      summary: Wait for a change to complete
+      description: |
+        Wait for the change to be finished.
+        
+        If the wait operation succeeds, the returned Change.Err string will be non-empty if the change itself had an error.
```
benhoyt, line 12: `Change.Err` is terminology from the Go client. This needs to say something like: "the result will have the `err` field set to an appropriate error message".

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+          description: The ID of the change to wait for.
+        - in: query
+          name: timeout
+          schema:
+            type: string
+          description: |
+            Optional timeout duration.
+            If specified, wait till change is ready or timeout occurs, whichever is first.
+            Defaults to zero if not specified, in which case the request will wait indefinitely until the change is ready.
+            Read more on the format of the timeout [here](#timeout).
```
benhoyt, line 30: Slight tweaks:

```suggestion
            Optional timeout (a [duration](link)).
            If specified, wait till the change is ready or a timeout occurs, whichever is first.
            If not specified or zero, wait indefinitely until the change is ready.
```

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+                $ref: "#/components/schemas/ErrorResponse"
+        '504':
+          description: Gateway Timeout - timeout waiting for change.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+  /v1/services:
+    get:
+      summary: Get services
```
benhoyt, line 95:
```suggestion
      summary: List services
```

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+      description: Fetch information about specific services (or all of them), ordered by service name.
```
benhoyt, line 96:
```suggestion
      description: Fetch information about specific services (or all of them). Return results ordered by service name.
```

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+      tags:
+        - services
+      parameters:
+        - in: query
+          name: names
+          description: The names of the services to get, a comma-separated string. If empty, get all services.
```
benhoyt, line 102: I believe this supports both comma-separated and multiple parameters. Is there a way to specify "this parameter can occur multiple times" in OAS? If so, let's use that. In any case, let's not mention the comma-separated version as I think it's better to let HTTP do that, so something like:

```suggestion
          description: The names of the services to get. To get multiple services, specify this parameter multiple times. If not set, get all services.
```

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+          schema:
+            type: string
+          examples:
```
benhoyt, line 105: I don't see where these examples show up. Let's just drop them.

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                action:
+                  type: string
+                  description: The action to perform. One of "start", "stop", "restart", "replan", "autostart".
```
benhoyt, line 147: No need to repeat the possible values in the description -- let's just keep these in the enum.

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+                  enum: ["start", "stop", "restart", "replan", "autostart"]
```
benhoyt, line 148: How about we standardise on ordering alphabetically for these enum values?

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/PostServices"
+              examples:
+                accepted:
+                  value:
+                    type: async
+                    status-code: 202
+                    status: Accepted
+                    change: 1
```
benhoyt, line 188: This is always a string, so:
```suggestion
                    change: "1"
```

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+          $ref: "#/components/schemas/errorResult"
+    errorResult:
+      type: object
+      properties:
+        message:
+          type: string
+          description: Error message.
+        kind:
+          type: string
+          description: Error kind, possible values are "login-required", "no-default-services", "not-found", "permission-denied", "generic-file-error", "system-restart", and "daemon-restart".
```
benhoyt, line 220: Ah yes, if we're moving the error description to their own section (not in the spec), let's list these possible kinds there.

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+    GetServices:
```
benhoyt, line 221: Because this is only referenced once, can this just be inline in the /v1/services part?

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+      type: object
+      properties:
+        type:
+          type: string
+          description: Response type, one of "sync", "async", "error".
+          enum: ["sync", "async", "error"]
```
benhoyt, line 227: For `GET /v1/services`, this is always "sync".

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+        status-code:
+          type: integer
+          description: HTTP response status code.
+        status:
+          type: string
+          description: The text for the HTTP status code, see https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml.
+        result:
```
benhoyt, line 234: The fields above `result` will be present in every response -- is there a way to define those fields only in one place?

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+          $ref: "#/components/schemas/serviceInfo"
```
benhoyt, line 235: I think this is incorrect, and should be a list of serviceInfo -- how do we specify that?

docs/specs/openapi.yaml
```diff
@@ -0,0 +1,335 @@
...
+    serviceInfo:
+      type: object
+      properties:
+        name:
+          type: string
```
benhoyt, line 240: This is the actually-interesting part of the response, so we definitely want (short) descriptions on each of these fields, eg:

* name: Name of the service.
* startup: Configured `startup` value (enum: enabled, disabled).
* current: Current status of the service (enum: active, backup, error, inactive).
* current-since: Time the service transitioned to the current status.

benhoyt (commented):
This is a significant improvement -- thanks. Overall I really don't love Swagger UI, but it seems "just good enough" for now.

I've left a bunch of comments. I think the introduction part needs some significant improvement, and the details of the OpenAPI spec need some work too.

It think it'd be good to focus on getting this PR very polished, to set the style and tone for future PRs, which will then just be basically rote work (except maybe the file API, which is multipart).

benhoyt:
I don't think we should switch to fully manual, but just for reference/comparison, the snapd folks use manual markdown to document their API: https://snapcraft.io/docs/snapd-api

benhoyt:
I updated the title (removed PoC) now that we've agreed on this basic approach (Swagger UI with YAML spec).

docs/reference/api.md
IronCore864: OK.

FYI, I removed the "possibly signed" part because it doesn't make sense for the timeout (or any parameter that is a duration in Pebble) to be negative. I hope this makes sense.

docs/specs/openapi.yaml
IronCore864: This is because of one detailed CSS customisation that we shouldn't do, I will revert it and keep the examples, otherwise, a random example will be generated by swagger automatically according to the schema, which makes less sense. However, I will reduce the example to just one.

docs/specs/openapi.yaml
IronCore864: After some investigation I have figured out a way to create a base response component and inherit from it.

IronCore864:
On the "updating openapi.yaml then refreshing the web page doesn't update" issue, I have done some research and testing and I don't think we can fix it, this is because we used a [symlink](https://github.com/canonical/pebble/pull/532/files#diff-00e5793290dc4e546926ee914f88f8fa15cf2ba99a0a8d7b4bdc9754836c4682R1) for the [html_extra_path](https://github.com/canonical/pebble/pull/532/files#diff-52f0fa43d5e9c03425bf8ce918e72d04ccda47904902a7a3a8a3774a33464f86R195) configuration (this method was copied from the LXD doc so that the real `openapi.yaml` file can be put at a more logical place), and I think the files for `html_extra_path` are [copied during build time](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path). I have tested it with `sphinx-autobuild` (the previous `tox -e serve` command) and it doesn't refresh the web page either.

I think the only solution is to watch the file, if it's changed, copy it to the build output directory, something like:

```
[testenv:docs]
description = Build the Sphinx docs
allowlist_externals =
    rm
    bash
deps = pip-tools
commands_pre =
    pip-sync {toxinidir}/requirements.txt
commands =
    rm -rf __pycache__ _build/
    sphinx-build -W --keep-going -b dirhtml . _build/html
    bash -c 'fswatch specs/openapi.yaml | xargs -n1 -I{} cp specs/openapi.yaml _build/html/ &'
    python -m http.server -b 127.0.0.1 -d _build/html/
```

It seems like a hack and I'm not sure if we could do this.

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Structure of the API
+
+Here is a breakdown of the general structure of most API endpoints in Pebble.
+
+### API endpoint URL
+
+Pebble API uses standard HTTP URLs to get actions and resources. The URLs are made up of a base URL followed by a path and optional query parameters.
+
+- Base URL: The consistent foundation of all API endpoints (for example, `http://localhost:4000/v1/`).
```
dwilding, line 15: I think we can leave out the description "The consistent foundation of all API endpoints" and directly say what the base URL is likely to be in Pebble, especially because most endpoints use the socket. How about something like this:

```suggestion
- Base URL: If you're accessing the API over the Unix socket, the base URL is not applicable. If you're accessing the API over TCP, the base URL is `http://localhost:4000` by default (you can configure the TCP address using `pebble run --http`.
```

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+- Path: Specifies the resource or collection being accessed (for example, `/changes`, `/services`). Paths are hierarchical and use forward slashes as separators.
```
dwilding, line 16: I think we should always include `/v1` in paths, since our concept of base URL is a bit different from normal

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+- Query parameters: Optional key-value pairs appended to the URL after a question mark (?) and separated by ampersands (&). They provide additional filtering or control over the request (for example, `/services?name=svc1`).
+
+Example:
+
+```bash
+http://localhost:4000/v1/services?name=svc1
+```
```
dwilding, line 23: I feel we can leave this out. It only works over the socket anyway (if I understood correctly). I'm happy to leave out full example URLs because we have the examples sections elsewhere

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Request
+
+Key components of a request include:
+
+- HTTP Method: The desired action (for example, `GET` for retrieving data, `POST` for creating data, `PUT` for updating data, and `DELETE` for deleting data).
```
dwilding, line 29: We could drop PUT and DELETE

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+- Headers: Metadata about the request, such as content type, basic auth, and accepted response formats (for example, `Authorization: Basic <credentials>`, `Content-Type: application/json`).
```
dwilding, line 30: I guess `Authorization` will be needed in the future :smile:, but I think for now we don't need to mention it, since we don't support that header

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+- Body: For requests that send data to the server (`POST`, `PUT`), the body contains the data in JSON format.
```
dwilding, line 31:
```suggestion
- Body: For requests that send data to the server using `POST`, the body contains the data in JSON format.
```

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Response
+
+After processing a request, Pebble sends back a response. Key components of a response are:
+
+- HTTP status code: the result of the request.
+- Headers: metadata about the response, such as content type and caching information.
```
dwilding, line 38: Does Pebble include caching info?

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+- Body: data returned by Pebble server in JSON format.
+
+### JSON format
+
+Pebble API uses JSON for data exchange. Both request and response bodies are formatted as JSON objects or arrays.
```
dwilding, line 43: Very minor nit, to remove risk of misinterpretation

```suggestion
Pebble API uses JSON for data exchange. Request and response bodies are formatted as JSON objects or arrays.
```

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+
+Example JSON Response:
```
dwilding, line 45: Shall we say which endpoint this response comes from?

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+
+The API uses HTTP over a Unix socket, with access to the API controlled by user ID. If `pebble run` is started with the `--http <address>` option, Pebble exposes a limited set of open-access endpoints (see {ref}`api-access-levels`) using the given TCP address.
+
+## Structure of the API
+
+Here is a breakdown of the general structure of most API endpoints in Pebble.
+
+### API endpoint URL
+
+Pebble API uses standard HTTP URLs to get actions and resources. The URLs are made up of a base URL followed by a path and optional query parameters.
```
dwilding, line 13: I think we can standardise on "The API" throughout the page, instead of "Pebble API"

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+### Request types
+
+Pebble API supports both synchronous and asynchronous requests.
+
+- Synchronous requests: The client sends a request and waits for the server to return a response before continuing.
+- Asynchronous requests: For long-running operations, Pebble API offers asynchronous processing. The client sends a request and gets an acknowledgment immediately, but the actual processing is still running on the server side. The client can poll for the result. Specific endpoints supporting asynchronous operations are documented below with `type` as `async`.
+
+## Using cURL
```
dwilding, line 67: I think "cURL" is styled as "curl" now, although I also thought it was supposed to be "cURL" at first...

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+client.start_services(['svc1', 'svc2'])
+```
+
+For more information, see:
+
+- [Ops library documentation for the Python client](https://ops.readthedocs.io/en/latest/reference/pebble.html#ops.pebble.Client)
+- [Source code of the Python client](https://github.com/canonical/operator/blob/main/ops/pebble.py)
+- [Pebble in the context of Juju charms](https://juju.is/docs/sdk/interact-with-pebble)
+
+For more examples, see "How to use the Pebble API". <!-- [David] Link to the how-to guide -->
```
dwilding, line 130: Will we have Python examples in the how-to doc? Maybe we could put this link at the end of the curl section and the Go section instead

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+  },
+}
+```
+
+Possible values for `status-code`:
+
+- 400: Bad request. For example, a badly-formatted parameter, like when starting a service, the service name is not provided.
+- 401: Unauthorized.
+- 404: Not found. For example, a change with the specified ID does not exist.
+- 500: Internal server error, for example, the Pebble database is corrupt. If these errors keep happening, please [open an issue](https://github.com/canonical/pebble/issues/new).
```
dwilding, line 174: A couple of suggestions for structuring each list item the same way:

```suggestion
- 400: Bad request. For example, if a parameter is missing or badly-formatted, such as trying to start a service without providing a the service name.
- 401: Unauthorized.
- 404: Not found. For example, if a change with the specified ID does not exist.
- 500: Internal server error. For example, if the Pebble database is corrupt. If these errors keep happening, please [open an issue](https://github.com/canonical/pebble/issues/new).
```

dwilding:
Thanks @IronCore864! I added some specific comments on api.md - hopefully they are helpful.

A couple more general thoughts:

- How about moving the "Using..." sections to the top of the page? I feel that if someone is familiar with HTTP APIs, they probably want to skip to those sections
- Is it possible to use CSS to entirely hide the "Pebble API (v1)" heading coming from Swagger UI? I like your in-page "API endpoints" heading better

docs/reference/api.md
benhoyt: Agreed on the text change. However, "the base URL" *path* part is applicable, just not the scheme and hostname. Perhaps "the scheme, hostname, and port are not applicable; only the `/v1/` path is important."

docs/reference/api.md
benhoyt: Agreed. I think we can just change this to something like:

> Headers: the only header necessary for Pebble requests the content type. This should be `Content-Type: application/json`.

docs/reference/api.md
benhoyt: No, again, the only header Pebble sends back is `Content-Type: application/json`.

benhoyt (approved):
I agree with David's comments and have added my own. I think it's overall quite close -- my major concern is that we make the "Structure of the API" section more Pebble-specific.

One other thing: it seems the input boxes in the GET endpoints have re-appeared -- is that expected? Ideally we can get rid of them, as they're just noise (and they're disabled anyway). I think they were gone in an earlier revision.

Even though I think this PR still needs work, I'm approving in advance to avoid blocking future work on top of it.

docs/reference/api.md
IronCore864: For completeness, let's include Python examples in the how-to doc.

Here I didn't mean "for more Python examples (since this is the end of the Python section), see ...", but rather:

```
## Accessing the API
### curl
### go
### python
### more examples - link to how-to
```

But since we removed the `Accessing the API` subsection because of previous comments, now the link is in an awkward position. If we put it under curl, it means "for more curl examples".

How about:

```
## Accessing Pebble API
Here are some methods to access the API. For more examples, see ...
### curl
### go
### python
```

docs/reference/api.md
benhoyt: I think David meant use just "The API", whereas I believe you've changed it to "Pebble API".

docs/reference/api.md
```diff
@@ -0,0 +1,214 @@
...
+
+    ```
+    curl ...
+    ```
+
+## Using the Go client
+
+To use the Go client to access API endpoints over the Unix socket, first create a client using `New`, and then call the methods on the returned Client to interact with the API. Example:
+
+```go
```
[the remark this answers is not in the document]
benhoyt: This is much nicer. However, let's just un-indent it to the top level to avoid the extra indentation.

benhoyt (approved):
This is really good now, thanks! We'll discuss any remaining concerns on the call shortly.

dwilding (approved):

IronCore864:
Latest changes according to discussion:

- Remove the input boxes in the parameters.
- Put curl/go/python as subsections under the section "Using the API".

PR #536, opened 2024-12-16
docs: add an overview of security

dwilding:
This PR adds a new page to summarize security considerations for Pebble. Preview: https://canonical-pebble--536.com.readthedocs.build/en/536/explanation/security/

### DIFF
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -12,7 +12,7 @@ For more information, see [](api-and-clients.md) and [](../how-to/manage-identit
 
 ## The Pebble directory
 
-Pebble stores its configuration, internal state, and Unix socket in the `$PEBBLE` directory. If `$PEBBLE` is not set, Pebble uses the directory `/var/lib/pebble/default`.
+Pebble stores its configuration, internal state, and Unix socket in the directory specified by the `PEBBLE` environment variable. If `$PEBBLE` is not set, Pebble uses the directory `/var/lib/pebble/default`.
 
 The `$PEBBLE` directory must be readable and writable by the UID of the pebble process. Make sure that no other UIDs can read or write to the $PEBBLE directory.
 
@@ -21,4 +21,4 @@ The file `$PEBBLE/.pebble.state` contains the internal state of the Pebble daemo
 
 ## Security updates
 
-There are several ways to install Pebble. The easiest way to ensure that you get security updates is to install the snap. See [](#install_pebble_snap).
+There are several ways to install Pebble. The easiest way to ensure that you get security updates is to [install the snap](#install_pebble_snap).
--- a/docs/explanation/security.md
+++ b/docs/explanation/security.md
@@ -1,0 +1,24 @@
+# Security
+
+
+## Access to the API
+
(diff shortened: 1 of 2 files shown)


docs/explanation/security.md
```diff
@@ -0,0 +1,24 @@
...
+The Pebble daemon exposes an API that enables remote clients to interact with the daemon. The API uses HTTP over a Unix socket, with access to the API controlled by user ID (UID). If you want to grant a specific access level to a user, you can define an "identity" for the user.
+
+If you use the `--http` option when starting the daemon, Pebble exposes a limited set of open-access API endpoints over TCP. No authentication is required to connect to the open-access endpoints.
+
+For more information, see [](api-and-clients.md) and [](../how-to/manage-identities.md).
+
+
+## The Pebble directory
+
+Pebble stores its configuration, internal state, and Unix socket in the `$PEBBLE` directory. If `$PEBBLE` is not set, Pebble uses the directory `/var/lib/pebble/default`.
```
benhoyt, line 15: What about making it slightly more explicit this is an environment variable:

```suggestion
Pebble stores its configuration, internal state, and Unix socket in the directory specified by the `PEBBLE` environment variable. If `$PEBBLE` is not set, Pebble uses the directory `/var/lib/pebble/default`.
```

docs/explanation/security.md
```diff
@@ -0,0 +1,24 @@
...
+
+The `$PEBBLE` directory must be readable and writable by the UID of the pebble process. Make sure that no other UIDs can read or write to the $PEBBLE directory.
+
+The file `$PEBBLE/.pebble.state` contains the internal state of the Pebble daemon. You shouldn't try to edit this file or change its permissions.
+
+
+## Security updates
+
+There are several ways to install Pebble. The easiest way to ensure that you get security updates is to install the snap. See [](#install_pebble_snap).
```
benhoyt, line 24: Can we just link "install the snap" directly?

```suggestion
There are several ways to install Pebble. The easiest way to ensure that you get security updates is to [install the snap](LINK).
```

benhoyt (approved):
This is great, thanks. A couple of nit comments.

docs/explanation/security.md
dwilding: We can, and I think I did at one point in the editing process :smile:

IronCore864 (approved):

docs/explanation/security.md
```diff
@@ -0,0 +1,24 @@
+# Security
+
+
+## Access to the API
```
IronCore864, line 4: Just to be sure: Do we want to move this section to the beginning of the API spec, or at least have a link there pointing to this?

dwilding: My idea is that we'll keep the "Access to the API" section in the new Security doc, as an entry-point for people looking for security info. We should be able to replace [API and clients](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/api-and-clients/) by content in the API reference, so when the API reference is ready we can change the link at the end of "Access to the API" to point to the API reference instead of [API and clients](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/api-and-clients/).

Does it seem reasonable?

PR #537, opened 2024-12-16
docs: clarify the replan behaviour with regards to stopped services

tonyandrewmeyer:
A `replan` command will start any services that have been stopped (with a `stop` command), if the service has `startup: enabled` set. Clarify that in the documentation.

### DIFF
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -721,7 +721,7 @@ current plan.
 When you update service configuration (by adding a layer), the services changed won't be automatically restarted. `pebble replan ` restarts them and brings the service state in sync with the new configuration.
 
 - The "replan" operation restarts `startup: enabled` services whose configuration have changed between when they started and now; if the configuration hasn't changed, replan does nothing.
-- Replan also starts `startup: enabled` services that have not yet been started.
+- Replan also starts `startup: enabled` services that have not yet been started, or that have been manually stopped.
 
 ### Examples
 
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -720,7 +720,12 @@ current plan.
 
 When you update service configuration (by adding a layer), the services changed won't be automatically restarted. `pebble replan ` restarts them and brings the service state in sync with the new configuration.
 
-- The "replan" operation restarts `startup: enabled` services whose configuration have changed between when they started and now; if the configuration hasn't changed, replan does nothing.
+For `startup: enabled` services that are running:
+
+- If the service hasn't changed configuration since it started, replan does nothing to the service.
+- If the service has changed configuration since it started, replan restarts the service.
+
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Thx!

docs/reference/cli-commands.md
```diff
@@ -721,7 +721,7 @@ current plan.
 When you update service configuration (by adding a layer), the services changed won't be automatically restarted. `pebble replan ` restarts them and brings the service state in sync with the new configuration.
 
 - The "replan" operation restarts `startup: enabled` services whose configuration have changed between when they started and now; if the configuration hasn't changed, replan does nothing.
```
dwilding, line 723: Your addition looks good. I've got a suggestion which I think would make the bullets easier to read:

```suggestion
The "replan" operation:

- Restarts `startup: enabled` services that have changed configuration since they started. If the configuration of a service hasn't changed, replan does nothing.

- Starts `startup: enabled` services that have not yet been started, or that have been manually stopped.
```

dwilding (approved):

docs/reference/cli-commands.md
tonyandrewmeyer: I do like this more, except that the "if the configuration of a service hasn't changed, replan does nothing" seems stronger without the "also" that's there now.

What do you (and @benhoyt) think about dropping that sentence? It's not true: the very next bullet counters it - and I think there's an implied "to that service" at the end, but it's not great to be implied rather than explicit.

I think the point is to reassure people that you can replan really often and it won't cause unnecessary restarts. Maybe it would be better to lead with that, and then list the cases where something does happen?

dwilding: How about something like this?

```suggestion
For `startup: enabled` services that are running:

- If the service hasn't changed configuration since it started, replan does nothing to the service.
- If the service has changed configuration since it started, replan restarts the service.

Replan also starts any `startup: enabled` services that have not yet been started, or that have been manually stopped.
```

docs/reference/cli-commands.md
```diff
@@ -720,8 +720,13 @@ current plan.
...
 
-- The "replan" operation restarts `startup: enabled` services whose configuration have changed between when they started and now; if the configuration hasn't changed, replan does nothing.
-- Replan also starts `startup: enabled` services that have not yet been started.
+For `startup: enabled` services that are running:
+
+- If the service hasn't changed configuration since it started, replan does nothing to the service.
+- If the service has changed configuration since it started, replan restarts the service.
+
+Replan also starts any `startup: enabled` services that have not yet been started, or that have been manually stopped.
+- Replan also starts `startup: enabled` services that have not yet been started, or that have been manually stopped.
```
tonyandrewmeyer, line 729:
```suggestion
Replan also starts any `startup: enabled` services that have not yet been started, or that have been manually stopped.
```

benhoyt (approved):

PR #542, opened 2024-12-23
docs: add API reference sections to the TOC

dwilding:
This PR adds JavaScript code to create TOC entries for the sections (tags) in the API reference.

**[Preview build](https://canonical-pebble--542.com.readthedocs.build/en/542/reference/api/)**

**The tags need to be hardcoded in _api.md_.** It would be possible for JavaScript to determine the tags from the rendered Swagger UI elements, but there's no event to tell us when the Swagger UI has rendered. So I think it's not worth the effort - simpler to hard code the tags.

**Behavior**
- Adds a link for each tag inside the last top-level entry in the TOC.
- When a tag link is clicked (with no modifier keys), the tag section scrolls into view. The tag section opens if it's not open.

**Why do this in JavaScript?** We already run some JavaScript, because Swagger UI requires JavaScript. And even if we inserted the tag links at build-time in Sphinx, we'd still need JavaScript to automatically open the tag sections.

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -175,5 +175,47 @@ window.onload = function() {
   // End Swagger UI call region
 
   window.ui = ui
+
+  function addSwaggerTagsToTOC(tags) {
+    // Find the last H2 entry in the TOC and insert a 'ul' element for the tag list
+    const tocContainer = document.querySelector(
+      ".toc-tree > ul > li > ul > li:last-child"
+    );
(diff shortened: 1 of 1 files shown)


dwilding:
@IronCore864 I managed to get the TOC working :smiley: What do you think? https://canonical-pebble--542.com.readthedocs.build/en/542/reference/api/

IronCore864 (approved):

PR #543, opened 2024-12-23
docs: add more api specs

IronCore864:
Adding more endpoints to the OpenAPI spec.

## Refactor

- Refactor existing endpoint parameter examples and remove them.
- Refactor existing request body examples, but keep only one.
- Refactor non-directly used component names as lowercase.
- Reorder API groups according to `pebble help` output.

## Style updates

- Similar to hiding the input box for parameters, hide the select drop-down (because one newly added endpoint has an enum).

## Newly added specs

- `/v1/system-info`. The code is [here](https://github.com/canonical/pebble/blob/master/internals/daemon/api.go#L122), and the possible boot-id value is [here](https://github.com/canonical/pebble/blob/master/internals/overlord/overlord.go#L243).
- `/v1/health`. The code is [here](https://github.com/canonical/pebble/blob/master/internals/daemon/api_health.go#L31).
- `/v1/changes`, code is [here](https://github.com/canonical/pebble/blob/master/internals/daemon/api_changes.go#L118).
- `/v1/changes/{id}`, both GET and Post. Code: [get](https://github.com/canonical/pebble/blob/master/internals/daemon/api_changes.go#L173), [post](https://github.com/canonical/pebble/blob/master/internals/daemon/api_changes.go#L226).
- `/v1/services/{name}`, not implemented in the [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_services.go#L223)
- `/v1/checks`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_checks.go#L34).

Sorry for this big PR, but it's only logical::

- `/v1/system-info` is more of a miscellaneous endpoint that doesn't belong to any group/tag.
- Since `/v1/health` is another public API besides `/v1/system-info`, I included it as well, only to find out later that `/v1/changes` belongs to the same group so I included it in the same PR.
- Since we already have some services and changes spec, to make the services and changes group/tag complete, I included other endpoints of the same group.

### DIFF
--- a/docs/specs/openapi.yaml
+++ b/docs/specs/openapi.yaml
@@ -16,7 +16,7 @@ paths:
           schema:
             type: string
       responses:
-        '200':
+        "200":
           description: List of services
           content:
             application/json:
@@ -65,7 +65,7 @@ paths:
                   action: start
                   services: ["svc1"]
       responses:
-        '202':
+        "202":
           description: Accepted - asynchronous operation started. Returns a change ID.
           content:
             application/json:
@@ -79,17 +79,6 @@ paths:
                     status: Accepted
                     change: "1"
                     result: null
-  /v1/services/{name}:
-    get:
-      summary: Get service information
-      tags:
-        - services
-      description: Not implemented yet.
-    post:
-      summary: Operate a service
-      tags:
-        - services
-      description: Not implemented yet.
   /v1/health:
     get:
       summary: Health of specified checks
@@ -109,7 +98,7 @@ paths:
           schema:
             type: string
       responses:
-        '200':
+        "200":
           description: Check is healthy
           content:
             application/json:
@@ -118,7 +107,7 @@ paths:
               examples:
                 healthy:
                   value:
-                    type: async
+                    type: sync
                     status-code: 200
                     status: OK
                     result:
@@ -142,7 +131,7 @@ paths:
           schema:
             type: string
       responses:
-        '200':
+        "200":
           description: Get checks
           content:
             application/json:
@@ -151,7 +140,7 @@ paths:
               examples:
                 get checks:
                   value:
-                    type: async
+                    type: sync
                     status-code: 200
                     status: OK
                     result:
@@ -168,7 +157,7 @@ paths:
       parameters:
         - name: select
           in: query
-          description: Filter changes by status. Defaults to "in-progress".
+          description: Filter changes by status.
           schema:
             type: string
             enum: [all, in-progress, ready]
@@ -179,37 +168,59 @@ paths:
           schema:
             type: string
       responses:
-        '200':
+        "200":
           description: Get changes
           content:
             application/json:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/GetChanges'
-              examples:
-                get changes:
-                  value:
-                    type: sync
-                    status-code: 200
-                    status: OK
-                    result:
-                      - id: 42
-                        kind: perform-check
-                        summary: "Perform HTTP check \"svc1-up\""
-                        status: Doing
-                        tasks: 
-                          - id: 42
-                            kind: perform-check
-                            summary: "Perform HTTP check \"svc1-up\""
-                            status: Doing,
-                            progress:
-                              label: ""
-                              done: 1
-                              total: 1
-                            spawn-time: "2024-12-22T19:53:14.85060309+08:00"
-                            ready: false,
-                        spawn-time: "2024-12-22T19:53:14.850613632+08:00"
+                  $ref: "#/components/schemas/GetChanges"
+              example:
+                {
+                  "type": "sync",
+                  "status-code": 200,
+                  "status": "OK",
+                  "result": [
+                      {
+                          "id": "4",
+                          "kind": "stop",
+                          "summary": "Stop service \"svc1\" and 1 more",
+                          "status": "Done",
+                          "tasks": [
+                              {
+                                  "id": "7",
+                                  "kind": "stop",
+                                  "summary": "Stop service \"svc1\"",
+                                  "status": "Done",
+                                  "progress": {
+                                      "label": "",
+                                      "done": 1,
+                                      "total": 1
+                                  },
+                                  "spawn-time": "2024-12-24T10:08:14.399194229+08:00",
+                                  "ready-time": "2024-12-24T10:08:14.429319813+08:00"
+                              },
+                              {
+                                  "id": "8",
+                                  "kind": "stop",
+                                  "summary": "Stop service \"svc2\"",
+                                  "status": "Done",
+                                  "progress": {
+                                      "label": "",
+                                      "done": 1,
+                                      "total": 1
+                                  },
+                                  "spawn-time": "2024-12-24T10:08:14.399199354+08:00",
+                                  "ready-time": "2024-12-24T10:08:14.432387271+08:00"
+                              }
+                          ],
+                          "ready": true,
+                          "spawn-time": "2024-12-24T10:08:14.399202521+08:00",
+                          "ready-time": "2024-12-24T10:08:14.432389313+08:00"
+                      }
+                  ]
+              }
   /v1/changes/{id}:
     get:
       summary: Get a specific change
@@ -224,12 +235,12 @@ paths:
           schema:
             type: string
       responses:
-        '200':
+        "200":
           description: Successful operation
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/GetChange'
+                $ref: "#/components/schemas/ChangeObject"
               examples:
                 get change:
                   value:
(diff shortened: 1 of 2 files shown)


docs/specs/openapi.yaml
```diff
@@ -183,6 +426,41 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/BaseResponse'
+        - type: object
+          properties:
+            change:
+              type: string
+              description: The Change ID of the asynchronous change.
+            result:
+              type:
+                - object
```
dwilding, line 439: When can `POST /v1/services` return an object for `result`? Looking at `v1PostServices()` in api_services.go, I only see `return AsyncResponse(nil, change.ID())` for returning a result.

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+                      ready: true
+                      spawn-time: "2024-12-12T12:50:29.743570175+08:00"
+                      ready-time: "2024-12-12T12:50:30.751094634+08:00"
+  /v1/system-info:
+    get:
+      summary: Get system information
+      description: |
+        Get Pebble daemon's version and boot-id of the system.
+      tags:
+        - misc
```
dwilding, line 373: How about `system` instead of `misc`?

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+  /v1/changes:
+    get:
+      summary: Get changes
+      tags:
+        - changes
+      description: Fetch information for the specified changes.
+      parameters:
+        - name: select
+          in: query
+          description: Filter changes by status. Defaults to "in-progress".
```
dwilding, line 171: I think we don't need to mention the default value in the description, because the default value is already shown (thanks to `default: in-progress` in the yaml)

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+            type: string
+      responses:
+        '200':
+          description: Get changes
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/GetChanges'
```
dwilding, line 189: Should this be an array? From the example and also the code, it seems that it should be like this:
```suggestion
                $ref: '#/components/schemas/GetChanges'
```

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+                    status: OK
+                    result:
+                      - id: 42
+                        kind: perform-check
+                        summary: "Perform HTTP check \"svc1-up\""
+                        status: Doing
+                        tasks: 
+                          - id: 42
+                            kind: perform-check
+                            summary: "Perform HTTP check \"svc1-up\""
```
dwilding, line 204: Would a task inside a change normally have the same ID and summary as the change? I'm thinking it might be helpful to show a different ID and summary, to distinguish the task from the change

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+                    status: OK
+                    result:
+                      id: 76
+                      kind: perform-check
+                      summary: "Perform HTTP check \"svc1-up\""
+                      status: Abort
+                      tasks: 
+                        - id: 76
+                          kind: perform-check
+                          summary: "Perform HTTP check \"svc1-up\""
```
dwilding, line 302: Same question about whether the change ID & summary would be the same as the task ID & summary

docs/specs/openapi.yaml
```diff
@@ -167,13 +401,22 @@ components:
...
       allOf:
         - $ref: '#/components/schemas/BaseResponse'
         - type: object
           properties:
             result:
-              $ref: '#/components/schemas/ChangeInfo'
+              type: array
+              items:
+                $ref: '#/components/schemas/changeInfo'    
+    GetChange:
```
dwilding, line 413: Since this name will display in the rendered output, as the response schema for several endpoints, it might be clearer to rename it to `ChangeObject`

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+                          spawn-time: "2024-12-12T12:50:29.743512342+08:00"
+                          ready-time: "2024-12-12T12:50:30.751092759+08:00"
+                      ready: true
+                      spawn-time: "2024-12-12T12:50:29.743570175+08:00"
+                      ready-time: "2024-12-12T12:50:30.751094634+08:00"
+  /v1/system-info:
+    get:
+      summary: Get system information
+      description: |
+        Get Pebble daemon's version and boot-id of the system.
```
dwilding, line 371: Minor suggestion for consistency between descriptions:
```suggestion
        Get the version of the Pebble daemon and the boot ID of the system.
```

docs/specs/openapi.yaml
```diff
@@ -284,3 +550,44 @@ components:
...
+      properties:
+        version:
+          type: string
+          description: Version of the Pebble daemon
+        boot-id:
+          type: string
+          description: |
+            Boot ID of the system, `/proc/sys/kernel/random/boot_id`.
+            Or `/proc/sys/kernel/random/boot_id` if pebble is PID 1 since in this case we are most likely running in a container
+            so we don't care about `/proc/sys/kernel/random/boot_id`.
```
dwilding, line 564: I don't understand this part, is there a typo? From the text, it seems like one of the paths should be different, but all the paths are `/proc/sys/kernel/random/boot_id`

docs/specs/openapi.yaml
```diff
@@ -167,13 +401,22 @@ components:
         status:
           type: string
           description: The text for the HTTP status code, see https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml.
```
dwilding, line 403: We should be able to use Markdown for this. I'd suggest a couple of very minor edits:
```suggestion
          description: |
            The description of the HTTP status code.
            See the [IANA list](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
```

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+        '200':
+          description: Check is healthy
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/GetHealth"
+              examples:
+                healthy:
+                  value:
+                    type: async
```
dwilding, line 121: I think this should be `sync`, not `async`

docs/specs/openapi.yaml
```diff
@@ -153,6 +79,314 @@ paths:
...
+        '200':
+          description: Get checks
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/GetChecks"
+              examples:
+                get checks:
+                  value:
+                    type: async
```
dwilding, line 154: I think this should be `type: sync`, not `type: async`

dwilding (approved):
Thanks! Approving so that you can merge later. I've added specific questions/suggestions (see the comments) and I have a couple more thoughts:

- For the two unsupported endpoints, I think it's best to exclude them from the doc if possible. To make it easier to navigate around the other endpoints

- I'm a bit confused about `GET /v1/health` - Does this return a 200 response for `healthy: true` and `healthy: false`? From reading the code, it seems like the response for `healthy: false` is `http.StatusBadGateway` (code 502). Maybe I misunderstood, but don't we need two different responses for this endpoint?

docs/specs/openapi.yaml
IronCore864: Changes is a list, according to the code [here](https://github.com/canonical/pebble/blob/master/internals/daemon/api_changes.go#L163).

To double-check this, I did some tests. Run: `curl -X GET -H "Content-type: application/json" -H "Accept: application/json" "localhost:4000/v1/changes?select=all"` and we get a list as the result:

```json
{
    "type": "sync",
    "status-code": 200,
    "status": "OK",
    "result": [
        {
            "id": "1",
            "kind": "autostart",
            "summary": "Autostart service \"svc1\" and 1 more",
            "status": "Done",
            "tasks": [
                {
                    "id": "1",
                    "kind": "start",
                    "summary": "Start service \"svc1\"",
                    "status": "Done",
                    "progress": {
                        "label": "",
                        "done": 1,
                        "total": 1
                    },
                    "spawn-time": "2024-12-24T10:02:57.342163161+08:00",
                    "ready-time": "2024-12-24T10:02:58.349252454+08:00"
                },
                {
                    "id": "2",
                    "kind": "start",
                    "summary": "Start service \"svc2\"",
                    "status": "Done",
                    "progress": {
                        "label": "",
                        "done": 1,
                        "total": 1
                    },
                    "spawn-time": "2024-12-24T10:02:57.342167536+08:00",
                    "ready-time": "2024-12-24T10:02:58.352393079+08:00"
                }
            ],
            "ready": true,
            "spawn-time": "2024-12-24T10:02:57.342169703+08:00",
            "ready-time": "2024-12-24T10:02:58.352395537+08:00"
        },
        {
            "id": "2",
            "kind": "stop",
            "summary": "Stop all running services",
            "status": "Done",
            "tasks": [
                {
                    "id": "3",
                    "kind": "stop",
                    "summary": "Stop service \"svc1\"",
                    "status": "Done",
                    "progress": {
                        "label": "",
                        "done": 1,
                        "total": 1
                    },
                    "spawn-time": "2024-12-24T10:03:45.422085643+08:00",
                    "ready-time": "2024-12-24T10:03:45.456315268+08:00"
                },
                {
                    "id": "4",
                    "kind": "stop",
                    "summary": "Stop service \"svc2\"",
                    "status": "Done",
                    "progress": {
                        "label": "",
                        "done": 1,
                        "total": 1
                    },
                    "spawn-time": "2024-12-24T10:03:45.422093309+08:00",
                    "ready-time": "2024-12-24T10:03:45.453272143+08:00"
                }
            ],
            "ready": true,
            "spawn-time": "2024-12-24T10:03:45.425286893+08:00",
            "ready-time": "2024-12-24T10:03:45.456316351+08:00"
        }
    ]
}

The example in the spec is also in the same format, see the preview [here](https://canonical-pebble--543.com.readthedocs.build/en/543/reference/api/#/changes/get_v1_changes).

I think it's safe to resolve this comment for now, if I made a mistake please reopen.
```

docs/specs/openapi.yaml
IronCore864: Yes, this makes sense. It's possible for one change to have multiple tasks, for example, if we do `pebble start svc1 svc2`, then starting svc1 and svc2 would be in different tasks. Let me update the example here.

docs/specs/openapi.yaml
IronCore864: Example updated to a real-world example with different IDs and also updated "examples" to a single example in JSON format.

docs/specs/openapi.yaml
IronCore864: I added a new empty line between the two lines of the description because I think you want to make them show up in different lines, right? Not just to improve the readability of the YAML spec. Feel free to open another small PR to fix this if it wasn't your intention.

docs/specs/openapi.yaml
IronCore864: Great, thanks for pointing this out. I made a mistake when copy-pasting. It originates from the [comment here](https://github.com/canonical/pebble/blob/5dc07864cf04032a4a2b5d2af48230abcc690cc2/internals/overlord/overlord.go#L237), and when Pebble is PID 1, it should be `random/uuid`, different from `random/boot_id` indeed.

docs/specs/openapi.yaml
IronCore864: Yes, you are absolutely right, after checking the code, the only valid value for "result" is null. There is a case where the result is not empty when there is an error but it's irrelevant here.

IronCore864:
> Thanks! Approving so that you can merge later. I've added specific questions/suggestions (see the comments) and I have a couple more thoughts:
> 
> * For the two unsupported endpoints, I think it's best to exclude them from the doc if possible. To make it easier to navigate around the other endpoints

Removed.

> * I'm a bit confused about `GET /v1/health` - Does this return a 200 response for `healthy: true` and `healthy: false`? From reading the code, it seems like the response for `healthy: false` is `http.StatusBadGateway` (code 502). Maybe I misunderstood, but don't we need two different responses for this endpoint?

Yes, it's 200 when healthy is true and 502 when false. However, in a discussion with Ben earlier, we agreed to only show responses for 200 and not 4xx/5xx responses, which are shown at the beginning of the doc, so I have only one response/example here.

PR #544, opened 2024-12-27
docs: finish pebble http api specs

IronCore864:
In this PR:

## Refactor

- Add 502 responses for `/v1/health` endpoint according to David's suggestion, because otherwise, people would think it's 200 even if healthy is false.
- Reformat multi-line descriptions.
- Change "examples" to "example" to make the spec shorter with less indentation
- Change the example to JSON format to make it less error-prone (no need to translate JSON response to YAML format in the spec).
- Rename component names to more appropriate ones because they will appear in the schema. Convention: `VerbResource[Description]Response`, for example, `GetPlanResponse`, `GetHealthOKResponse`.

## Newly added specs

- `/v1/plan`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_plan.go#L27).
- `/v1/layers`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_plan.go#L42).
- `/v1/logs`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_logs.go#L45).
- `/v1/signals`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_signals.go#L27).
- `/v1/files`, code: [GET](https://github.com/canonical/pebble/blob/master/internals/daemon/api_files.go#L40), [POST](https://github.com/canonical/pebble/blob/master/internals/daemon/api_files.go#L344).
- `/v1/tasks/{task-id}/websocket/{websocket-id}`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_files.go#L344).
- `/v1/exec`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_exec.go#L46).
- `/v1/notices`, code: [GET](https://github.com/canonical/pebble/blob/master/internals/daemon/api_notices.go#L45), [POST](https://github.com/canonical/pebble/blob/master/internals/daemon/api_notices.go#L181).
- `/v1/notices/{id}`, [code](https://github.com/canonical/pebble/blob/master/internals/daemon/api_notices.go#L242).
- `/v1/identities`,  code: [GET](https://github.com/canonical/pebble/blob/master/internals/daemon/api_identities.go#L24), [POST](https://github.com/canonical/pebble/blob/master/internals/daemon/api_identities.go#L33).

Preview: https://canonical-pebble--544.com.readthedocs.build/en/544/reference/api/.

### DIFF
--- a/docs/specs/openapi.yaml
+++ b/docs/specs/openapi.yaml
@@ -19,7 +19,7 @@ paths:
           required: true
       responses:
         "200":
-          description: Get plan.
+          description: The current plan in YAML format.
           content:
             application/json:
               schema:
@@ -143,11 +143,8 @@ paths:
                     Ignored for "replan" and "autostart" (resolved automatically for "autostart" to default services).
                   items:
                     type: string
-            examples:
-              start one service:
-                value:
-                  action: start
-                  services: ["svc1"]
+            example:
+              {"action": "start", "services": ["svc1"]}
       responses:
         "202":
           description: Accepted - asynchronous operation started. Returns a change ID.
@@ -171,6 +168,14 @@ paths:
       description: |
         Fetch previously-written logs from the given services,
         or request logs from the given services and follows them, in JSON lines format.
+
+        Example:
+
+        ```
+        {"time":"2024-12-31T02:11:09.361Z","service":"svc1","message":" * Serving Flask app 'main'"}
+        {"time":"2024-12-31T02:11:09.361Z","service":"svc1","message":" * Debug mode: off"}
+        {"time":"2024-12-31T02:11:09.382Z","service":"svc1","message":" * Running on http://127.0.0.1:5000"}
+        ```
       parameters:
         - name: services
           in: query
@@ -184,10 +189,7 @@ paths:
             type: string
         - name: follow
           in: query
-          description: |
-            Whether to follow the logs (keep streaming new logs).
-
-            Must be "true" or "false"; defaults to false.
+          description: Whether to follow the logs (keep streaming new logs).
           schema:
             type: string
             enum: ["true", "false"]
@@ -197,21 +199,19 @@ paths:
           description: |
             Number of log entries to retrieve.
 
-            If `follow` is true, this parameter is ignored and defaults to zero.
+            If `follow` is true, `n` is ignored.
 
-            In non-follow mode, the default value is server-defined, currently 30.
+            If `follow` is false:
 
-            If `n` is -1, all available logs are returned (up to a server-defined limit).
-
-            If `n` is 0 and `follow` is false, a server-defined default number of logs is returned.
-
-            If `n` is a positive integer, up to that many logs are returned.
+            - If `n` is -1, all available logs are returned (up to a server-defined limit).
+            - If `n` is 0 or not specified, a server-defined default number of logs is returned. The default is currently 30.
+            - If `n` is a positive integer, up to that many logs are returned.
           schema:
             type: integer
             format: int32
       responses:
         "200":
-          description: Get service logs in JSON lines format.
+          description: Service logs in JSON lines format.
           content:
             application/x-ndjson:
               schema:
@@ -247,11 +247,11 @@ paths:
               required:
                 - signal
                 - services
-              example:
-                {
-                  "signal": "SIGHUP",
-                  "services": ["svc1"]
-                }
+            example:
+              {
+                "signal": "SIGHUP",
+                "services": ["svc1"]
+              }
       responses:
         "200":
           description: Signal sent successfully
@@ -335,7 +335,7 @@ paths:
             type: string
       responses:
         "200":
-          description: Get checks
+          description: Information about health checks.
           content:
             application/json:
               schema:
@@ -386,7 +386,9 @@ paths:
             type: string
         - name: itself
           in: query
-          description: Whether to list the directory itself, rather than its contents for the "list" action.
+          description: |
+            For the "list" action, `itself` specifies whether to return information about the directory itself ("true")
+            or list the contents of the directory ("false").
           schema:
             type: string
             enum: ["false", "true"]
@@ -438,15 +440,15 @@ paths:
                     ]
                   }
     post:
-      summary: Create, write, remove files/directories.
+      summary: Create, write, remove files/directories
       tags:
         - files
       description: |
-        Write content to a path on the remote system;
-        
-        creates a directory or directory tree;
+        This endpoint can:
 
-        delete a file or directory.
+        - Write content to a path on the remote system. For this mode, use a multipart/form-data request body with JSON metadata in the first part. In the JSON metadata, set `action` to `write`.
+        - Create a directory or directory tree. For this mode, use an application/json request body with `action` set to `make-dirs`.
+        - Delete a file or directory. For this mode, use an application/json request body with `action` set to `remove`.
       requestBody:
         content:
           multipart/form-data:  # action: write
@@ -495,7 +497,7 @@ paths:
       summary: Execute a command
       tags:
         - files
-      description: Start a command with the given options, return a value representing the process.
+      description: Start a command with the given options and return a value representing the process.
       requestBody:
         required: true
         content:
@@ -581,7 +583,7 @@ paths:
             type: string
       responses:
         "200":
-          description: Get changes
+          description: Information about changes.
           content:
             application/json:
               schema:
@@ -682,7 +684,7 @@ paths:
                   }
                 }
     post:
-      summary: Perform an action on a change.
+      summary: Perform an action on a change
       tags:
         - changes
       description: Perform an action on a change. Currently the only supported action is "abort".
@@ -809,13 +811,13 @@ paths:
       parameters:
         - in: query
           name: user-id
-          description: Filter notices by user ID (admin only). Only one user ID can be specified.
+          description: Filter notices by user ID. Only one user ID can be specified. This parameter can only be used by admin users.
           schema:
             type: integer
             format: int32
         - in: query
           name: users
-          description: If set to "all", return notices for all users (admin only). Cannot be used with `user-id`.
+          description: If set to "all", return notices for all users. Cannot be used with `user-id`. This parameter can only be used by admin users.
           schema:
             type: string
             enum: ["all"]
@@ -848,7 +850,7 @@ paths:
             format: duration
       responses:
         "200":
-          description: Successful response.
+          description: Notices successfully retrieved.
           content:
             application/json:
               schema:
--- a/docs/specs/openapi.yaml
+++ b/docs/specs/openapi.yaml
@@ -3,6 +3,744 @@ info:
   title: Pebble API
   version: v1
 paths:
+  /v1/plan:
+    get:
+      summary: Get the current plan
+      tags:
+        - plan
+      description: Get the plan in YAML format.
+      parameters:
+        - name: format
+          in: query
+          description: The format of the plan.
+          schema:
+            type: string
+            enum: [yaml]
+          required: true
+      responses:
+        "200":
+          description: Get plan.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/GetPlanResponse"
+              example:
+                {
+                  "type": "sync",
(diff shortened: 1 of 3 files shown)


docs/specs/openapi.yaml
```diff
@@ -70,15 +154,120 @@ paths:
...
+
+            If not set, returns logs for all services.
+          schema:
+            type: string
+        - name: follow
+          in: query
+          description: |
+            Whether to follow the logs (keep streaming new logs).
+
+            Must be "true" or "false"; defaults to false.
```
dwilding, line 190: The rendered yaml already shows the available values as true and false, and also the default value of false. So I think we can remove this line

docs/specs/openapi.yaml
```diff
@@ -70,15 +154,120 @@ paths:
...
+
+            If `follow` is true, this parameter is ignored and defaults to zero.
+
+            In non-follow mode, the default value is server-defined, currently 30.
+
+            If `n` is -1, all available logs are returned (up to a server-defined limit).
+
+            If `n` is 0 and `follow` is false, a server-defined default number of logs is returned.
+
+            If `n` is a positive integer, up to that many logs are returned.
```
dwilding, line 208: I found this a little bit confusing. How about using this (if it's correct):

```suggestion
            If `follow` is true, `n` is ignored.

            If `follow` is false:
            - If `n` is -1, all available logs are returned (up to a server-defined limit).
            - If `n` is 0 or not specified, a server-defined default number of logs is returned. The default is currently 30.
            - If `n` is a positive integer, up to that many logs are returned.
```

docs/specs/openapi.yaml
```diff
@@ -70,15 +154,120 @@ paths:
...
+                    type: string
+                  minItems: 1  # At least one service is required.
+              required:
+                - signal
+                - services
+              example:
+                {
+                  "signal": "SIGHUP",
+                  "services": ["svc1"]
+                }
```
dwilding, line 256: I think this can be indented up one level (to be the same level as `schema`). Currently, Swagger UI is showing the example under "Schema" as well as under "Example Value". We only need the example to show under "Example Value"

docs/specs/openapi.yaml
```diff
@@ -136,18 +341,227 @@ paths:
...
+                    "status-code": 200,
+                    "status": "OK",
+                    "result": [
+                      {
+                        "path": "/home/ubuntu/bar"
+                      }
+                    ]
+                  }
+    post:
+      summary: Create, write, remove files/directories.
```
dwilding, line 443: Suggest removing period for consistency with other summaries:
```suggestion
      summary: Create, write, remove files/directories
```

docs/specs/openapi.yaml
```diff
@@ -238,44 +652,361 @@ paths:
...
+                        "ready-time": "2024-12-27T12:31:27.681780702+08:00"
+                      }
+                    ],
+                    "ready": true,
+                    "spawn-time": "2024-12-27T12:31:26.673297951+08:00",
+                    "ready-time": "2024-12-27T12:31:27.686371869+08:00"
+                  }
+                }
+    post:
+      summary: Perform an action on a change.
```
dwilding, line 687: Suggest removing period for consistency with other summaries:
```suggestion
      summary: Perform an action on a change
```

docs/specs/openapi.yaml
```diff
@@ -136,18 +341,227 @@ paths:
...
+          style: form # For handling comma-separated values in "read".
+          explode: true # For handling comma-separated values in "read".
+        - name: pattern
+          in: query
+          description: Glob pattern to filter files/directories for the "list" action.
+          schema:
+            type: string
+        - name: itself
+          in: query
+          description: Whether to list the directory itself, rather than its contents for the "list" action.
```
dwilding, line 389: At first I didn't understand what "list the directory itself" meant. How about this wording:
```suggestion
          description: For the "list" action, `itself` specifies whether to return information about the directory itself ("true") or list the contents of the directory ("false").
```

docs/specs/openapi.yaml
```diff
@@ -136,18 +339,227 @@ paths:
...
+    post:
+      summary: Create, write, remove files/directories.
+      tags:
+        - files
+      description: |
+        Write content to a path on the remote system;
+        
+        creates a directory or directory tree;
+
+        delete a file or directory.
```
dwilding, line 449: Suggest writing this as bullets:
```suggestion
        This endpoint can:
        - Write content to a path on the remote system        
        - Create a directory or directory tree
        - Delete a file or directory
```

docs/specs/openapi.yaml
```diff
@@ -136,18 +339,227 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/PostFilesResponse"
+              example:
+                {"type":"sync","status-code":200,"status":"OK","result":[{"path":"/home/ubuntu/foo"}]}
+  /v1/exec:
+    post:
+      summary: Execute a command
+      tags:
+        - files
+      description: Start a command with the given options, return a value representing the process.
```
dwilding, line 498: Minor readability suggestion:
```suggestion
      description: Start a command with the given options and return a value representing the process.
```

docs/specs/openapi.yaml
```diff
@@ -238,44 +650,361 @@ paths:
...
+  /v1/notices:
+    get:
+      summary: Get notices
+      tags:
+        - notices
+      description: Get a list of notices that match the filters, ordered by the last-repeated time.
+      parameters:
+        - in: query
+          name: user-id
+          description: Filter notices by user ID (admin only). Only one user ID can be specified.
```
dwilding, line 812: When I first read this, I thought it meant that the user ID must be the ID of an admin user. How about this wording:
```suggestion
          description: Filter notices by user ID. Only one user ID can be specified. This parameter can only be used by admin users.
```

docs/specs/openapi.yaml
```diff
@@ -238,44 +650,361 @@ paths:
...
+          schema:
+            type: integer
+            format: int32
+        - in: query
+          name: users
+          description: If set to "all", return notices for all users (admin only). Cannot be used with `user-id`.
```
dwilding, line 818: Similar suggestion as for the `user-id` parameter:
```suggestion
          description: If set to "all", return notices for all users. Cannot be used with `user-id`. This parameter can only be used by admin users.
```

docs/specs/openapi.yaml
```diff
@@ -238,44 +650,361 @@ paths:
...
+                      "type": "change-update",
+                      "key": "1",
+                      "first-occurred": "2024-12-27T09:55:13.393868798Z",
+                      "last-occurred": "2024-12-27T09:55:14.400978382Z",
+                      "last-repeated": "2024-12-27T09:55:14.400978382Z",
+                      "occurrences": 3,
+                      "last-data": {
+                        "kind": "autostart"
+                      },
+                      "expire-after": "168h0m0s"
```
dwilding, line 874: Should we also include `retry-after` in this example (and the example for POST /v1/notices/{id})? I don't think it's essential, I just wondered whether it's not here for some reason.

docs/specs/openapi.yaml
```diff
@@ -238,44 +650,361 @@ paths:
...
+            format: date-time
+        - in: query
+          name: timeout
+          description: The maximum time [duration](#duration) to wait for notices. If no notices are available within this time, an empty list is returned.
+          schema:
+            type: string
+            format: duration
+      responses:
+        "200":
+          description: Successful response.
```
dwilding, line 851: Suggest changing the description for consistency with GET /v1/notices/{id}:
```suggestion
          description: Notices successfully retrieved.
```

dwilding:
Thank you for preparing all of this! I've finished my review and have added comments on specific parts.

Some other comments:

- It looks like one `examples` still needs changing to `example`. [The one under POST /v1/services](https://github.com/canonical/pebble/blob/9f5275c1ffb02257ec5a7310e0bb3247418d8e09/docs/specs/openapi.yaml#L146-L150)

- Most descriptions have a period at the end, but a few are missing a period. I won't list them all here, but I used this regex to find them: `description:.*[^"|\.]$`

- A few 200 descriptions start with "Get ...". I think it would be clearer to describe the responses differently, as "Get" is more appropriate for a description of a request. How about these changes:
    - Get plan. **==>** The current plan in YAML format.
    - Get service logs in JSON lines format. **==>** Service logs in JSON lines format.
    - Get checks **==>** Information about health checks.
    - Get changes **==>** Information about changes.

- GET /v1/files appears more complicated than necessary because the operation does two different things, depending on the value of `action`. Would it be possible to document this as two separate endpoints?
    - GET /v1/files?action=list
    - GET /v1/files?action=read

    But if this is too messy to achieve in OpenAPI, I think it's OK to leave it as-is.

dwilding (approved):

docs/specs/openapi.yaml
IronCore864: Thanks for pointing this out, this is for sure a human error, it should only show under "example value".

docs/specs/openapi.yaml
IronCore864: Yes, here I should not have invented this description myself, it is not clear enough.

While I found a [comment from the code](https://github.com/canonical/pebble/blob/master/client/files.go#L44) on this parameter, it is not as good as the version you proposed in my eyes. So here I will use what you proposed.

docs/specs/openapi.yaml
IronCore864: I tried both change-update notices and custom notices, and neither returned `repeat-after`, so I left them out and used real responses from Pebble.

IronCore864:
> Thank you for preparing all of this! I've finished my review and have added comments on specific parts.
> 
> Some other comments:
> 
> * It looks like one `examples` still needs changing to `example`. [The one under POST /v1/services](https://github.com/canonical/pebble/blob/9f5275c1ffb02257ec5a7310e0bb3247418d8e09/docs/specs/openapi.yaml#L146-L150)
> * Most descriptions have a period at the end, but a few are missing a period. I won't list them all here, but I used this regex to find them: `description:.*[^"|\.]$`
> * A few 200 descriptions start with "Get ...". I think it would be clearer to describe the responses differently, as "Get" is more appropriate for a description of a request. How about these changes:
>   
>   * Get plan. **==>** The current plan in YAML format.
>   * Get service logs in JSON lines format. **==>** Service logs in JSON lines format.
>   * Get checks **==>** Information about health checks.
>   * Get changes **==>** Information about changes.
> * GET /v1/files appears more complicated than necessary because the operation does two different things, depending on the value of `action`. Would it be possible to document this as two separate endpoints?
>   
>   * GET /v1/files?action=list
>   * GET /v1/files?action=read
>   
>   But if this is too messy to achieve in OpenAPI, I think it's OK to leave it as-is.

I fixed most comments above. There was a human error (the example indentation), and other style improvement suggestions were all adopted.

There is one thing I did not change, which is the example for notices, it doesn't have `repeat-after` in the response because I could not get this from the Pebble server no matter with custom notice or change-update notice.

For the three points above:

1. The last "examples" under POST `/v1/services`: I kept it because I want to have a named example, the name shows on the rendered page. But after a second thought, I think consistency is better, so your suggestions above are also taken.

2. The 200 response description: Yes your proposal makes sense and I updated all of them.

3. On separating `/v1/files` into two endpoints, after some research, I did not find a very good solution.

I think if we can split the API into two, following RestAPI design, it might be better, like `/v1/files/{filepath}` for reading the file and `/v1/files` for list files. But we probably can't do this change because it's not backward compatible.

If we are not changing the URL, potentially, we could unify the response format. Even though the data returned is different, we can encapsulate it within JSON for both actions, for example:

```yaml
examples:
  list:
    value:
      {
        "type": "sync",
        "status-code": 200,
        "status": "OK",
        "result": {
          "action": "list",
          "files": [
            { "path": "/file1.txt", ... }
          ]
        }
      }
  read:
    value:
      {
        "type": "sync",
        "status-code": 200,
        "status": "OK",
        "result": {
          "action": "read",
          "files": [
            { "path": "/file1.txt", "content": "File content here", ... }
          ]
        }
      }
```

But this is also not backward compatible.

Given the current situation, I will have to leave it as it is.

PR #545, opened 2024-12-31
docs: how to use pebble api

IronCore864:


dwilding:
Capturing my initial thoughts before I do a more detailed review (not today):

- The doc is based around a scenario of dynamically setting up a service configuration in the context of CI. I think we should present this as the purpose of the doc. So perhaps the title could be "How to use the Pebble API for automated service testing". The short title could be "Automate service testing".

- If we present the doc this ^ way, the introduction should state that the purpose of the doc is to demonstrate how to programmatically set up a service configuration, run tests against the services, then tear down the configuration.

- There seems to be some overlap between the introduction and the "API and automated workflows" section. Since this is a how-to doc, I'd move as much of the explanation into the "API and automated workflows" section (keeping it brief) and make the introduction shorter & focused on the purpose of the doc.

- Before the examples of using the API, should we mention how to make sure that Pebble is available in the environment? What I mean is - in the context of CI, how do we recommend that Pebble gets installed?

- The doc talks about using Pebble as a more robust alternative to shell scripts, but then the first examples we give are curl. On the surface, it doesn't appear much more robust/structured than using shell scripts. Should we take a stronger position and position Go or Python as the recommended approach? We could still include curl, but I don't think we should lead with it.

- In the code examples, would it be possible to include a demo of how to run a couple of tests against the services? I know it might be difficult without specific services in mind... Maybe we could have a comment that said something like "Suppose that service 'svc1' exposes an HTTP API on port 5000" then include a code snippet of how to check that a particular endpoint such as `/version` returns 200. It doesn't need to be super realistic - just something to make this step more concrete.

IronCore864:
> Capturing my initial thoughts before I do a more detailed review (not today):
> 
> * The doc is based around a scenario of dynamically setting up a service configuration in the context of CI. I think we should present this as the purpose of the doc. So perhaps the title could be "How to use the Pebble API for automated service testing". The short title could be "Automate service testing".
> * If we present the doc this ^ way, the introduction should state that the purpose of the doc is to demonstrate how to programmatically set up a service configuration, run tests against the services, then tear down the configuration.
> * There seems to be some overlap between the introduction and the "API and automated workflows" section. Since this is a how-to doc, I'd move as much of the explanation into the "API and automated workflows" section (keeping it brief) and make the introduction shorter & focused on the purpose of the doc.
> * Before the examples of using the API, should we mention how to make sure that Pebble is available in the environment? What I mean is - in the context of CI, how do we recommend that Pebble gets installed?
> * The doc talks about using Pebble as a more robust alternative to shell scripts, but then the first examples we give are curl. On the surface, it doesn't appear much more robust/structured than using shell scripts. Should we take a stronger position and say that Go or Python is the recommended approach? We could still include curl, but I don't think we should lead with it.
> * In the code examples, would it be possible to include a demo of how to run a couple of tests against the services? I know it might be difficult without specific services in mind... Maybe we could have a comment that said something like "Suppose that service 'svc1' exposes an HTTP API on port 5000" then include a code snippet of how to check that a particular endpoint such as `/version` returns 200. It doesn't need to be super realistic - just something to make this step more concrete.

On the title:

I think we should base this how-to on a scenario, which is also what we discussed earlier when talking about Pebble how-to guide ideas, see the [doc here](https://docs.google.com/document/d/1HdpkWWE2Y-YipbeSq7fEOTReUfeDbZZ7WRV7K8DICAI/edit?tab=t.0) (internal only, sorry).

The scenario in the example is indeed "Automate service testing", but the same pattern can be used in anything that needs to be automated. So maybe using "Automate service testing" as the title would be too narrow, because if I'm a user of Pebble and I want to use Pebble for some automation work but not automated testing, seeing this title would not attract me and I might not click on it.

On making Pebble available in the env:

I think "how to make Pebble available in the CI env" is a tricky question to answer, mostly because there is no right answer because the CI runner could be a VM where Pebble can be installed in many ways: manual/configuration management/built in the custom image, or it could be an OCI image where Pebble is or is not the entrypoint. Maybe we want to recommend using a container CI runner with Pebble as the entrypoint (from Canonical's standpoint), but it is too opinionated and may only be preferred/known by/to rock/charm users.

So, how about we only slightly mention that it must be available as a prerequisite, list a few options, but don't recommend a specific one?

On adding more steps to run tests:

For similar reasons, some detailed steps (or even some mocks) to run some tests are not included, because there are many ways. Maybe the user doesn't want to trigger the tests in one script. Maybe the users can have a setup script, then run a test in a different step, then another step for the teardown. The applications could be endless.

What's more, I think the goal of a how-to is not to provide something the user can copy-paste and reproduce (that would be the goal of a tutorial), but to show how to use some feature and why this feature can be useful.

On the order of the ways to access the API:

Compared to systemd, the API is a strong selling point, but it's not logical to put curl in the first place. However, not leading with curl is a bit tricky because it's the most well-known and easy-to-start method for API access.

Thoughts? Shall we reorder it to something like Python-Go-curl? (Ranking Python before Go purely because of the [TIOBE index](https://www.tiobe.com/tiobe-index/)).

docs/how-to/use-pebble-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+For example, your CI workflow could start a service or even a group of inter-dependent services, verify services are running and health checks are OK, run integration tests against those services, and finally stop the services as a cleanup step. This automation reduces intervention, reduces risks of human errors, and ensures consistency and idempotency across different environments.
+
+The Pebble API offers a significant advantage over CLI-based approaches in CI/CD pipelines. First of all, the structured interface simplifies interactions, eliminating the need for parsing command output. Second, API enables more sophisticated error handling. Last but not least, using the API promotes code reusability and maintainability, because instead of hard-coding shell commands throughout the CI workflow definition, we can encapsulate service operations within dedicated functions and modules, which leads to cleaner and more maintainable workflows.
+
+---
```
dwilding, line 21: I think we can compress this part to make the points stand out more clearly and reduce duplication. I used your text and came up with this suggestion - hopefully I didn't remove too much. If you think an important point as been lost, please put me straight!

```suggestion
Pebble provides a command-line interface (CLI) for managing services, similar to systemd. However, Pebble distinguishes itself with its dedicated REST API.

This guide demonstrates how to use the Pebble API to programmatically manage services as part of an automated workflow.


## Automated workflows

As an example scenario, consider an automated workflow that starts and tests a group of inter-dependent services as part of a continuous integration pipeline. The tests could include verifying that the services are running, querying their status and health, and running integration tests. After testing the services, the workflow stops the services.

Although you could use shell scripts and CLI commands to implement the workflow, the Pebble API enables a more robust and maintainable approach:

- You don't need to parse command output.
- You can handle errors in a more sophisticated way.
- You can encapsulate service operations within reusable functions and modules.

This approach reduces manual intervention and the risk of human error. It also supports consistency and idempotency across different environments.
```

docs/how-to/use-pebble-api.md
```diff
@@ -0,0 +1,281 @@
...
+   :end-before: End: Pebble API overview
+```
+
+For more information, see [API](/reference/api).
+
+---
+
+## Using the API
+
+You can use different tools and clients to access the API. For more examples, see [API and clients](../explanation/api-and-clients) and the [API reference doc](../reference/api).
```
dwilding, line 36: I recommend combining this all into the "Using the API section". Like this:

````suggestion
## Using the API

```{include} /reuse/api.md
   :start-after: Start: Pebble API overview
   :end-before: End: Pebble API overview
```

For reference information about the API, see [](../explanation/api-and-clients) and [](../reference/api).
````

**Comment by dwilding** [comment]
I'm adding suggestions on specific parts.

To continue the discussion of the general comments:

- For the long title, how about "How to use the Pebble API to manage services". And for the short title "Use the Pebble API". I like what you said about having an attractive title.

- Thinking about what you said about installing Pebble, I think it's best if we don't add anything about this. We already have "Install Pebble" as the first item under "How-to guides" in the menu. Like you said for another point, this new doc isn't a tutorial - it's supposed to be focused on the API.

- Let's see what Ben thinks, but I also feel the order should be Python, Go, curl. (BTW, in the API reference doc it makes sense to put curl first, because people may be manually trying out different endpoints, and curl is quick to do that.)

File: docs/how-to/use-pebble-api.md [hunk 3]
```diff
@@ -0,0 +1,281 @@
...
+
+### curl
+
+At the moment (v1.17.0), only a handful of Pebble APIs are exposed via HTTP, most APIs are only available via the linux websocket. With curl, there is a `H       `--unix-socket <path>` parameter, which allows us to connect through this Unix domain socket instead of using the network. In this section, let's see how we can use curl to access Pebble API via the unix socket.
```
dwilding, line 40: I recommend removing that "at the moment" part. We can update this at a later date if the situation changes. For now, how about this:
```suggestion
Most APIs are only available via the Unix socket. With curl, there is a `H       `--unix-socket <path>` parameter, which allows us to connect through this Unix domain socket instead of using the network. In this section, let's see how we can use curl to access Pebble API via the unix socket.
```

docs/how-to/use-pebble-api.md
```diff
@@ -0,0 +1,281 @@
...
+curl --unix-socket $PEBBLE/.pebble.socket -XGET http://localhost/v1/services
+```
+
+And we should get the following JSON response showing that there is no service:
+
+```json
+{"type":"sync","status-code":200,"status":"OK","result":[]}
+```
+
+To add a simple layer with one service. run:
```
dwilding, line 63: Typo
```suggestion
To add a simple layer with one service, run:
```

docs/how-to/use-pebble-api.md
```diff
@@ -0,0 +1,281 @@
...
+    "ready": true,
+    "spawn-time": "2024-12-30T08:57:43.194065853+08:00",
+    "ready-time": "2024-12-30T08:57:44.210879687+08:00"
+  }
+}
+```
+
+If we try to get the services again, we can see it's active now.
+
+We can integrate these steps in our CI workflow to start services and dependencies, and now we can run some tests against them. Of course, after the CI workflow is finished, we want to do some cleanup like stopping those services.
```
dwilding, line 140: Some small suggestions for consistency and readability:
```suggestion
We can integrate these steps in our automated workflow to start services and their dependencies. Then we're able to run some tests against the services. As a final step in the workflow, we'll want to stop the services.
```

docs/how-to/use-pebble-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+To stop a service, run:
+
+```bash
+curl --unix-socket $PEBBLE/.pebble.socket -XPOST http://localhost/v1/services -d '{"action": "stop", "services": ["svc1"]}'
+```
+
+As aforementioned, this endpoint is `async`. We will get a response similar to:
```
dwilding, line 148: This could just be:
```suggestion
This endpoint is `async`, so we'll get a response similar to:
```

docs/how-to/use-pebble-api.md
```diff
@@ -0,0 +1,281 @@
...
+# wait for the change
+client.wait_change(changeID)
+```
+
+---
+
+## See more
+
+- [Go client](https://pkg.go.dev/github.com/canonical/pebble/client)
+- [Python client for Pebble API](https://ops.readthedocs.io/en/latest/reference/pebble.html).
```
dwilding, line 279: Removing extra period:
```suggestion
- [Python client for Pebble API](https://ops.readthedocs.io/en/latest/reference/pebble.html)
```

docs/how-to/index.md
```diff
@@ -37,3 +37,15 @@ Use named "identities" to allow additional users to access the API.
 
 Manage identities <manage-identities>
 ```
+
+
+## API
+
+To integrate Pebble with your automated workflows, you can use Pebble API.
```
dwilding, line 44: Minor nit:
```suggestion
To integrate Pebble with your automated workflows, you can use the Pebble API.
```

dwilding:
I've finished adding comments on specific parts. There are two main refactoring suggestions [here](https://github.com/canonical/pebble/pull/545#discussion_r1900633427) and [here](https://github.com/canonical/pebble/pull/545#discussion_r1900635960).

Copying the outstanding discussion points from earlier:
 - **Title** - For the long title, how about "How to use the Pebble API to manage services". And for the short title "Use the Pebble API". I like what you said about having an attractive title.
- **Order of examples** - Let's see what Ben thinks, but I also feel the order should be Python, Go, curl. (BTW, in the API reference doc it makes sense to put curl first.

benhoyt:
Note that the docs build is failing now.

IronCore864:
Refactored after our discussion. Changes:

- Reduced the introduction part.
- Removed Go and curl examples, and added a link to the API reference doc.
- Removed the wait change part in the Python example, and added a comment.
- Merged the latest master to pull in the API reference doc.

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,69 @@
...
+```{terminal}
+   :input: pebble run
+2024-12-30T01:07:10.275Z [pebble] Started daemon.
+2024-12-30T01:07:10.281Z [pebble] POST /v1/services 75.042µs 400
+2024-12-30T01:07:10.281Z [pebble] Cannot start default services: no default services
+```
+
+Here is an example in Python:
+
+```python
```
dwilding, line 27: I have a question about the Python (maybe I didn't understand it correctly): Is `client.wait_change(changeID)` needed at the end? If not, should we remove the `changeID` variable too (on lines 46 and 56), or would that variable be useful for other things?

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,69 @@
...
+Suppose we start the Pebble daemon with no default services and an empty layer:
+
+```{terminal}
+   :input: pebble run
+2024-12-30T01:07:10.275Z [pebble] Started daemon.
+2024-12-30T01:07:10.281Z [pebble] POST /v1/services 75.042µs 400
+2024-12-30T01:07:10.281Z [pebble] Cannot start default services: no default services
+```
+
+Here is an example in Python:
```
dwilding, line 25: I suggest making this a bit more descriptive of what follows. Maybe something like:

```suggestion
We can then interact with Pebble using a Python client:
```

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,69 @@
+# How to use the Pebble API to manage services
+
+Pebble provides a command-line interface (CLI) for managing services, similar to systemd. However, Pebble distinguishes itself with its dedicated REST API.
+
+This guide demonstrates how to use the Pebble API to programmatically manage services as part of an automated workflow where we can start, test, and stop interdependent services in a CI pipeline. While shell scripts could be used for this purpose, Pebble API offers a more robust approach with easier error handling and reusable components. This reduces manual work, errors, and ensures consistency across environments.
```
dwilding, line 5: Some small suggestions for reading flow:

```suggestion
This guide demonstrates how to use the Pebble API to programmatically manage services as part of an automated workflow that starts, tests, and stops interdependent services in a CI pipeline. While shell scripts could be used for this purpose, the Pebble API offers a more robust approach with easier error handling and reusable components. This reduces manual work, errors, and ensures consistency across environments.
```

docs/how-to/index.md
```diff
@@ -37,3 +37,15 @@ Use named "identities" to allow additional users to access the API.
...
+
+## API
+
+To integrate Pebble with your automated workflows, you can use the Pebble API.
+
+```{toctree}
+:titlesonly:
+:maxdepth: 1
+
+Use Pebble API <use-the-pebble-api>
```
dwilding, line 50: I think it's better to include "the" in the title:

```suggestion
Use the Pebble API <use-the-pebble-api>
```

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,69 @@
+# How to use the Pebble API to manage services
+
+Pebble provides a command-line interface (CLI) for managing services, similar to systemd. However, Pebble distinguishes itself with its dedicated REST API.
```
dwilding, line 3: Being extremely picky, we could remove "CLI" as that term isn't used anywhere else in the page. I'm actually good either way.

```suggestion
Pebble provides a command-line interface for managing services, similar to systemd. However, Pebble distinguishes itself with its dedicated REST API.
```

dwilding (approved):
Nice new doc, thanks! I've finished reviewing

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+This guide demonstrates how to use the Pebble API to programmatically manage services as part of an automated workflow where we can start, test, and stop interdependent services in a CI pipeline. While shell scripts could be used for this purpose, Pebble API offers a more robust approach with easier error handling and reusable components. This reduces manual work, errors, and ensures consistency across environments.
+
+## Using the API
```
dwilding, line 7: I think this should be "Use the API", for consistency with section headings in other how-to guides:

```suggestion
## Use the API
```

docs/how-to/use-the-pebble-api.md
IronCore864: This is an oversight, I will remove the wait change, add a comment instead, and also remove the variable.

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,66 @@
...
+This guide demonstrates how to use the Pebble API to programmatically manage services as part of an automated workflow that starts, tests, and stops interdependent services in a CI pipeline. While shell scripts could be used for this purpose, the Pebble API offers a more robust approach with easier error handling and reusable components. This reduces manual work, errors, and ensures consistency across environments.
+
+## Use the API
+
+```{include} /reuse/api.md
+   :start-after: Start: Pebble API overview
+   :end-before: End: Pebble API overview
+```
+
+For reference information about the API, see [](../explanation/api-and-clients) and [](../reference/api).
```
benhoyt, line 14: It's hard to tell in this context what the difference between these two links are. How about:

```suggestion
For an explanation of API access levels, see [](../explanation/api-and-clients). For the full API reference, see [](../reference/api).
```

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,66 @@
...
+assert len(services) == 0
+
+# add layer
+layerYAML = """
+services:
+  svc1:
+    override: replace
+    command: sleep 100
+"""
+client.add_layer(label="ci", layer=layerYAML, combine=True)
```
benhoyt, line 43: `label` and `layer` are required positional args, so you'd typically use them without keywords. How about:
```suggestion
client.add_layer("label1", layerYAML, combine=True)
```

docs/how-to/use-the-pebble-api.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+You can also use Go or curl to achieve the same result. For more information, see {ref}`api_go_client` and {ref}`api_curl`.
+
+## See more
+
+- [Go client](https://pkg.go.dev/github.com/canonical/pebble/client)
+- [Python client for Pebble API](https://ops.readthedocs.io/en/latest/reference/pebble.html)
+- [API and clients](../explanation/api-and-clients)
+- [API](../reference/api)
```
benhoyt, line 66: Again, it's hard to tell what distinguishes these last two links. Can we do something like:

- [Explanation of API access control](...)
- [API reference](...)

docs/reference/api.md
```diff
@@ -1,15 +1,17 @@
 # API
 
-Pebble exposes an HTTP API that remote clients can use to interact with the daemon. The API has endpoints for starting and stopping services, adding configuration layers to the plan, and so on.
-
-The API uses HTTP over a Unix socket, with access to the API controlled by user ID. If `pebble run` is started with the `--http <address>` option, Pebble exposes a limited set of open-access endpoints (see {ref}`api-access-levels`) using the given TCP address.
+```{include} /reuse/api.md
```
benhoyt, line 3: I don't think we should use re-use here. I think we should leave this detailed explanation in the api.md reference, and just have a one or two-sentence summary in the how-to, with a link to the API reference for more info. They're different kinds of docs, so I don't think we should re-use reference content directly in the how-to.

benhoyt (commented):
A couple of minor comments, and one comment about not trying to reuse content between reference and how-to. But this is looking very good now, thanks.

docs/how-to/use-the-pebble-api.md
dwilding: Side question - Are we keeping [API and clients](https://canonical-pebble.readthedocs-hosted.com/en/latest/explanation/api-and-clients/) as-is? I thought we had intended to move some (all?) of it into [API](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/api/)

benhoyt: Oooh, yes, I think that's part of the confusion. I think that's a good idea. I think most of the introduction in "API and clients" is repeated in the API reference anyway. And the identities stuff is over in References > Identities. So the only thing that actually needs to move IMO is the "API access levels" section.

Probably best to do that in a separate PR, though.

docs/how-to/use-the-pebble-api.md
```diff
@@ -6,12 +6,9 @@ This guide demonstrates how to use the Pebble API to programmatically manage ser
 
 ## Use the API
 
-```{include} /reuse/api.md
-   :start-after: Start: Pebble API overview
-   :end-before: End: Pebble API overview
-```
+Pebble's API allows remote client interaction with the daemon. It uses HTTP over a Unix socket, with access controlled by user ID.
```
benhoyt, line 9: Prefer a more active voice:

```suggestion
Pebble's API allows clients to interact remotely with the daemon. It uses HTTP over a Unix socket, with access controlled by user ID.
```

benhoyt (approved):
One nit suggestion, but this looks good to me -- approved!

PR #549, opened 2025-01-08
refactor: update client code+tests to use Requester interface

IronCore864:
Refactor client test framework to use the requester interface.

Closes https://github.com/canonical/pebble/issues/315.

See more information here in https://github.com/canonical/pebble/pull/310.

### DIFF
--- a/client/changes.go
+++ b/client/changes.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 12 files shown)


IronCore864:
> I noticed there are some different usages like [`client.Requester().Do`](https://github.com/search?q=repo%3Acanonical%2Fpebble%20client.Requester().Do&type=code) and [`client.doSync`](https://github.com/search?q=repo%3Acanonical%2Fpebble%20client.doSync&type=code). Should we update them all to `client.doSync` to make them consistent? Or the other way around?

According to the discussion in the daily sync, we decided to use `client.Requester().Do`.

benhoyt:
> According to the discussion in the daily sync, we decided to use client.Requester().Do.

Yep, thanks. It is a bit more verbose, but it's also more consistent and easier to read, because of the field names instead of calling functions that take 6 positional arguments.

benhoyt (approved):
All looks good, thanks!

flotter (approved):
Removing the doAsync and doSync private helpers aligns it with how we use it in derivative projects, as we do not get access to these from outside. Just a pity the same code is copied in so many places.

PR #555, opened 2025-01-10
docs: update api spec according to suggestions from ben

### LINKED ISSUE #554: Suggestions for API reference doc
The new [API reference page](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/api/) is really good, thanks @IronCore864 for your efforts here. Here are a bunch of suggested minor improvements after a closer look:

- [x] On reflection, it would be great to have the curl, Go, and Python examples all do the same thing (and state as much under "Using the API"). Per discussion, let's use the Go client example: stop a single service "mysvc". So let's change the Python example to use `client.stop_services(['mysvc'])`.
- [x] The curl equivalent would be:

```
$ curl --unix-socket /path/to/.pebble.socket \
       --data '{"action":"stop","services":["mysvc"]}' \
       http://_/v1/services
{"type":"async","status-code":202,"status":"Accepted","change":"42","result":null}
$ curl --unix-socket /path/to/.pebble.socket \
       http://_/v1/changes/42/wait
{"type":"sync","status-code":200,"status":"OK",...}
```

- [x] Remove the unnecessary indentation in the Go example.
- [x] Let's make the socket path `/path/to/.pebble.socket` like the Python one in all cases:
- [x] I'm pretty sure RFC 3339 times should have *either* a `Z` or a time zone offset, not both. Two of the examples have both.
- [x] Typo "without providing a the service name" -> "without providing a service name".
- [x] Is "Work-in-progress" still necessary? It looks like they're all there now, which is great.
- [x] The order of the categories and URLs within a category is unclear to me. Can we order the categories alphabetically, and the URLs within a category alphabetically?
- [x] The websocket-task endpoint should probably be under "changes", and that category renamed to "changes and tasks". (I realise it's only used for exec right now.)
- [x] I think the Manage services description "Services operations, like start/stop/restart a service/services, auto-start default services, and replan." is a bit unwieldy. How about "Perform a service operation such as start, stop, or replan. " (Autostart is kind of deprecated, and we don't need to list all of them here -- that's in the enum.)
- [x] I don't think "format: int32" is helpful on /v1/logs `n`. In fact, I think we just drop the "format: int32" everywhere.
- [x] The description of /v1/health needs to say how the "healthy: true/false" value is calculated. Maybe just a one-sentence summary and link to https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/health-checks/#health-endpoint?
- [x] On "Health check level", would be good to add "If omitted, aggregate healthy status of checks with any (or no) level.". Similar for /v1/checks "level".
- [x] For GET /v1/files, it would be good to provide an example of a full raw multipart response for the "read" operation. Is that possible, probably in the description? You can see some examples in ops [test_pebble.py](https://github.com/canonical/operator/blob/cece866bada2d9ada7a240846123a1c67faf48ee/test/test_pebble.py#L2239).
- [x] For POST /v1/files action "write", could we have a raw multipart example too?
- [x] For POST /v1/exec, would be good to say in the description that you then need to use /v1/tasks/{task-id}/control and /v1/tasks/{task-id}/stdio (and /stderr if split-stderr true), where the task-id comes from, and so on.
- [x] Under /v1/tasks/... "websocket-id", could list possible enum values: "control", "stdio", "stderr".
- [x] We could fully document the websocket protocol, but IMO we don't need to do that now -- realistically no one will need it at that level. I suggest we just cheat and say something like "The full websocket protocol is not documented. For details, see the [Python client code](https://github.com/canonical/operator/blob/cece866bada2d9ada7a240846123a1c67faf48ee/ops/pebble.py#L1818)."
- [x] notice.user_id should be notice.user-id
- [x] For most parameters we prefer "To specify multiple X, specify X multiple times." However, for GET /v1/notices "types" and "keys" we say "Multiple types can be specified as a comma-separated list." These can be specified multiple times too, which IMO is preferable, and good to be consistent. Can we change to "To specify multiple types, include this parameter multiple times." (or similar)
- [x] Can we link to Data Formats > Time in the GET /v1/notices "after" description?
- [x] For the description under POST /v1/identities, could link to https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/identities/ for the format of the identities data?
- [x] For GET /v1/system-info "boot-id"'s description, I think we should just say what's in the Go client doc comment: "A unique string that represents this boot of the server."

I realise that's a lot of items, but they should all be pretty minor!

CC: @dwilding

IronCore864:
Pebble HTTP API spec refactor.

Closes #554.

Preview: https://canonical-pebble--555.com.readthedocs.build/en/555/reference/api/

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -15,15 +15,15 @@ For more examples, see "How to use Pebble API". <!-- [David] Link to the how-to 
 To access the API endpoints over the Unix socket, use the `--unix-socket` option of `curl`. For example:
 
 ```{terminal}
-   :input: curl --unix-socket /path/to/.pebble.socket -XPOST http://localhost/v1/services -d '{"action": "stop", "services": ["svc1"]}'
+   :input: curl --unix-socket /path/to/.pebble.socket http://_/v1/services --data '{"action": "stop", "services": ["svc1"]}'
 {"type":"async","status-code":202,"status":"Accepted","change":"42","result":null}
 ```
 
 <br />
 
 ```{terminal}
-   :input: curl --unix-socket /path/to/.pebble.socket -XGET localhost:4000/v1/changes/42/wait
-{"type":"sync","status-code":200,"status":"OK","result":{"id":"42","kind":"stop","summary":"Stop service \"svc1\"","status":"Done","tasks":[{"id":"42","kind":"stop","summary":"Stop service \"svc1\"","status":"Done","progress":{"label":"","done":1,"total":1},"spawn-time":"2025-01-10T10:59:53.903366896+08:00","ready-time":"2025-01-10T10:59:53.930409771+08:00"}],"ready":true,"spawn-time":"2025-01-10T10:59:53.903373188+08:00","ready-time":"2025-01-10T10:59:53.930410979+08:00"}}
+   :input: curl --unix-socket /path/to/.pebble.socket http://_/v1/changes/42/wait
+{"type":"sync","status-code":200,"status":"OK","result":{...}}
 ```
 
 ### Go client
@@ -56,8 +56,7 @@ The Ops library for writing and testing Juju charms includes a [Python client fo
 ```python
 import ops
 client = ops.pebble.Client("/path/to/.pebble.socket")
-changeID = client.stop_services(["mysvc"])
-client.wait_change(changeID)
+client.stop_services(["mysvc"])  # Python client also waits for change to finish
 ```
 
 For more information, see:
@@ -110,7 +109,7 @@ Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 A timestamp is a string in the [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format with sub-second precision. Here are some examples:
 
 - "1985-04-12T23:20:50.52Z"
-- "1996-12-19T16:39:57-08:00"
+- "1996-12-19T16:39:57.123456789-08:00"
 
 ## Errors
 
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -185,7 +185,8 @@ window.onload = function() {
       // Create an 'a' element for the tag link
       const tocLink = document.createElement("a");
       tocLink.classList.add("reference", "internal");
-      tocLink.href= `#/${tag}`;
+      const urlFriendlyTag = tag.replace(/ /g, "-");
+      tocLink.href = `#/${urlFriendlyTag}`;
       tocLink.innerText = tag;
       tocLink.addEventListener("click", event => {
         if (event.shiftKey || event.ctrlKey || event.altKey || event.metaKey) {
@@ -194,7 +195,8 @@ window.onload = function() {
         // When the tag link is clicked with no modifier keys:
         // - Scroll the tag section into view
         // - If the tag section is closed, open it (by simulating a click)
-        const swaggerHeading = document.getElementById(`operations-tag-${tag}`);
+        const operationsTag = tag.replace(/ /g, "_");
+        const swaggerHeading = document.getElementById(`operations-tag-${operationsTag}`);
         swaggerHeading.scrollIntoView({
           behavior: "smooth"
         });
@@ -211,13 +213,13 @@ window.onload = function() {
 
   // Make sure to match the tags defined in openapi.yaml
   addSwaggerTagsToTOC([
-    "plan",
-    "services",
+    "changes and tasks",
     "checks",
     "files",
-    "changes",
-    "notices",
     "identities",
+    "notices",
+    "plan",
+    "services",
     "system"
   ]);
 }
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -14,35 +14,35 @@ For more examples, see "How to use Pebble API". <!-- [David] Link to the how-to 
 
 To access the API endpoints over the Unix socket, use the `--unix-socket` option of `curl`. For example:
 
-* TODO
+```{terminal}
+   :input: curl --unix-socket /path/to/.pebble.socket -XPOST http://localhost/v1/services -d '{"action": "stop", "services": ["svc1"]}'
+{"type":"async","status-code":202,"status":"Accepted","change":"42","result":null}
+```
 
-    ```
-    curl ...
-    ```
+<br />
 
-* TODO
-
-    ```
-    curl ...
-    ```
+```{terminal}
+   :input: curl --unix-socket /path/to/.pebble.socket -XGET localhost:4000/v1/changes/42/wait
+{"type":"sync","status-code":200,"status":"OK","result":{"id":"42","kind":"stop","summary":"Stop service \"svc1\"","status":"Done","tasks":[{"id":"42","kind":"stop","summary":"Stop service \"svc1\"","status":"Done","progress":{"label":"","done":1,"total":1},"spawn-time":"2025-01-10T10:59:53.903366896+08:00","ready-time":"2025-01-10T10:59:53.930409771+08:00"}],"ready":true,"spawn-time":"2025-01-10T10:59:53.903373188+08:00","ready-time":"2025-01-10T10:59:53.930410979+08:00"}}
+```
 
 ### Go client
 
 To use the [Go client](https://pkg.go.dev/github.com/canonical/pebble/client) to access API endpoints over the Unix socket, first create a client using `New`, and then call the methods on the returned `Client` struct to interact with the API. For example, to stop a service named `mysvc`:
 
 ```go
-    pebble, err := client.New(&client.Config{Socket: ".pebble.socket"})
-    if err != nil {
-        log.Fatal(err)
-    }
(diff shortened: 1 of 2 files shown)


docs/reference/api.md
```diff
@@ -14,35 +14,35 @@ For more examples, see "How to use Pebble API". <!-- [David] Link to the how-to
 
 To access the API endpoints over the Unix socket, use the `--unix-socket` option of `curl`. For example:
 
-* TODO
-
-    ```
-    curl ...
-    ```
+```{terminal}
+   :input: curl --unix-socket /path/to/.pebble.socket -XPOST http://localhost/v1/services -d '{"action": "stop", "services": ["svc1"]}'
```
benhoyt, line 18: I think we should consistency use `--long` arguments here (`--request` instead of `-X` and `--data` instead of `-d`). Also, should we drop `-X/--request` given that POST is implied with `-d/--data`?

benhoyt, line 18: Let's also change the host from `localhost` to `_` to clearly show the host is ignore with `--unix-socket`.

docs/reference/api.md
```diff
@@ -14,35 +14,35 @@ For more examples, see "How to use Pebble API". <!-- [David] Link to the how-to
...
+```
 
-* TODO
+<br />
 
-    ```
-    curl ...
-    ```
+```{terminal}
+   :input: curl --unix-socket /path/to/.pebble.socket -XGET localhost:4000/v1/changes/42/wait
```
benhoyt, line 25: I think we should drop `-XGET` as it's the default. Also, let's use `http://_/...` for consistency.

docs/reference/api.md
```diff
@@ -14,35 +14,35 @@ For more examples, see "How to use Pebble API". <!-- [David] Link to the how-to
...
+{"type":"sync","status-code":200,"status":"OK","result":{"id":"42","kind":"stop","summary":"Stop service \"svc1\"","status":"Done","tasks":[{"id":"42","kind":"stop","summary":"Stop service \"svc1\"","status":"Done","progress":{"label":"","done":1,"total":1},"spawn-time":"2025-01-10T10:59:53.903366896+08:00","ready-time":"2025-01-10T10:59:53.930409771+08:00"}],"ready":true,"spawn-time":"2025-01-10T10:59:53.903373188+08:00","ready-time":"2025-01-10T10:59:53.930410979+08:00"}}
```
benhoyt, line 26: Do you think we need the whole response here? I think it's better to use `...`, like:

```
{"type":"sync","status-code":200,"status":"OK","result":{...}}
```

As the result isn't interesting or important here (when you're waiting for a known change).

docs/reference/api.md
```diff
@@ -55,8 +55,9 @@ The Ops library for writing and testing Juju charms includes a [Python client fo
 
 ```python
 import ops
-client = ops.pebble.Client('/path/to/.pebble.socket')
-client.start_services(['svc1', 'svc2'])
+client = ops.pebble.Client("/path/to/.pebble.socket")
+changeID = client.stop_services(["mysvc"])
+client.wait_change(changeID)
```
benhoyt, line 60: `wait_change` isn't needed here, as the Python client's `stop_services` (and similar) does it automatically. See [here](https://github.com/canonical/operator/blob/cece866bada2d9ada7a240846123a1c67faf48ee/ops/pebble.py#L2242). Let's remove it but mention that explicitly in a comment:
```suggestion
client.stop_services(["mysvc"])  # Python client also waits for change to finish
```

docs/reference/api.md
```diff
@@ -108,9 +109,8 @@ Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 
 A timestamp is a string in the [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format with sub-second precision. Here are some examples:
 
-- "2006-01-02T15:04:05Z07:00"
-- "2006-01-02T15:04:05.999999999Z07:00"
-- "2024-12-18T10:45:29.919907089+08:00"
+- "1985-04-12T23:20:50.52Z"
+- "1996-12-19T16:39:57-08:00"
```
benhoyt, line 113: Let's add back the nanoseconds, though, as these are included in responses:
```suggestion
- "1996-12-19T16:39:57.123456789-08:00"
```

benhoyt (changes requested):
These look good, thanks. A few minor changes suggested to the curl examples, and please finish the unchecked changes listed https://github.com/canonical/pebble/issues/554 per comments on that thread.

dwilding:
On the order of the categories (I'll refer to them as "tags" to match the OpenAPI terminology): Since we decided not to match the output of `pebble help`, I think we should make the tags more granular and align with the endpoint URLs as much as possible

The idea is to make it easier to locate a particular endpoint without guessing where to find it. When all the endpoints are closed, it's easy enough to scan up and down to review all the endpoints. But when some/all are open, they take up a lot of space, and we have to rely on the TOC

So here's what I'm thinking:

```
changes and tasks:
  - GET /v1/changes
  - GET /v1/changes/{id}
  - POST /v1/changes{id}
  - GET /v1/changes/{id}/wait
  - GET /v1/tasks/{task-id}/websocket/{websocket-id}

checks:
  - GET /v1/checks

exec:
  - POST /v1/exec

files:
  - GET /v1/files
  - POST /v1/files

health:
  - GET /v1/health

identities:
  - GET /v1/identities
  - POST /v1/identitis

layers:
  - POST /v1/layers

logs:
  - GET /v1/logs

notices:
  - GET /v1/notices
  - POST /v1/notices
  - GET /v1/notices/{id}

plan:
  - GET /v1/plan

services:
  - GET /v1/services
  - POST /v1/services

signals:
  - POST /v1/signals

system-info:
  - GET /v1/system-info
```

I kept "changes and tasks" as one tag as a special case, because of how often these terms are used together. Also note that I changed "system" to "system-info" to match the endpoint URL.

What do you think?

dwilding:
@IronCore864 I see this error box rendered underneath the "API endpoints" heading:

```
Errors

Resolver error at responses.200.content.multipart/form-data.schema.$ref
Could not resolve reference: JSON Pointer evaluation failed while evaluating token "PostFileResponse" against an ObjectElement

Resolver error at responses.200.content.multipart/form-data.schema.properties.response.$ref
Could not resolve reference: JSON Pointer evaluation failed while evaluating token "PostFileResponse" against an ObjectElement

Resolver error at responses.200.content.multipart/form-data.schema.properties.response.$ref
Could not resolve reference: JSON Pointer evaluation failed while evaluating token "PostFileResponse" against an ObjectElement
```

IronCore864:
> @IronCore864 I see this error box rendered underneath the "API endpoints" heading:
> 
> ```
> Errors
> 
> Resolver error at responses.200.content.multipart/form-data.schema.$ref
> Could not resolve reference: JSON Pointer evaluation failed while evaluating token "PostFileResponse" against an ObjectElement
> 
> Resolver error at responses.200.content.multipart/form-data.schema.properties.response.$ref
> Could not resolve reference: JSON Pointer evaluation failed while evaluating token "PostFileResponse" against an ObjectElement
> 
> Resolver error at responses.200.content.multipart/form-data.schema.properties.response.$ref
> Could not resolve reference: JSON Pointer evaluation failed while evaluating token "PostFileResponse" against an ObjectElement
> ```

Thanks, fixed.

benhoyt (approved):
Looks good, approving. Just one more nit: alphabetically, /v1/exec should come before /v1/files.

IronCore864:
Re-categorized and re-ordered.

dwilding (approved):

docs/specs/openapi.yaml
```diff
@@ -474,227 +506,111 @@ paths:
...
-              command: ["/bin/ls", "-l"]
+              oneOf:  # Use oneOf since only one action can be performed at a time.
+                - $ref: "#/components/schemas/PostFilesMakeDirsRequest"
+                - $ref: "#/components/schemas/PostFilesRemovePathsRequest"
+            description: JSON payload for "make-dirs" or "remove" actions.
       responses:
-        "202":
-          description: Command execution initiated.
+        "200":
+          description: Post files successfully. The result in the response is a JSON array of the file result object containing path and (optional) errors.
```
dwilding, line 520: Since this endpoint can do several different operations, I think a more generic description would be better - saying "post files successfully" suggests to me that a file has been sent, which won't necessarily be the case. How about saying this (matching another endpoint):
```suggestion
          description: Successful operation. The result in the response is a JSON array of the file result object containing path and (optional) errors.
```

docs/specs/openapi.yaml
```diff
@@ -3,127 +3,139 @@ info:
...
+        - name: id
+          in: path
+          required: true
+          description: ID of the change.
           schema:
             type: string
       responses:
         "200":
-          description: List of services
+          description: Get change successfully.
```
dwilding, line 92: For this one, I think it would be better to directly say what the response contains. How about:
```suggestion
          description: Information about the change.
```

IronCore864:
All comments were resolved, also did some formatting stuff about the descriptions to make them consistent according to David's suggestion. All are updated and reviewed, merging now.

PR #556, opened 2025-01-14
docs: add doc style tips to HACKING.md

dwilding:
This PR adds a few style tips to the Docs section of HACKING.md. I've tried to keep the list short and (hopefully!) memorable

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -228,15 +228,15 @@ To pull in the latest style and dependencies from the starter pack, clone the [C
 
 ### Writing a great doc
 
-- Use short sentences, ideally with one or two clauses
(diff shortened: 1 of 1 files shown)


IronCore864:
I agree with all the items. Just one comment: Maybe each bullet point in the list should end with a period? They are all full sentences after all; also to be consistent with other lists in the same doc.

IronCore864 (approved):

dwilding:
Good call @IronCore864, thanks! (I have a bit of a habit of leaving periods off...)

benhoyt (approved):
Excellent, thanks for this.

PR #557, opened 2025-01-14
poc: add run-check to execute a check immediately

IronCore864:
Some notes (will delete later):

- [x] use req.Context()
- [x] something on the check manager
- [x] RunCheck(ctx context.Context, checkName string) error
- [x] get the config from plan.checks
- [x] POST /v1/checks {"action": "run", "check": "chk1"}
- [x] subcommand run-check (stop-checks/start-checks)
- [x] parse plan manager to check manager `PlanManager.Plan() *Plan`

### DIFF
--- a/client/checks.go
+++ b/client/checks.go
@@ -104,15 +104,15 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
 	return checks, nil
 }
 
-type RunCheckPayload struct {
+type RunCheckOptions struct {
 	Check string
 }
 
-// RunCheck runs a specific health check immediately and return the status.
-func (client *Client) RunCheck(opts *RunCheckPayload) (string, error) {
+// RunCheck runs a specific health check immediately and returns the error.
+func (client *Client) RunCheck(opts *RunCheckOptions) error {
 	body, err := json.Marshal(opts)
 	if err != nil {
-		return "", fmt.Errorf("cannot marshal checks payload: %w", err)
+		return fmt.Errorf("cannot marshal checks payload: %w", err)
 	}
 
 	resp, err := client.Requester().Do(context.Background(), &RequestOptions{
@@ -122,9 +122,8 @@ func (client *Client) RunCheck(opts *RunCheckPayload) (string, error) {
 		Body:   bytes.NewBuffer(body),
 	})
 	if err != nil {
-		return "", err
+		return err
 	}
 	var status string
-	err = resp.DecodeResult(&status)
-	return status, err
+	return resp.DecodeResult(&status)
 }
--- a/client/checks.go
+++ b/client/checks.go
@@ -15,6 +15,10 @@
 package client
 
 import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
 	"net/url"
 )
 
@@ -89,4 +93,30 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
 		return nil, err
 	}
 	return checks, nil
+}
+
+type CheckPayload struct {
+	Action string
+	Check  string
+}
+
+// RunCheck runs a specific health check immediately and return the status.
+func (client *Client) RunCheck(opts *CheckPayload) (string, error) {
+	body, err := json.Marshal(opts)
+	if err != nil {
+		return "", fmt.Errorf("cannot marshal checks payload: %w", err)
+	}
+
+	resp, err := client.Requester().Do(context.Background(), &RequestOptions{
+		Type:   SyncRequest,
+		Method: "POST",
+		Path:   "/v1/checks",
+		Body:   bytes.NewBuffer(body),
+	})
+	if err != nil {
+		return "", err
+	}
+	var status string
+	err = resp.DecodeResult(&status)
+	return status, err
 }
--- a/client/checks.go
+++ b/client/checks.go
@@ -104,13 +104,12 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
 	return checks, nil
 }
 
-type CheckPayload struct {
-	Action string
-	Check  string
+type RunCheckPayload struct {
+	Check string
 }
 
 // RunCheck runs a specific health check immediately and return the status.
-func (client *Client) RunCheck(opts *CheckPayload) (string, error) {
+func (client *Client) RunCheck(opts *RunCheckPayload) (string, error) {
 	body, err := json.Marshal(opts)
 	if err != nil {
 		return "", fmt.Errorf("cannot marshal checks payload: %w", err)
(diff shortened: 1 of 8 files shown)


client/checks.go
```diff
@@ -100,3 +103,29 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
...
 }
+
+type CheckPayload struct {
+	Action string
+	Check  string
+}
+
+// RunCheck runs a specific health check immediately and return the status.
+func (client *Client) RunCheck(opts *CheckPayload) (string, error) {
+	body, err := json.Marshal(opts)
```
benhoyt, line 114: Have a look at how the other actions work. The `RunCheckOptions` struct should be separate from the `runCheckPayload` struct that's marshalled for the API.

internals/daemon/api.go
```diff
@@ -89,9 +89,11 @@ var API = []*Command{{
 	WriteAccess: AdminAccess{},
 	POST:        v1PostSignals,
 }, {
-	Path:       "/v1/checks",
-	ReadAccess: UserAccess{},
-	GET:        v1GetChecks,
+	Path:        "/v1/checks",
+	ReadAccess:  UserAccess{},
+	WriteAccess: UserAccess{},
```
benhoyt, line 94: Should be `AdminAccess`?

internals/daemon/api_checks.go
```diff
@@ -66,3 +67,34 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+		}
+
+		err := c.d.overlord.CheckManager().RunCheck(r.Context(), check)
+		if err != nil {
+			return SyncResponse(err)
+		}
+		return SyncResponse("check passed")
+
+	default:
+		return BadRequest("action %s not supported", payload.Action)
```
benhoyt, line 98: In other places we use `"invalid action %q"` -- let's just be careful to be consistent for things like this.

internals/daemon/api_checks.go
```diff
@@ -66,3 +67,34 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+	}
+	switch payload.Action {
+	case "run":
+		if payload.Check == "" {
+			return BadRequest("no check to %s provided", payload.Action)
+		}
+
+		check, ok := c.d.overlord.PlanManager().Plan().Checks[payload.Check]
+		if !ok {
+			return SyncResponse("check not found")
```
benhoyt, line 88: Should be an error response, see other similar code. `return NotFound("cannot find check with name %q", payload.Check)`

internals/daemon/api_checks.go
```diff
@@ -66,3 +67,34 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+		Check  string `json:"check"`
+	}
+	decoder := json.NewDecoder(r.Body)
+	if err := decoder.Decode(&payload); err != nil {
+		return BadRequest("cannot decode data from request body: %v", err)
+	}
+	switch payload.Action {
+	case "run":
+		if payload.Check == "" {
+			return BadRequest("no check to %s provided", payload.Action)
```
benhoyt, line 83: Let's again be consistent in our error messages with other code. Looks like it's usually like "must specify check name"

internals/daemon/api_checks.go
```diff
@@ -66,3 +67,34 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+		}
+
+		check, ok := c.d.overlord.PlanManager().Plan().Checks[payload.Check]
```
benhoyt, line 86: Coding style to make it easier to read: let's split the `.Plan()` call and the `.Checks[foo]` lookup onto two lines.

internals/daemon/api_checks.go
```diff
@@ -66,3 +67,34 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+		if !ok {
+			return SyncResponse("check not found")
+		}
+
+		err := c.d.overlord.CheckManager().RunCheck(r.Context(), check)
```
benhoyt, line 91: Again, let's split onto two lines so it's easier to see what's being called. Similar code in v1GetChecks:

```go
	checkMgr := c.d.overlord.CheckManager()
	checks, err := checkMgr.Checks()
	if err != nil {
		return InternalError("%v", err)
	}
```

internals/overlord/checkstate/handlers.go
```diff
@@ -184,3 +184,9 @@ func pluralise(n int, singular, plural string) string {
 	}
 	return fmt.Sprintf("%d %s", n, plural)
 }
+
+func (m *CheckManager) RunCheck(ctx context.Context, check *plan.Check) error {
+	logger.Debugf("Performing check %q", check.Name)
```
benhoyt, line 189: Probably don't need this log as the POST /v1/checks/run will already be recorded, and each checker prints a debug log too.

internals/daemon/api_checks.go
tonyandrewmeyer: Hmm, I used this wording too, because we already use it for services and changes:

```
tameyer@tam-canoncial-1:~/code/pebble$ rg 'is unsupported'
internals/daemon/api_changes.go
246:		return BadRequest("change action %q is unsupported", reqData.Action)

internals/daemon/api_checks.go
106:		return BadRequest("action %q is unsupported", payload.Action)

internals/daemon/api_services.go
185:		return BadRequest("action %q is unsupported", payload.Action)
```

internals/daemon/api_checks.go
tonyandrewmeyer: "must specify" does seem more common here, but services has it like the above:

```
		if len(payload.Services) == 0 {
			return BadRequest("no services to %s provided", payload.Action)
		}
```

internals/daemon/api_checks.go
benhoyt: Hmm, true -- thanks for the detailed info. I guess we should pick one. I'd vote for `invalid action %q` as it's a bit more common, and "invalid" seems more right than "unsupported" to me (that implies it might be supported on another system/version).

@IronCore864, can you please change the "unsupported" ones to `invalid action %q`?

internals/daemon/api_checks.go
benhoyt: Hmm, yeah, we should pick one. Looks like "must specify" is significantly more common:

```
api_exec.go:		return BadRequest("must specify command")
api_files.go:			return BadRequest("must specify one or more paths")
api_files.go:			return BadRequest(`must accept multipart/form-data`)
api_files.go:			return BadRequest("must specify path")
api_files.go:		return BadRequest("must specify one or more files")
api_services.go:			return BadRequest("no services to %s provided", payload.Action)
api_signals.go:		return BadRequest("must specify one or more services")
```

In addition, "no services to start provided" reads awkwardly to me -- "provided" seems like it's in the wrong place. Let's change that one to `must specify services for %s action`.

benhoyt:
Closing, per spec discussion last night. We've decided to change the approach significantly: add a new `pebble check <name>` command with a `--refresh` option.

PR #558, opened 2025-01-14
feat(workloads): add plumbing to support workloads

anpep:
This is a preliminary patch that adds support for workloads in plan, plan manager and service manager by passing the workloads from the daemon.New() constructor all the way down to servstate manager. This also requires splitting the plan.ReadDir() and (*Plan).Validate() methods, and passing the workloads as an argument to Validate(), though this is a temporary measure to get the workloads feature working before refactoring the plan and service manager code to make services a plan section, rather than an ad-hoc member of the Plan struct.

In the future, this patch needs to be refactored to store the workloads in a future service extension, when the services layer in the plan is refactored to be a plan section. This way, we can pass the workloads from a plan extension and keep the current plan, daemon, and overlord APIs intact.


internals/overlord/planstate/manager.go
```diff
@@ -249,7 +255,7 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 		LogTargets: combined.LogTargets,
 		Sections:   combined.Sections,
 	}
-	err = p.Validate()
+	err = p.Validate(m.workloads)
```
flotter, line 258: Just a note for reviewers: Once services and checks are moved over to use the plan extension system, the workload definition can be supplied by the extension itself, and this signature change can be reverted so that the plan library is unaware of workloads.

internals/overlord/servstate/handlers.go
```diff
@@ -346,9 +346,16 @@ func (s *serviceData) startInternal() error {
 	s.cmd.Dir = s.config.WorkingDir
 
 	// Start as another user if specified in plan.
-	uid, gid, err := osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
-	if err != nil {
-		return err
+	w := s.manager.workloads[s.config.Workload]
```
flotter, line 349: Should we deal the a failure lookup here, even though in theory validation already happened ?

internals/overlord/servstate/manager.go
```diff
@@ -13,8 +13,12 @@ import (
 	"github.com/canonical/pebble/internals/overlord/state"
 	"github.com/canonical/pebble/internals/plan"
 	"github.com/canonical/pebble/internals/servicelog"
+	"github.com/canonical/pebble/internals/workload"
 )
 
+// ServicesField is the top-level string key used in the Pebble plan.
+const ServicesField = "services"
+
```
flotter, line 21: Used any more ?

internals/overlord/servstate/manager.go
```diff
@@ -30,7 +34,8 @@ type ServiceManager struct {
 	randLock sync.Mutex
 	rand     *rand.Rand
 
-	logMgr LogManager
+	workloads map[string]workload.Workload
```
flotter, line 37: Since workloads are the mechanism which will evolve over time to provide service level access permissions and confinement features, for me it makes sense that the Service Manager owns the configuration.

internals/plan/plan.go
```diff
@@ -974,13 +977,25 @@ func (layer *Layer) Validate() error {
...
+				return &FormatError{
+					Message: fmt.Sprintf(`service %q cannot run in workload %q because workloads are not supported in %v`, name, service.Workload, cmd.DisplayName),
+				}
+			}
+			if _, ok := workloads[service.Workload]; !ok {
+				return &FormatError{
+					Message: fmt.Sprintf(`service %q cannot run in non-existing workload %q`, name, service.Workload),
+				}
+			}
+		}
```
flotter, line 998: Should we enforce the rule here that if workloads are defined, uid, user, gid and group properties of the service must be unset ?

internals/workload/workload.go
flotter, line 1: Perhaps, until this gets more flesh (i.e. app armor and friends), this file could simply live inside servstate. However, looking at the dependency list, until we remove workloads from the plan into the servstate extension, this will not work as servstate imports plan already, and we would end up with a circular import between the plan and servstate.

internals/workload/workload.go
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workload
+
+type Workload struct {
+	UID, GID    *int
```
paul-rodriguez, line 18: Seems to me those aren't optional, so I don't see why we'd use a `*int` here instead of `int`.

internals/overlord/servstate/handlers.go
```diff
@@ -346,9 +346,16 @@ func (s *serviceData) startInternal() error {
...
+	uid, gid := w.UID, w.GID
+	if (uid == nil) != (gid == nil) {
+		panic("both uid and gid must be provided by workload")
```
paul-rodriguez, line 352: This panic will be triggered if either `uid` or `gid` is `nil` but not both. The panic text doesn't match.

internals/overlord/servstate/handlers.go
```diff
@@ -346,9 +346,16 @@ func (s *serviceData) startInternal() error {
...
+	}
+	if uid == nil {
+		uid, gid, err = osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
```
paul-rodriguez, line 355: Why reassign `gid` here if we know we have an explicitly set value already?

internals/plan/plan.go
```diff
@@ -974,13 +977,25 @@ func (layer *Layer) Validate() error {
...
 	for name, service := range p.Services {
 		if service.Command == "" {
 			return &FormatError{
 				Message: fmt.Sprintf(`plan must define "command" for service %q`, name),
 			}
 		}
+		if service.Workload != "" {
+			if workloads == nil {
+				return &FormatError{
+					Message: fmt.Sprintf(`service %q cannot run in workload %q because workloads are not supported in %v`, name, service.Workload, cmd.DisplayName),
```
paul-rodriguez, line 990: Is this always an accurate message in this case? What if we just didn't specify any workload but are using a program that supports them?

benhoyt (commented):
I think this is worth further discussion. I can see how some wiring is necessary (the service manager needs to know the workload info when starting a service), but I wonder if there's a way to achieve it without quite so much plumbing.

Maybe we could have add a servstate.GetWorkloadFor(serviceName string) that we override for projects that need it.

Anyway, happy to discuss further on our call.

benhoyt:
Closing this as we're settling on a different approach (including workloads in the plan). Feel free to open a new PR at that point, or reopen this.

PR #559, opened 2025-01-16
docs: change Diataxis categories to the standard order

dwilding:
The Diataxis categories in Pebble are currently ordered: Tutorial, How-to guides, Explanation, Reference. The standard order should be Tutorial, How-to guides, _Reference, Explanation_.

The 2x2 grid of cards on the homepage actually has the correct layout (matching the [Diataxis map](https://diataxis.fr/_images/diataxis.png)):

```
Tutorial     How-to guides
Explanation  Reference
```

This PR:

- Reorders the Diataxis categories to Tutorial, How-to guides, Reference, Explanation.
- Uses `:reverse:` in the grid source to swap Reference and Explanation, instead of ordering them as Explanation > Reference in the grid source. We need to do it this way so that Reference stacks above Explanation if the browser window is narrow.

In short, we're matching how Ops is set up. See https://github.com/canonical/operator/blob/main/docs/index.md?plain=1#L54

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,17 +24,17 @@ Pebble is useful for developers who are building [Juju charms on Kubernetes](htt
 ````
 
 ````{grid} 1 1 2 2
-```{grid-item-card} [Explanation](explanation/index)
-**Discussion and clarification** of key topics
-- [General model](explanation/general-model.md)
-- [API and clients](explanation/api-and-clients.md)
-```
-
 ```{grid-item-card} [Reference](reference/index)
 **Technical information**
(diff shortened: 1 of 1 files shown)


dwilding:
@benhoyt, @IronCore864, I was speaking with Daniele yesterday and he pointed out that the ordering of Diataxis categories in Pebble needs to be adjusted. Please could you review this small PR when you get a chance? Thanks!

benhoyt (approved):

benhoyt:
Thanks for this. No need for further review -- merging.

dwilding:
Thanks, Ben!

PR #560, opened 2025-01-16
feat: add a startup field for checks, and start-checks and stop-checks actions

tonyandrewmeyer:
Add support for starting and stopping checks, in a very similar manner to starting and stopping services. This includes a new optional layer field for checks, `startup`, that determines whether or not the check will be started when added to the plan.

I've manually tested this outside of the edge cases (for example, I've added a plan with the three `startup` values and checked they behave as expected both then and in a replan, I've started and stopped those checks both when all in the 'wrong' state, all already in the desired state, and a mixture of those, I've tried to start/stop checks that don't exist, I've changed the period for started and stopped checks and made sure the behaviour is correct, and so forth). I've done the majority of this with a locally running Pebble and the CLI, and the core checks also with a locally running Pebble and the Python Pebble client with the draft support for this.

### Layer

The layer specification gains a new `startup` field for checks, which must be either `enabled` or `disabled`, and defaults to `enabled` for backwards compatibility.

The impact of this field is that when the plan is updated (including when Pebble first starts) or a layer is added, new checks where `startup` is set to `disabled` will not be started. If a check's configuration changes (for example, the timeout or period), and it is set to `startup: disabled` but is currently running, then it will be restarted as previously. Any stopped checks where `startup` is changed to `enabled` will also be started when the plan changes.

This field is implemented via the existing check manager `PlanChanged` function. That function is registered as a listener for when the plan changes, which includes new checks being added, and check fields (including `startup`) changing. The function is adjusted in this PR to only start new checks if `startup` is enabled.

### CLI

Minor change: since `replan` now impacts checks as well as services, it is moved in the help listing from the `services` section to the `plan` section.

Two new commands are added: `start-checks` and `stop-checks`. These take one or more check names as arguments, and start or stop the specified checks. Starting a running check or stopping a stopped check is a no-op. The command response informs the user which checks actually required stopping or starting.

The `checks` command output is altered to include a new "Startup" column, which is simply the value of the field from the combined plan for that check, and to show `inactive` in the "Status" column for any checks that have been stopped (these also get a `-` value for both "Change" and "Failures").

### API

The existing `/v1/checks` endpoint is extended to also support `POST`, requiring admin access. The structure for this endpoint is similar to posting to `/v1/services`, in that it requires a list of check names and an action to perform (currently either `start` or `stop`). The response is a `SyncResponse` that contains a payload that has a `changed` list of checks that did require starting/stopping. The API waits for the checks to start/stop before returning.

In addition, the `replan` command of the `/v1/services` endpoint now calls the `Replan` function in the check manager, which effectively does an "autostart" for checks.

[Spec - OP052](https://docs.google.com/document/d/1KpgFR0eaxkrwt_Vd0rTZzb0D57dQVtAv7t1b9iuMUQw/edit?tab=t.0) (internal link only, sorry).


tonyandrewmeyer:
@benhoyt as discussed, if you could give this a "pre-review", that would be great, thanks!

client/checks.go
```diff
@@ -100,3 +119,55 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
...
+	response, err = client.doMultiCheckAction("stop", opts.Names)
+	return response, err
+}
+
+type multiCheckActionData struct {
+	Action string   `json:"action"`
+	Checks []string `json:"checks"`
+}
+
+func (client *Client) doMultiCheckAction(actionName string, checks []string) (changeID string, err error) {
```
benhoyt, line 142: This might become moot, but:
```suggestion
func (client *Client) doMultiCheckAction(actionName string, checks []string) (response string, err error) {
```

client/checks.go
```diff
@@ -100,3 +119,55 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
...
+	action := multiCheckActionData{
+		Action: actionName,
+		Checks: checks,
+	}
+	data, err := json.Marshal(&action)
+	if err != nil {
+		return "", fmt.Errorf("cannot marshal multi-check action: %w", err)
+	}
+	headers := map[string]string{
+		"Content-Type": "application/json",
```
benhoyt, line 152: Oddly, we don't normally provide the content type for requests (even though it's almost always JSON, except for files, some of which are multipart). We do in `POST /v1/services` though, so I don't mind specifying it here for consistency with that.

I do wonder if we should get the client to add it by default (unless it's overridden in `Headers`). Would that cause any issues?

I realise this is not related to this PR.

client/checks.go
```diff
@@ -100,3 +119,55 @@ func (client *Client) Checks(opts *ChecksOptions) ([]*CheckInfo, error) {
...
+	}
+
+	resp, err := client.Requester().Do(context.Background(), &RequestOptions{
+		Type:    SyncRequest,
+		Method:  "POST",
+		Path:    "/v1/checks",
+		Query:   nil,
```
benhoyt, line 159: It's typical and expected to omit zero values like nil (kinda like default keyword args in Python).

internals/daemon/api_checks.go
```diff
@@ -66,3 +70,49 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+	if err != nil {
+		return BadRequest("cannot %s checks: %v", payload.Action, err)
+	}
+
+	st := c.d.overlord.State()
+	st.EnsureBefore(0) // start and stop tasks right away
+
+	var result string
+	if len(checks) == 0 {
+		result = fmt.Sprintf("No checks needed to %s", payload.Action)
```
benhoyt, line 111: I know this is somewhat parallel to what services does with the change summary, but there we're limited by the structure of changes. Here, however, it seems like putting display logic or formatting in the API. I think it'd be better to return the slice of check names started/stopped directly, and let the client do what it wants to with them.

However, rather than return a `[]string` directly, let's be a bit more future-proof and return a `*ChecksActionResult` with a `Checks []string` field.

On the client side I'd probably simplify (drop mention of "queued") and just say something like:

- "Checks already started." (or stopped)
- "Checks started: foo, bar, baz."

internals/daemon/api_checks.go
```diff
@@ -66,3 +70,49 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+	if err := decoder.Decode(&payload); err != nil {
+		return BadRequest("cannot decode data from request body: %v", err)
+	}
+
+	if len(payload.Checks) == 0 {
+		return BadRequest("must specify checks for %s action", payload.Action)
+	}
+
+	var err error
+	var checks []string
```
benhoyt, line 90: Nit: I'd probably move these to just above the `switch` block, as they're not needed before that.

internals/overlord/checkstate/manager.go
```diff
@@ -349,10 +375,75 @@ type CheckInfo struct {
...
+	CheckStatusInactive CheckStatus = "inactive"
 )
 
 type checker interface {
 	check(ctx context.Context) error
 }
+
+// StartChecks starts the specified checks by creating a new perform-checks
+// change for each check that is in the current plan and not already running.
+func (m *CheckManager) StartChecks(currentPlan *plan.Plan, checks []string) ([]string, error) {
```
benhoyt, line 389: I think I'd prefer that these methods fetched the plan via the manager, rather than the caller having to pass in the plan. Can we drop it from the signatures here, and instead have `NewManager` take a reference to and store the `*PlanManager`? That seems reasonable, as the check manager now needs to know about the plan for replanning purposes.

internals/overlord/checkstate/manager.go
```diff
@@ -349,10 +375,75 @@ type CheckInfo struct {
...
+	CheckStatusDown     CheckStatus = "down"
+	CheckStatusInactive CheckStatus = "inactive"
 )
 
 type checker interface {
 	check(ctx context.Context) error
 }
+
+// StartChecks starts the specified checks by creating a new perform-checks
+// change for each check that is in the current plan and not already running.
```
benhoyt, line 388: I think we should probably drop the "by creating a new perform-checks change" (seems too much like implementation details), and add what the return value represents.

internals/overlord/checkstate/manager.go
```diff
@@ -349,10 +375,75 @@ type CheckInfo struct {
...
+		}
+	}
+	var started []string
+	for _, name := range checks {
+		check := currentPlan.Checks[name] // We know this is ok because we checked it above.
+		info, ok := m.checks[name]
+		if !ok {
+			panic(fmt.Sprintf("check %s is in the plan but not known to the manager", name))
+		}
+		// If the check is already running, skip it.
```
benhoyt, line 406: Let's be sure to consider what happens in the race conditions: 1) if the plan changes between fetching it and using it here, and 2) if `m.checks` changes between fetching it and creating starting it (creating the perform-check change).

It doesn't matter *too* much IMO, as long as it doesn't crash or do something weird.

In addition, I think there's a data race here: `m.checks` needs a mutex around access, in case another goroutine modifies it. There's the existing `checksLock` for that, though I haven't throught through what the scope of the lock/unlock should be here (perhaps around the whole fetch / update operation to avoid that race?).

internals/overlord/checkstate/manager.go
```diff
@@ -349,10 +375,75 @@ type CheckInfo struct {
...
+		if _, ok := currentPlan.Checks[name]; !ok {
+			return nil, fmt.Errorf("cannot find check %q in plan", name)
+		}
+	}
+	var started []string
+	for _, name := range checks {
+		check := currentPlan.Checks[name] // We know this is ok because we checked it above.
+		info, ok := m.checks[name]
+		if !ok {
+			panic(fmt.Sprintf("check %s is in the plan but not known to the manager", name))
```
benhoyt, line 404: I suspect this can happen if the plan was just updated? Or can that not happen because of how the updates and the plan-changed listener is called inside a mutex? I'm not sure off the top of my head, just want to make sure we've thought through it.

internals/daemon/api_services.go
```diff
@@ -95,10 +95,35 @@ func v1PostServices(c *Command, r *http.Request, _ *UserState) Response {
 		payload.Services = services
 	default:
 		if len(payload.Services) == 0 {
-			return BadRequest("no services to %s provided", payload.Action)
+			return BadRequest("must specify services for %s action", payload.Action)
 		}
 	}
 
+	// TODO (ping @benhoyt) This is clearly the wrong place for this - it's out
```
benhoyt, line 102: Hmm, yes, this is a tricky one. I don't have the perfect solution off the top of my head, and it's close to EOD, so let's pair tomorrow and discuss an approach.

benhoyt (changes requested):

client/checks.go
```diff
@@ -15,19 +15,23 @@
...
+	"encoding/json"
+	"fmt"
 	"net/url"
 )
 
 type ChecksOptions struct {
 	// Level is the check level to query for. A check is included in the
 	// results if this field is not set, or if it is equal to the check's
-	// level.
+	// level. This field is ignored for start and stop actions.
```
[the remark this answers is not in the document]
tonyandrewmeyer: Sure. I went through a bunch of iterations with this - first like it is now, then I didn't like that it had the level and didn't use it, then added in filtering by level so it did, then didn't think that really added value, so split it off to a separate one as you suggest, then decided that it seemed overkill to have a separate one with just one field, so went back to where I started :)

client/checks.go
tonyandrewmeyer: Open an issue for someone to make this change in the future. For now, since this isn't technically required and in the future it'll be the default, leave it out here.

internals/overlord/checkstate/manager.go
tonyandrewmeyer: Think this through in detail to be sure, but e.g. it might be that instead of the panicking it should be more defensive instead (like continuing and logging what happened instead).

internals/daemon/api_services.go
tonyandrewmeyer: Discussed this and decided that api_services can directly call a Replan function on the check manager - we will keep it simple for now, and if it gets complicated later we could add a replan listener type system then. The API function already has a state lock, so the Replan method needs to not wait for that - this can be mentioned in the documentation for the Replan function (other functions already do this).

client/checks.go
tonyandrewmeyer: https://github.com/canonical/pebble/issues/561

client/checks_test.go
```diff
@@ -47,11 +49,63 @@ func (cs *clientSuite) TestChecksGet(c *check.C) {
...
 		"level": {"alive"},
-		"names": {"chk1", "chk3"},
+		"names": {"chk1", "chk3", "chk5"},
 	})
 }
+
+func (cs *clientSuite) TestStartChecks(c *check.C) {
+	cs.rsp = `{
+	    "result": {"changed": ["chk1", "chk2"]},
+		"status": "OK",
```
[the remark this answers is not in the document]
tonyandrewmeyer: Mix of tabs & spaces so wasn't immediately visible in my IDE. Thanks!

internals/daemon/api_checks.go
```diff
@@ -66,3 +69,47 @@ func v1GetChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+	switch payload.Action {
+	case "start":
+		changed, err = checkmgr.StartChecks(payload.Checks)
+	case "stop":
+		changed, err = checkmgr.StopChecks(payload.Checks)
+	default:
+		return BadRequest("invalid action %q", payload.Action)
+	}
+	if err != nil {
+		return BadRequest("cannot %s checks: %v", payload.Action, err)
```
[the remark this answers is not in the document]
tonyandrewmeyer: I've added a `CheckNotFound` specific error. The naming ("Check" vs. "Checks") makes me wonder: at the moment (both before and after it was a specific type) it gives up as soon as it finds any check that isn't in the plan. That's fine if there's only one problem, but if you (for example) do `start-checks chk1 chk2 chk3` and `chk2` *and* `chk3` aren't in the plan, then you only get told about `chk2`, until you (presumably) fix that and re-run to get the `chk3` error.

@benhoyt I could change it so that it finds all the missing ones instead - it would take (a presumably unnoticeable amount of time) longer, but you'd get all errors of this type at once. Would that be the pattern we'd usually do here?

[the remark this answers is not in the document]
benhoyt: Yeah, it's always better to report all such errors in one hit. I've done that for example in the identities code, where it loops through all the identities and keeps a list of existing ones and reports a single error at the end: https://github.com/canonical/pebble/blob/7bd950b5f67f800703faaaf4455334bed1b2d2f1/internals/overlord/state/identities.go#L138-L141

However, you'll notice we're inconsistent in Pebble, and a lot of the time we just do the simple thing and return a single error early. Even in this identities code, I'm returning any "identity invalid" errors singly (just a few lines above!).

However, it seems like an easy change to make here, so we might as well. How about `ChecksNotFound` with a `[]string` and the `Error()` method uses `strings.Join`.

internals/overlord/checkstate/manager_test.go
```diff
@@ -584,3 +612,259 @@ func changeData(c *C, st *state.State, changeID string) map[string]string {
...
+	err := s.planMgr.AppendLayer(origLayer, false)
+	c.Assert(err, IsNil)
+	waitChecks(c, s.manager, []*checkstate.CheckInfo{
+		{Name: "chk1", Startup: "enabled", Status: "up", Threshold: 3},
+	})
+
+	changed, err := s.manager.StartChecks([]string{"chk1", "chk2"})
+	_, ok := err.(*checkstate.ChecksNotFound)
+	c.Assert(ok, Equals, true)
+	c.Assert(err.(*checkstate.ChecksNotFound).Names, DeepEquals, []string{"chk2"})
```
benhoyt, line 704: Nit: you can avoid the second type assertion by assigning the value here:

```suggestion
	notFoundErr, ok := err.(*checkstate.ChecksNotFound)
	c.Assert(ok, Equals, true)
	c.Assert(notFoundErr.Names, DeepEquals, []string{"chk2"})
```

benhoyt: In fact, it's probably better to use [`errors.As`](https://pkg.go.dev/errors#As) these days, which also detects errors which "wrap" `ChecksNotFound`:

```go
var notFoundErr *checkstate.ChecksNotFound
c.Assert(errors.As(err, &notFoundErr), Equals, true)
c.Assert(notFoundErr.Names, DeepEquals, []string{"chk2"})
```

benhoyt (approved):
Nice! The new tests look good -- thanks. One nit / Go style comment, but I'm happy with this.

PR #563, opened 2025-01-23
feat: add a new basic type identity

IronCore864:
A new "basic" type of identity and "metrics" type of access are added for the upcoming metrics feature.

To expose the upcoming metrics feature over HTTP, we need a certain level of authentication to protect the endpoint. We decided to use HTTP basic authentication for this purpose. A new type of "basic" identity needs to be implemented for this to work, and the access level would be `metrics`. The basic identity looks like this:

```yaml
identities:
  bob:
    access: metrics
    basic:
      password: hashed-password-with-salt
```

Where the password is generated by `openssl passwd -6`. The hashed password will be stored in the state, and when the user accesses the metrics endpoint, they need to set the `Authorization` header accordingly. Pebble daemon will sha512 hash the password and compare it to the identity stored in the state.

**To test this manually**, make the following code change to add a fake `/v1/metrics` endpoint before running `pebble run`:

```diff
diff --git a/internals/daemon/api.go b/internals/daemon/api.go
index 9d5199d..ad5e4ae 100644
--- a/internals/daemon/api.go
+++ b/internals/daemon/api.go
@@ -110,6 +110,10 @@ var API = []*Command{{
        WriteAccess: AdminAccess{},
        GET:         v1GetIdentities,
        POST:        v1PostIdentities,
+}, {
+       Path:       "/v1/metrics",
+       ReadAccess: MetricsAccess{},
+       GET:        v1GetMetrics,
 }}

 var (
@@ -131,3 +135,9 @@ func v1SystemInfo(c *Command, r *http.Request, _ *UserState) Response {
        }
        return SyncResponse(result)
 }
+
+func v1GetMetrics(c *Command, r *http.Request, _ *UserState) Response {
+       return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+               writer.Write([]byte(`pebble_service_start_count{service="test1"} 1` + "\n"))
+       })
+}
```

Then use `openssl passwd -6` to generate a password, and create an identities file with an identity to add, for example `identities.yaml`:

```
identities:
    web:
        access: metrics
        basic:
            # "foobar"
            password: $6$ggLo8P2qHXnogNBh$NUANaA3QilhGwrDjJR9U6HKP1mSsEP1.YBaWF4m3YNhyZloNjo/Q1IY84d7qiGz/TBs2DYC8CWa4mtVJuWAw40
```

Finally, add the new "basic" identity:

```
$ pebble add-identities --from identities.yaml
```

Finally, run `pebble run --http=:4000` in one terminal, then hit the metrics endpoint via curl in another:

```
$ curl -u web:foobar http://localhost:4000/v1/metrics
pebble_service_start_count{service="test1"} 1
$ curl -u web:wrongpass http://localhost:4000/v1/metrics
```

For more details, see [the spec here](https://docs.google.com/document/d/1jezGM7saLaBrOoegG7d4r0YTfypPUpCFUE2pJdKJQOM/edit?tab=t.0#heading=h.lwbdfpio7s9z).


benhoyt:
Let's please add a few more details about the feature to the PR description, and link to the spec (internal, I know).

benhoyt:
Just for the record, I'm pasting my little bit of research into the available sha-crypt libraries I could find. I think GehirnInc/crypt is the right choice.

```
https://github.com/GehirnInc/crypt
- 76 stars
- project last updated 2y ago
- file last updated 7y ago
- file last updated TODO
- originally by Jeramey Crawford
- https://github.com/GehirnInc/crypt/blob/master/sha512_crypt/sha512_crypt.go

https://github.com/tredoe/osutil
- 76 stars
- project last updated 6mo ago
- file last updated 9y ago
- originally by Jeramey Crawford, somewhat similar to GehirnInc/crypt impl
- https://github.com/tredoe/osutil/blob/master/user/crypt/sha512_crypt/sha512_crypt.go

https://github.com/ncw/pwhash
- 5 stars
- project last updated 10y ago
- file last updated 10y ago
- originally by Jeramey Crawford, somewhat similar to GehirnInc/crypt impl
https://github.com/ncw/pwhash/blob/master/sha512_crypt/sha512_crypt.go

https://github.com/go-crypt/crypt
- 16 stars
- project last updated 5 days ago
- dir last updated 2y ago (but impl is in separate xcrypt repo)
- a *lot* more API surface and complexity of implementation
- actual crypt impl is in https://github.com/go-crypt/x: 1 star, 6mo ago
- https://github.com/go-crypt/crypt/blob/master/algorithm/shacrypt/hasher.go
- https://github.com/go-crypt/x/blob/master/crypt/crypt.go
```

internals/overlord/state/identities.go
```diff
@@ -275,6 +317,12 @@ func (s *State) IdentityFromInputs(userID *uint32) *Identity {
 			if identity.Local.UserID == *userID {
 				return identity
 			}
+		case identity.Basic != nil && username != "" && identity.Name == username && password != "":
```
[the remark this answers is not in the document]
benhoyt: Hmmm, actually we need to think about this logic some more. Even now, we kind of want the highest access level allowed to be returned... or maybe we need to return a slice of matching identities.

But then what would `userFromRequest` do with that? It wants an access level and the associated user-id (or username), not a mix of both. Hmm, will have to think more later.

internals/daemon/daemon.go
```diff
@@ -213,7 +218,8 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// not good: https://github.com/canonical/pebble/pull/369
 	var user *UserState
 	if _, isOpen := access.(OpenAccess); !isOpen {
-		user, err = userFromRequest(c.d.state, r, ucred)
+		basicAuthUsername, basicAuthPassword, _ := r.BasicAuth()
```
benhoyt, line 221: Given the context / scope of these variables, just `username` and `password` would be fine:
```suggestion
		username, password, _ := r.BasicAuth()
```

internals/daemon/access.go
```diff
@@ -68,3 +68,18 @@ func (ac UserAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Re
...
+
+// MetricsAccess allows requests over the HTTP from authenticated users.
+type MetricsAccess struct{}
+
+func (ac MetricsAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	if user == nil {
+		return Unauthorized(accessDenied)
+	}
+	switch user.Access {
+	case state.MetricsAccess, state.AdminAccess:
```
benhoyt, line 80: Should `ReadAccess` be allowed here too?
```suggestion
	case state.MetricsAccess, state.ReadAccess, state.AdminAccess:
```

internals/overlord/state/identities.go
```diff
@@ -47,27 +52,46 @@ type LocalIdentity struct {
...
-			d.Access, AdminAccess, ReadAccess, UntrustedAccess)
+		return fmt.Errorf("invalid access value %q, must be %q, %q, %q, or %q",
+			d.Access, AdminAccess, ReadAccess, MetricsAccess, UntrustedAccess)
 	}
 
 	switch {
 	case d.Local != nil:
+		return nil
+	case d.Basic != nil:
+		if d.Basic.Password == "" {
```
benhoyt, line 88: Per our spec discussion, I believe that for now we only wanted to allow `access: metrics` (or I guess `untrusted`) with basic authentication, not read or admin.

benhoyt (changes requested):
Thanks for this. Some minor stuff, but a couple of things worth further discussion, particularly how `IdentityFromInputs` will need to change.

internals/overlord/state/identities.go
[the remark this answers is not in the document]
IronCore864: I don't have a good solution here.

First, I don't think `IdentityFromInputs` should return a slice. Even if an identity has two types, it's still one identity, not two separate ones. It's also stored as a single identity in the state, I verified this. So it can't return a slice, it has to return the whole identity with two types.

But this raises more questions, for example, what if there is only `userID` without username/password, what does it mean? Do we return the identity although the user doesn't specify user/pass? It's the user who added the identity in the first place, they know it has two types, and they only want to use userID to fetch it because they can't be bothered to enter the user/pass. Or, they mean only to get the access level from the basic type, since they didn't specify user/pass, implying not using the basic type even if it exists? Either way, we have to guess the user's intention.

Then, it would mean `userFromRequest` needs to figure out which access level to use. The higher one? Or check input, if user/pass isn't specified, use the access level from local. Again, we have to guess the user's intention.

For another example, what if both userID and user/pass exist? Does the user mean if any matches, use that type? Or does the user mean all must match? It's like when logging in to a website, you enter two sets of user/pass, do we allow them to log in even if one set of password is invalid?

And the problem would be bigger if we add yet another identity type in the future.

Maybe all issues would go away if only one type is allowed per identity because this neither would break backward compatibility (since we only have one type now) nor does it affect user experience since they can have two separate identities "bob-local" and "bob-basic" if they want. Or, maybe we can set a rule that basic type takes precedence (since it needs explicit setting)? In this way, there is no guessing the user's intention.

If we must support multiple identities, we might do something like this:

```go
func (s *State) IdentityFromInputs(userID *uint32, username, password string) *Identity {
	if userID != nil && username == "" && password == "" {
		// local
		for _, identity := range s.identities {
			if identity.Local != nil && identity.Local.UserID == *userID {
				return &identity
			}
		}
		return nil
	} else if userID == nil && username != "" && password != "" {
		// basic
		for _, identity := range s.identities {
			if identity.Basic != nil && identity.Basic.Password == password && identity.Name == username {
				// check password
				return &identity
			}
		}
		return nil
	} else if userID != nil && username != "" && password != "" {
		// both types
		for _, identity := range s.identities {
			// all matching or just one match is OK?
			if identity.Local != nil && identity.Local.UserID == *userID && identity.Basic != nil && identity.Basic.Password == password && identity.Name == username {
				// check password
				return &id
			}
		}
		return nil
	}

	return nil
}
```

Then figure out how to decide which access level to use in `userFromRequest`.

internals/overlord/state/identities.go
IronCore864: Is this comment intended for the `CheckAccess` method, the same as the previous comment?

internals/overlord/state/identities.go
```diff
@@ -47,27 +52,46 @@ type LocalIdentity struct {
...
+	Password string // Note: In a real application, store a password hash, not the plaintext password.
+}
+
 // validate checks that the identity is valid, returning an error if not.
 func (d *Identity) validate() error {
 	if d == nil {
 		return errors.New("identity must not be nil")
 	}
 
+	if d.Name != "" {
```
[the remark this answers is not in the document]
IronCore864: I thought about this, at first sight, `validateExcludingName` doesn't sound very nice, but I didn't come up with a better solution, because although `validate` and `validateName` sound better, they don't fit the use cases where unmarshaling JSON doesn't need to validate the names. So I will go with your suggestion.

IronCore864:
In the latest commit, I have fixed most comments except for the access level part and multiple identity type part. See above comment for more information.

internals/overlord/state/identities.go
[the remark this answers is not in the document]
benhoyt: Per our voice discussion, I think we have this problem whether we allow identity definitions to have multiple types or not, because a user could connect on the socket (UID-based auth) and using a basic auth `Authorization` header, which could match multiple named identities. Consider:

```
identities:
	alice:
		access: admin
		local:
			user-id: 1234

	bob:
		access: metrics
		basic:
			password: hash('pass1')

	charlie:
		access: read
		basic:
			password: hash('pass2')
		local:
			user-id: 4321

Connection                               Identities      Access
----------------------------------------------------------------
socket uid 1234                          alice           admin
tcp Auth bob:pass1                       bob             metrics
socket uid 4321                          charlie         read
socket uid 4321, tcp Auth charlie:pass2  charlie         read
socket uid 4321, tcp Auth charlie:wrong  charlie         read   # weird, but they have access via uid
socket uid 1234, tcp Auth charlie:pass2  alice, charlie  admin, read  # pick max(access) = alice/admin
socket uid 1234, tcp Auth charlie:wrong  alice           admin
```

So I think our algorithm should be:

- keep a list of matching identities
- loop through all identities
- if it matches request via any type (UID or basic-auth password), add it to the list
- at the end, if there's 1) no matching identities, no access (untrusted), if there's 2) one matching identity, return it, or if there's 3) multiple matching identities, return the one with the highest access, because they clearly have at least one set of credentials that let them in at that access level.

I *think* that's reasonable. Though I do want to get a couple of other pairs of eyes on this before we merge, as modifying our API access controls is sensitive stuff.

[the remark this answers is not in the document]
hpidcock: We shouldn't try to be clever here.

If an `Authorization` header has been supplied (as in, the header key is included, even if it is empty), then uid auth should be discarded.

The reason behind this is that an `Authorization` header is a per-request, client controlled property. The uid auth is a server controlled, connection level property. If the client specified an http `Authorization` header, it did that with intent.

If the client then provided an invalid `Authorization` header, simply return a `403 Forbidden`.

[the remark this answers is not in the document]
hpidcock: As with the duplicate UID or Username, the state code should prevent such ambiguity from occurring.

internals/overlord/state/identities.go
```diff
@@ -18,8 +18,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
+
+	"github.com/GehirnInc/crypt/sha512_crypt"
```
flotter, line 25: I there any reason why we use this library and not ```crypto/sha512``` for hashing the password ?

internals/overlord/state/identities.go
```diff
@@ -28,8 +31,9 @@ type Identity struct {
 	Access IdentityAccess
 
 	// One or more of the following type-specific configuration fields must be
-	// non-nil (currently the only type is "local").
+	// non-nil.
 	Local *LocalIdentity
+	Basic *BasicIdentity
```
flotter, line 36: This may have already been debated, but in the global scoped context of identities, would it not be a bit clearer that this refers to an HTTP authentication scheme specifically, and not a basic identity scheme in the generic sense? Or is that fact that this is simply a user / password mapping deliberately made available as a generic type identity? My first thoughts were ```BasicHTTP```, ```BasicHTTPIdentity``` ...

flotter (commented):
This is great. Looking forward to getting this in Pebble.

internals/overlord/state/identities.go
benhoyt: Yes, they're different things: `crypto/sha512` is the straight SHA-512 hash algorithm, but `GehirnInc/crypt/sha512_crypt` implements the [SHA-crypt](https://www.akkadia.org/drepper/SHA-crypt.txt) algorithm using SHA-512 hashes and a whole bunch of "rounds" and other stuff I don't understand :-) to make it good for use with passwords (slow and secure, basically).

`GehirnInc/crypt/sha512_crypt` [uses](https://github.com/GehirnInc/crypt/blob/8cc1b52080c5761aa65952b1c39dd889bf54d76c/sha512_crypt/sha512_crypt.go#L65) the stdlib's `crypt/sha512` in its implementation.

internals/overlord/state/identities.go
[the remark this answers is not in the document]
benhoyt: Yep, that's a good point, @hpidcock -- thanks. That's simpler and less error-prone. @IronCore864, let's go ahead with the simpler approach Harry suggested.

IronCore864:
Some manual tests:

### 1 Two Types of Identity

```yaml
identities:
    bob:
        access: metrics
        basic:
            password: $6$F9cFSVEKyO4gB1Wh$8S1BSKsNkF.jBAixGc4W7l80OpfCNk65LZBDHBng3NAmbcHuMj4RIm7992rrJ8YA.SJ0hvm.vGk2z483am4Ym1
        local:
            user-id: 1001
```

```bash
$ ./pebble add-identities --from ./identity.yaml
Added 1 new identity.
$ ./pebble identities
Name  Access   Types
bob   metrics  basic,local
```

### 2 No Password for Basic Type

```yaml
identities:
    nancy:
        access: metrics
        basic:
            password: 
```

```bassh
$ ./pebble add-identities --from ./identity.yaml
error: cannot decode request body: basic identity must specify password (hashed)
```

### 3 Invalid Access Level for Basic Type

```yaml
identities:
    bob:
        access: admin
        basic:
            password: $6$F9cFSVEKyO4gB1Wh$8S1BSKsNkF.jBAixGc4W7l80OpfCNk65LZBDHBng3NAmbcHuMj4RIm7992rrJ8YA.SJ0hvm.vGk2z483am4Ym1
```

```bassh
$ ./pebble add-identities --from ./identity.yaml
error: cannot decode request body: basic identity can only have "metrics" access, got "admin"
```

internals/overlord/state/identities.go
flotter: Thanks! Makes sense.

internals/overlord/state/identities.go
flotter: Resolving.

flotter (approved):

internals/overlord/state/identities.go
```diff
@@ -102,15 +152,22 @@ func (d *Identity) UnmarshalJSON(data []byte) error {
...
+	if ai.Local != nil {
 		if ai.Local.UserID == nil {
 			return errors.New("local identity must specify user-id")
 		}
 		identity.Local = &LocalIdentity{UserID: *ai.Local.UserID}
 	}
+	if ai.Basic != nil {
+		if ai.Basic.Password == "" {
+			return errors.New("basic identity must specify password (hashed)")
+		}
```
benhoyt, line 165: I think this is unnecessary because is checked in `validateExcludingName`, which is called below, right?

internals/overlord/state/identities.go
```diff
@@ -130,10 +187,11 @@ func (s *State) AddIdentities(identities map[string]*Identity) error {
 		if _, ok := s.identities[name]; ok {
 			existing = append(existing, name)
 		}
-		err := identity.validate()
+		err := identity.validate(name)
 		if err != nil {
 			return fmt.Errorf("identity %q invalid: %w", name, err)
 		}
+		identity.Name = name
```
benhoyt, line 194: We should remove this here, as we still check for further errors below (and it's set below anyway).

internals/overlord/state/identities.go
```diff
@@ -266,17 +324,43 @@ func (s *State) Identities() map[string]*Identity {
...
+// If no matching identity is found for the given inputs, nil is returned.
+func (s *State) IdentityFromInputs(userID *uint32, username, password string) *Identity {
 	s.reading()
 
-	for _, identity := range s.identities {
-		switch {
-		case identity.Local != nil && userID != nil:
-			if identity.Local.UserID == *userID {
+	switch {
+	case username != "" || password != "":
```
benhoyt, line 341: I think we probably want `&&` here (which matches the comment above too - "if *both* username and password are provided").

internals/overlord/state/identities.go
```diff
@@ -266,17 +324,43 @@ func (s *State) Identities() map[string]*Identity {
 
 // IdentityFromInputs returns an identity with the given inputs, or nil
 // if there is none.
-func (s *State) IdentityFromInputs(userID *uint32) *Identity {
+//
+// Identity priority:
+//  1. If both username and password are provided, the function attempts to
+//     match a basic type identity. The userID is ignored in this case. If
```
benhoyt, line 330: Nit: just to clarify it's a value and not just a normal word, let's add quotes:
```suggestion
//     match a "basic" type identity. The userID is ignored in this case. If
```

internals/overlord/state/identities.go
```diff
@@ -266,17 +324,43 @@ func (s *State) Identities() map[string]*Identity {
...
+//     a matching username is found but the password verification fails, nil
+//     is returned immediately.
+//  2. If username and password are not both provided, the function attempts to
+//     match a local type identity using the userID.
```
benhoyt, line 334:
```suggestion
//     match a "local" type identity using the userID.
```

internals/overlord/state/identities.go
```diff
@@ -266,17 +324,43 @@ func (s *State) Identities() map[string]*Identity {
...
-			if identity.Local.UserID == *userID {
+	switch {
+	case username != "" || password != "":
+		// Prioritize username/password if provided, because they come from HTTP
+		// Authorization header, a per-request, client controlled property. If set
+		// by the client, it's intentional, so it should have a higher priority.
+		for _, identity := range s.identities {
+			if identity.Basic != nil && identity.Name == username {
+				crypt := sha512_crypt.New()
+				err := crypt.Verify(identity.Basic.Password, []byte(password))
```
benhoyt, line 348: How about we put `passwordBytes := []byte(password)` above the for loop to avoid allocating/copying each time (maybe Go's compiler is smart enough to avoid this, but I doubt it).

internals/overlord/state/identities_test.go
```diff
@@ -553,28 +634,81 @@ func (s *identitiesSuite) TestIdentityFromInputs(c *C) {
...
+	c.Assert(identity, IsNil)
+
+	identity = st.IdentityFromInputs(nil, "nancy-wrong-username", "wrong-password")
+	c.Assert(identity, IsNil)
+
+	// userID, username, password all provided, two matching identities, prioritize basic type.
+	userID = 42
+	identity = st.IdentityFromInputs(&userID, "nancy", "test")
+	c.Assert(identity, NotNil)
+	c.Check(identity.Name, Equals, "nancy")
```
benhoyt, line 692: Should we check that identity.UID is nil here?

benhoyt (commented):
This looks very close now - just a few nit comments.

I've marked it as "request changes" as a reminder to add documentation. We need to update at least the [Identities reference page](https://documentation.ubuntu.com/pebble/reference/identities/).

In addition, I'll ask for review from some careful Go reviewers as this is sensitive security stuff.

internals/daemon/access.go
IronCore864: I thought I left a comment somewhere and didn't resolve this, but I guess it was in the other metrics PR. I will remove `ReadAccess` and keep `AdminAccess` - I think "admin" should have the highest level of access, including "read" metrics.

internals/overlord/state/identities.go
IronCore864: Yes, it doesn't look good. I will avoid allocation in each loop.

Out of curiosity, I did a benchmark test and it seems it doesn't make a difference, they are more or less the same and even allocation might be faster sometimes. But I agree, it doesn't read very nice because it at least feels expensive :) I will move it out of the loop.

internals/overlord/state/identities.go
IronCore864: Per Harry's comment, I think we want to be strict and on the safer side: since both user/pass are intentionally set by the user if any is set, it means the user doesn't want to use the "basic" type identity, and it's likely that they used a empty password. If we change it to `&&`, the user might get a basic type identity, if it also has read access, it will be granted metrics access.

For example, if the identities are:

```
"bob": {
	Access: state.ReadAccess,
	Local:  &state.LocalIdentity{UserID: 42},
},
"nancy": {
	Access: state.MetricsAccess,
	Basic:  &state.BasicIdentity{Password: "hash"},
},
```

And if the user input is (42, "nancy", ""), it will match bob although the user's intention is to use nancy since they set the username explicitly.

What do you think?

benhoyt: Ah yes, that makes sense. In that case, let's just update the doc comment to match.

client/identities.go
```diff
@@ -47,6 +48,12 @@ type LocalIdentity struct {
 	UserID *uint32 `json:"user-id" yaml:"user-id"`
 }
 
+// BasicIdentity holds identity configuration specific to the "basic" type
+// (for HTTP basic authentication).
```
TheSignPainter98, line 52: This doc comment uses the word 'basic' quite a lot, is there anywhere in this repo where we explain what 'basic' means? I think it could be good to add an explanation of what it entails, or at least link to somewhere which does this.

Also should it be _http basic_ authentication or _basic http_ authentication?

client/identities.go
```diff
@@ -47,6 +48,12 @@ type LocalIdentity struct {
...
+type BasicIdentity struct {
+	Password string `json:"password" yaml:"password"`
```
TheSignPainter98, line 54: This may be overkill, but is this password hashed and salted? If so, I think it may be worth noting this in a doc comment (like in the `BasicIdentity` in internals). Especially as this is a public field on a public type, there's a small but horrifying possibility that this secret could end up in logs or plaintext on disk

I also notice that the internals `BasicIdentity` has the same layout as `BasicIdentity`. Is this intentional and something which needs to be maintained? If so, is it worth adding a comments on each to state that if one is changed, the other should be too? It may also be possible to assert that they have the same layout at compile time

internals/daemon/access.go
```diff
@@ -68,3 +68,18 @@ func (ac UserAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Re
 	// An identity explicitly set to "access: untrusted" isn't allowed.
 	return Unauthorized(accessDenied)
 }
+
+// MetricsAccess allows requests over the HTTP from authenticated users.
```
TheSignPainter98, line 72: Although an acronym (well, technically an initialism), I think 'http' is in such wide use that it's become its own proper noun, so the 'the' can probably be dropped

internals/daemon/access.go
```diff
@@ -68,3 +68,18 @@ func (ac UserAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Re
...
+
+func (ac MetricsAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	if user == nil {
+		return Unauthorized(accessDenied)
+	}
+	switch user.Access {
+	case state.MetricsAccess, state.ReadAccess, state.AdminAccess:
+		return nil
+	}
+	// An identity explicitly set to "access: untrusted" isn't allowed.
```
TheSignPainter98, line 83: This is a curious comment as it assumes that 'untrusted' is the only other access level, which could be subject to change in the future. I think it could be nice to either drop this, or possibly acknowledge that all other access levels (in particular 'untrusted' are denied)

internals/daemon/access.go
```diff
@@ -68,3 +68,18 @@ func (ac UserAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Re
...
+type MetricsAccess struct{}
+
+func (ac MetricsAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	if user == nil {
+		return Unauthorized(accessDenied)
+	}
+	switch user.Access {
+	case state.MetricsAccess, state.ReadAccess, state.AdminAccess:
+		return nil
+	}
```
TheSignPainter98, line 82: Nit: I think it could be nice to put the code below in a `default` condition, that way we can force the compiler to complain if a new codepath doesn't return a value. Although the above `switch` is short so it isn't too big of an issue here, being explicit could be a nice precedent

internals/overlord/state/identities.go
```diff
@@ -47,28 +52,63 @@ type LocalIdentity struct {
 	UserID uint32
 }
 
+// BasicIdentity holds identity configuration specific to the "basic" type
+// (for HTTP basic authentication).
+type BasicIdentity struct {
+	Password string // Note: this is the sha512-crypt hashed password.
```
TheSignPainter98, line 58: I don't recall, will a comment after the field like this appear in docs? I suspect no, in which case I think it's probably worth putting this before and refactoring into the usual style. Would something like this work?
```
// Password holds the user's sha512-crypt-hashed password.
```

internals/overlord/state/identities.go
```diff
@@ -47,28 +52,63 @@ type LocalIdentity struct {
...
+}
+
+// This is used to ensure we send a well-formed identity Name.
+var identityNameRegexp = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9_\-]*$`)
```
TheSignPainter98, line 62: I'm not sure how stringent we want to be with this check, but is it worth also worth forbidding that the name doesn't end in a `_` or `-`? For ease of implementation, I think it's probably pretty reasonable to assert that identity names must be `>=3` characters.

Also, is it worth enforcing either only `_`s or `-`s? If we don't express an opinion, this can easily get messy later. (I don't know whether the time has already passed for this, though)

As a bonus, is it possible to prevent ugly sequences like `--`/`__`? Probably cleanest to do this with a simple `strings.Contains(..., "--")` rather than making the regex hard to read

internals/overlord/state/identities.go
```diff
@@ -47,28 +52,63 @@ type LocalIdentity struct {
...
+
+	if !identityNameRegexp.MatchString(name) {
+		return fmt.Errorf("identity name %q invalid: must start with an alphabetic character and only contain alphanumeric characters, underscore, and hyphen", d.Name)
+	}
+
+	return d.validateExcludingName()
+}
+
+// validateExcludingName checks that the identity is valid, returning an error if not.
+func (d *Identity) validateExcludingName() error {
```
TheSignPainter98, line 78: The name of this function feels quite cumbersome and if there's something else we need to exclude in future, things could get extremely messy. Could the name be inverted to highlight what it _does_ check? Would `validateAccess` work?

internals/overlord/state/identities.go
```diff
@@ -47,28 +52,63 @@ type LocalIdentity struct {
...
-	switch {
-	case d.Local != nil:
-		return nil
-	default:
-		return errors.New(`identity must have at least one type ("local")`)
+	gotType := false
+	if d.Local != nil {
+		gotType = true
+	}
+	if d.Basic != nil {
```
TheSignPainter98, line 101: Is it worth validating that `Basic` and `Local` are mutually exclusive? It sounds like a logic error, but then again, we can't prevent all the user's creative ways of breaking things

internals/overlord/state/identities.go
```diff
@@ -266,17 +320,44 @@ func (s *State) Identities() map[string]*Identity {
...
-		case identity.Local != nil && userID != nil:
-			if identity.Local.UserID == *userID {
+	switch {
+	case username != "" || password != "":
+		// Prioritize username/password if provided, because they come from HTTP
+		// Authorization header, a per-request, client controlled property. If set
+		// by the client, it's intentional, so it should have a higher priority.
+		passwordBytes := []byte(password)
+		for _, identity := range s.identities {
+			if identity.Basic != nil && identity.Name == username {
```
TheSignPainter98, line 343: This section is quite indented, could it be flattened a little? Would something like this work?
```go 
for _, identity := range s.identities {
    if identity.Basic == nil || identity.Name != username {
        continue;
    }
    ...
    if err == nil {
        return identity
    }
    return nil
}
```

internals/overlord/state/identities.go
```diff
@@ -266,17 +320,44 @@ func (s *State) Identities() map[string]*Identity {
...
+//
+// If no matching identity is found for the given inputs, nil is returned.
+func (s *State) IdentityFromInputs(userID *uint32, username, password string) *Identity {
 	s.reading()
 
-	for _, identity := range s.identities {
-		switch {
-		case identity.Local != nil && userID != nil:
-			if identity.Local.UserID == *userID {
+	switch {
```
TheSignPainter98, line 336: This seems a little strange, is there a reason we're using a `switch` here rather than an `if`?

TheSignPainter98 (changes requested):
This is looking pretty nice! Just some minor notes

internals/overlord/state/identities.go
IronCore864: An identity may have multiple types, this is by design, see the [Pebble identities spec](https://docs.google.com/document/d/1nASgUt-piV94i1cpFsbEPRk_xFbZt8AskgL97gh1Dgg/edit?tab=t.0) for more details.

client/identities.go
IronCore864: Thanks for pointing this out, I will add more comments in the code where it mentions basic type identities.

Regarding the name, I do agree "HTTP basic" is better than simply "basic", but the name name "basic" has been decided in the spec review meeting with Gustavo, so I will keep it as it is for now.

client/identities.go
IronCore864: Yes, it's hashed and salted. A comment is added elsewhere but worth mentioning here, thanks for pointing it out!

As for the internal/duplication of the same struct: Yes, it's intentional, because the client has a separate definition of it and should not depend on the internal one.

internals/overlord/state/identities.go
IronCore864: Yes, a comment like this will be put in the docs as verified by Ben. However I do think your suggested line is better (and it fits the one on the client better as well), so I will adopt this. Thanks!

internals/overlord/state/identities.go
IronCore864: After an internal discussion with Ben, we think this is not a security issue, so we will leave it as is: users should be allowed to use underscores or dashes, and maybe two-character names make sense, for example, "Ed".

internals/overlord/state/identities.go
IronCore864: From my point of view, the benefit is if we add more types in the future, `switch` will be better. And, even if an identity has multiple types, we should return just one here.

internals/overlord/state/identities.go
TheSignPainter98: Ah I see, thanks for the explanation!

internals/overlord/state/identities.go
TheSignPainter98: This is true, but do we know whether we'll add more types in the (reasonably) near future? Until that happens, this code will look a little odd, but either way seems fine

docs/explanation/api-and-clients.md
```diff
@@ -16,12 +16,15 @@ In addition to the Go client, there's also a [Python client](https://github.com/
...
 
-API endpoints fall into one of three access levels, from least restricted to most restricted:
+API endpoints fall into one of four access levels, from least restricted to most restricted:
 
 * **Open-access** - Allowed from any user, even unauthenticated users using the HTTP-over-TCP listener.
     * `GET /v1/system-info`, which returns the Pebble version and other information
     * `GET /v1/health`, which returns a boolean to indicate whether Pebble's health checks are all healthy
 
+* **Metrics-access** - Allowed from any authenticated user.
+    * `GET /v1/metrics`, which returns metrics for services and health checks in OpenTelemetry exposition format.
```
dwilding, line 26: Is "exposition format" the name of a specific format? I can't find much by searching for this term. So it would probably be fine to say "OpenTelemetry format".

docs/reference/identities.md
```diff
@@ -14,16 +14,18 @@ identities:
         # - untrusted: has access to open-access endpoints only
         # - read: has access to read-access endpoints
         # - admin: has access to all endpoints
-        access: untrusted | read | admin
+        access: untrusted | read | metrics | admin
```
dwilding, line 17: Should the order be `untrusted`, `metrics`, `read`, `admin` to match the least-restricted-to-most-restricted order used in the API and clients doc?

docs/reference/identities.md
```diff
@@ -35,3 +37,16 @@ identities:
...
+```yaml
+identities:
+    alice:
+        access: metrics
+        basic:
+            # The password is sha512-hashed.
+            password: <password hash>
+```
+
+Note that for `basic` type identity, the password is sha512-hashed. Use `openssl passwd -6` to generate a hashed password.
```
dwilding, line 52: I think the final sentence can be shortened a bit. Password is only relevant for basic auth, so I think we can say:

```
The password must be sha512-hashed. Use `openssl passwd -6` to generate a hashed password.
```

internals/cli/cmd_add-identities.go
```diff
@@ -39,6 +39,16 @@ For example, to add a local admin named "bob", use YAML like this:
 >         access: admin
 >         local:
 >             user-id: 42
+
+To add an identity named "alice" with metrics access using HTTP basic authentication, use YAML like this:
```
dwilding, line 43: I would remove "use YAML like this", and have:

```
To add an identity named "alice" with metrics access using HTTP basic authentication:
```

docs/reference/cli-commands.md
```diff
@@ -67,6 +67,17 @@ For example, to add a local admin named "bob", use YAML like this:
 >         local:
 >             user-id: 42
 
+To add an identity named "alice" with metrics access using HTTP basic
+authentication, use YAML like this:
```
dwilding, line 71: I would remove "use YAML like this", and have:

```
To add an identity named "alice" with metrics access using HTTP basic authentication:
```

docs/explanation/api-and-clients.md
IronCore864: I think "exposition format" and "format" are interchangeable. Based on Google searches, there are many uses of "Prometheus format" and "Prometheus exposition format." I will then choose the simpler "OpenTelemetry format" and stick with it throughout our docs.

For the endpoint - thanks, I forgot to update the OpenAPI spec, will do, and add a link there.

IronCore864: I will resolve this comment for now and implement the OpenAPI spec in the other metrics PR.

docs/explanation/api-and-clients.md
```diff
@@ -55,4 +63,15 @@ identities:
...
 ```
 
+An example identity named "alice" with `basic` type and `metrics` access level is shown below:
+
+```yaml
+identities:
+    alice:
+        access: metrics
+        basic:
+            # The password is sha512-hashed.
```
benhoyt, line 73: It's actually hashed using the sha512-crypt algorithm, which is a bunch of rotations and salt and stuff on top of bare sha512:
```suggestion
            # The password is hashed using sha512-crypt, as generated by "openssl passwd -6".
```

docs/how-to/manage-identities.md
```diff
@@ -20,9 +20,10 @@ identities:
         local:
             user-id: 42
     alice:
-        access: read
-        local:
-            user-id: 2000
+        access: metrics
+        basic:
+            # The password is sha512-hashed.
```
benhoyt, line 25: Same as above: sha512-crypt
```suggestion
            # The password is hashed using sha512-crypt, as generated by "openssl passwd -6".
```

docs/reference/cli-commands.md
```diff
@@ -67,6 +67,17 @@ For example, to add a local admin named "bob", use YAML like this:
...
+To add an identity named "alice" with metrics access using HTTP basic
+authentication:
+
+> identities:
+>     alice:
+>         access: metrics
+>         basic:
+>             password: <password hash>
+
+Use "openssl passwd -6" to generate a hashed password.
```
benhoyt, line 79:
```suggestion
Use "openssl passwd -6" to generate a hashed password (sha512-crypt format).
```

docs/reference/identities.md
```diff
@@ -12,18 +12,21 @@ identities:
...
-        # Currently the only suported authentication type is "local". Other
-        # types may be added in future, at which point you may configure an
-        # identity with one or more authentication types.
+        # Currently the supported authentication types are "local" and "basic".
+        # You may configure an identity with one or more authentication types.
         local:
             # (Required) Peer credential UID.
             user-id: <uid>
+        basic:
+            # (Required) Sha512-hashed password
```
benhoyt, line 28:
```suggestion
            # (Required) Hashed password in sha512-crypt format.
```

docs/reference/identities.md
```diff
@@ -35,3 +38,16 @@ identities:
...
 ```
+
+For another example, a basic identity named "alice" that is granted `metrics` access would be defined as follows:
+
+```yaml
+identities:
+    alice:
+        access: metrics
+        basic:
+            # The password is sha512-hashed.
```
benhoyt, line 49: Same comment here as above.

docs/reference/identities.md
```diff
@@ -35,3 +38,16 @@ identities:
...
+            password: <password hash>
+```
+
+The password must be sha512-hashed. Use `openssl passwd -6` to generate a hashed password.
```
benhoyt, line 53: Similar here.

benhoyt (commented):
Just a bunch of comments about sha512-crypt and how we document that.

hpidcock (approved):
Awesome work. Crypt module choice appears sound.

benhoyt:
I'm going to merge this shortly, to make https://github.com/canonical/pebble/pull/519 easier to review.

TheSignPainter98 (approved):
Looks great, thanks for the tweaks!

holmanb (commented):
I was in the process of reviewing this but it merged before I was done. Approving retroactively, thanks!

PR #567, opened 2025-02-03
chore: bump documentation dependencies

tonyandrewmeyer:
Primarily this pulls in security fixes for starlette, jinja, urllib3, and certifi, but it updates all the dependencies while we're doing this. The docs build with no errors and look ok with a cursory review.


benhoyt (approved):
Looks good to me, thanks.

PR #568, opened 2025-02-05
chore: replace interface{} with any

benhoyt:
`any` is (as of Go 1.18) a builtin alias for `interface{}`, equivalent in every respect. It was introduced in Go 1.18+, and our go.mod says we're on 1.22. A similar thing was done to the Go project and stdlib a while back, see for example https://pkg.go.dev/encoding/json#Marshal

Commands used to generate this PR (based on the [same thing done](https://go-review.googlesource.com/c/protobuf/+/585735) to the Go project):

```
$ sed -i 's,interface{},any,g' **/*.go
$ sed -i 's,interface{},any,g' **/**/*.go
$ sed -i 's,interface{},any,g' **/**/**/*.go
# I'm not actually sure why the last two were needed here and not for Go,
# but the first one only did the client package

go fmt ./...
```

This also adds a GitHub Action (part of the format checks) that runs grep to ensure we don't have any remaining `interface{}`, for example from in-flight PRs. See [failing example run](https://github.com/canonical/pebble/actions/runs/13263762684/job/37026049090?pr=568).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,6 +62,10 @@ jobs:
         go fmt ./...
         git diff --exit-code
 
+    - name: Ensure no use of empty interface (should be any)
+      run: |
+        ! egrep -R --exclude-dir .git 'interface\{\}'
+
   docs:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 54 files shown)


benhoyt:
@flotter @pedronis What do you think about doing this to snapd and other codebases derived from Pebble, seeing we like to keep in sync?

flotter:
@benhoyt I do not have a strong opinion other than it does not bother me, and if you do it, we will follow.

IronCore864 (approved):

PR #570, opened 2025-02-12
chore: update doc links in README

dwilding:
A quick PR to update the README to use the new URLs of Pebble docs. We've already set up redirects, so this is just to keep things tidy.

@benhoyt / @IronCore864 are you able to update the link in the **About** section of the repo? I don't think I have access to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,15 +11,15 @@ _Take control of your internal daemons!_
 
 Pebble's key features:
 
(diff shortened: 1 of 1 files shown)


IronCore864 (approved):

benhoyt (approved):
Thanks!

PR #571, opened 2025-02-12
feat: add workloads to the plan

anpep:
Workloads are a way to optionally group services and apply shared run configuration (user, group, environment) to services, in a way that may be extended in the future to support confinement.

```yaml
services:
    my-service:
        override: replace
        command: /bin/my-service
        workload: my-workload

    my-other-service:
        override: replace
        command: /bin/my-other-service
        workload: my-workload

workloads:
    my-workload:
        user: foo
        group: foo
        environment:
            SHARED_VARIABLE: bar
```


internals/overlord/servstate/handlers.go
```diff
@@ -338,15 +358,22 @@ func (s *serviceData) startInternal() error {
...
 	// Start as another user if specified in plan.
-	uid, gid, err := osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
+	var uid, gid *int
+	if s.config.UserID != nil || s.config.GroupID != nil || s.config.User != "" || s.config.Group != "" {
+		// User/group config from the service takes precedence
+		uid, gid, err = osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
+	} else if s.workload != nil {
+		// Take user/group config from workload
+		uid, gid, err = osutil.NormalizeUidGid(s.workload.UserID, s.workload.GroupID, s.workload.User, s.workload.Group)
+	}
```
paul-rodriguez, line 376: What about the `else` case?

internals/overlord/servstate/handlers.go
```diff
@@ -338,15 +358,22 @@ func (s *serviceData) startInternal() error {
...
 
 	// Start as another user if specified in plan.
-	uid, gid, err := osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
+	var uid, gid *int
+	if s.config.UserID != nil || s.config.GroupID != nil || s.config.User != "" || s.config.Group != "" {
+		// User/group config from the service takes precedence
+		uid, gid, err = osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
+	} else if s.workload != nil {
+		// Take user/group config from workload
+		uid, gid, err = osutil.NormalizeUidGid(s.workload.UserID, s.workload.GroupID, s.workload.User, s.workload.Group)
```
paul-rodriguez, line 375: To avoid any bad surprises in the future, I would move the `if err != nil {return err}` inside the two `if` blocks.

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+
+const WorkloadsField = "workloads"
+
+var _ plan.Section = (*WorkloadsSection)(nil)
+
+type WorkloadsSection struct {
+	Entries map[string]*Workload `yaml:",inline"`
+}
+
+func (ws *WorkloadsSection) IsZero() bool {
```
paul-rodriguez, line 95: Suggestion to rename to `IsEmpty`

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+			return &plan.FormatError{
+				Message: fmt.Sprintf("workload %q %v", name, err),
+			}
+		}
+	}
+	return nil
+}
+
+func (ws *WorkloadsSection) combine(other *WorkloadsSection) error {
+	if len(other.Entries) != 0 && ws.Entries == nil {
```
paul-rodriguez, line 116: `!other.IsEmpty() && ws.Entries == nil`

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+		ws.Entries = make(map[string]*Workload)
```
paul-rodriguez, line 117: Sounds like it would make sense to reserve `len(other.Entries)` slots in the `make` call.

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+	User        string            `yaml:"user,omitempty"`
+	GroupID     *int              `yaml:"group-id,omitempty"`
+	Group       string            `yaml:"group,omitempty"`
+}
+
+func (w *Workload) validate() error {
+	if w.Name == "" {
+		return errors.New("cannot have an empty name")
+	}
+	// Value of Override is checked in the (*WorkloadSection).combine() method
```
paul-rodriguez, line 161: My intuition is that `validate` should do that check regardless.

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+	return nil
```
paul-rodriguez, line 162: Additionally, there could be a lot of invalid stuff in the `Environment` map. Would it make sense to check here?

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	if w.GroupID != nil {
+		copied.GroupID = copyIntPtr(w.GroupID)
+	}
+	return &copied
+}
+
+func (w *Workload) merge(other *Workload) {
+	if len(other.Environment) != 0 && w.Environment == nil {
+		w.Environment = make(map[string]string)
```
paul-rodriguez, line 184: Same here: reserving the same space is probably better

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+	// Basic details
+	Name     string        `yaml:"-"`
+	Override plan.Override `yaml:"override,omitempty"`
+
+	// Options for command execution
+	Environment map[string]string `yaml:"environment,omitempty"`
+	UserID      *int              `yaml:"user-id,omitempty"`
+	User        string            `yaml:"user,omitempty"`
+	GroupID     *int              `yaml:"group-id,omitempty"`
+	Group       string            `yaml:"group,omitempty"`
```
paul-rodriguez, line 154: Considering we allow specifying _both_ user-id and username, or _both_ group-id and group name, there's a possibility of a workload where the group-id doesn't actually match the group name and same for users. Where would we deal with that? I expect this will only get detected for sure in kernos bootstrap. Correct?

paul-rodriguez (commented):
Except for a couple of minor details this looks good to me. I'm still a bit fuzzy on how this will actually play out in derivative projects.

internals/overlord/servstate/handlers.go
anpep: `uid`, `gid` are `nil`, and `setCmdCredential()` is not called, resulting in the services being run as the current user

internals/overlord/servstate/handlers.go
```diff
@@ -378,6 +405,12 @@ func (s *serviceData) startInternal() error {
 		}
 	}
 
+	if s.workload != nil && len(s.workload.Environment) != 0 {
+		for k, v := range s.workload.Environment {
+			environment[k] = v
```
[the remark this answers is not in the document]
anpep: Yes, we decided nothing against that when discussing the spec, but I'm happy to explore this. Personally, I don't think there's anything wrong per se with allowing every environment variable to be changed, but if you have arguments against that I'm happy to make them part of the spec discussion

internals/overlord/servstate/workloads.go
anpep: Feels more fitting, however `IsZero()` is already part of the `Section` interface in the plan:

https://github.com/canonical/pebble/blob/d8a630a02868fa5347bae81494cc17007b4ef231/internals/plan/plan.go#L53-L59

internals/overlord/servstate/workloads.go
anpep: This would potentially result in a nil pointer dereference in sneaky ways, since it's up to the implementation of `IsZero()` to dereference or not the pointer receiver. `len()` is special in that `len(nil) == 0` for slices and maps.

internals/overlord/servstate/workloads.go
anpep: Good call, thanks!

internals/overlord/servstate/workloads.go
anpep: Thanks!

internals/overlord/servstate/workloads.go
anpep: This is detected in `servstate` when calling `osutil.NormalizeUidGid()`. An error may be returned if `user`/`user-id` or `group`/`group-id` does not match, as well as when specifying a `user`/`group` that does not exist in the system.

internals/overlord/servstate/workloads.go
anpep: What does "invalid" mean in this sense?

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+	return ws, nil
+}
+
+func (ext *WorkloadsSectionExtension) ParseSection(data yaml.Node) (plan.Section, error) {
+	ws := &WorkloadsSection{}
+	// The following issue prevents us from using the yaml.Node decoder
+	// with KnownFields = true behavior. Once one of the proposals get
+	// merged, we can remove the intermediate Marshal step.
```
flotter, line 49: Was there an issue reference in the original comment ?

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+	if len(data.Content) != 0 {
+		yml, err := yaml.Marshal(data)
+		if err != nil {
+			return nil, fmt.Errorf(`internal error: cannot marshal "workloads" section: %w`, err)
```
flotter, line 53: I've seen people quote with ```%q``` which is a single quote. Would it make sense to follow that convention using ```'workloads'``` and just use normal double quotes on the outside ? Maybe I am missing something.

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+	return ws, nil
+}
+
+func (ext *WorkloadsSectionExtension) ValidatePlan(p *plan.Plan) error {
+	ws, ok := p.Sections[WorkloadsField].(*WorkloadsSection)
+	if !ok {
+		return fmt.Errorf("internal error: invalid section type %T", ws)
+	}
+	for name, service := range p.Services {
+		_, ok := ws.Entries[service.Workload]
```
flotter, line 77: nit: For me it feels a little strange looking up an empty workload name, which by workload design I would imagine is not a valid workload key, so can never match? Would it not be worth the extra indentation to pre-check if a workload exist for the service before validating it ?

flotter: Once we migrate servstate to the extension system, it would be better to let services validate its workload field, as that feels like a more accurate dependency direction.

internals/overlord/servstate/workloads.go
flotter: Which comes from https://pkg.go.dev/gopkg.in/yaml.v3#IsZeroer :)

internals/overlord/servstate/workloads.go
flotter: https://github.com/canonical/pebble/blob/e3b81e139fc188e8ef01311e9fb13fe26402f816/internals/plan/extensions_test.go#L601

Another way is to use the for/range below as your hint you need to allocate, as it would only enter if ```other``` has something.

see ```makeMapIfNil```

internals/overlord/servstate/workloads.go
flotter: Could it make sense to check this as part of the layer Workloads plan validation ? That would allow us to detect issues automatically at earlier points of entry (i.e. when the plan is loaded or someone attempts to modify it over the API) ?

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,209 @@
...
+		if service.Workload != "" && !ok {
+			return &plan.FormatError{
+				Message: fmt.Sprintf(`plan service %q cannot run in unknown workload %q`, name, service.Workload),
+			}
+		}
+	}
```
flotter, line 83: Should we add uid/gid/user/group validation here?

internals/overlord/servstate/workloads.go
flotter: The way is has been implemented in the plan library is that the combine logic checks this, and therefore the extension followed the same flow. Since both layer creation and merging of the layer has to be successful for a new plan to be updated, it does not really matter in practice where its done, the issue will be caught.

Having said that, since ```plan.Override``` is the same type being used everywhere, it would be super simple to get the type to validate itself as part of layer validation instead, and it could make the ```combine``` a little simpler.

internals/overlord/servstate/handlers.go
```diff
@@ -167,27 +180,34 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
-			manager: m,
-			state:   stateInitial,
-			config:  config.Copy(),
-			logs:    servicelog.NewRingBuffer(maxLogBytes),
-			started: make(chan error, 1),
-			stopped: make(chan error, 2), // enough for killTimeElapsed to send, and exit if it happens after
+			manager:  m,
+			state:    stateInitial,
+			config:   config.Copy(),
+			workload: w,
```
flotter, line 199: Can this line not just look like the one above for ```config plan.Service``` ?

internals/overlord/servstate/handlers.go
```diff
@@ -167,27 +180,34 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
+			started:  make(chan error, 1),
+			stopped:  make(chan error, 2), // enough for killTimeElapsed to send, and exit if it happens after
 		}
 		m.services[config.Name] = service
 		return service, ""
 	}
 
 	// Ensure config is up-to-date from the plan whenever the user starts a service.
 	service.config = config.Copy()
+	service.workload = w
```
flotter, line 210: Why not copy the line above ? :)

internals/overlord/servstate/manager.go
```diff
@@ -264,7 +264,15 @@ func (m *ServiceManager) Replan() ([][]string, [][]string, error) {
 			if config.Equal(s.config) {
 				continue
 			}
-			s.config = config.Copy() // update service config from plan
+			// Update service config and workload from plan
+			s.config = config.Copy()
+			if s.config.Workload != "" {
```
flotter, line 269: If workload changed from "foo" to "" with the previous update, I think you do want to update ```s.workload``` to not contain a copy of the previous workload ?

flotter: Also if the plan.Service is the same, and the workload changed, replan will not restart the service.

flotter (commented):
Great start Angel. Let's see what @benhoyt thinks about the concept :)

benhoyt:
Sorry for the delay -- I hope to get to this tomorrow.

internals/overlord/servstate/workloads.go
anpep: It would be trivial to add a `func (plan.Override) Validate() error` method, but it would serve no practical purpose because the validation here is done at the same time as we perform the logic for combining the layers

internals/overlord/servstate/handlers.go
anpep: Because `workload` here might be `nil`, and calling `workload.copy()` could potentially dereference this `nil` pointer. I made the code a bit more easy to follow though, thanks!

internals/overlord/servstate/workloads.go
anpep: Maybe I did not understand correctly, but `%q` uses double quotation marks. I personally don't like mixing up single/double quotes, and also found this in existing code:

https://github.com/canonical/pebble/blob/master/internals/plan/plan.go#L732-L736

internals/overlord/servstate/manager.go
anpep: Nice catch!

internals/overlord/servstate/workloads.go
anpep: I will think about it; however, I decided not to initially because the validation of the services layer does not do it like that, and instead relies on doing that when the service start request is received

internals/overlord/servstate/handlers.go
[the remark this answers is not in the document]
paul-rodriguez: Fine by me

internals/overlord/servstate/workloads.go
paul-rodriguez: Can `combine` be called with `other == nil`? In that case I would expect the first line to be `if other == nil { return nil }` or something.

If it can't and we know `other != nil` then calling `other.IsEmpty()` should never fail, even if `other.Entries` itself is `nil`.

internals/overlord/servstate/handlers.go
paul-rodriguez: Ok, please resolve

internals/overlord/servstate/workloads.go
paul-rodriguez: Ok, please resolve

internals/overlord/servstate/workloads.go
paul-rodriguez: The most broken examples are probably adding an environment variable that contains a `=` or a `\0` character in its name. Those two are strictly forbidden by POSIX. Realistically there are a bunch of things that are legal to do in golang strings that will almost certainly break compatibility with certain standard tools if present in an environment variable name (or even value, in some cases).

internals/overlord/servstate/workloads.go
paul-rodriguez: Seems fine to me, at least for now.

internals/overlord/servstate/workloads.go
anpep: While that is true, Pebble currently does not enforce any kind of sanity over the environment a service has

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,223 @@
...
+	if w.GroupID != nil {
+		copied.GroupID = copyPtr(w.GroupID)
+	}
+	return &copied
+}
+
+func (w *Workload) merge(other *Workload) {
+	w.Environment = makeMapIfNil(w.Environment)
+	for k, v := range other.Environment {
+		w.Environment[k] = v
```
benhoyt, line 194: I'd prefer to do the `makeMapIfNil` inside the loop here (like `Service.Merge` and others do) so that if both maps are nil, the result is nil. Only allocate if we have to.

Or maybe we can use `maps.Copy`:

```go
if len(other.Environment) > 0 {
    w.Environment = makeMapIfNil(w.Environment)
    maps.Copy(w.Environment, other.Environment)
}
```

Then again, that dance is kind of annoying, so maybe a `copyMap` helper that does this. I'll give it some more thought separately to this PR.

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (ws *WorkloadsSection) IsZero() bool {
+	return len(ws.Entries) == 0
+}
+
+func (ws *WorkloadsSection) Validate() error {
+	for name, workload := range ws.Entries {
+		if workload == nil {
+			return &plan.FormatError{
+				Message: fmt.Sprintf("workload %q has a null value", name),
```
benhoyt, line 114:
```suggestion
				Message: fmt.Sprintf("workload %q cannot have a null value", name),
```

internals/overlord/servstate/workloads_test.go
```diff
@@ -0,0 +1,373 @@
...
+	layers: []string{`
+workloads:
+    default:
+        override: merge
+        environment:
+            foo: bar
+            bar: baz
+        user-id: 1337
+        user: alyssa
+        group-id: 1337
```
benhoyt, line 102: Let's make the user, user-id and group, group-id different in these tests, to ensure the code isn't accidentally swapping the two.

benhoyt (commented):
I realise this is still a draft PR, but this will also need documentation: for now, at least changes to the [layer spec](https://documentation.ubuntu.com/pebble/reference/layer-specification/) in `docs/reference/layer-specification.md`.

internals/overlord/planstate/manager.go
```diff
@@ -253,6 +253,8 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 	if err != nil {
 		return nil, err
 	}
+	// Validate workloads & services here?
+	// We can make a switch in personality for KernOS/Pebble
```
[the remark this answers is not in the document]
anpep: That should definitely not made it to a commit--was a comment I put there during a conversation about something not meant for this PR, sorry!

internals/overlord/servstate/workloads.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func (ws *WorkloadsSection) combine(other *WorkloadsSection) error {
+	ws.Entries = makeMapIfNil(ws.Entries)
+	for name, workload := range other.Entries {
+		switch workload.Override {
```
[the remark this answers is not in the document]
anpep: I've tried this but I found no way to do it without resorting to reflection, because here we access the `Entries map[string]*Workload` field, whereas in checks is `Checks map[string]*Check` from the layer struct, etc. I say let's do this when we eventually refactor the plan to consist on `SectionExtension`s and move `Services`, `Checks`, `LogTargets` from the `plan.Layer` struct onto their own section extensions, what do you think?

internals/overlord/servstate/handlers.go
```diff
@@ -115,13 +118,21 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
...
-	config, ok := currentPlan.Services[request.Name]
-	if !ok {
+	config, configFound := currentPlan.Services[request.Name]
+	if !configFound {
 		return fmt.Errorf("cannot find service %q in plan", request.Name)
 	}
 
+	var workload *workloads.Workload
+	if config.Workload != "" {
+		ws := currentPlan.Sections[workloads.WorkloadsField].(*workloads.WorkloadsSection)
```
paul-rodriguez, line 128: This will panic if the type isn't right. I know that such case would be an unrecoverable internal error anyway, so panicking isn't wrong necessarily. What I'm suggesting is to make it explicit i.e. own the panic instead of leaving it up to interpretation whether this is meant to panic in the bad case.

internals/overlord/servstate/handlers.go
```diff
@@ -338,17 +355,31 @@ func (s *serviceData) startInternal() error {
...
 
 	s.cmd.Dir = s.config.WorkingDir
 
 	// Start as another user if specified in plan.
-	uid, gid, err := osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
-	if err != nil {
-		return err
+	var uid, gid *int
+	if s.config.UserID != nil || s.config.GroupID != nil || s.config.User != "" || s.config.Group != "" {
+		// User/group config from the service takes precedence if any of them are set
```
paul-rodriguez, line 372: Do we have an explicit specification about mixed cases? 

As in, what should happen if the service config specifies: 1) a user and 2) a workload but 3) no group? As a user I think I would expect that the service would then use the group from its workload and the user from the service config. In the code as it is, in such a case the user written in the service config would be silently ignored.

internals/overlord/servstate/manager_test.go
```diff
@@ -1776,6 +1784,79 @@ services:
...
+	c.Assert(err, ErrorMatches, `plan service "test1" workload not defined: "non-existing"`)
+}
+
+func (s *S) tryPlanAddLayer(c *C, layerYAML string) error {
+	cnt := len(s.plan.Layers)
+	layer, err := plan.ParseLayer(cnt, fmt.Sprintf("test-plan-layer-%v", cnt), []byte(layerYAML))
+	if err != nil {
+		return err
+	}
+	// Resolve {{.NotifyDoneCheck}}
```
paul-rodriguez, line 1840: What's this? Is this just a comment or does it do anything?

internals/workloads/workloads.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return ws, nil
+}
+
+func (ext *PlanExtension) ParseSection(data yaml.Node) (plan.Section, error) {
+	ws := &WorkloadsSection{}
+	// The following issue prevents us from using the yaml.Node decoder
+	// with KnownFields = true behavior. Once one of the proposals get
+	// merged, we can remove the intermediate Marshal step.
+	// https://github.com/go-yaml/yaml/issues/460
```
paul-rodriguez, line 52: The code that this comment refers to has been moved into `SectionDecode`

internals/workloads/workloads.go
```diff
@@ -0,0 +1,230 @@
...
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/canonical/pebble/internals/osutil"
+	"github.com/canonical/pebble/internals/plan"
+)
+
+var _ plan.SectionExtension = (*PlanExtension)(nil)
+
+type PlanExtension struct{}
```
paul-rodriguez, line 31: Consider moving everything related to `PlanExtension` into a different file in this package.

internals/workloads/workloads.go
```diff
@@ -0,0 +1,230 @@
...
+	// https://github.com/go-yaml/yaml/issues/460
+	if len(data.Content) != 0 {
+		if err := plan.SectionDecode(&data, ws); err != nil {
+			return nil, &plan.FormatError{
+				Message: fmt.Sprintf(`cannot parse the "workloads" section: %v`, err),
+			}
+		}
+	}
+	for name, workload := range ws.Entries {
+		if workload != nil {
```
paul-rodriguez, line 61: Is this condition ever false? There would be an entry in the `ws.Entries` map that has a `name` key that points to nothing?

internals/overlord/servstate/handlers.go
benhoyt: I believe I suggested we don't need special handling here (to simplify the code). I can see what Paul's saying, but I think that would be a slippery slope, leading to extraneous `if`s and explicit `panic`s in a ton of places if we followed that everywhere. How about a compromise: leave the code as is but add a comment like `// Will panic if we used an incorrect type (programmer error)`?

internals/overlord/servstate/handlers.go
```diff
@@ -338,17 +355,31 @@ func (s *serviceData) startInternal() error {
...
-	environment := make(map[string]string)
-	for k, v := range s.config.Environment {
-		environment[k] = v
+	var environment map[string]string
+	if s.workload != nil && len(s.workload.Environment) > 0 {
+		environment = maps.Clone(s.workload.Environment)
+	} else if len(s.config.Environment) > 0 {
+		environment = maps.Clone(s.config.Environment)
+	} else {
+		environment = make(map[string]string)
```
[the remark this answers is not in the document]
benhoyt: I don't think so quite. If `maps.Clone` is passed a nil map, it returns nil, and you won't be able to write to that map. We could have a more complicated `if` condition, but it's probably clearer this way.

internals/overlord/servstate/workloads.go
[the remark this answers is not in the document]
benhoyt: I think that's a reasonable compromise. There's probably a decent way to factor it out with an interface or something.

internals/workloads/workloads.go
```diff
@@ -0,0 +1,230 @@
...
+		if err := workload.Validate(); err != nil {
+			return &plan.FormatError{
+				Message: fmt.Sprintf("workload %q %v", name, err),
+			}
+		}
+	}
+	return nil
+}
+
+func (ws *WorkloadsSection) combine(other *WorkloadsSection) error {
```
benhoyt, line 121: I've been messing around with this, but can't get it to compile yet - but this is kinda what I'm thinking, add a `GetOverride` method so everything we need is on a (generic) interface:

```go
type entryType[T any] interface {
	GetOverride() plan.Override
	Copy() *T
	Merge(other *T)
}

func combineEntries[E entryType[T], T any](otherEntries map[string]E) (map[string]E, error) {
	var entries map[string]E
	for name, entry := range otherEntries {
		if entries == nil {
			entries = make(map[string]E, len(otherEntries))
		}
		switch entry.GetOverride() {
		case plan.MergeOverride:
			if current, ok := entries[name]; ok {
				copied := current.Copy()
				copied.Merge(entry)
				entries[name] = copied
				break
			}
			fallthrough
		case plan.ReplaceOverride:
			entries[name] = entry.Copy()
		case plan.UnknownOverride:
			return nil, &plan.FormatError{
				Message: fmt.Sprintf(`workload %q must define an "override" policy`, name),
			}
		default:
			return nil, &plan.FormatError{
				Message: fmt.Sprintf(`workload %q has an invalid "override" policy: %q`, name, entry.GetOverride()),
			}
		}
	}
	return entries, nil
}
```

I'll play with it some more tomorrow, but if you're a Go generics expert, maybe you know why it's not working? :-)

internals/workloads/workloads.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	if other.GroupID != nil {
+		w.GroupID = copyPtr(other.GroupID)
+	}
+	if other.Group != "" {
+		w.Group = other.Group
+	}
+}
+
+func (w *Workload) Equal(other *Workload) bool {
```
benhoyt, line 198: TODO ben: up to here in review.

benhoyt (commented):
I got half way through reviewing this today - hope to finish tomorrow, but posting the comments I have for now.

internals/workloads/workloads.go
anpep: Yes, in the following case:

```yaml
workloads:
    default:
```

internals/overlord/servstate/manager_test.go
anpep: It's just a comment, this was copy-pasted from `func (s *S) planAddLayer()`. `tryPlanAddLayer()` does almost the same, except it will return errors instead of asserting they are `nil`

internals/overlord/servstate/handlers.go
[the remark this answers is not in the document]
anpep: According to the spec:

```yaml
workloads:
        ...

        # (Optional) When specified, the environment of the service running in
        # this workload is updated to include these values.
        environment:
            <env var name>: <env var value>
```

benhoyt:
@anpep It looks like you may have resolved a bunch of comments but forgotten to commit/push?

internals/workloads/workloads.go
anpep: Thanks for working on this! I briefly discussed with Fred the other day and we were also close, but I agree we should ring fence this PR to workloads

internals/workloads/sectionextension.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+var _ plan.SectionExtension = (*SectionExtension)(nil)
+
+type SectionExtension struct{}
+
+func (ext *SectionExtension) CombineSections(sections ...plan.Section) (plan.Section, error) {
+	ws := &WorkloadsSection{}
+	for _, section := range sections {
+		// The following will panic if any of the supplied section is not a WorkloadsSection
```
flotter, line 33: typo: ```... any if the supplied sections are ... ```

I wonder how useful this comment is ? This will only occur in the case of some coding error, and then the behavior will be obvious enough, maybe ?

internals/workloads/sectionextension.go
```diff
@@ -0,0 +1,135 @@
...
+	for name, workload := range ws.Entries {
+		if workload != nil {
+			workload.Name = name
+		}
+	}
+	return ws, nil
+}
+
+func (ext *SectionExtension) ValidatePlan(p *plan.Plan) error {
+	// The following will panic if the "workloads" section is not a WorkloadsSection
```
flotter, line 58: nit: Not sure this comment is needed ?

internals/workloads/sectionextension.go
```diff
@@ -0,0 +1,135 @@
...
+	ws := p.Sections[WorkloadsField].(*WorkloadsSection)
+	for name, service := range p.Services {
+		if service.Workload == "" {
+			continue
+		}
+		if _, ok := ws.Entries[service.Workload]; !ok {
+			return &plan.FormatError{
+				Message: fmt.Sprintf("plan service %q workload not defined: %q", name, service.Workload),
```
flotter, line 66: nit: I would put the workload name inline with the message, and typically only the error after the ```:```. In this case it would drop the colon.

internals/workloads/sectionextension.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+		}
+	}
+	for name, workload := range ws.Entries {
+		if _, _, err := osutil.NormalizeUidGid(workload.UserID, workload.GroupID, workload.User, workload.Group); err != nil {
+			return &plan.FormatError{
+				Message: fmt.Sprintf("plan workload %q %v", err, name),
+			}
+		}
```
flotter, line 75: No error will be returned if we pass ```nil, nil, "", "" ```. Is that OK for the generic case here ?

internals/workloads/sectionextension.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return nil
+}
+
+func (ws *WorkloadsSection) combine(other *WorkloadsSection) error {
+	for name, workload := range other.Entries {
+		if ws.Entries == nil {
+			ws.Entries = make(map[string]*Workload, len(other.Entries))
+		}
```
flotter, line 112: I think you can replace this with ```makeMapIfNil``` as already included in ```workloads.go```

internals/workloads/workloads.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// Options for command execution
+	Environment map[string]string `yaml:"environment,omitempty"`
+	UserID      *int              `yaml:"user-id,omitempty"`
+	User        string            `yaml:"user,omitempty"`
+	GroupID     *int              `yaml:"group-id,omitempty"`
+	Group       string            `yaml:"group,omitempty"`
+}
+
+func (w *Workload) Validate() error {
```
flotter, line 39: nit: I think this can be private ?

internals/workloads/workloads.go
```diff
@@ -0,0 +1,108 @@
...
+	if w.Name == "" {
+		return errors.New("cannot have an empty name")
+	}
+	// Value of Override is checked in the (*WorkloadSection).combine() method
+	return nil
+}
+
+func (w *Workload) Copy() *Workload {
```
flotter, line 47: Can be private ?

internals/workloads/workloads.go
```diff
@@ -0,0 +1,108 @@
...
+	copied := *w
+	copied.Environment = maps.Clone(w.Environment)
+	copied.UserID = copyPtr(w.UserID)
+	copied.GroupID = copyPtr(w.GroupID)
+	return &copied
+}
+
+func (w *Workload) Merge(other *Workload) {
```
flotter, line 55: Can be private ?

internals/workloads/workloads.go
```diff
@@ -0,0 +1,108 @@
...
+		return reflect.DeepEqual(w, other)
+	}
+	otherUID, otherGID, err := osutil.NormalizeUidGid(other.UserID, other.GroupID, other.User, other.Group)
+	if err != nil {
+		return reflect.DeepEqual(w, other)
+	}
+	if uid != nil && gid != nil && otherUID != nil && otherGID != nil {
+		return *uid == *otherUID && *gid == *otherGID
+	}
+	return reflect.DeepEqual(w, other)
```
flotter, line 92: The strategy currently for services are that ```plan.Service.Equal``` will do a ```DeepEqual``` between the old and new service to decide if a restart is needed. The resolution of the effective uid/gid only happens at the point it is started.

Equally, this method is only used during ```Replan``` - same place where we check of the service config changed. At this point we already know both workloads (old and new) are valid (```osutil.NormalizeUidGid``` is currently run during plan validation).

Would it not be simpler to just do a DeepEqual between the two workloads like services ? The only penalty if the odd case where someone went through all the effort to make a change to the relevant workload, file leaving the effective uid/gid the same.

internals/overlord/servstate/handlers.go
```diff
@@ -115,13 +118,21 @@ func (m *ServiceManager) doStart(task *state.Task, tomb *tomb.Tomb) error {
 	}
 
 	currentPlan := m.getPlan()
-	config, ok := currentPlan.Services[request.Name]
-	if !ok {
+	config, configFound := currentPlan.Services[request.Name]
+	if !configFound {
```
flotter, line 122: nit: Was the name originally used lower down? Just the rename on its own does not seem to add much value, especially with the error below describing it ?

internals/overlord/servstate/handlers.go
```diff
@@ -177,17 +190,23 @@ func (m *ServiceManager) serviceForStart(config *plan.Service) (service *service
 		service = &serviceData{
 			manager: m,
 			state:   stateInitial,
-			config:  config.Copy(),
 			logs:    servicelog.NewRingBuffer(maxLogBytes),
 			started: make(chan error, 1),
 			stopped: make(chan error, 2), // enough for killTimeElapsed to send, and exit if it happens after
 		}
+		service.config = config.Copy()
```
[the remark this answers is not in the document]
flotter: This is my view as well, than ```plan.Service``` should never be mutated once created, only the whole plan can be replaced, but that still leave previous instances valid and untouched.

internals/overlord/servstate/handlers.go
```diff
@@ -337,16 +354,38 @@ func (s *serviceData) startInternal() error {
...
-	// Copy environment to avoid updating original.
-	environment := make(map[string]string)
-	for k, v := range s.config.Environment {
-		environment[k] = v
+	var environment map[string]string
+	if len(s.config.Environment) > 0 {
+		environment = maps.Clone(s.config.Environment)
+	} else {
+		environment = make(map[string]string)
+	}
```
flotter, line 362: I think you can replace this block with:

```
environment := make(map[string]string)
maps.Copy(environment, s.config.Environment)
```

Copy seems to handle a nil source.

internals/overlord/servstate/handlers.go
```diff
@@ -337,16 +354,38 @@ func (s *serviceData) startInternal() error {
...
+	if s.config.GroupID != nil {
+		gid = s.config.GroupID
+	}
+	if s.config.User != "" {
+		username = s.config.User
+	}
+	if s.config.Group != "" {
+		groupname = s.config.Group
+	}
```
flotter, line 387: Should a workload not replace the service definition of user/group. 

1. Did we not say that if a workgroup is supplied, the service uid/gid,user and group should be undefined ? As in, a validation error must be reported if both the service and the workload supplies user information?

2. The design document mentions the concept of an implicit "default" workload for Pebble. Or is this optional ?

flotter (commented):
Looks close from my perspective @anpep!

internals/workloads/sectionextension.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return nil
+}
+
+const WorkloadsField = "workloads"
+
+var _ plan.Section = (*WorkloadsSection)(nil)
+
+type WorkloadsSection struct {
```
flotter, line 84: nit: This is not important for me, but I could not help to notice:

```
ws := currentPlan.Sections[workloads.WorkloadsField].(*workloads.WorkloadsSection)
```
My observation is that I think having Section in every concrete section type feels a bit unnecessary. Would this not read a little nicer like this:

```
workloads := currentPlan.Sections[workloads.WorkloadsField].(*workloads.Workloads)
```

The section is a Plan concept referring to the interface. What you have above is the concrete type backing the Plan section. It's just a fine line, but at the end I wonder if it will not be nicer to just work with the concrete type inside the service manager, and not still dragging the Plan section concept with you all the way :)

internals/workloads/sectionextension.go
anpep: About the internal error, I literally copied that from you, but Ben rightfully pointed out that it introduced a bit of noise that gave little to no value, so I just let the type assertion panic because the developer (us) is meant to know when to use it properly

flotter: Sorry my comment was unclear. I do not think the internal error is worth it, but I was just asking if the comment about an type assertion causing a panic (as apposed many other situations in code that can also cause panics) is worth explicitly documenting.

internals/overlord/servstate/handlers.go
flotter: @benhoyt Hi Ben! @anpep and myself was talking about this preference where user/group from the service would always replace the workload.

Do you think this is an important starting position to take? Once we expose that behavior, it feels like it would be difficult to go back on it later, as we would teach people that a workload combined with service user/group information can coexist.

I am worried about whether this idea is definitely future proof. If we start to see confinement features getting added to services, through workloads (as in two services may run in the same sandbox if they belong to the same workload), would it be a safe and desirable feature to have to allow service users to override the user/group of one service ?

My view would be why do we not start of preventing service level user / group  (```nil, nil, "", ""```) if a workload is defined, by failing validation ? This we we teach people you have to decide. I cannot see a clear case today why we need supporting both at the same time.

Any insights / ideas ?

benhoyt: I think you're on the right track. However, one concern I have is that you can't unset a workload in a subsequent layer, so you couldn't override a previous layer's workload with service-level config even if you wanted to. That doesn't seem great. Maybe have `workload: none` be a special value meaning unset the workload (and "none" would not be a valid workload name).

But then again, it does seem to me that at least the environment variables should be additive, because that's how environment variable overrides usually work. In other words, I think service.environment should override service.workload.environment.

In any case, if we prevent i now (validation error if you have workload and one of environment/user/group specified), that doesn't prevent us relaxing that later, so I'm open to that approach too.

benhoyt: I just now see @anpep's [comment](https://github.com/canonical/pebble/pull/571#discussion_r1973664249) that according to the spec, when the workload is specified, the environment of the service running is this workload is updated to include the workload values.

But user/group work the other way around (service overrides workload).

This seems inconsistent to me. I would have thought the more specific (service) should always override the less specific (workload). But at the least the environment and user/group should work the same, no?

flotter: @benhoyt Maybe just a quick brainstorm together with @anpep will be easier to find a solution here?

> However, one concern I have is that you can't unset a workload in a subsequent layer,

This is true, but you should be able to remove a service from a workload, or place a service into another workload, which I think gives you the same control ?

I see a workload as a sandbox owned by a user/group. Once this sandbox exists, it may contain valuable data from past services that run inside it. So a service can decide to run inside this sandbox, or it can later decide to run in another one, or no-one.

We treat this the same as "system users". Once it has been observed, we do not delete at, as data may exist on a system that is owned by it.

> But at the least the environment and user/group should work the same, no?

This is a good question. For me, since environment variables have an inheritance model at the process level (some comes from the parent, others are added later), the idea of the workload supplying some, and the service some more, feels natural to me ?

The user/group feels different to me, but I could view this wrong. The main issue for me is that the process user/group will affect the ownership of data created, and ownership information for access to other parts of the system. I am worried that a process running inside a sandbox, yet specifying its own user/group (especially since this is the most known way for existing Pebble users), could create setups that cannot be migrated to workloads).

For example, I had some idea that if you run two processes in the same workload, they can simple share data. Perhaps this is naive, but not allowing service level overrides could make this "just work".

internals/workloads/sectionextension.go
anpep: I added it as a suggestion from (I think) Ben

internals/workloads/sectionextension.go
anpep: That would be equivalent to the following workload:

```yaml
workload:
    default:
```

We definitely should support this, meaning "don't override service user/group information, and run as the Pebble user if the service does not include user/group information"

internals/overlord/servstate/handlers.go
benhoyt: To summarize our voice discussion, we've agreed on the following:

- the service environment would override the workload environment
- it would be a validation error to specify user/user-id/group/group-id on the service (in the combined plan) if a workload is specified
- for now at least, we won't document this feature in Pebble (so that Pebble "cloud" users don't start using it yet) but consider adding it as a first-class feature later, after projects importing Pebble have proved it out
- we will prohibit changes to workload via the add-layer API

anpep: Resolving this, as this was fixed yesterday (:

internals/workloads/workloads.go
```diff
@@ -0,0 +1,211 @@
...
+		default:
+			return &plan.FormatError{
+				Message: fmt.Sprintf(`workload %q: has an invalid "override" policy: %q`, name, workload.Override),
+			}
+		}
+	}
+	return nil
+}
+
+func (*Workloads) CombineSections(sections ...plan.Section) (plan.Section, error) {
```
flotter, line 139: This change where ```Workloads``` is now the extension type, and also creates ```Workloads``` seems a bit strange.

internals/overlord/planstate/manager.go
```diff
@@ -253,6 +256,10 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 	if err != nil {
 		return nil, err
 	}
+	newWorkloads, ok := p.Sections[workloads.WorkloadsField].(*workloads.Workloads)
+	if ok && !reflect.DeepEqual(newWorkloads, m.plan.Sections[workloads.WorkloadsField]) {
+		return nil, errors.New("plan workloads cannot be changed")
+	}
```
flotter, line 262: This can work - just wondering if there is not a way to hide this completely in the workloads extension system. I have an idea ...

benhoyt:
@anpep Is this ready for final review from me? Looks like it also has conflicts that need to be merged from master and resolved.

internals/overlord/servstate/manager_test.go
```diff
@@ -408,6 +414,70 @@ services:
...
+    test6:
+        override: replace
+        startup: enabled
+        command: /bin/test6
+        workload: default
+workloads:
+    default:
+        override: replace
+        user: nobody
+        group: nogroup
```
benhoyt, line 433: Wouldn't these fail to start unless running the tests as root? IIRC, a non-root user can't start a process as another user/group.

benhoyt (commented):
Just a couple of minor questions/comments from me before approval.

internals/overlord/servstate/handlers.go
```diff
@@ -357,32 +357,36 @@ func (s *serviceData) startInternal() error {
...
-		environment = maps.Clone(s.workload.Environment)
-	} else if len(s.config.Environment) > 0 {
-		environment = maps.Clone(s.config.Environment)
-	} else {
-		environment = make(map[string]string)
+	environment := make(map[string]string)
+	if s.workload != nil {
+		maps.Copy(environment, s.workload.Environment)
 	}
+	maps.Copy(environment, s.config.Environment)
```
paul-rodriguez, line 364: Is it the intended behavior to merge the two environment maps in case both the one from the workload and the config are non-empty?

internals/workloads/workloads.go
```diff
@@ -120,98 +108,104 @@ func (ws *WorkloadsSection) Validate() error {
...
+	if ext.currentPlan != nil {
+		currentWorkloads := ext.currentPlan.Sections[WorkloadsField].(*WorkloadsSection)
+		if !reflect.DeepEqual(currentWorkloads.Entries, ws.Entries) {
+			return &plan.FormatError{
+				Message: "cannot change workloads once the plan has been loaded",
+			}
+		}
 	}
-	return reflect.DeepEqual(w, other)
+	ext.currentPlan = p
```
paul-rodriguez, line 207: I would not have guessed that the method called "ValidatePlan" also mutates the `ext` object. 

I suggest either 
* changing the name of the function to reflect this, because ultimately it's the most important thing that's happening in this function or 
* separating the implementation of validation and this assignment into a different function.

paul-rodriguez (approved):
Just a little structure comment

internals/overlord/planstate/manager.go
```diff
@@ -55,6 +56,11 @@ func NewManager(layersDir string) (*PlanManager, error) {
 // the case of a non-existent layers directory, or no layers in the layers
 // directory, an empty plan is announced to change subscribers.
 func (m *PlanManager) Load() error {
+	if !reflect.DeepEqual(m.plan, &plan.Plan{}) {
```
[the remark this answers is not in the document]
anpep: We can't compare to `nil` because the initial value for a valid plan (see `New`) is `&plan.Plan{}` and this seems to be an assumption that's made throughout the Pebble codebase (changing it is breaking a lot of stuff).

I don't think it's a problem to call `Load()` twice, but I made an observation that this is called only once in the overlord `New()` function, and that it could be useful for other applications extending Pebble that if a plan was somehow loaded before Pebble gets the chance to, we could effectively control the plan that the overlord consumes on creation.

To this aim, I introduced the `(*PlanManager).Init(*plan.Plan)` function, equivalent to `Load()` but supplying a plan instance. The idea is that if the Plan has been `Load`ed or `Init`ed already, it won't be overriden by the `Load()` call in the overlord `New()` function.

internals/overlord/planstate/manager.go
```diff
@@ -68,6 +74,19 @@ func (m *PlanManager) Load() error {
 	return nil
 }
 
+// InitializePlan loads the plan from an existing instance and announced to
+// change subscribers.
+func (m *PlanManager) InitializePlan(p *plan.Plan) {
+	if !reflect.DeepEqual(m.plan, &plan.Plan{}) {
```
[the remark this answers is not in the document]
anpep: (See comment above)

internals/overlord/servstate/manager_test.go
anpep: That is true, however, for this test specifically I do not start the services--only replan and test whether the services would be started after changing workload configuration.

internals/overlord/planstate/manager.go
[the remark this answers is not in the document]
benhoyt: Got it, thanks for the explanation.

benhoyt (approved):
Tests are failing (data race?), but code looks good to me now -- approving.

internals/cli/cmd_enter_test.go
```diff
@@ -172,6 +175,7 @@ func (s *PebbleSuite) TestEnterExecListDir(c *C) {
 }
 
 func (s *PebbleSuite) TestEnterExecReadServiceOutputFile(c *C) {
+	plan.UnregisterSectionExtension(workloads.WorkloadsField)
```
flotter, line 178: It is not clear to me when this is needed, or why this is needed. It is not consistently needed, which makes me think I do not understand. Can this not be, for example, moved to the test teardown with an explanation why this is needed ?

internals/overlord/planstate/manager.go
```diff
@@ -55,19 +56,41 @@ func NewManager(layersDir string) (*PlanManager, error) {
 // the case of a non-existent layers directory, or no layers in the layers
 // directory, an empty plan is announced to change subscribers.
 func (m *PlanManager) Load() error {
+	m.planLock.Lock()
+	if !reflect.DeepEqual(m.plan, &plan.Plan{}) {
```
flotter, line 60: nit: This is a lot of code to effectively determine if we have ```Init``` or ```Load``` before, and it is also slightly odd in the sense that if someone explicitly init it to ```plan.Plan{}```, that is not recognized, and it will allow the ```Load``` to replace the explicitly set empty plan.

Would it not be simpler to just add a flag in the plan manager tracking this instead ? ```planInitialized bool``` ?

flotter:
```go
	plan, err := plan.ReadDir(m.layersDir)
	if err != nil {
		return err
	}

	m.planLock.Lock()
        if !m.planInitialized {
	    m.plan = plan
	    m.planInitialized = true            
	}
	m.planLock.Unlock()

	m.callChangeListeners(plan)
	return nil
```

flotter:
```
func (m *PlanManager) Load() (err error) {
    m.planLock.Lock()
    defer func() {
        m.planLock.Unlock()
        if err != nil {
            m.callChangeListeners(plan)
        }
    }
    
    plan, err := plan.ReadDir(m.layersDir)
    if err != nil {
	return err
    }
    if !m.planInitialized {
        m.plan = plan
        m.planInitialized = true
    }
    return nil
```

internals/overlord/planstate/manager_test.go
```diff
@@ -588,3 +589,31 @@ func (ps *planSuite) TestAppendLayersWithInner(c *C) {
 		c.Assert(plan.Layers[i].Label, Equals, layer.label)
 	}
 }
+
+func (ps *planSuite) TestAppendWorkloadLayer(c *C) {
```
flotter, line 593: Should you also test the combine ?

internals/overlord/servstate/handlers.go
```diff
@@ -340,16 +357,34 @@ func (s *serviceData) startInternal() error {
...
 
 	s.cmd.Dir = s.config.WorkingDir
 
 	// Start as another user if specified in plan.
-	uid, gid, err := osutil.NormalizeUidGid(s.config.UserID, s.config.GroupID, s.config.User, s.config.Group)
+	var uid, gid *int
+	var username, groupname string
+	if s.workload != nil {
```
flotter, line 371: As per our previous conversation:

```
[Ben's notes in the comment below] it would be a validation error to specify user/user-id/group/group-id on the service (in the combined plan) if a workload is specified
```

Is this true in the implementation ? The code looks like service uid/gid would replace workloads. Is it prevented at a validation phase ?

flotter (commented):
Final small couple of nits

internals/overlord/servstate/handlers.go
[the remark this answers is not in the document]
flotter: Last point discussed. Resolving.

internals/overlord/planstate/manager.go
```diff
@@ -36,8 +36,9 @@ func (e *LabelExists) Error() string {
 type PlanManager struct {
 	layersDir string
 
-	planLock sync.Mutex
-	plan     *plan.Plan
+	planLock      sync.Mutex
+	plan          *plan.Plan
+	isInitialized bool
```
flotter, line 41: ```isLoaded```

internals/overlord/planstate/manager.go
```diff
@@ -54,18 +55,44 @@ func NewManager(layersDir string) (*PlanManager, error) {
...
 // directory, an empty plan is announced to change subscribers.
-func (m *PlanManager) Load() error {
-	plan, err := plan.ReadDir(m.layersDir)
+func (m *PlanManager) Load() (err error) {
+	var p *plan.Plan
+
+	m.planLock.Lock()
+	defer func() {
+		m.planLock.Unlock()
+		if err == nil && p != nil {
```
flotter, line 64: remove ```p != nil```

flotter: replace with local ```!isInitialized```

internals/overlord/servstate/handlers.go
```diff
@@ -340,16 +357,39 @@ func (s *serviceData) startInternal() error {
...
 
-	// Copy environment to avoid updating original.
 	environment := make(map[string]string)
-	for k, v := range s.config.Environment {
-		environment[k] = v
+	if s.workload != nil {
+		maps.Copy(environment, s.workload.Environment)
 	}
+	// If both workload and service provides the same environment variable,
+	// the workload environment prevails.
```
flotter, line 365: comment: I think we agreed the service environment must trump. Is that what the comment is saying ? It sounds inverted to me?

flotter (approved):
Thanks @anpep Looks good to me.

flotter:
@benhoyt We made 3 tweaks I would love to get your re-approval on:

1. Please see planstate ```Load()``` - We ignore load if ```Init()``` was called previously, now based on a flag.

2. Please see planstate ```Init()``` - We ignore init if ```Load()``` was called previously, now based on a flag, only for symmetry.

3. Please see plan ```NewPlan()``` - we uncovered a bug. The design assumption for the plan and extensions is that a plan **always** contains all the required content present (even if empty), which simplifies layer merging and callback receiver code, especially when extensions comes in. This rule was broken in the case where no layers directory existed, in which case we returned an empty plan without the extension sections. This broke ```pebble plan``` and also ```pebble replan```.

flotter (approved):

benhoyt (approved):
Updates look reasonable to me, thanks.

PR #572, opened 2025-02-13
feat: sort the changed checks in the API response

tonyandrewmeyer:
The PR changes the response to a `/v1/checks` POST to provide the list of names of started or stopped checks as sorted alphabetically.

A test is added to confirm this, which also tests the API post method more generally, which was missed in #560

This will also provide a sorted list of check names in the client (since it uses the API) and the CLI (since it uses the client). The *manager* continues to return the list of check names in the order in which they were started/stopped - this feels more correct to me (in terms of management, what matters is what happened; in terms of response to a request, convenience is important), but I could be convinced that the sorting should move up a level.

### DIFF
--- a/internals/daemon/api_checks.go
+++ b/internals/daemon/api_checks.go
@@ -17,6 +17,7 @@ package daemon
 import (
 	"encoding/json"
 	"net/http"
+	"sort"
 
 	"github.com/canonical/x-go/strutil"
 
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Thanks, that placement (in the API) sounds right to me too. Thanks for the additional test.

PR #574, opened 2025-02-19
feat: add pebble check command

IronCore864:
This PR adds a new "pebble check" command, which takes a single check name as a required positional argument and returns the status of the check synchronously.

This new subcommand is needed for the upcoming `pebble check <check> --refresh` feature where a check will run immediately. See [OP055](https://docs.google.com/document/d/1fDfb8FL9urUWhcvExYEWUXDTUq--a22V-v0GaRuLyFk/edit?tab=t.0) (internal link) for more details. Changes:

- Reuse existing `/v1/checks` API endpoint. A new endpoint will be added for the `--refresh` feature in a separate PR.
- Docs and unit tests are updated.
- Note: The `level` argument/filter from `pebble checks` (plural) is not copied to this new subcommand because this command only gets a single check by name, so the "filter by level" feature probably is not needed, return the check by name in any case.

Some manual tests:

```bash
$ pebble check
error: the required argument `<check>` was not provided

$ pebble check chk1 chk2
error: too many arguments for command

$ ./pebble check chk1
No matching health checks.

$ ./pebble check check1
name: check1
level: ""
startup: enabled
status: up
failures: 0
threshold: 3
changeid: "1"
```


internals/cli/cmd_help.go
```diff
@@ -201,7 +201,7 @@ var HelpCategories = []HelpCategory{{
 }, {
 	Label:       "Checks",
 	Description: "manage health checks",
-	Commands:    []string{"checks", "start-checks", "stop-checks", "health"},
+	Commands:    []string{"check", "checks", "start-checks", "stop-checks", "health"},
```
benhoyt, line 204: I know it's simpler so it sort of makes sense for it to be first, but I think we should have `checks` first, then `check`, as the summary provided by `checks` is what most people should use first I think, and these are ordered logically (at least we attempt to).

internals/cli/cmd_check.go
```diff
@@ -0,0 +1,74 @@
...
+
+	opts := client.ChecksOptions{
+		Names: []string{cmd.Positional.Check},
+	}
+	checks, err := cmd.client.Checks(&opts)
+	if err != nil {
+		return err
+	}
+	if len(checks) == 0 {
+		fmt.Fprintln(Stderr, "No matching health checks.")
```
benhoyt, line 63: Because they're asking for a single check by name, I think this should be an error (which will return nonzero). Similar to `cmdIdentity`:

```go
		return fmt.Errorf("cannot find check %q", cmd.Positional.Check)
```

internals/cli/cmd_check.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil
+	}
+
+	check := checks[0]
+	data, err := yaml.Marshal(check)
```
benhoyt, line 68: I think we should have a separate `type checkInfo struct` for this marshalling in this package, and copy the fields over. That way we can add the `yaml:` tags and (later) the error details needed for `--refresh`.

internals/cli/cmd_check_test.go
```diff
@@ -0,0 +1,70 @@
...
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
+	c.Check(s.Stdout(), check.Equals, `
+name: chk1
+level: ""
+startup: enabled
+status: up
+failures: 0
+threshold: 3
+changeid: "1"
```
benhoyt, line 49: Notice how the change ID is coming through. Should be `change-id`. Per my previous example that I wrote in the spec, I think this should be:

```
name: chk1
level: alive      # omitted if empty
startup: enabled
status: up
failures: 0       # included if 0
threshold: 3
change-id: "42"   # omitted if empty
```

So level and change-id would be marked `omitempty`. I think it's good to keep "failures" without omitempty, so that the user can clearly see there were 0 failures if so.

benhoyt (changes requested):
Nice and simple, thanks. Just a few comments.

benhoyt (approved):

PR #576, opened 2025-02-25
feat: verbose mode using environment variable

IronCore864:
This PR adds a new environment variable "PEBBLE_VERBOSE", which if set to "1", Pebble will print service logs to Pebble's standard output. See [this spec](https://docs.google.com/document/d/1-NCelXSb4dVjZuEGDp_jsTnzV0idq6BXqZ4kGCkO9Nw/edit?tab=t.0) (internal only) for more details.

Summary:

- For `pebble run`, allow verbose logging mode (log all output from services to stdout) to be enabled when starting the daemon by setting the environment variable `PEBBLE_VERBOSE=1`.  This is equivalent to `pebble run --verbose`. Even if `PEBBLE_VERBOSE` is set to "0" explicitly, if the `--verbose` flag is still specified, the verbose logging mode is still enabled: the `--verbose` flag has higher priority.
- For `pebble enter exec`, the `--verbose` flag is currently disabled. However, `pebble enter` (including `pebble enter exec`) would still respect `PEBBLE_VERBOSE=1`.
- Some tests are added to cover the function of this environment variable.
- A new reference doc is created to document all existing environment variables for Pebble.

### DIFF
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -1,56 +1,33 @@
 # Pebble Environment Variables
 
-## NOTIFY_SOCKET
-
-If the `NOTIFY_SOCKET` environment variable is set, Pebble will send state string notifications to systemd. Specifically:
-
-- When Pebble daemon starts, it sends "READY=1", which tells systemd that Pebble startup is finished.
-- When Pebble daemon stops, it sends "STOPPING=1", which tells systemd that Pebble is beginning its shutdown. 
-
 ## PEBBLE
 
-Pebble's configuration directory. Defaults to "/var/lib/pebble/default" if not specified.
+Pebble's configuration directory. Defaults to `/var/lib/pebble/default` if not specified.
 
 The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of configuration files. See [general model](../explanation/general-model) for more information.
 
 ## PEBBLE_COPY_ONCE
 
-If specified, when Pebble daemon starts, Pebble copies the directory specified by `PEBBLE_COPY_ONCE` to the Pebble directory once.
+To initialise the `$PEBBLE` directory with the contents of another, in a one-time copy, set the `PEBBLE_COPY_ONCE` environment variable to the source directory.
+
+This will only copy the contents if the target directory, `$PEBBLE`, is empty.
 
 ## PEBBLE_DEBUG
 
-If set to "1", debug logs will be printed.
-
-## PEBBLE_REBOOT_DELAY
-
-A duration string for the delay of reboot. Used for tests.
-
-Reboot delay defaults to 1 minute if `PEBBLE_REBOOT_DELAY` is not set.
+If set to "1", debug logs will be printed to stderr.
 
 ## PEBBLE_SOCKET
 
-Pebble socket path. Defaults to `$PEBBLE/.pebble.socket` if not specified.
+Pebble socket path. Defaults to `$PEBBLE/.pebble.socket` if not specified, or `/var/lib/pebble/default/.pebble.socket` if `PEBBLE` is not set.
 
 ## PEBBLE_VERBOSE
 
-If set to "1", write service logs to Pebble's stdout.
+If set to "1", the Pebble daemon writes service logs to stdout.
 
-For `pebble run`, allow verbose logging mode (log all output from services to stdout) to be enabled when starting the daemon by setting the environment variable `PEBBLE_VERBOSE=1`.  This is in addition to the existing way of enabling verbose mode with a command line argument: `pebble run --verbose`.
+For `pebble run`, either `PEBBLE_VERBOSE=1` or the `--verbose` flag turns on verbose logging, with the command line flag overriding the environment variable.
 
-For pebble enter exec, the `--verbose flag` is currently disabled. However, `pebble enter` (including `pebble enter exec`) would still respect `PEBBLE_VERBOSE=1`: the user should know how their applications behave and it’s okay to use with verbose logging turned on.
-
-## SPREAD_SYSTEM
-
-If set, Pebble will think we are running tests, and the progress bar for commands `autorestart`, `replan`, `restart`, `start` and `stop` will be disabled.
-
-## UNSAFE_IO
-
-If set to "1," sync for testing is disabled. This brings massive improvements on certain filesystems (like btrfs) and noticeable improvements in all unit tests in general.
-
-## WATCHDOG_USEC
-
-If the `WATCHDOG_USEC` environment variable is set, systemd expects notifications from Pebble. Systemd will usually terminate a service when it does not get a notification message within the specified time after startup and after each previous message. It is recommended to send a keep-alive notification message every half of the time (in microseconds) specified by `WATCHDOG_USEC`. Notification messages are sent with `sd_notify` with a message string of "WATCHDOG=1".
+For `pebble enter exec`, the `--verbose` flag is currently disallowed. However, `pebble enter` (including `pebble enter exec`) still respects the `PEBBLE_VERBOSE=1` environment variable: the user should know how their applications behave, and that they're okay to use with verbose logging turned on.
 
 ## XDG_CONFIG_HOME
 
-Pebble CLI state path. Defaults to "$HOME/.config" if not specified.
+The [XDG configuration directory](https://specifications.freedesktop.org/basedir-spec/latest/#basics). Certain Pebble CLI commands create or use data files in `$XDG_CONFIG_HOME/pebble`. Defaults to `$HOME/.config` if not specified.
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -4,7 +4,7 @@
 
 Pebble's configuration directory. Defaults to `/var/lib/pebble/default` if not specified.
 
-The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of configuration files. See [general model](../explanation/general-model) for more information.
+The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of configuration files. See [general model](../explanation/general-model) and [How to use layers](../how-to/use-layers) for more information.
 
 ## PEBBLE_COPY_ONCE
 
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -1,0 +1,56 @@
+# Pebble Environment Variables
+
+## NOTIFY_SOCKET
+
+If the `NOTIFY_SOCKET` environment variable is set, Pebble will send state string notifications to systemd. Specifically:
+
+- When Pebble daemon starts, it sends "READY=1", which tells systemd that Pebble startup is finished.
+- When Pebble daemon stops, it sends "STOPPING=1", which tells systemd that Pebble is beginning its shutdown. 
+
+## PEBBLE
+
+Pebble's configuration directory. Defaults to "/var/lib/pebble/default" if not specified.
+
+The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of configuration files. See [general model](../explanation/general-model) for more information.
+
+## PEBBLE_COPY_ONCE
+
+If specified, when Pebble daemon starts, Pebble copies the directory specified by `PEBBLE_COPY_ONCE` to the Pebble directory once.
+
+## PEBBLE_DEBUG
+
+If set to "1", debug logs will be printed.
+
+## PEBBLE_REBOOT_DELAY
+
+A duration string for the delay of reboot. Used for tests.
+
+Reboot delay defaults to 1 minute if `PEBBLE_REBOOT_DELAY` is not set.
+
+## PEBBLE_SOCKET
+
+Pebble socket path. Defaults to `$PEBBLE/.pebble.socket` if not specified.
+
+## PEBBLE_VERBOSE
+
+If set to "1", write service logs to Pebble's stdout.
+
+For `pebble run`, allow verbose logging mode (log all output from services to stdout) to be enabled when starting the daemon by setting the environment variable `PEBBLE_VERBOSE=1`.  This is in addition to the existing way of enabling verbose mode with a command line argument: `pebble run --verbose`.
+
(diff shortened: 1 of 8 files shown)


IronCore864:
Since `enter exec` prints strings as stdout instead of `servicelog.Entry`, in the current draft, a new helper function is created to return channels of string. There is some duplicated code, I did another draft to unify `pebble run` and `pebble enter`, something like:


```go
//go:build integration

// Copyright (c) 2024 Canonical Ltd
//
// This program is free software: you can redistribute it and/or modify
// it under the terms of the GNU General Public License version 3 as
// published by the Free Software Foundation.
//
// This program is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU General Public License for more details.
//
// You should have received a copy of the GNU General Public License
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

package tests

import (
	"bufio"
	"flag"
	"fmt"
	"io"
	"os"
	"os/exec"
	"path/filepath"
	"strings"
	"testing"
	"time"

)

var pebbleBin = flag.String("pebblebin", "", "Path to the pre-built Pebble binary")

// TestMain builds the pebble binary of `-pebblebin` flag is not set
// before running the integration tests.
func TestMain(m *testing.M) {
	flag.Parse()

	if *pebbleBin == "" {
		goBuild := exec.Command("go", "build", "-o", "../pebble", "../cmd/pebble")
		if err := goBuild.Run(); err != nil {
			fmt.Println("Cannot build pebble binary:", err)
			os.Exit(1)
		}
		*pebbleBin = "../pebble"
	} else {
		// Use the pre-built Pebble binary provided by the pebbleBin flag.
		fmt.Println("Using pre-built Pebble binary at:", *pebbleBin)
	}

	exitCode := m.Run()
	os.Exit(exitCode)
}

// createLayer creates a layer file with layerYAML under the directory "pebbleDir/layers".
func createLayer(t *testing.T, pebbleDir, layerFileName, layerYAML string) {
	t.Helper()

	layersDir := filepath.Join(pebbleDir, "layers")
	err := os.MkdirAll(layersDir, 0o755)
	if err != nil {
		t.Fatalf("Cannot create layers directory: %v", err)
	}

	layerPath := filepath.Join(layersDir, layerFileName)
	err = os.WriteFile(layerPath, []byte(layerYAML), 0o755)
	if err != nil {
		t.Fatalf("Cannot create layers file: %v", err)
	}
}

// OutputReader defines an interface for reading output, allowing us to handle
// both log entries and strings (for pebble enter exec output).
type OutputReader interface {
	Read(t *testing.T) (interface{}, bool)
}

// LogReader reads servicelog.Entry from a channel.
type LogReader struct {
	logsCh <-chan servicelog.Entry
}

func (l *LogReader) Read(t *testing.T) (interface{}, bool) {
	log, ok := <-l.logsCh
	if !ok {
		t.Error("log channel closed before all expected logs were received")
	}
	return log, ok
}

// StringReader reads strings from a channel.
type StringReader struct {
	textCh <-chan string
}

func (s *StringReader) Read(t *testing.T) (interface{}, bool) {
	text, ok := <-s.textCh
	if !ok {
		t.Error("string channel closed before expected text was received")
	}
	return text, ok
}

// pebbleCommand runs a pebble command (run or enter) with optional arguments,
// and returns two channels for standard output and standard error.
func pebbleCommand(t *testing.T, pebbleDir, command string, args ...string) (stdoutReader, stderrReader OutputReader) {
	t.Helper()

	cmd := exec.Command(*pebbleBin, append([]string{command}, args...)...)
	cmd.Env = append(os.Environ(), "PEBBLE="+pebbleDir)

	stdoutPipe, err := cmd.StdoutPipe()
	if err != nil {
		t.Fatalf("Cannot create stdout pipe: %v", err)
	}
	stderrPipe, err := cmd.StderrPipe()
	if err != nil {
		t.Fatalf("Cannot create stderr pipe: %v", err)
	}

	err = cmd.Start()
	if err != nil {
		t.Fatalf("Error starting 'pebble %s': %v", command, err)
	}

	stopStdout := make(chan struct{})
	stopStderr := make(chan struct{})

	t.Cleanup(func() {
		err := cmd.Process.Signal(os.Interrupt)
		if err != nil {
			t.Errorf("Error sending SIGINT/Ctrl+C to pebble: %v", err)
		}
		cmd.Wait()
		close(stopStdout)
		close(stopStderr)
	})

	if command == "run" {
		stdoutCh := make(chan servicelog.Entry)
		stderrCh := make(chan servicelog.Entry)
		go readLogs(t, stdoutPipe, stdoutCh, stopStdout)
		go readLogs(t, stderrPipe, stderrCh, stopStderr)
		return &LogReader{stdoutCh}, &LogReader{stderrCh}

	} else if command == "enter" { // Assuming "enter" is the other command
		stdoutCh := make(chan string)
		stderrCh := make(chan string)
		go readLines(t, stdoutPipe, stdoutCh, stopStdout)
		go readLines(t, stderrPipe, stderrCh, stopStderr)
		return &StringReader{stdoutCh}, &StringReader{stderrCh}
	}

	return nil, nil // Or handle other commands appropriately
}

func readLogs(t *testing.T, reader io.Reader, ch chan servicelog.Entry, stop <-chan struct{}) {
	parser := servicelog.NewParser(reader, 4*1024)
	for parser.Next() {
		if err := parser.Err(); err != nil {
			t.Errorf("Cannot parse Pebble logs: %v", err)
		}
		select {
		case ch <- parser.Entry():
		case <-stop:
			return
		}
	}
}

func readLines(t *testing.T, reader io.Reader, ch chan string, stop <-chan struct{}) {
	scanner := bufio.NewScanner(reader)
	for scanner.Scan() {
		select {
		case ch <- scanner.Text():
		case <-stop:
			return
		}
	}
	if err := scanner.Err(); err != nil {
		t.Errorf("Error reading output: %v", err)
	}
}

// waitForOutput waits for specific output, whether a log entry or string.
func waitForOutput(t *testing.T, reader OutputReader, expected interface{}, timeout time.Duration) {
	t.Helper()

	timeoutCh := time.After(timeout)
	for {
		select {
		case <-timeoutCh:
			t.Fatalf("timed out after %v waiting for output: %#v", timeout, expected)
		default:
			output, ok := reader.Read(t)
			if !ok {
				return
			}

			switch e := expected.(type) {
			case string:
				if o, ok := output.(string); ok && strings.Contains(o, e) {
					return
				}
			case servicelog.Entry:
				if o, ok := output.(servicelog.Entry); ok {
					if o.Service == e.Service && strings.Contains(o.Message, e.Message) {
						return
					}
				}
			default:
				t.Fatalf("unexpected expected type: %T", expected)
			}
		}
	}
}

// waitForFile waits until a file exists, or fails the test after a specified timeout
// if the file still doesn't exist.
func waitForFile(t *testing.T, file string, timeout time.Duration) {
	t.Helper()

	timeoutCh := time.After(timeout)
	ticker := time.NewTicker(time.Millisecond)
	for {
		select {
		case <-timeoutCh:
			t.Fatalf("timeout waiting for file %s", file)

		case <-ticker.C:
			stat, err := os.Stat(file)
			if err == nil && stat.Mode().IsRegular() {
				return
			}
		}
	}
}

// runPebbleCommand runs a pebble command and returns the standard output.
func runPebbleCommand(t *testing.T, pebbleDir string, args ...string) string {
	t.Helper()

	cmd := exec.Command(*pebbleBin, args...)
	cmd.Env = append(os.Environ(), "PEBBLE="+pebbleDir)

	output, err := cmd.CombinedOutput()
	if err != nil {
		t.Fatalf("error executing pebble command: %v", err)
	}

	return string(output)
}

```

I think it's a bit too complicated and maybe not easy to use just for the sake of removing duplicated code, so I didn't use this version in this PR.

tests/enter_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"time"
+)
+
+func TestEnterExec(t *testing.T) {
+	pebbleDir := t.TempDir()
+
+	script := `
+#!/bin/sh
+echo "hello world"
+sleep 2
```
benhoyt, line 32: This makes the test take 2+ seconds to run. Is that necessary? In my tests I can change it to sleep 0.1 and the test still works fine (I guess the start change would failing but that's probably okay?). Or worst case, if there's some reason that's bad, we could make it `sleep 1.1` (just over 1s).

tests/run_test.go
```diff
@@ -142,6 +142,53 @@ services:
...
+	waitForLog(t, stderrCh, "pebble", "Started daemon", 3*time.Second)
+	waitForLog(t, stdoutCh, "svc1", "hello world", 3*time.Second)
+}
+
+// TestVerboseFlagOverrideEnvVar tests that Pebble logs all output from services to stdout
+// with the environment variable `PEBBLE_VERBOSE` set to "0" but also with the `--verbose`
+// option.
+func TestVerboseFlagOverrideEnvVar(t *testing.T) {
+	os.Setenv("PEBBLE_VERBOSE", "0")
+	defer os.Setenv("PEBBLE_VERBOSE", "")
```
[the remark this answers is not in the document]
benhoyt: Similar elsewhere.

tests/enter_test.go
benhoyt: Similar below.

tests/enter_test.go
```diff
@@ -0,0 +1,65 @@
...
+	`
+	path := filepath.Join(pebbleDir, "test.sh")
+	err := os.WriteFile(path, []byte(script), 0755)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	stdoutCh, _ := pebbleEnter(t, pebbleDir, "exec", "/bin/sh", "-c", path)
+	waitForText(t, stdoutCh, "hello world", 1*time.Second)
```
benhoyt, line 41: We use `3*time.Second` in the run tests, I guess to give slow CI a bit more of a chance. Let's stick to that timeout throughout.

tests/main_test.go
```diff
@@ -131,6 +133,64 @@ func pebbleRun(t *testing.T, pebbleDir string, args ...string) (stdoutCh chan se
 	return stdoutCh, stderrCh
 }
 
+// pebbleEnter runs `pebble enter` with optional arguments
+// and returns two channels for standard output and standard error.
+func pebbleEnter(t *testing.T, pebbleDir string, args ...string) (stdoutCh chan string, stderrCh chan string) {
```
benhoyt, line 138: I saw your comment about trying to merge this with `pebbleRun` with an interface -- but I think it's a lot simpler than that. We can basically use what you have as `pebbleEnter` for both run and enter (I've called it `pebbleDaemon`), so it returns lines. And then change `waitForLog` to use `servicelog.Parse` on each line.

https://gist.github.com/benhoyt/4cfe74acaa366cdc8a4fc7f5e3a1d9eb

benhoyt (changes requested):
Thanks for this. Love how we have an integration test framework now -- thanks for adding those.

benhoyt:
@rebornplusplus Would you mind double-checking the modified logic here in `cmd_run.go` and `cmd_enter.go`?

tests/run_test.go
```diff
@@ -142,6 +142,53 @@ services:
...
+	createLayer(t, pebbleDir, layersFileName, layerYAML)
+
+	stdoutCh, stderrCh := pebbleRun(t, pebbleDir)
+	waitForLog(t, stderrCh, "pebble", "Started daemon", 3*time.Second)
+	waitForLog(t, stdoutCh, "svc1", "hello world", 3*time.Second)
+}
+
+// TestVerboseFlagOverrideEnvVar tests that Pebble logs all output from services to stdout
+// with the environment variable `PEBBLE_VERBOSE` set to "0" but also with the `--verbose`
+// option.
```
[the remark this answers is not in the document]
IronCore864: Agreed, the func name is already quite self-explanatory. I will remove this.

tests/enter_test.go
IronCore864: I made it 1s initially and met some issues so I increased it. I will do more tests with 1.1 and verify again.

docs/reference/environment-variables.md
```diff
@@ -0,0 +1,33 @@
+# Pebble Environment Variables
+
+## PEBBLE
+
+Pebble's configuration directory. Defaults to `/var/lib/pebble/default` if not specified.
+
+The `$PEBBLE` directory must contain a `layers/` subdirectory that holds a stack of configuration files. See [general model](../explanation/general-model) for more information.
```
benhoyt, line 7: Nit: would a better place to link be the layers how-to?

docs/reference/index.md
```diff
@@ -41,6 +42,13 @@ The `pebble` command has several subcommands.
 * [CLI commands](cli-commands)
 
 
+## Pebble environment variables
```
benhoyt, line 45: Let's drop the the word Pebble here:
```suggestion
## Environment variables
```

tests/main_test.go
```diff
@@ -91,7 +94,7 @@ func pebbleRun(t *testing.T, pebbleDir string, args ...string) (stdoutCh chan se
 
 	err = cmd.Start()
 	if err != nil {
-		t.Fatalf("Error starting 'pebble run': %v", err)
+		t.Fatalf("Error starting 'pebble enter': %v", err)
```
benhoyt, line 97:
```suggestion
		t.Fatalf("Error starting 'pebble %s': %v", runOrEnter, err)
```

tests/main_test.go
```diff
@@ -154,6 +157,29 @@ func waitForLog(t *testing.T, logsCh <-chan servicelog.Entry, expectedService, e
...
+// specified timeout if the expectedText is still not found.
+func waitForText(t *testing.T, textCh <-chan string, expectedText string, timeout time.Duration) {
+	t.Helper()
+
+	timeoutCh := time.After(timeout)
+	for {
+		select {
+		case text, ok := <-textCh:
+			if !ok {
+				t.Error("channel closed before expected text was received")
```
benhoyt, line 170: This should probably fail the test immediately, that is, be `t.Fatal`.

benhoyt (approved):
This is looking really good now, thanks. Just a couple of minor suggestions before merging.

PR #577, opened 2025-02-26
feat: pebble check --refresh

IronCore864:
Run a check immediately with `pebble check --refresh <check>`.

A new API endpoint `/v1/checks/refresh` with `POST` is added with the admin access level. 

Some manual tests:

```bash
$ # check success
$ pebble check chk1
name: chk1
startup: enabled
status: up
failures: 0
threshold: 3
change-id: "1"

$ # check failures > 0
$ pebble check chk2
name: chk2
startup: enabled
status: down
failures: 8
threshold: 3
change-id: "4"
logs: |
    2025-03-05T16:35:02+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:07+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:12+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:17+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:22+08:00 ERROR non-2xx status code 500; Health check failed

$ # refresh success
$ pebble check chk1 --refresh
name: chk1
startup: enabled
status: up
failures: 0
threshold: 3
change-id: "1"

$ # refresh failure
$ pebble check chk2 --refresh
name: chk2
startup: enabled
status: down
failures: 11
threshold: 3
change-id: "4"
error: non-2xx status code 500
logs: |
    2025-03-05T16:35:02+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:07+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:12+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:17+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:22+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:27+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:32+08:00 ERROR non-2xx status code 500; Health check failed
    2025-03-05T16:35:35+08:00 ERROR non-2xx status code 500; Health check failed
$ echo $?
0

$ # refresh some other error
$ pebble check chk3 --refresh
name: chk3
startup: enabled
status: up
failures: 2
threshold: 3
change-id: "3"
error: check timed out after 1s
logs: |
    2025-03-05T16:38:32+08:00 ERROR check timed out after 1s
```


client/checks.go
```diff
@@ -172,3 +172,38 @@ func (client *Client) doMultiCheckAction(actionName string, checks []string) (*C
 
 	return results, nil
 }
+
+type RefreshCheckOptions struct {
+	// Name of the check to refresh, required.
```
benhoyt, line 177: Nit for style and consistency:
```suggestion
	// Name of the check to refresh (required).
```

client/checks.go
```diff
@@ -172,3 +172,38 @@ func (client *Client) doMultiCheckAction(actionName string, checks []string) (*C
...
+	Name string
+}
+
+type RefreshCheckResult struct {
+	Info *CheckInfo `json:"info"`
+	// The error message, empty if success.
+	Error string `json:"err"`
```
benhoyt, line 184: Let use the JSON tag "error":
```suggestion
	Error string `json:"error"`
```

client/checks.go
```diff
@@ -172,3 +172,38 @@ func (client *Client) doMultiCheckAction(actionName string, checks []string) (*C
...
 	return results, nil
 }
+
+type RefreshCheckOptions struct {
+	// Name of the check to refresh, required.
+	Name string
+}
+
+type RefreshCheckResult struct {
+	Info *CheckInfo `json:"info"`
```
benhoyt, line 182: Should this be a non-pointer, seeing it's always present, and essentially embedding the CheckInfo in this struct?
```suggestion
	Info CheckInfo `json:"info"`
```

client/checks.go
```diff
@@ -172,3 +172,38 @@ func (client *Client) doMultiCheckAction(actionName string, checks []string) (*C
...
+	// The error message, empty if success.
```
benhoyt, line 183:
```suggestion
	// The error message from running the check; empty string on success.
```

client/checks_test.go
```diff
@@ -109,3 +109,39 @@ func (cs *clientSuite) TestStopChecks(c *check.C) {
...
+		"result": {
+			"info": {
+				"name": "chk1",
+				"startup": "enabled",
+				"status": "up",
+				"failures": 0,
+				"threshold": 3,
+				"change-id": "42"
+			},
+			"error": ""
```
benhoyt, line 124: Per my suggestion below, if we make this field `omitempty` on the server, let's remove it here.

Also, let's add another test with an actual error message.

docs/specs/openapi.yaml
```diff
@@ -348,6 +348,50 @@ paths:
...
+                  "result": {
+                    "info": {
+                      "name": "check1",
+                      "startup": "enabled",
+                      "status": "up",
+                      "failures": 0,
+                      "threshold": 3,
+                      "change-id": "1"
+                    },
+                    "err": ""
```
benhoyt, line 392: Per comments elsewhere:
```suggestion
                    "error": ""
```

docs/specs/openapi.yaml
```diff
@@ -1301,6 +1345,20 @@ components:
...
+          properties:
+            result:
+              type: object
+              properties:
+                info:
+                  $ref: "#/components/schemas/checkInfo"
+                  type: object
+                error:
+                  type: string
+                  description: An error message if the check fails.
```
benhoyt, line 1361:
```suggestion
                  description: The error message if the check failed; empty string on success.
```

internals/cli/cmd_check.go
```diff
@@ -62,6 +70,41 @@ func (cmd *cmdCheck) Execute(args []string) error {
...
+			Level:     string(res.Info.Level),
+			Startup:   string(res.Info.Startup),
+			Status:    string(res.Info.Status),
+			Failures:  res.Info.Failures,
+			Threshold: res.Info.Threshold,
+			ChangeID:  res.Info.ChangeID,
+		}
+		if res.Error != "" {
+			checkInfo.Error = res.Error
+			logs, err := cmd.taskLogs(checkInfo.ChangeID)
```
benhoyt, line 93: `ChangeID` can be empty in some edge cases, so we should only call this if `ChangeID != ""`. See similar code in `cmdChecks.changeInfo`.

internals/cli/cmd_check_test.go
```diff
@@ -66,3 +67,111 @@ func (s *PebbleSuite) TestCheckNotFound(c *C) {
...
+		c.Check(body, check.DeepEquals, map[string]any{
+			"Name": "chk1",
+		})
+		fmt.Fprint(w, `
+{
+    "type": "sync",
+    "status-code": 200,
+    "result": {
+        "info": {"name": "chk1", "startup": "enabled", "status": "up", "threshold": 3, "change-id": "1"},
+        "error": ""
```
benhoyt, line 85: This is going to be `omitempty` now (see below), so:

```suggestion
```

internals/daemon/api_checks.go
```diff
@@ -114,6 +114,48 @@ func v1PostChecks(c *Command, r *http.Request, _ *UserState) Response {
...
+		Failures:  result.Failures,
+		Threshold: result.Threshold,
+		ChangeID:  result.ChangeID,
+	}
+	errStr := ""
+	if err != nil {
+		errStr = err.Error()
+	}
+
+	type responsePayload struct {
```
benhoyt, line 152: It's a bit annoying that there's a top-level type called `responsePayload` (that was probably me; we could make that one defined locally too). In any case, probably just `payload` is enough here. Could do `refreshPayload`, but I think the `refresh` bit wouldn't be needed in this context.

benhoyt (commented):
Thanks for this. Several minor comments, but also a suggestion to refactor the perform/recover logic a bit. I've opened a PR with the (slightly modified) refactoring in a [new PR](https://github.com/canonical/pebble/pull/582) that will hopefully make the logic changes here easier to review.

internals/daemon/api_checks.go
IronCore864: There already is a `payload` (request payload), so I'm gonna use `refreshPayload`.

internals/daemon/api_checks_test.go
```diff
@@ -221,3 +221,61 @@ func (s *apiSuite) postChecks(c *C, body string) *resp {
...
 }
+
+func (s *apiSuite) TestPostChecksRefresh(c *C) {
+	writeTestLayer(s.pebbleDir, `
+checks:
+    chk1:
+        override: replace
+        level: ready
+        exec:
+            command: echo "hello"
```
benhoyt, line 232: How do we know the check was actually run/refreshed? Should we have it create a file in a temp dir that we assert exists, like we do in some other tests?

internals/overlord/checkstate/handlers.go
```diff
@@ -42,6 +42,13 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 	ticker := time.NewTicker(config.Period.Value)
 	defer ticker.Stop()
 
+	// Retrieve checkData and its channels
```
benhoyt, line 45: This should now just be "channel". As is, I don't think this comment pays for itself -- should we drop it?

internals/overlord/checkstate/handlers.go
```diff
@@ -42,6 +42,13 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
...
+	m.checksLock.Lock()
+	m.ensureCheck(config.Name)
+	checkData := m.checks[details.Name]
+	refresh := checkData.refresh
```
benhoyt, line 49: `ensureCheck` returns the data, so how about this:

```go
data := m.ensureCheck(config.Name)
refresh := data.refresh
```

Or even this (but up to you which is clearer):

```go
refresh := m.ensureCheck(config.Name).refresh
```

internals/overlord/checkstate/handlers.go
```diff
@@ -94,17 +101,36 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 				m.state.Unlock()
 			}
 		}
-		return false, nil
+		return false, err
```
benhoyt, line 104: I think it'd be simpler/clearer to change the `if err != nil` / `else` to just an `if` with an early `return false, err`. And then remove and un-indent the `else` block, with `return false, nil` at the end. This would make it the same structure as `recoverCheck`.

internals/overlord/checkstate/handlers.go
```diff
@@ -137,6 +163,13 @@ func (m *CheckManager) doRecoverCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 	ticker := time.NewTicker(config.Period.Value)
 	defer ticker.Stop()
 
+	// Retrieve checkData and its channels
+	m.checksLock.Lock()
+	m.ensureCheck(config.Name)
+	checkData := m.checks[details.Name]
+	refresh := checkData.refresh
```
benhoyt, line 170: Similar to above:
```suggestion
	data := m.ensureCheck(config.Name)
	refresh := data.refresh
```

internals/overlord/checkstate/manager.go
```diff
@@ -419,6 +428,8 @@ type checkData struct {
 	changeID     string
 	successCount int64
 	failureCount int64
+	refresh      chan refreshInfo
+	result       chan error
```
benhoyt, line 432: What I intended here is that this `result` field would be completely removed from here, and only be in the `refreshInfo` struct, and a new channel created there each time a refresh is requested (because the result "belongs" to the refresh).

internals/overlord/checkstate/manager.go
```diff
@@ -633,3 +644,50 @@ func (m *CheckManager) Replan() {
 		m.updateCheckData(check, changeID, 0)
 	}
 }
+
+// RefreshCheck runs a check immediately.
+func (m *CheckManager) RefreshCheck(ctx context.Context, check *plan.Check) (*CheckInfo, error) {
+	m.checksLock.Lock()
+	checkData := m.ensureCheck(check.Name)
+	refresh := checkData.refresh
+	result := checkData.result
```
benhoyt, line 653: Per above, this line will go away.

internals/overlord/checkstate/manager.go
```diff
@@ -633,3 +644,50 @@ func (m *CheckManager) Replan() {
 		m.updateCheckData(check, changeID, 0)
 	}
 }
+
+// RefreshCheck runs a check immediately.
+func (m *CheckManager) RefreshCheck(ctx context.Context, check *plan.Check) (*CheckInfo, error) {
+	m.checksLock.Lock()
+	checkData := m.ensureCheck(check.Name)
```
benhoyt, line 651: Nit: let's rename this local variable to just `data` throughout. This is all in the context of "check", and it would avoid shadowing the type name.

internals/overlord/checkstate/manager.go
```diff
@@ -633,3 +644,50 @@ func (m *CheckManager) Replan() {
...
+	refresh := checkData.refresh
+	result := checkData.result
+	m.checksLock.Unlock()
+
+	getCheckInfo := func() *CheckInfo {
+		m.checksLock.Lock()
+		checkData := m.ensureCheck(check.Name)
+		m.checksLock.Unlock()
```
benhoyt, line 659: I believe we need to lock for this whole function, right? Because this is a pointer to `checkData`, it could be modified while we're looking at it (non-atomic). So I recommend the "lock...defer unlock" pattern.

internals/overlord/checkstate/manager.go
```diff
@@ -633,3 +644,50 @@ func (m *CheckManager) Replan() {
...
+			ChangeID:  checkData.changeID,
+		}
+		return &info
+	}
+
+	// If the check is stopped, run the check directly without using changes and tasks.
+	if checkData.changeID == "" {
+		chk := newChecker(check)
+		err := runCheck(ctx, chk, check.Timeout.Value)
+		return getCheckInfo(), err
```
benhoyt, line 676: This looks nice and simple, thanks!

internals/overlord/checkstate/manager.go
```diff
@@ -633,3 +644,50 @@ func (m *CheckManager) Replan() {
...
+			Status:    checkData.status,
+			Failures:  checkData.failures,
+			Threshold: checkData.threshold,
+			ChangeID:  checkData.changeID,
+		}
+		return &info
+	}
+
+	// If the check is stopped, run the check directly without using changes and tasks.
+	if checkData.changeID == "" {
```
benhoyt, line 673: I think we need to fetch the `changeID` here with the lock held, just like we're fetching the `refresh` channel.

internals/overlord/checkstate/manager.go
```diff
@@ -633,3 +644,50 @@ func (m *CheckManager) Replan() {
...
+	}
+	if refresh == nil || result == nil {
+		panic(fmt.Sprintf("internal error: refresh channels not initialized for check %q", checkData.name))
+	}
+
+	select {
+	case refresh <- refreshInfo{ctx, result}:
+	case <-ctx.Done():
+		return getCheckInfo(), ctx.Err()
+	}
```
benhoyt, line 686: Here's where we'd make the result channel:
```suggestion
	result := make(chan error)
	select {
	case refresh <- refreshInfo{ctx, result}:
	case <-ctx.Done():
		return getCheckInfo(), ctx.Err()
	}
```

internals/overlord/checkstate/manager_test.go
```diff
@@ -1000,3 +1001,123 @@ pebble_check_failure_count{check="chk1"} 0
...
+
+func (s *ManagerSuite) TestRefreshCheck(c *C) {
+	chk1 := &plan.Check{
+		Name:      "chk1",
+		Override:  "replace",
+		Period:    plan.OptionalDuration{Value: time.Second},
+		Threshold: 3,
+		Exec:      &plan.ExecCheck{Command: "echo chk1"},
+	}
+	origLayer := &plan.Layer{
```
benhoyt, line 1013: Why is this called "orig" layer? In what sense it is "original"?

benhoyt:
Hmm, oddly, `go test -race` is failing in CI (but not locally, at least for me), [see failing run](https://github.com/canonical/pebble/actions/runs/13760571086/job/38526307613?pr=577). I've re-run it but it still fails, so I think we should investigate more deeply. It doesn't *look* like it's related to these changes, but maybe it is.

Otherwise, I'm happy with this PR. I do have a design question about the error details in the `--refresh` case, because when the check is disabled it doesn't go via changes&tasks, so no logs are shown, and the error message is less helpful. We need to figure out a way to make it helpful all the time. I have some ideas I've been playing with locally, but let's chat after daily sync.

hpidcock:
> Hmm, oddly, `go test -race` is failing in CI (but not locally, at least for me), [see failing run](https://github.com/canonical/pebble/actions/runs/13760571086/job/38526307613?pr=577). I've re-run it but it still fails, so I think we should investigate more deeply. It doesn't _look_ like it's related to these changes, but maybe it is.
> 
> Otherwise, I'm happy with this PR. I do have a design question about the error details in the `--refresh` case, because when the check is disabled it doesn't go via changes&tasks, so no logs are shown, and the error message is less helpful. We need to figure out a way to make it helpful all the time. I have some ideas I've been playing with locally, but let's chat after daily sync.

Using:
```
$ go install golang.org/x/tools/cmd/stress
$ go test -c -race
$ stress ./daemon.test
```

I get the race reliably on go 1.24.1.

internals/overlord/checkstate/handlers.go
```diff
@@ -82,29 +87,49 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
...
 			shouldExit, err := performCheck()
+			select {
+			case info := <-refresh:
+				// If refresh requested while running check, send result.
+				select {
+				case info.result <- err:
+				case <-info.ctx.Done():
+				}
+			default:
+			}
```
hpidcock, line 129: This is an interesting optimisation that highlights a potential issue with multiple requests to refresh the same check.

Since all refreshes to this same check will be queued, there is a case where multiple requests are queued up causing this to behave rather strangely.

I suggest something like this:
```
	respondToRequests := func(result error) {
		timeout := time.NewTimer(time.Millisecond)
		for {
			select {
			case <-timeout.C:
				return
			case info := <-refresh:
				select {
				case info.result <- result:
				case <-info.ctx.Done():
				}
			case <-tomb.Dying():
				timeout.Stop()
				return
			}
		}
	}

	for {
		select {
		case info := <-refresh:
			// Reset ticker on refresh.
			ticker.Reset(config.Period.Value)
			shouldExit, err := performCheck()
			select {
			case info.result <- err:
			case <-info.ctx.Done():
			}
			respondToRequests(err)
			if shouldExit {
				return err
			}
		case <-ticker.C:
			shouldExit, err := performCheck()
			respondToRequests(err)
			if shouldExit {
				return err
			}
		case <-tomb.Dying():
			return checkStopped(config.Name, task.Kind(), tomb.Err())
		}
	}
```

hpidcock (commented):
Looks pretty good, mostly happy with the approach. It is a little strange that a stopped check is processed in a different way. But I guess, fundamentally they both serve the same purpose, just in one case the user may be unaware that the check is stopped.

IronCore864:
> > Hmm, oddly, `go test -race` is failing in CI (but not locally, at least for me), [see failing run](https://github.com/canonical/pebble/actions/runs/13760571086/job/38526307613?pr=577). I've re-run it but it still fails, so I think we should investigate more deeply. It doesn't _look_ like it's related to these changes, but maybe it is.
> > Otherwise, I'm happy with this PR. I do have a design question about the error details in the `--refresh` case, because when the check is disabled it doesn't go via changes&tasks, so no logs are shown, and the error message is less helpful. We need to figure out a way to make it helpful all the time. I have some ideas I've been playing with locally, but let's chat after daily sync.
> 
> Using:
> 
> ```
> $ cd internals/daemon
> $ go install golang.org/x/tools/cmd/stress@latest
> $ go test -c -race
> $ stress ./daemon.test
> ```
> 
> I get the race reliably on go 1.24.1.

Thanks for the help! According to the log, the race appears in the `TestStateChange` test, so I only tried to stress it and related tests and could not reproduce it. If I stress the whole daemon test suite, I can also reproduce it now.

After checking the code, I think the race is because `timeNow` is a global var accessed by both `changeStatus` (read, and in a few other functions) and tests (write, in `FakeTime`) without synchronization. The simplest solution is to add something like `timeNoeMutex` and lock it in `FakeTime`?

internals/overlord/checkstate/handlers.go
benhoyt: Interesting idea to empty the refresh request queue (or at least try for 1ms). I'm not entirely sure it's worth it, though. It seems like realistically there will only ever be 1 waiting, because it's likely from a CLI user, and if there's more, it gets run again.

But if we do this, what about just looping till the channel is empty, something like this?

```go
emptyRefreshQueue := func(result error) {
	for {
		select {
		case info := <-refresh:
			select {
			case info.result <- result:
			case <-info.ctx.Done():
			}
		default:
			return
		}
	}
}
```

benhoyt:
> If I stress the whole daemon test suite, I can also reproduce it now.
>
> After checking the code, I think the race is because timeNow is a global var accessed by both changeStatus (read, and in a few other functions) and tests (write, in FakeTime) without synchronization. The simplest solution is to add something like timeNowMutex and lock it in FakeTime?

I can't even repro it locally under stress when running the whole daemon test suite. However, that's a clue -- it is because `timeNow` is a global var being accessed by the tests and by `changeStatus` (via the overlord loop). The overlord loop is only being started manually by some tests -- see `apiSuite.startOverlord`. If I add `s.startOverlord()` right under the `d := s.daemon(c)` line in `TestStateChange`, then just running that individual test (not under stress) with `-race` reproduces the data race pretty reliably for me:

```
go test -race -v ./internals/daemon/ -check.vv -check.f '^TestStateChange$'
```

That means that it must be one of the other daemon tests starting the overlord loop interfering with this test. I don't quite understand why, as `apiSuite.TearDownTest` stops the overlord and waits for the overlord loop to finish. Oh wait, I guess it's because `TaskRunner.run` fires up another goroutine ... and that's the one that's calling `changeStatus`, and we're not waiting for *that* goroutine to finish.

Either way, it'd be nice if we can solve it by have the other tests clean up properly after themselves, rather than wrapping all the `timeNow()` calls in a mutex.

internals/overlord/checkstate/handlers.go
benhoyt: I guess that makes sense: they're really asking for a new refresh to start after they asked. Okay, that's simplest anyway -- can you please make that change @IronCore864?

IronCore864:
After some investigation, I think other tests, in theory, shouldn't interfere with `TestStateChange` (and tests that call `FakeTime`/write `timeNow` for that matter) as long as the overlord loop is by `startOverlord`, because in this case, a flag is set, then in `apiSuite.TearDownTest`, `Overlord.Stop` is called, which in turn calls `TaskRunner.Stop`, meaning the clean up is already done properly.

Adding `startOverlord` in `TestStateChange` (in the same test where `FakeTime` is called) can reproduce the race condition because, in this case, the deferred write on `timeNow` happens before the teardown when the overlord/taskrunner is stopped, so in theory, the taskrunner can still `changeStatus` when `timeNow` is being written.

I did find two test cases in the same package/test suite that start the overlord loop without using `startOverlord`, which I hope can fix the issue. I'm not 100% sure about this, though, because the original issue doesn't always occur, even in GitHub Actions. Anyway, with a lot of stress (50 min, 6 failures, all unrelated to the race condition) and `-race` tests (50 runs) locally, and a few reruns in GitHub Actions, I couldn't reproduce it for now.

internals/overlord/checkstate/handlers.go
IronCore864: I have refactored it according to the above discussion.

Regarding the block issue, I don't think it's a big problem because as Ben mentioned it's only used with CLI, and even if there are two refreshes with CLI, both can get the result, it's just that the first request would block the second, so the second takes longer to be responded. In fact, at the beginning, I thought of something like a request ID to support multiple refreshes at the same time, but in the end, I think it doesn't add much value, so I used an unbuffered channel.

benhoyt:
> I did find two test cases in the same package/test suite that start the overlord loop without using startOverlord, which I hope can fix the issue. 

Ah yes, that would do it, good find! Tests that were starting the overlord loop without using `startOverlord` means that those loops could still be running, because the teardown wouldn't have stopped the loop.

internals/daemon/api_metrics_test.go
```diff
@@ -33,7 +33,7 @@ services:
         command: sleep 10
 `)
 	d := s.daemon(c)
-	d.overlord.Loop()
+	s.startOverlord()
```
benhoyt, line 36: Thanks for this test flakiness fix!

benhoyt (approved):
This is looking really nice now. I'm approving this PR, and we can merge. However, let's add a new PR documenting it -- a short section to the documentation describing this feature. Perhaps a short section "Testing a check" after the "Checks command" section at https://documentation.ubuntu.com/pebble/reference/health-checks/? And we'll ask David to help with review.

tonyandrewmeyer:
Speaking of new PRs, please also remember to add an issue or PR in canonical/operator to add this there. Maybe `refresh=False` as an argument to `get_checks` that errors if you pass more than one check, or maybe there should be a new `get_check`, which would more closely match Pebble.

benhoyt:
Opened, thanks @tonyandrewmeyer: https://github.com/canonical/operator/issues/1625 -- though it should probably be `refresh_check` to match the API level.

PR #580, opened 2025-02-28
fix: avoid health checks URL panic; add staticcheck

### LINKED ISSUE #579: HTTP checker fails to verify error returned by http.NewRequestWithContext
The error returned by this line is not being checked:
https://github.com/canonical/pebble/blob/65446713a1f212801a10158595b715580bac53bb/internals/overlord/checkstate/checkers.go#L54

While the most likely error scenario would be invalid URLs (which wouldn't be a problem if the caller is validating the URL beforehand), `http.NewRequestWithContext` unfortunately makes no guarantees as to when it will returns errors: https://pkg.go.dev/net/http#NewRequestWithContext

benhoyt:
This fixes #579, which results in a panic if an HTTP check URL was malformed. Thanks for the report and the test at https://github.com/canonical/tako/pull/24, @alnvdl-work.

I also fixed another similar "forgot to check err" issue in `cmd_warnings.go`.

This PR also adds [staticcheck](https://staticcheck.dev/) to the linting, which found this issue, and a few other minor issues. Fix the other issues and add a `staticcheck.conf` and a GitHub Action to run it. The other issues are mostly not problems, but include:

- Use of deprecated functions (these are okay, just `lint:ignore`)
- Error messages shouldn't end with "."
- `strconv.ParseFloat` second arg is not the base (10), so should be 32 or 64. I actually [tried](https://go-review.googlesource.com/c/tools/+/267600) to add this check to `go vet` years ago.
- Various unused functions and struct fields.
- `time.Now().Sub(t)` -> `time.Since(t)`
- Use of `rand.Seed` (deprecated) in tests

Fixes #579

### DIFF
--- a/internals/cli/cmd_help.go
+++ b/internals/cli/cmd_help.go
@@ -96,10 +96,6 @@ func addHelp(parser *flags.Parser) error {
 	return nil
 }
 
-func (cmd *cmdHelp) setParser(parser *flags.Parser) {
-	cmd.parser = parser
-}
-
 // manfixer is a hackish way to fix drawbacks in the generated manpage:
 // - no way to get it into section 8
 // - duplicated TP lines that break older groff (e.g. 14.04), lp:1814767
@@ -150,7 +146,7 @@ func (cmd cmdHelp) Execute(args []string) error {
 	}
 	if cmd.All {
 		if len(cmd.Positional.Subs) > 0 {
-			return fmt.Errorf("help accepts a command, or '--all', but not both.")
+			return fmt.Errorf("help accepts a command, or '--all', but not both")
 		}
 		printLongHelp(cmd.parser)
 		return nil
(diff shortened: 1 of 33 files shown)


internals/cli/cmd_help.go
```diff
@@ -150,7 +146,7 @@ func (cmd cmdHelp) Execute(args []string) error {
 	}
 	if cmd.All {
 		if len(cmd.Positional.Subs) > 0 {
-			return fmt.Errorf("help accepts a command, or '--all', but not both.")
+			return fmt.Errorf("help accepts a command, or '--all', but not both")
```
alnvdl-work, line 149: So, in the other repo, this was actually deemed *not* to be a problem, so I reverted the change there. WDYT?

staticcheck.conf
```diff
@@ -0,0 +1,6 @@
+checks = [
+	"inherit",
+	"-SA5008",  # duplicate struct tag (cli/cmd_*.go use this for "choice")
+	"-ST1012",  # error var should have name of the form errFoo
```
alnvdl-work, line 4: Out of curiosity, what is the rationale for disabling this one?

internals/cli/cmd_help.go
benhoyt: It's not a problem, but "error messages with the '.'" is our usual style too, and none of the other `fmt.Errorf` messages in Pebble end in a `.`, so I "fixed" it.

staticcheck.conf
benhoyt: Because these "errors" were actually not errors but sentinel values, so shouldn't really be named `errFoo`. However, I think this is a reasonable rule in general, so I ended up using line-ignores for those and re-enabling this rule in staticcheck.conf -- thanks.

IronCore864 (approved):

PR #585, opened 2025-03-11
chore: update to use Go 1.24 now that 1.22 is end-of-life

benhoyt:
- fix vet issues
- use os.CopyFS instead of our vendored version


internals/daemon/api_changes_test.go
```diff
@@ -210,7 +210,7 @@ func (s *apiSuite) TestStateChange(c *check.C) {
 	stateChangeCmd := apiCmd("/v1/changes/{id}")
 
 	// Execute
-	req, err := http.NewRequest("POST", "/v1/change/"+ids[0], nil)
+	req, err := http.NewRequest("GET", "/v1/change/"+ids[0], nil)
```
benhoyt, line 213: Drive-by fix: doesn't actually fix anything as `v1GetChange` is called directly below, but it's actually a GET and this was confusing to me when debugging something else.

IronCore864 (approved):

PR #588, opened 2025-03-13
docs: add a section on pebble check refresh

IronCore864:
Add docs for how to test health checks with `pebble check --refresh`.

### DIFF
--- a/docs/reference/health-checks.md
+++ b/docs/reference/health-checks.md
@@ -138,6 +138,68 @@ Health checks are implemented using two change kinds:
 
 When a check is stopped, the active `perform-check` or `recover-check` change is aborted. When a stopped (inactive) check is started, a new `perform-check` change is created for the check.
 
+## Testing a check
+
+You can run a check immediately and get the result in YAML format using the `pebble check <name> --refresh` command, which is helpful when developing, testing and debugging a check. For example:
+
+```{code-block} bash
+$ pebble check chk1 --refresh
+name: chk1
+startup: enabled
+status: up
+failures: 0
+threshold: 3
+change-id: "1"
+```
+
+If running the check fails, it will also show the error and logs:
+
+```{code-block} bash
+:emphasize-lines: 8-10
+
+$ pebble check chk1 --refresh
+name: chk1
+startup: enabled
+status: up
+failures: 1
+threshold: 3
+change-id: "1"
+error: non-2xx status code 500; Health check failed
+logs: |
+    2025-03-13T10:05:45+08:00 ERROR non-2xx status code 500; Health check failed
+```
+
+You can run a check even if it's inactive (for example, `startup: disabled`, or stopped by [`pebble stop-checks`](#reference_health_checks_start_stop_command). This will only run the check immediately but won't start the check. If running a stopped check fails, it will not change the status or the failures count; it will only show the error but no logs:
+
+```{code-block} bash
+:emphasize-lines: 4,5,7
+
+$ pebble check chk1 --refresh
+name: chk1
+startup: disabled
+status: inactive
+failures: 0
+threshold: 3
+error: non-2xx status code 500; Health check failed
+```
+
+Without the optional `--refresh` flag, you can use the command `pebble check <name>` to get the details of a check in YAML format without running it, and if the check has failed previously, it will also show logs from the task that ran the check:
+
+```{code-block} bash
+:emphasize-lines: 8,9
+
+$ pebble check chk1
+name: chk1
+startup: enabled
+status: up
+failures: 1
+threshold: 3
+change-id: "1"
+logs: |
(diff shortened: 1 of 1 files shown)


docs/reference/health-checks.md
```diff
@@ -138,6 +138,68 @@ Health checks are implemented using two change kinds:
 
 When a check is stopped, the active `perform-check` or `recover-check` change is aborted. When a stopped (inactive) check is started, a new `perform-check` change is created for the check.
 
+## Testing a check
+
+You can run a check immediately and get the result in YAML format using the `pebble check <name> --refresh` command, which is helpful when developing, testing and debugging a check. For example:
```
dwilding, line 143: A couple of minor suggestions for readability:

```suggestion
You can run a check immediately using the `pebble check <name> --refresh` command, which is helpful when developing, testing, and debugging a check. The result of the check is in YAML format. For example:
```

docs/reference/health-checks.md
```diff
@@ -138,6 +138,68 @@ Health checks are implemented using two change kinds:
...
+status: up
+failures: 1
+threshold: 3
+change-id: "1"
+error: non-2xx status code 500; Health check failed
+logs: |
+    2025-03-13T10:05:45+08:00 ERROR non-2xx status code 500; Health check failed
+```
+
+You can run a check even if it's inactive (for example, `startup: disabled`, or stopped by [`pebble stop-checks`](#reference_health_checks_start_stop_command). This will only run the check immediately but won't start the check. If running a stopped check fails, it will not change the status or the failures count; it will only show the error but no logs:
```
dwilding, line 172: Some suggestions for readability:

```suggestion
You can run a check even if it's inactive (for example, if the check has `startup: disabled` or was stopped by [`pebble stop-checks`](#reference_health_checks_start_stop_command)). This will run the check immediately but won't start the check. If running a stopped check fails, Pebble won't change the status or the failure count, and the result won't include logs:
```

docs/reference/health-checks.md
```diff
@@ -138,6 +138,68 @@ Health checks are implemented using two change kinds:
...
+$ pebble check chk1 --refresh
+name: chk1
+startup: disabled
+status: inactive
+failures: 0
+threshold: 3
+error: non-2xx status code 500; Health check failed
+```
+
+Without the optional `--refresh` flag, you can use the command `pebble check <name>` to get the details of a check in YAML format without running it, and if the check has failed previously, it will also show logs from the task that ran the check:
```
dwilding, line 186: Small suggestions for readability:

```suggestion
Without the `--refresh` flag, `pebble check <name>` gets the details of a check without running the check. If the check has failed previously, the result will show logs from the task that ran the check. For example:
```

docs/reference/health-checks.md
```diff
@@ -138,6 +138,68 @@ Health checks are implemented using two change kinds:
...
+
+```{code-block} bash
+:emphasize-lines: 8,9
```
dwilding, line 189: I don't feel the highlighting is needed on this one, as we're not comparing against anything else, and the logging is not the only think we're demonstrating here - we're demonstrating the whole of `pebble check chk1`.

```suggestion
```

docs/reference/health-checks.md
```diff
@@ -138,6 +138,68 @@ Health checks are implemented using two change kinds:
...
+$ pebble check chk1 --refresh
+name: chk1
+startup: enabled
+status: up
+failures: 0
+threshold: 3
+change-id: "1"
+```
+
+If running the check fails, it will also show the error and logs:
```
dwilding, line 155:
```suggestion
If running the check fails, the result will also show the error and logs:
```

dwilding (approved):
Nice one @IronCore864, I just added some small suggestions. The new section helped me easily understand how the feature works. And I like the way you put the non-refresh example at the end, as putting `--refresh` first does feel more appropriate for a "testing" section.

About the highlighting: I recommend removing the highlighting from example number 4 (the non-refresh example) because I don't think it adds much there. Then you could do this:
- For example 1 and 4, use the `{terminal}` style.
- For example 2 and 3, remove the `$ pebble check chk1 --refresh` line and keep the current style.

But I also feel that what you have now is fine.

benhoyt (changes requested):
This is great content, thanks. Regarding the styling, I find it strange and confusing why some are black+green and others in the same section are gray with a larger font. I think at least within the same section we should be consistent. I don't actually think the highlight lines are particularly needed here -- it's obvious that "error" and "logs" are new, for example.

I think we should make them consistent. My vote would be for all `{terminal}` style (to be consistent with the "Checks command" section and other docs), but I'm okay with all non-terminal style if you strongly prefer that.

benhoyt (approved):
Thanks! I do think this looks consistent. Please just update the PR description before merging.

PR #591, opened 2025-03-16
feat(tls): add an X509 keypair manager

flotter:
In order to extend API access to include HTTPS (TLS), we need a manager that can take care of:

1. Selecting a valid X509 keypair on every new HTTPS/TLS session
2. Generate a new valid self-signed X509 keypair when no suitable one exists
3. Publish the valid public certificates on the API to allow clients trusting the server to update certificates

### DIFF
--- a/internals/overlord/certstate/manager.go
+++ b/internals/overlord/certstate/manager.go
@@ -45,22 +45,24 @@ type CertManager struct {
 // x509KeyPair represents the X509 public certificate and the private
 // key that signed it.
 type x509KeyPair struct {
-	order       int
-	private     any
-	certificate *x509.Certificate
+	Order       int
+	Private     any
+	Certificate *x509.Certificate
 }
 
-func NewManager(tlsDir string) *CertManager {
+func NewManager(tlsDir string) (*CertManager, error) {
 	manager := &CertManager{tlsDir: tlsDir}
-	manager.loadX509KeyPairs(tlsDir)
-	return manager
+	if err := manager.loadX509KeyPairs(); err != nil {
+		return nil, err
+	}
+	return manager, nil
 }
 
-// TLSCertificate returns a valid TLS certificate instance that
+// X509Keypair returns a valid tls.Certificate instance that
 // can be used for starting a TLS based network listener. A call
-// to this function may generate a new TLS certificate if none
-// of the available certificates are valid.
-func (c *CertManager) TLSCertificate() (*tls.Certificate, error) {
+// to this function may generate a new X509 keypair if none
+// of the available pairs are valid.
+func (c *CertManager) X509Keypair() (*tls.Certificate, error) {
 	c.x509KeyPairsLock.Lock()
 	defer c.x509KeyPairsLock.Unlock()
 
@@ -68,7 +70,7 @@ func (c *CertManager) TLSCertificate() (*tls.Certificate, error) {
 
 	// Try to find an existing valid X509
 	for i, keypair := range c.x509KeyPairs {
-		if !isCertExpired(keypair.certificate) {
+		if !isCertExpired(keypair.Certificate) {
 			selected = &c.x509KeyPairs[i]
 			break
 		}
@@ -76,7 +78,7 @@ func (c *CertManager) TLSCertificate() (*tls.Certificate, error) {
 
 	// Generate a keypair if no one exists.
 	if selected == nil {
-		selected, err := c.generateX509ECP256Keypair()
+		selected, err := c.newX509ECP256Keypair()
 		if err != nil {
 			return nil, fmt.Errorf("cannot generate X509 keypair: %w", err)
 		}
@@ -85,14 +87,14 @@ func (c *CertManager) TLSCertificate() (*tls.Certificate, error) {
 		}
 
 		// Notify subscribers of the new certificate in use
-		c.callChangeListeners(selected.certificate)
+		c.callChangeListeners(selected.Certificate)
 	}
 
-	pemCert, err := x509CertToPEM(selected.certificate)
+	pemCert, err := x509CertToPEM(selected.Certificate)
 	if err != nil {
 		return nil, err
 	}
-	pemKey, err := privateKeyToPEM(selected.private)
+	pemKey, err := privateKeyToPEM(selected.Private)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+	keypair, err := generateX509ECP256Keypair(notBefore, notAfter)
+	if err != nil {
+		return nil, err
+	}
+	keypair.Order = order
+	return keypair, nil
+}
+
+func (c *CertManager) persistX509Keypair(keypair *x509KeyPair) error {
+	return writeX509Keypair(keypair, c.tlsDir)
```
anpep, line 159: Let's inline this function call (drop `writeX509Keypair()` in favor of `persistX509Keypair()`), since the former is only called here and in the test setup (and this test setup call can be replaced by `persistX509Keyapir(*x509KeyPair)`

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+// changing the state of the system (it does not create new keypairs).
+func (c *CertManager) loadX509KeyPairs() error {
+	// If the directory does not exist that simply means no keypairs
+	// are currently available, and that is OK.
+	_, err := os.Stat(c.tlsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+	}
```
anpep, line 201: nit:
```suggestion
	if _, err := os.Stat(c.tlsDir); err != nil && os.IsNotExist(err) {
		return nil
	}
```

I'm always inclined to inline function calls returning errors in the `if` clause, when the line does not get too long. However, you can do as you please. I think we should enforce this style automatically, or at least have a team-wide convention on this specifically (because right now the Project Crystal codebase is inconsistent in this, largely my fault)

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+			// 3. The next entry is not the expected order matching private key
+			return fmt.Errorf("cannot find the expected X509 certificate or its private key")
+		}
+
+		certPath := filepath.Join(c.tlsDir, entry.Name())
+		keyPath := filepath.Join(c.tlsDir, fmt.Sprintf("%3d-key.pem", order))
+
+		err = expectPermission(certPath, 0o644)
+		if err != nil {
+			return fmt.Errorf("cannot verify X509 certificate permission: %w", err)
```
anpep, line 251: nit: here and below
```suggestion
			return fmt.Errorf("cannot use X509 certificate with too open permissions: %w", err)
```

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+	certificate, err := x509.ParseCertificate(derBytes)
+	if err != nil {
+		return nil, err
+	}
+	return &x509KeyPair{Private: privateKey, Certificate: certificate}, nil
+}
+
+func writeX509Keypair(keypair *x509KeyPair, tlsDir string) error {
+	keyPath := filepath.Join(tlsDir, fmt.Sprintf("%03d-key.pem", keypair.Order))
+	keyOut, err := os.OpenFile(keyPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
anpep, line 317:
```suggestion
	keyOut, err := os.OpenFile(keyPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
```

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+		return err
+	}
+	if _, err = keyOut.Write(keyPEM); err != nil {
+		return err
+	}
+	if err := keyOut.Close(); err != nil {
+		return err
+	}
+	certPath := filepath.Join(tlsDir, fmt.Sprintf("%03d-cert.pem", keypair.Order))
+	certOut, err := os.OpenFile(certPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
anpep, line 332:
```suggestion
	certOut, err := os.OpenFile(certPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o644)
```

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+	if err != nil {
+		return err
+	}
```
anpep, line 335:
```suggestion
	if err != nil {
		return err
	}
	defer certOut.Close()
```

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+	certPEM, err := x509CertToPEM(keypair.Certificate)
+	if err != nil {
+		return err
+	}
+	if _, err = certOut.Write(certPEM); err != nil {
+		return err
+	}
+	if err := certOut.Close(); err != nil {
+		return err
+	}
```
anpep, line 345:
```suggestion
```

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+	info, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+	actualPerm := info.Mode().Perm()
+	if actualPerm != perm {
+		return fmt.Errorf("expected permission 0o%o (got 0o%o) for %q", perm, actualPerm, path)
+	}
+	return nil
+}
```
anpep, line 429: This is great. However, I'd be more inclined to test group/other read/write bits against zero rather than expecting a concrete octal value here, and fail with something like `permissions <...> too open. Try 0X00`

internals/overlord/certstate/manager_test.go
```diff
@@ -0,0 +1,56 @@
...
+	keypairs []map[int]struct {
+		order     int
+		notBefore time.Time
+		notAfter  time.Time
+		generated *certstate.X509KeyPair
+	}
+	errorResult   string
+	selectedOrder int
+}{{
+	summary:     "Driectory does not exist",
```
anpep, line 54:
```suggestion
	summary:     "Directory does not exist",
```

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+		return nil, err
+	}
+	return manager, nil
+}
+
+// X509Keypair returns a valid tls.Certificate instance that
+// can be used for starting a TLS based network listener. A call
+// to this function may generate a new X509 keypair if none
+// of the available pairs are valid.
+func (c *CertManager) X509Keypair() (*tls.Certificate, error) {
```
flotter, line 65: Add the thread that creates the candidate certificates, before the current one expires.

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+
+// x509KeyPair represents the X509 public certificate and the private
+// key that signed it.
+type x509KeyPair struct {
+	Order       int
+	Private     any
+	Certificate *x509.Certificate
+}
+
+func NewManager(tlsDir string) (*CertManager, error) {
```
flotter, line 53: The candidate generation thread should not always run.

1. Optionally start certstate
2. Only "wake up certmanager" if its needed.

internals/overlord/certstate/manager.go
```diff
@@ -0,0 +1,429 @@
...
+// isCertExpired checks if a X509 certificate is expired.
+func isCertExpired(cert *x509.Certificate) bool {
+	now := time.Now()
+	if now.Before(cert.NotBefore) {
+		return true
+	}
+	if now.After(cert.NotAfter) {
+		return true
+	}
+	return false
```
anpep, line 414:
```suggestion
	return now.Before(cert.NotBefore) || now.After(cert.NotAfter)
```

flotter:
This will be replaced by a new PR chain. Closing, but keeping for comment reference.

PR #596, opened 2025-04-07
feat(tls): add TLS keypair manager

flotter:
This is the first in a number of patches that will add daemon HTTPS support.

Add a TLS keypair manager, which will look after the following tasks:

1. Load existing TLS keypairs when Pebble starts up (on manager creation)

```go
    tlsMgr, err := tlsstate.NewManager(tlsDir)
```

3. Select a suitable TLS keypair on demand for every new HTTPS session
4. Provide access to all valid TLS public certificates (used by clients during rotation)
5.


internals/overlord/tlsstate/export_test.go
```diff
@@ -0,0 +1,132 @@
...
+// FakeCertificateValidityPeriod allows setting how long a new certificate will be valid for.
+func (m *TLSManager) FakeCertificateValidityPeriod(valid time.Duration) (restore func()) {
+	old := certificateValidityPeriod
+	certificateValidityPeriod = valid
+	return func() {
+		certificateValidityPeriod = old
+	}
+}
+
+// AddNewUnsupportedTLSKeyPair installs an unsupported ECDSA NIST P-224
```
paul-rodriguez, line 71: I'm kind of surprised we choose to use a NIST algorithm when we could be using ed25519 instead. Is there a rationale behind this choice somewhere?

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+// ~/> openssl genpkey -algorithm EC -pkeyopt ec_paramgen_curve:prime256v1 -out ec-p256.pem
+// ~/> openssl req -new -x509 -key ec-p256.pem -out ec-p256-cert.pem -days 365
+//
+// The name of the file must match the SHA512/384 hash of the raw public certificate
+// encoded in base32 (without padding), and contain both the private and certificate
+// blocks in a single PEM file:
+//
+// ~/> FINGERPRINT=$(openssl x509 -in ec-p256-cert.pem -outform DER | openssl dgst -sha384 -binary | base32 -w0 | tr -d '=')
+// ~/> cat ec-p256.pem > ${FINGERPRINT}.pem
+// ~/> cat ec-p256-cert.pem >> ${FINGERPRINT}.pem
```
paul-rodriguez, line 77: Appending one file to the other sounds like a weird ad-hoc file format, which is neither a certificate nor a key and is not the expected content of a `.pem`. Is this a standard practice? If not, is there really a reason to do this instead of having two files: `${FINGERPRINT}.pem` and `${FINGERPRINT}-cert.pem` ?

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	if err := manager.loadTLSKeyPairs(); err != nil {
+		return nil, err
+	}
+	return manager, nil
+}
+
+// TLSKeyPair returns the currently preferred keypair. This method may generate
+// a suitable TLS keypair if none exists as the time of the call. The
+// generated keypair will be installed (persisted) in the TLS directory.
+func (m *TLSManager) TLSKeyPair() (*TLSKeyPair, error) {
```
paul-rodriguez, line 95:
```suggestion
func (m *TLSManager) PreferredKeyPair() (*TLSKeyPair, error) {
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	// certificateValidityPeriod determines how long a newly generated
+	// TLS keypair is valid for.
+	certificateValidityPeriod = (10 * 365 * 24 * time.Hour)
+)
+
+type TLSManager struct {
+	// tlsDir is the location of the PEM keypair files.
+	tlsDir         string
+	tlsKeyPairLock sync.Mutex
+	tlsKeyPairs    []*TLSKeyPair
```
paul-rodriguez, line 55: The order that this slice is assumed to be in should be precisely documented here.
It's a very tight coupling in the design, so it should be hard to miss.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+		return a.Compare(b)
+	})
+	return keypair, nil
+}
+
+// genECP256Keypair generates a new keypair based on ECDSA NIST P-256
+// private key and X509 public certificate. The certificate is signed with
+// the same private key (self-signed), which means that the public key
+// in the certificate can be used to verify the signature.
+func (m *TLSManager) genECP256Keypair(notBefore time.Time, notAfter time.Time) (*TLSKeyPair, error) {
```
paul-rodriguez, line 282:
```suggestion
func newECP256Keypair(notBefore time.Time, notAfter time.Time) (*TLSKeyPair, error) {
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+
+// addFingerprint adds the base32 string encoded fingerprint of the public certificate
+// so we can access it later as a cached value. No fingerprint will be added if
+// this method is called before a certificate was added.
+func (t *TLSKeyPair) addFingerprint() {
+	if t.Certificate != nil {
+		hashBytes := sha512.Sum384(t.Certificate.Raw)
+		t.fingerprint = base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(hashBytes[:])
+	}
+}
```
paul-rodriguez, line 511: This function doesn't need a keypair as input, only the certificate.

```suggestion
func fingerprint(cert *x509.Certificate) string {
  hashBytes := sha512.Sum384(cert.Raw)
  return base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(hashBytes[:])
}
```

And I think you could just inline this code into the function that creates a new keypair. It doesn't make sense to call this function anywhere else anyway.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+	if err := pem.Encode(&pemBuffer, pemCertificateBlock); err != nil {
+		return nil, fmt.Errorf("cannot convert x509 certificate to PEM: %w", err)
+	}
+	return pemBuffer.Bytes(), nil
+}
+
+// fromPEMFile loads a PKCS8 private key and x509 certificate
+// from a PEM formatted file.
+func (t *TLSKeyPair) fromPEMFile(path string) error {
```
paul-rodriguez, line 464:
```suggestion
func loadKeyPair(path string) (*TLSKeyPair, error) {
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+// filenameRegexp matches PEM files based on an SHA512/384 hash of the
+// x509 raw certificate, encoded in base32 (77 digits), with an ".pem" suffix.
+var filenameRegexp = regexp.MustCompile(`^[ABCDEFGHIJKLMNOPQRSTUVWXYZ234567]{77}\.pem$`)
+
+// withPEMFile inspects the TLS directory, locates all the PEM files and
+// call the 'do' function on each valid file. A non-existant or empty TLS
+// directory is valid, but a TLS directory or PEM file with invalid
+// permissions will result in an error. Files in the TLS directory not
+// matching the naming convention will be ignored.
+func (m *TLSManager) withPEMFile(do func(path string, fingerprint string) error) error {
```
paul-rodriguez, line 180: I don't think the callback is necessary, especially considering this function is only used in one context.

You could also have one function that returns a slice of all the fingerprints found in the certificate directory. Then in a different function you can iterate on that slice, opening the files one by one.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+// over the daemon API.
+type TLSKeyPair struct {
+	privateKey  any
+	Certificate *x509.Certificate
+	fingerprint string
+}
+
+// Compare compares two keypairs and returns an int as used by sorting
+// functions.
+func (t TLSKeyPair) Compare(other *TLSKeyPair) int {
```
paul-rodriguez, line 367: This function compares keypairs in a very specific way, and I don't think we should consider that _the_ canonical way to compare keypairs. In fact, I'm not sure it's correct in our use case to compare by start time then by end time. What about end time first then start time? Either way there are situations where some clients will lose access even though they still hold a "valid" certificate.

In short, I think this function should have a name that suggest that it is specifically for ordering keypairs in order of creation (I suppose that's the intent) and not a universally valid comparison.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+		return notAfterSort
+	}
+	// Fingerprint last.
+	return strings.Compare(t.Fingerprint(), other.Fingerprint())
+}
+
+// Fingerprint returns an unpadded base32 encoded SHA512/384 hash of the raw
+// x509 certificate.
+func (t TLSKeyPair) Fingerprint() string {
+	return t.fingerprint
```
paul-rodriguez, line 385: You could simplify your life and just calculate the fingerprint right here, though this might have a performance impact. Alternatively, you could calculate it lazily (i.e. leave it `nil` until someone calls this function, and only calculate if the field is nil), combining simplicity and performance.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	pemPath := filepath.Join(m.tlsDir, fmt.Sprintf("%v.pem", keypair.Fingerprint()))
+	pemFile, err := os.OpenFile(pemPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = errors.Join(err, pemFile.Sync())
+		err = errors.Join(err, pemFile.Close())
```
paul-rodriguez, line 341: Calling both `Sync` and `Close` shouldn't be necessary. `Close` will implicitly synchronize.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+func (m *TLSManager) persistTLSKeyPair(keypair *TLSKeyPair) (err error) {
+	// If the TLS directory does not yet exist, create it.
+	_, err = os.Stat(m.tlsDir)
+	if os.IsNotExist(err) {
+		// Create the leaf directory node with 0700 permissions.
+		err = os.Mkdir(m.tlsDir, 0700)
+	}
+	if err != nil {
+		return fmt.Errorf("cannot create directory leaf of path %v: %w", m.tlsDir, err)
+	}
```
paul-rodriguez, line 332: I think you should just use `MkdirAll` which unlike this code, will error out if `m.tlsDir` exists but is not a directory. And because the permissions really have to be `0700` always, it's ok to possibly change them to that.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+		}
+	}
+	if t.privateKey == nil || t.Certificate == nil {
+		return fmt.Errorf("PEM file does not contain the expected keypair")
+	}
+	return nil
+}
+
+// isExpired returns true if the X509 certificate has expired.
+func (t TLSKeyPair) isExpired() bool {
```
paul-rodriguez, line 495: This is misleading as it will return true if the certificate isn't valid yet.

internals/overlord/tlsstate/manager.go
flotter: PEM is a multi block container format, so that part is not unusual.

We can split it in two, but I am not sure that makes the solution safer from a security standpoint. It does make the code significantly more complex (I had it like that before). I think the next step would be to move the private key into a hardware backed keystore, as part of a secure boot capable system.

I am happy to split it in two if that is the consensus.

internals/overlord/tlsstate/export_test.go
flotter: I will change the private key / certificate algorithm to Ed25519.

```
Feature	RSA	ECDSA	Ed25519
Key Size (bits)	2048–4096	256–521	256
Signature Size (bytes)	~256–512	~64–132	64
Speed (sign/verify)	Slow/Moderate	Fast/Fast	Very Fast
TLS Support	Excellent	Very Good	Limited
Certificate Size	Large	Small	Small
Security Efficiency	Moderate	High	Very High
CA Availability	Universal	High	Low
Curve Trust Issues	N/A	Possible	None
```

internals/overlord/tlsstate/manager.go
flotter: I will add a comment.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+		Bytes: t.Certificate.Raw,
+	}
+	var pemBuffer bytes.Buffer
+	if err := pem.Encode(&pemBuffer, pemPrivateBlock); err != nil {
+		return nil, fmt.Errorf("cannot convert key to PEM: %w", err)
+	}
+	if err := pem.Encode(&pemBuffer, pemCertificateBlock); err != nil {
+		return nil, fmt.Errorf("cannot convert x509 certificate to PEM: %w", err)
+	}
+	return pemBuffer.Bytes(), nil
```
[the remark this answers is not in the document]
flotter: Yes, thanks.

internals/overlord/tlsstate/manager.go
paul-rodriguez: I've skimmed through https://stackoverflow.com/questions/5355046/where-is-the-pem-file-format-specified which suggests that just concatenating the two pieces of data is probably fine. And you're right that it makes the code simpler.

anpep (changes requested):
Great job! I see how a lot of design effort went into this patch. Do review receiver types throughout the code (pointer vs. non-pointer receivers in methods): even when a struct is three, four fields that are actually pointers, it would be great that we didn't make copies of copies unnecessarily, specially if we are calling non-pointer receiver methods on loops (see `TLSKeyPair.Compare`). Also, as Paul pointed out, we are on `tlsstate` so removing `TLS` from many identifiers would make the code read less verbose while retaining clarity. Thank you!

internals/overlord/tlsstate/package_test.go
```diff
@@ -0,0 +1,172 @@
...
+		TLSConfig: &tls.Config{
+			MinVersion:       tls.VersionTLS13,
+			MaxVersion:       tls.VersionTLS13,
+			CurvePreferences: []tls.CurveID{tls.CurveP256},
+			GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+				cert := certSource()
+				tls, err := cert.TLSCertificate()
+				c.Assert(err, IsNil)
+				return tls, nil
+			},
```
hpidcock, line 157: We should use `GetConfigForClient` and return a `tls.Config` with all the certificates.

hpidcock (commented):
My comments are all over the place here as I've understood more what you are trying to achieve.

My suggestion is to break this down a bit further, reuse as much of the standard library as you can.

internals/overlord/tlsstate/manager.go
flotter: This is a pattern in Go used for sorting. We agreed with CTO we will order by start time for now.

internals/overlord/tlsstate/manager.go
flotter: Close does not imply sync last time I checked. It allows some users how do not care about lazy cache draining to leave it to the OS and continue with more important stuff.

internals/overlord/tlsstate/manager.go
flotter: The intent here is that I only create a leaf directory for myself and I assume the parent is owned and created by someone else. I prefer to be precise in what I expect, and what I should do and will do (especially when it comes to TLS permissions). I do not like ```MkdirAll``` in this case as you cannot possibly define beforehand what it will do. I want to be certain that if the parent directory does not exist, I fail. This is not a condition I support.

internals/overlord/tlsstate/export_test.go
```diff
@@ -0,0 +1,132 @@
...
+	return m.addNewTLSKeyPair(notBefore, notAfter)
+}
+
+func (m TLSManager) IsTLSKeypairSupported(keypair *TLSKeyPair) bool {
+	return m.isTLSKeyPairSupported(keypair)
+}
+
+func (t TLSKeyPair) IsExpired() bool {
+	return t.isExpired()
+}
```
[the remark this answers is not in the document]
flotter: I use pointer receivers when the object is updated, and non-pointer when its read-only. I will review this approach.

flotter:
image

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// tlsstate manages TLS keypairs on behalf of the daemon's HTTPS server.
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Resolving.

internals/overlord/tlsstate/export_test.go
flotter: Addressed. Closing.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+// tlsDir directory.
+//
+// The TLS manager will generate keypairs on demand, so no external keypair
+// generation or lifecycle management of existing keypairs will be required.
+//
+// However, is possible to pre-seed the tlsDir with TLSv1.3 supported keypairs.
+// The following steps provide an example of how to create a TLSv1.3 supported
+// self-signed TLS keypair (the private key must be encoded in PKCS8 format):
+//
+// ~/> openssl genpkey -algorithm EC -pkeyopt ec_paramgen_curve:prime256v1 -out ec-p256.pem
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	// Return the TLS keypair.
+	return selectedKeypair, nil
+}
+
+// TLSKeyPairs returns the slice of currently valid and supported tlsKeyPairs.
+func (m *TLSManager) TLSKeyPairs() []*TLSKeyPair {
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+			break
+		}
+		switch block.Type {
+		case "PRIVATE KEY":
+			t.privateKey, err = x509.ParsePKCS8PrivateKey(block.Bytes)
+			if err != nil {
+				return err
+			}
+		case "CERTIFICATE":
+			t.Certificate, err = x509.ParseCertificate(block.Bytes)
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+	return nil
+}
+
+// TLSKeyPair represents a TLS key pair (see tls.LoadX509KeyPair), which
+// includes both the x509 certificate and the private key (unexported). This
+// type can directly be converted to a tls.Certificate for configuring a TLS
+// listener, or exported as a PEM encoded x509 certificate for easy transport
+// over the daemon API.
+type TLSKeyPair struct {
```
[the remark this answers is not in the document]
flotter: Migrated to tls.Certificate as the type. Resolving.

internals/overlord/tlsstate/manager_test.go
```diff
@@ -0,0 +1,363 @@
...
+
+	for i, testOrder := range expectedOrder {
+		c.Assert(keypairs[i].Fingerprint(), Equals, tests[testOrder].keypair.Fingerprint())
+		c.Assert(keypairs[i].IsExpired(), Equals, tests[testOrder].expired)
+		c.Assert(mgr.IsTLSKeypairSupported(keypairs[i]), Equals, tests[testOrder].supported)
+	}
+}
+
+// TestBenchmarkTLSCreate checks how long it takes to generate a new TLS
+// keypair on demand. Run with "go test -check.vv" to see output.
```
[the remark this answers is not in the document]
flotter: Fixed.

internals/overlord/tlsstate/export_test.go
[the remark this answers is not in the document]
flotter: Resolving for now.

internals/overlord/tlsstate/export_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+// FakeSystemTime allows faking the time for the TLS manager.
+func (m *TLSManager) FakeSystemTime(date string, offset time.Duration) (restore func(), clock time.Time) {
```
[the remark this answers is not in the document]
flotter: Fixed. Resolving.

internals/overlord/tlsstate/export_test.go
```diff
@@ -0,0 +1,132 @@
...
+	systemTime = func() time.Time {
+		return now
+	}
+	return func() {
+		systemTime = old
+	}, now
+}
+
+// FakeCertificateValidityPeriod allows setting how long a new certificate will be valid for.
+func (m *TLSManager) FakeCertificateValidityPeriod(valid time.Duration) (restore func()) {
```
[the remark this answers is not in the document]
flotter: Fixed. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	if notAfterSort != 0 {
+		return notAfterSort
+	}
+	// Fingerprint last.
+	return strings.Compare(t.Fingerprint(), other.Fingerprint())
+}
+
+// Fingerprint returns an unpadded base32 encoded SHA512/384 hash of the raw
+// x509 certificate.
+func (t TLSKeyPair) Fingerprint() string {
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	return nil
+}
+
+// TLSKeyPair represents a TLS key pair (see tls.LoadX509KeyPair), which
+// includes both the x509 certificate and the private key (unexported). This
+// type can directly be converted to a tls.Certificate for configuring a TLS
+// listener, or exported as a PEM encoded x509 certificate for easy transport
+// over the daemon API.
+type TLSKeyPair struct {
+	privateKey  any
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	tlsKeyPairs    []*TLSKeyPair
+}
+
+// NewManager creates a new TLS manager and loads existing keypairs from the
+// tlsDir directory.
+//
+// The TLS manager will generate keypairs on demand, so no external keypair
+// generation or lifecycle management of existing keypairs will be required.
+//
+// However, is possible to pre-seed the tlsDir with TLSv1.3 supported keypairs.
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return nil, err
+	}
+	template := x509.Certificate{
+		SerialNumber: serialNumber,
+		NotBefore:    notBefore,
+		NotAfter:     notAfter,
+		Subject: pkix.Name{
+			Organization: []string{"Canonical Ltd."},
```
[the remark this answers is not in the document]
flotter: Removed, thank you. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// Load a clean X509 certificate from the DER encoded bytes.
+	certificate, err := x509.ParseCertificate(derBytes)
+	if err != nil {
+		return nil, err
+	}
+	keypair := &TLSKeyPair{privateKey: privateKey, Certificate: certificate}
+	keypair.addFingerprint()
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: Did exactly that. Thanks. Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: N/A anymore. Resolving

internals/overlord/tlsstate/package_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	certPool := x509.NewCertPool()
+	if verifyCert != nil {
+		// The root CA pool contains certificates which can verify the
+		// the incoming certificate signature. Since we are using self-signed
+		// certificates, we can actually verify the signature with the public
+		// key part. In practise we will not use the incoming certificate to
+		// verify itself. Instead, a previously saved copy obtained during a
+		// trust exchange routine (i.e. pairing) will be used to make sure
+		// the incoming certificate was signed by the same private key.
```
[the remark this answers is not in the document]
flotter: I have redesigned this with a master identity keypair, which signs this TLS keypair. The master identity key certificate is the CA. Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+
+// TLSKeyPairs returns the slice of currently valid and supported tlsKeyPairs.
+func (m *TLSManager) TLSKeyPairs() []*TLSKeyPair {
+	m.tlsKeyPairLock.Lock()
+	defer m.tlsKeyPairLock.Unlock()
+
+	var activeTLSKeyPairs []*TLSKeyPair
+	for _, keypair := range m.tlsKeyPairs {
+		if !keypair.isExpired() && m.isTLSKeyPairSupported(keypair) {
+			activeTLSKeyPairs = append(activeTLSKeyPairs, keypair)
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+		}
+	}
+	return activeTLSKeyPairs
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	// The signing algorithm will match the EC private key length.
+	certSigningSupported := false
+	switch keypair.Certificate.SignatureAlgorithm {
+	case x509.ECDSAWithSHA256, x509.ECDSAWithSHA384, x509.ECDSAWithSHA512:
+		certSigningSupported = true
+	}
+	// We only do ECDSA for now, as is evident from the above.
+	certPublicKeySupported := keypair.Certificate.PublicKeyAlgorithm == x509.ECDSA
+
+	return keySupported && certSigningSupported && certPublicKeySupported
```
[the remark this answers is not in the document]
flotter: PR reworked. N/A for current PR. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+// NewManager create a new TLS keypair manager. The tlsDir must be a
+// directory of which only the leaf element, at most, does not exist. The
+// leaf directory will be created with 0o700 permissions. The signer
+// must implement a crypto.Signer interface. The signer represents the
+// identity key of the machine, container or device. The signer will be
+// used to sign TLS keypairs and its certificate (CA) is included in the
+// TLS certificate chain, allowing a client to pin the identity
+// certificate once trust has been manually established as part of a
+// trust exchange procedure.
+func NewManager(tlsDir string, signer crypto.Signer) (*TLSManager, error) {
```
anpep, line 87: nit:
```suggestion
func NewManager(tlsDir string, signer crypto.Signer) *TLSManager {
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	return m.tlsCert, nil
+}
+
+// Ensure implements StateManager.Ensure.
+func (m *TLSManager) Ensure() error {
+	return nil
+}
+
+func (m *TLSManager) createTLSCert() error {
+	_, privateKey, err := ed25519.GenerateKey(rand.Reader)
```
anpep, line 137: nit: `nil` is equivalent to `crypto/rand.Reader`
```suggestion
	_, privateKey, err := ed25519.GenerateKey(nil)
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+			if blockCount == 1 {
+				return nil, fmt.Errorf("unexpected PEM block")
+			}
+			cert, err = x509.ParseCertificate(block.Bytes)
+			if err != nil {
+				return nil, err
+			}
+			blockCount += 1
+		default:
+			return nil, fmt.Errorf("unexpected PEM block")
```
anpep, line 265: nit: as in the other PR, how about we ignore unrecognized block types, and make sure to fail if no `CERTIFICATE` block was found?

anpep (approved):

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,511 @@
...
+	for _, keypair := range m.tlsKeyPairs {
+		if !keypair.isExpired() && m.isTLSKeyPairSupported(keypair) {
+			activeTLSKeyPairs = append(activeTLSKeyPairs, keypair)
+		}
+	}
+	return activeTLSKeyPairs
+}
+
+// Ensure implements StateManager.Ensure.
+func (m *TLSManager) Ensure() error {
```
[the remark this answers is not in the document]
flotter: Resolving.

internals/overlord/tlsstate/manager.go
flotter: Removed. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+		NotBefore:             systemTime(),
+		NotAfter:              systemTime().Add(idCertValidity),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+		// We can only sign leaf certificates with this.
+		MaxPathLen:     0,
+		MaxPathLenZero: true,
+	}
```
ghislainbourgeois, line 320: The subject field seems to be missing. I am not sure what go does in this case, but according to the RFC, the subject field is required: https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.6

In this case, it might be unnecessary if the client and server are from the same code base, but if it is expected to have other clients, this might break validation. I would suggest populating it at least with something like a uuid.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+		return err
+	}
+	template := x509.Certificate{
+		SerialNumber:          serialNumber,
+		NotBefore:             systemTime(),
+		NotAfter:              systemTime().Add(tlsCertValidity),
+		KeyUsage:              x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+	}
```
ghislainbourgeois, line 153: I left a similar comment on the CA generation, and would probably add a subject field here.

ghislainbourgeois (changes requested):
I have some concerns about the lack of subject fields. Otherwise I think this implementation looks good.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	// The identity certificate loaded from disk.
+	idCert *x509.Certificate
+	// The identity key used for signing TLS certificates.
+	signer crypto.Signer
+}
+
+// NewManager create a new TLS keypair manager. The tlsDir must be a
+// directory of which only the leaf element, at most, does not exist. The
+// leaf directory will be created with 0o700 permissions. The signer
+// must implement a crypto.Signer interface. The signer represents the
```
benhoyt, line 81: This sentence is unnecessary, given the type:

```suggestion
// leaf directory will be created with 0o700 permissions. The signer represents the
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	return manager, nil
+}
+
+// GetCertificate returns an identity signed TLS certificate. The certificate chain includes
+// both the TLS leaf certificate, as well as the self signed identity CA certificate. If
+// either the identity or TLS certificate nears expiry, this functions creates new
+// certificates on demand. Note that even if the identity certificate is re-created, this
+// does not mean that the identity key changed (the key itself has no expiry).
+func (m *TLSManager) GetCertificate(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+	var err error
```
benhoyt, line 101: Why is this needed, instead of the more common `err := foo()` pattern below?

internals/overlord/tlsstate/manager.go
benhoyt: Per comment on other PR, probably best to keep this explicit in security-sensitive code.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return err
+	}
+	template := x509.Certificate{
+		SerialNumber:          serialNumber,
+		NotBefore:             systemTime(),
+		NotAfter:              systemTime().Add(tlsCertValidity),
```
benhoyt, line 149: For sanity, it's probably best to call `now := systemTime()` once, and then use that one value for both of these, rather than have two values a little bit out.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	return os.Mkdir(m.tlsDir, 0o700)
+}
+
+// getIDCert verifies that the identity certificate exists, and that the
+// certificate matches the supplied identity key signer. If the identity key
+// has changed, or the certificate has expired, the identity certificate is
+// replaced with a new identity certificate signed with the identity key.
+func (m *TLSManager) getIDCert() error {
+	idPath := filepath.Join(m.tlsDir, idCertFile)
+	exists, err := pathExists(idPath)
```
benhoyt, line 194: Do we need to call this if `m.idCert != nil`?

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Load the identity certificate it not yet loaded.
```
benhoyt, line 199:
```suggestion
	// Load the identity certificate if not yet loaded.
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+func createIDCert(signer crypto.Signer) (*x509.Certificate, error) {
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return nil, err
+	}
+	template := &x509.Certificate{
+		SerialNumber:          serialNumber,
+		NotBefore:             systemTime(),
+		NotAfter:              systemTime().Add(idCertValidity),
```
benhoyt, line 312: Similar to above, let's fetch `systemTime()` only once.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	return true
+}
+
+// isCertDerived checks if the public key in the certificate matches
+// the signer public key. This confirms the certificate matches the
+// identity key.
+func isCertDerived(cert *x509.Certificate, signer crypto.Signer) bool {
+	signerPublic := signer.Public().(ed25519.PublicKey)
+	certPublic := cert.PublicKey.(ed25519.PublicKey)
+	return bytes.Equal(signerPublic, certPublic)
```
benhoyt, line 354: Should we use [subtle.ConstantTimeCompare](https://pkg.go.dev/crypto/subtle#ConstantTimeCompare) for this (and any similar code)?

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	return nil
+}
+
+// pathExists returns true of the path exists, false if it does not. If the
+// operation reports an unrelated error, the error is returned.
+func pathExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
```
benhoyt, line 376: Per the docs, this predates `errors.Is`, and new code should use `errors.Is(err, fs.ErrNotExist)` (similar in the other PR).

internals/overlord/tlsstate/package_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Hook up check.v1 into the "go test" runner.
+func Test(t *testing.T) { TestingT(t) }
+
+type tlsSuite struct{}
+
+var _ = Suite(&tlsSuite{})
+
+// GetFakeTime converts a simple string date to the time.Time type for testing purposes.
+func (ts *tlsSuite) GetFakeTime(c *C, date string) time.Time {
```
benhoyt, line 40: This would be better as a top-level function, as it isn't related to `ts` at all. I think `getFakeTime`.

internals/overlord/tlsstate/package_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if ca != nil {
+		certPool.AddCert(ca)
+	}
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				MinVersion:         tls.VersionTLS13,
+				MaxVersion:         tls.VersionTLS13,
+				InsecureSkipVerify: true,
+				VerifyConnection: func(ts tls.ConnectionState) error {
```
benhoyt, line 72: This is unfortunate shadowing of `ts *tlsSuite` above. Can we call this something else, `connState` maybe?

internals/overlord/tlsstate/package_test.go
```diff
@@ -0,0 +1,138 @@
...
+	c.Assert(err, IsNil)
+	c.Assert(string(body), Equals, "TLS 1.3!")
+
+	return serverCerts, nil
+}
+
+func (ts *tlsSuite) testTLSServer(c *C, getCertificate func(*tls.ClientHelloInfo) (*tls.Certificate, error)) (shutdown func()) {
+	// First start a listener so we buffer client requests while the HTTPS
+	// server routine is starting up.
+	listener, err := net.Listen("tcp", ":8888")
```
benhoyt, line 111: Can we use the trick (from the other PR) where Listen finds a free port and we get the port via the `Addr` method? `net.Listen("tcp", ":")` I think.

internals/overlord/tlsstate/package_test.go
```diff
@@ -0,0 +1,138 @@
...
+		err := server.ServeTLS(listener, "", "")
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
+			c.Fatalf("Cannot start server: %v", err)
+		}
+	}()
+
+	// The server must be shutdown when client tests are completed.
+	return func() {
+		server.Shutdown(context.Background())
+		listener.Close()
```
benhoyt, line 136: I think this should wait for the goroutine to finish, either using channels or a tomb. Otherwise `c.Fatalf` might be called when `c` is not valid anymore.

benhoyt (commented):
I've reviewed this, though I must say I'm a bit out of my depth. Would love thorough review from @ghislainbourgeois on the TLS/security aspects.

internals/overlord/tlsstate/manager.go
flotter: Fixed, Resolving.

internals/overlord/tlsstate/manager.go
flotter: Fixed (matching first PR). Resolving.

internals/overlord/tlsstate/manager.go
flotter: @ghislainbourgeois 

I have now addressed this. By default, I now generate at least a Subject Common Name (which Go appears to copy to the Issuer Common Name if the Issuer is left empty.

The Common Name now looks like this:

```<Program Name>-<Fingerprint[:8]>```

which means for Pebble it will look something like this:

```pebble-A453BD6C```

For derivative, the program name will change. But also, there is a ```SetX509Templates``` method now added that will allow templates to be provided externally, which can be used to supply ```Subject```, ```Issuer``` and the ```SAN``` fields (```DNSNames``` and ```EmailAddresses```) for now.

I think this should be sufficient for now. What do you think ?

File: internals/overlord/tlsstate/manager.go (hunk 39 above)
Comment (flotter) [reply internals/overlord/tlsstate/manager.go]: This has been address as part of the previous comment, so I will resolve this one, and use the other one for future conversation. Resolving.

File: internals/overlord/tlsstate/export_test.go [hunk 52]
```diff
@@ -0,0 +1,64 @@
...
+package tlsstate
+
+import (
+	"time"
+)
+
+// FakeSystemTime fakes the system time for the TLS manager.
+func FakeSystemTime(date string, offset time.Duration) (restore func(), clock time.Time) {
+	layout := "2006-01-02"
+	now, err := time.Parse(layout, date)
```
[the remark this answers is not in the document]
flotter: Vastly simplified as per your next comment proposal. Resolving.

internals/overlord/tlsstate/export_test.go
```diff
@@ -0,0 +1,64 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package tlsstate
+
+import (
+	"time"
+)
+
+// FakeSystemTime fakes the system time for the TLS manager.
+func FakeSystemTime(date string, offset time.Duration) (restore func(), clock time.Time) {
```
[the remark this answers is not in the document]
flotter: Yes, thanks. I cleaned this up and now it looks much cleaner. Resolving.

```go
func FakeTimeNow(t time.Time) (restore func()) {
        old := timeNow
        timeNow = func() time.Time {
                return t
        }
        return func() {
                timeNow = old
        }
}
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	systemTime = time.Now
+
+	// idCertValidity sets the x509 identity key validity period. The
+	// certificate is managed by the TLS manager as it is not directly
+	// supported by a crypto.Signer interface, as its an implementation
+	// detail related to TLS (not a signing key). Expiry of this
+	// certificate does not change the identity, but changes the client
+	// pinned identity certificate, forcing the client to re-pair with
+	// the server.
+	idCertValidity = 10 * 365 * 24 * time.Hour
```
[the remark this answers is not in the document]
flotter: I've settled for just adding a day (366). Getting this actually correct depends on leap days and also leap seconds, which feels like making this more complex than what is needed.

I've added a line in the description saying its just over 10 years. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Fixed, thanks you. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,382 @@
...
+	"math/big"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+)
+
+var (
+	// systemTime can be faked during testing.
+	systemTime = time.Now
```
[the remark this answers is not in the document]
flotter: Fixed, thanks. looks much better :) Resolving.

internals/overlord/tlsstate/manager.go
flotter: This was a leftover line. Removing. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Yes. I fixed this in two places, thanks. Resolving.

internals/overlord/tlsstate/manager.go
flotter: I've reshuffled it a little bit, and I think its looks better now. The path is now only checked for the case where no certificate was previously loaded. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Fixed. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Fixed. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Sounds good! Done. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Fixed. Resolving.

internals/overlord/tlsstate/package_test.go
flotter: Deleted after refactoring test time. Resolving.

internals/overlord/tlsstate/package_test.go
flotter: Since it is only used once on the very next line, I just made it ```cs```. Resolving.

internals/overlord/tlsstate/package_test.go
flotter: Yes, thanks. Done. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+
+	// Create the leaf directory node with 0o700 permissions.
+	return os.Mkdir(m.tlsDir, 0o700)
+}
+
+// getIDCert verifies that the identity certificate exists, and that the
+// certificate matches the supplied identity key signer. If the identity key
+// has changed, or the certificate has expired, the identity certificate is
+// replaced with a new identity certificate signed with the identity key.
+func (m *TLSManager) getIDCert() error {
```
hpidcock, line 234: This does not return a certificate, it should have a better name.
```suggestion
func (m *TLSManager) ensureIDCert() error {
```

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Can we use the loaded identity certificate?
+	if m.idCert != nil {
+		isDerived := isCertDerived(m.idCert, m.signer)
+		isActive := isCertActive(m.idCert)
```
hpidcock, line 255: I am concerned about this. You will need to make sure the time resynchronisation implementation is robust. If we rely on plain old NTP, we should probably disable this regeneration (i.e. make it a documented manual process if time is not considered acquired securely/robustly).

hpidcock (approved):
Happy to +1 this, there is still work that can be done, but I think it is in a better place.

I've added a few nits, and one concerning point, but these should not block us.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+	prefixLen := len(cn.String())
+	// 8 bytes for the identity fingerprint.
+	for _, r := range fingerprint {
+		cnLen := len(cn.String())
+		runeLen := len(string(r))
+		if (cnLen + runeLen) > (prefixLen + 8) {
+			break
+		}
+		cn.WriteRune(r)
+	}
```
ghislainbourgeois, line 450: Because we control the fingerprint format and it is base32, we do not need this loop here and can just take a slice.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+	var cn strings.Builder
+	// <= 55 bytes for the prefix (program name)
+	for _, r := range cmd.ProgramName {
+		cnLen := len(cn.String())
+		runeLen := len(string(r))
+		if (cnLen + runeLen) > 55 {
+			break
+		}
+		cn.WriteRune(r)
+	}
```
ghislainbourgeois, line 438: This could be computed on startup to avoid the loop every time we generate a new certificate.

ghislainbourgeois (changes requested):

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+	return cert, nil
+}
+
+// isCertActive checks of the current time now is within the start and end time
+// validity period as defined by the certificate.
+func isCertActive(cert *x509.Certificate) bool {
+	now := timeNow()
+	// Note that we shorten the NotAfter timestamp by a tlsCertRenewWindow
+	// duration to avoid a TLS handshake race against the expiry time.
+	if now.Before(cert.NotBefore) || now.After(cert.NotAfter.Add(-tlsCertRenewWindow)) {
```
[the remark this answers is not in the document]
flotter: This is a time package curve-ball. Sub calculates a duration, not a new time.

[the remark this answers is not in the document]
ghislainbourgeois: Oh, TIL, thanks!

benhoyt (approved):
This looks good to me now, in terms of Go code structure and from a non-TLS-expert point of view -- so please ensure to resolve all comments from Harry and Ghislain before merging.

internals/overlord/tlsstate/manager.go
flotter: Resolving this comment.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var cert *x509.Certificate
+	pemData, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	block, _ := pem.Decode(pemData)
```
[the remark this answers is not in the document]
flotter: I thought about this a bit, and I hope you are OK with my solution:

This file is officially under our control. So, I've opted for a slightly stricter option - I will error if I detect unexpected bytes at the end of the block. I think this approach is fine because in practice, there is really only one case where we would see extra stuff at the end, and this is if some workflow uses PEBBLE_COPY_ONCE to provision a specific identity and certificate from a Pebble install. In this case, let's teach people from day one how to do it correctly, rather than warning, that may be missed. The cost is low, since it will fail on the very first attempt at doing this.

The PEM decoder is quite flexible, so in practice this means that:

1. If there is nonsense preamble before the first block, it will be OK since the decoder will look for the start of the first block.
2. The first block must be of type PRIVATE KEY.
3. A newline after the block is OK, but anything more is not OK.

This feels fair to me. Do you feel happy with this ?

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+	pemFile, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = errors.Join(err, pemFile.Sync())
+		err = errors.Join(err, pemFile.Close())
+	}()
+
+	if _, err = pemFile.Write(pemBytes); err != nil {
```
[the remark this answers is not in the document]
flotter: Doh, yes thank you. I am now encoding directly into the file. Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+// pathExists returns true of the path exists, false if it does not. If the
+// operation reports an unrelated error, the error is returned.
+func pathExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+		return false, err
+	}
```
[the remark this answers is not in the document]
flotter: Tweaked as per suggestion, thank you. Resolving.

internals/overlord/tlsstate/manager.go
flotter: Just to summarize:

1. TLS code uses time.Now() for certificate validity checking unless we supply it with a different time source in the tls.Config when we create the HTTPS server.

2. This manager will use time.Now() as well for creating certificates (as Harry points out), and deciding on the validity range based on the current time.

So I would say that in order to solve this, we need a secure system time source, perhaps support for NTS. At the same time I think it would be great to have a functional solution without NTS (or some equivalent secure system time), for use cases where hardware support or reliable internet is simply is not available.

I would say let's start with this, knowing the risks, and I will look at an optional NTS secure time source as a next step?

https://pkg.go.dev/github.com/beevik/nts

internals/overlord/tlsstate/manager.go
flotter: Fixed, thank you. Resolving.

internals/overlord/tlsstate/manager_test.go
```diff
@@ -0,0 +1,507 @@
...
+	// identity certificate by our client.
+	certs, err := ts.testTLSInsecureClient(c, testBaseTime)
+	c.Assert(err, IsNil)
+	tlsCert1 := certs[0]
+	idCert := certs[1]
+	restoreTime()
+
+	// Set the system time to 5 seconds before the actual timeout, but
+	// within the renewal window.
+	renewWindow := 5 * time.Second
```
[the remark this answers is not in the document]
flotter: I change the test point to the first second (After) entering the window, which is the first time it will renew it. So it now tests for 59 second before expiry, which is immediately after the 60 seconds window.

I also updated the description of the tlsCertRenewWindow field to explain it does only renew _after_ entering the window, not exactly on this point in time.

Resolving.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,470 @@
...
+	return true
+}
+
+// isCertDerived checks if the public key in the certificate matches
+// the signer public key. This confirms the certificate matches the
+// identity key.
+func isCertDerived(cert *x509.Certificate, signer crypto.Signer) bool {
+	signerPublic := signer.Public().(ed25519.PublicKey)
+	certPublic := cert.PublicKey.(ed25519.PublicKey)
+	return subtle.ConstantTimeCompare(signerPublic, certPublic) == 1
```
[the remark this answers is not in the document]
flotter: The original bytes.Equal implementation was taken from the Go crypto TLS package, its used everywhere in that package, so let me fall back to what I had previously.

Resolving.

internals/overlord/tlsstate/manager.go
flotter: True. I simplified the implementation to assume that. Thank you. Resolving.

internals/overlord/tlsstate/manager.go
flotter: @ghislainbourgeois 

The current design does everything on-demand. This allows the identity cert and TLS keypairs only to exist if someone actually uses HTTPS (starts up Pebble with --https and a session is started). Also, in theory, if the identity did not come from a file, it could change at runtime (maybe this is not possible in practice, but I could swap out YubiKeys), and the system would currently ensure the identity certificate and public key match, and if not, would update the identity certificate. Since the fingerprint is essentially the public identity key, the current design is too dynamic to simply create it at startup. I have measured the impact, and since this only happens on an actual first time cert create, or renewal, it feels fairly light weight.

Will this be OK for you for as a starting point?

ghislainbourgeois: I think it's OK, this comment was really only a suggestion, it can always be improved in the future if required.

ghislainbourgeois (approved):
LGTM

internals/overlord/tlsstate/manager.go
hpidcock: Yep, my intent was to highlight the risk here. How you proceed is up to you. Regenerating the CA certificate because of a potentially externally influenced variable (e.g. network time), means this is a potentially useful tool to deny service. Clients should never just assume they trust the new CA certificate, and automatically trusting it because it has the same signer, is also not valid, since that makes it impossible to regenerate leaked PKs.

We need to ensure clients must re-verify the authenticity of the certificate/PK. Right now, manually is the only option. But in doing so, this opens up devices with non-robust time or exploitable time, to attacks which degrade a human's vigilance. Maybe at some point, someone accepts a PK/Certificate that is controlled by an attacker.

So, back to why I suggest we don't automatically regenerate the Certificate. If a client, connects to a pebble instance, with an expired CA, but one we have seen before, we can regenerate that CA, in collaboration with the client as a witness. But this is for the future. For now, disabling the automatic CA regeneration + adding a flag to clients to connect, but allow that the last seen CA can be expired, would allow someone to work through this manually.

internals/overlord/tlsstate/manager.go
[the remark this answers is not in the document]
flotter: Resolving following agreement.

internals/overlord/tlsstate/manager.go
flotter: Thanks for the input. Agreed, let's see how this performs in practice and adapt.

internals/overlord/tlsstate/manager.go
flotter: @hpidcock If possible, I just want to get your feedback on this topic in person. I will reach out to you to see if that will be possible.

flotter: Thanks for the chat @hpidcock. We will do one small tweak for now:

>> Disable one re-generation scenario, where the identity certificate expires.

The identity certificate will still generate on demand, and it will still generate if the identity private key changes.

The logic here is that we are limiting scope of time based attacks on the identity certificate. This does mean that the client may encounter a device with expired identity certificate (in 10years), and we need to add the ability for the client or the server to safely generate a new identity certificate in a more controlled fashion.

flotter: Resolving.

internals/overlord/tlsstate/manager.go
```diff
@@ -0,0 +1,487 @@
...
+// certificates. The fingerprint must not include non-ASCII characters, but
+// this is not the case (see: idkey package).
+func defaultCertSubject(fingerprint string) pkix.Name {
+	// The common name must not exceed 64 bytes (RFC 5280).
+	var cn strings.Builder
+	// <= 55 bytes for the prefix (program name). The program name
+	// will never have non-ASCII characters (r < 128).
+	for _, r := range cmd.ProgramName {
+		cn.WriteRune(r)
+		if len(cn.String()) == 55 {
```
flotter, line 459: replace with dn.Len()

flotter: In fact, redo this assuming ascii.

PR #598, opened 2025-04-08
docs: document how to tail file logs

dimaqq:
The proposal to add file tailing functionality directly to Pebble was rejected.
This PR documents the workaround.

### DIFF
--- a/docs/how-to/capture-logs-from-files.md
+++ b/docs/how-to/capture-logs-from-files.md
@@ -1,4 +1,4 @@
-# How to tail file logs
+# How to capture logs from files
 
 Some applications do not log to `stdout` or `stderr`, but instead write logs directly to
 files such as `access.log`, `audit.log` or `error.log`. While Pebble does not natively
--- a/docs/how-to/capture-logs-from-files.md
+++ b/docs/how-to/capture-logs-from-files.md
@@ -27,8 +27,9 @@ services:
     startup: enabled
 ```
 
-The separate helper service `tail -F` starts before the main service.
-The `-F` option ensures that logs are captured reliably, even if log files are rotated.
+The helper service `foo-error-log` starts before the main service.
+The `-F` option of the `tail` command ensures that logs are captured reliably,
+even if log files are rotated.
 
 This setup handles a variety of common corner cases:
 - the log file doesn't exist when the service starts
--- a/docs/how-to/capture-logs-from-files.md
+++ b/docs/how-to/capture-logs-from-files.md
@@ -14,8 +14,6 @@ To use `tail`, we'll define a layer that has a main service (`foo`) and a helper
 for `tail`:
 
 ```{code-block} yaml
-  :emphasize-lines: 5, 7, 10
-
 services:
   foo:
     command: foo
@@ -30,6 +28,10 @@ services:
 ```
 
 The helper service `foo-error-log` starts before the main service.
+Likewise, if the services are disabled, the helper service `foo-error-log` will be
+stopped after the main service `foo`.
+This ensures that logs emitted during both startup and termination are captured.
+
 The `-F` option of the `tail` command ensures that logs are captured reliably,
 even if log files are rotated.
 
@@ -37,7 +39,6 @@ This setup handles common corner cases:
 - The log file doesn't exist when the service starts
 - The log file is truncated, deleted or replaced
 - A safety net if the log file is large at startup
-- Controlled stop order if services are disabled
 
 A separate helper service is required for each log file.
 
@@ -64,4 +65,4 @@ To learn more about Rockcraft and `rockcraft.yaml` files, see the
 ## See more
 
 - After capturing logs, you can [forward the logs to Loki](./forward-logs-to-loki).
-- If `tail` is insufficient, consider using Promtail instead of `tail`.
+- If `tail` is insufficient, consider using the Promtail binary instead.
--- a/docs/how-to/capture-logs-from-files.md
+++ b/docs/how-to/capture-logs-from-files.md
@@ -3,13 +3,13 @@
 Some applications do not log to `stdout` or `stderr`, but instead write logs directly to
 files such as `access.log`, `audit.log` or `error.log`. While Pebble does not natively
 support tailing files, you can use `tail` from `coreutils` in a separate Pebble service
-to route file contents into Pebble’s log stream.
+to route file contents into Pebble's log stream.
 
 This guide shows how to set that up.
 
 ## Example Pebble layer
 
-Here’s how to tail a log file alongside your main service:
+Here's how to tail a log file alongside your main service:
 
 ```{code-block} yaml
   :emphasize-lines: 5, 7, 10
--- a/docs/how-to/capture-logs-from-files.md
+++ b/docs/how-to/capture-logs-from-files.md
@@ -1,11 +1,12 @@
 # How to capture logs from files
 
-Some applications do not log to `stdout` or `stderr`, but instead write logs directly to
-files such as `access.log`, `audit.log` or `error.log`. While Pebble does not natively
-support tailing files, you can use `tail` from `coreutils` in a separate Pebble service
-to route file contents into Pebble's log stream.
+Pebble stores the most recent `stdout` and `stderr` from each service. However, some
+applications don't log to `stdout` or `stderr`. Instead, they write logs directly to
+files such as `access.log`, `audit.log`, or `error.log`.
 
-This guide shows how to set that up.
+This guide demonstrates how to use the `tail` command to print log files to `stdout`,
+so that Pebble can capture the logs. The `tail` command is provided by the `coreutils`
+package.
 
 ## Define a layer
 
--- a/docs/how-to/capture-logs-from-files.md
+++ b/docs/how-to/capture-logs-from-files.md
@@ -41,7 +41,10 @@ A separate helper service is required for each log file.
 
 ## Include `tail` in a rock image
 
-If your workload is packaged as a rock, you may need to explicitly add `tail`:
+If your main service runs inside a container image created by Rockcraft,
+you might need to explicitly include `tail` when you build the image.
+
+To include `tail`, add the following part to your `rockcraft.yaml` file:
 
 ```yaml
 parts:
@@ -52,6 +55,9 @@ parts:
     organize:
       usr/bin/tail: usr/bin/tail
 ```
+
+To learn more about Rockcraft and `rockcraft.yaml` files, see the
+[Rockcraft documenttion](https://documentation.ubuntu.com/rockcraft/en/stable/).
 
 ## See more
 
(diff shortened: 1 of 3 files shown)


dwilding:
Thanks @dimaqq! I'm starting to review this and will leave suggestions or push commits as you recommended. I initially found the title a bit cryptic, so I'm proposing "Capture logs from files" instead - I've just pushed a commit to change that, but would be happy to further revise :slightly_smiling_face: 

BTW, here's the link to the preview build: https://canonical-ubuntu-documentation-library--598.com.readthedocs.build/pebble/how-to/capture-logs-from-files/

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+so that Pebble can capture the logs. The `tail` command is provided by the `coreutils`
+package.
+
+## Define a layer
+
+To use `tail`, we'll define a layer that has a main service (`foo`) and a helper service
+for `tail`:
+
+```{code-block} yaml
+  :emphasize-lines: 5, 7, 10
```
dwilding, line 17: I didn't find the emphasis particularly helpful. I think we can remove it, especially as the example is pretty short.

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+The helper service `foo-error-log` starts before the main service.
+The `-F` option of the `tail` command ensures that logs are captured reliably,
+even if log files are rotated.
+
+This setup handles common corner cases:
+- The log file doesn't exist when the service starts
+- The log file is truncated, deleted or replaced
+- A safety net if the log file is large at startup
```
dwilding, line 39: "A safety net if the log file is large at startup" - Is the point here that we avoid dumping a large log before the service has done anything meaningful?

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+- Controlled stop order if services are disabled
```
dwilding, line 40: "Controlled stop order if services are disabled" - I feel we could leave this one out.

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+      usr/bin/tail: usr/bin/tail
+```
+
+To learn more about Rockcraft and `rockcraft.yaml` files, see the
+[Rockcraft documentation](https://documentation.ubuntu.com/rockcraft/en/stable/).
+
+## See more
+
+- After capturing logs, you can [forward the logs to Loki](./forward-logs-to-loki).
+- If `tail` is insufficient, consider using Promtail instead of `tail`.
```
dwilding, line 67: Is this Promtail? https://grafana.com/docs/loki/latest/send-data/promtail/. I see that it's deprecated. Do you think it would still be OK to recommend? If so, we should link to more details.

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
+# How to capture logs from files
+
+Pebble stores the most recent `stdout` and `stderr` from each service. However, some
+applications don't log to `stdout` or `stderr`. Instead, they write logs directly to
+files such as `access.log`, `audit.log`, or `error.log`.
+
+This guide demonstrates how to use the `tail` command to print log files to `stdout`,
+so that Pebble can capture the logs. The `tail` command is provided by the `coreutils`
```
dwilding, line 8: Do you think we need to link somewhere for `coreutils`? Perhaps not, as it's so well known.

dwilding (approved):
I think this is going to be a helpful new doc, thanks again. I added a few minor comments and pushed a bunch more commits - please take a look to make sure you're happy with my changes.

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+```{code-block} yaml
+  :emphasize-lines: 5, 7, 10
+
+services:
+  foo:
+    command: foo
+    startup: enabled
+    requires:
+      - foo-error-log
+    after:
```
IronCore864, line 25: What happens if it's the first time the app starts and the log file doesn't exist yet?

IronCore864 (approved):

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+      - foo-error-log
+  foo-error-log:
+    command: tail -F /var/log/foo/error.log
```
james-garner-canonical, line 28: WDYT about using or documenting the long form here? e.g.
```suggestion
    command: tail --retry --follow=/var/log/foo/error.log
```

james-garner-canonical: Based purely on my looking at the manual and seeing `-F` documented as "same as `--follow=name --retry`"

benhoyt: I'm usually in favour of long args, but I'm not sure about this one. I think `tail -f` and `tail -F` are *much* more well-known than `--follow` and co, so I think it'd actually be less clear to many. I could be wrong. In addition, `tail -F` is a bit more portable (not that that matters here), for example macOS https://ss64.com/mac/tail.html

docs/how-to/capture-logs-from-files.md
```diff
@@ -0,0 +1,67 @@
...
+    startup: enabled
+```
+
+The helper service `foo-error-log` starts before the main service.
+The `-F` option of the `tail` command ensures that logs are captured reliably,
+even if log files are rotated.
```
james-garner-canonical, line 34:
```suggestion
even if log files are rotated, because in addition to following a specific file, it will check
and retry if the file is renamed or rotated, or does not exist yet.
```

docs/how-to/capture-logs-from-files.md
james-garner-canonical: Makes sense, maybe just a little explanation then, new suggestion below.

docs/how-to/capture-logs-from-files.md
dimaqq: Sure, removing is fine.

docs/how-to/capture-logs-from-files.md
dimaqq: Yes that's exactly that.

docs/how-to/capture-logs-from-files.md
dimaqq: Ah! There's Promtail and there's Promtail...

They are the same Promtail, but used differently.

I've hit that page too and was confused too. That page talks about "Promtail [as a data source for Grafana] is now deprecated", note the heading above, **Promtail agent**. So they recommend their Alloy (agent) which is much larger (logs and metrics and something) instead.

Meanwhile the Promtail (binary) is here:
https://github.com/grafana/loki/tree/main/clients/cmd/promtail
https://github.com/grafana/loki/tree/main/clients/pkg/promtail

Would it help if this read "consider using Promtail binary" instead?

docs/how-to/capture-logs-from-files.md
dimaqq: Well this was called out in the spec review:

> I have concerns around termination. The `tail` must stop after the log producing service. Otherwise the `tail` service will miss the death throes of the log producing service.

Maybe this line could be more precise?

docs/how-to/capture-logs-from-files.md
dimaqq: `tail -F` handles that.
- tail will log one line that file is missing
- when files is created, it will start tailing

docs/how-to/capture-logs-from-files.md
dimaqq: nah, well-known.

docs/how-to/capture-logs-from-files.md
dimaqq: +1 on what Ben said.

And for a good reason: you got it wrong in your suggestion, James!

`--follow` accepts one of two **literals** `name` or `descriptor`, meaning strategy to to use when a file is moved, not the file name.

I know it's not intuitive... that's why I didn't want to suggest the long form.

dimaqq:
Wow, this reads so much better now, thank you @dwilding

docs/how-to/capture-logs-from-files.md
dimaqq: I'm not sure about this, there's a list of things that `tail -F` does right blow.
Spelling it out in a sentence seems redundant.

I am however looking for ideas on how to structure this better, because there's a mix of concerns:
- emitting log lines as they are added
- handling the log file being created/moved/truncated/deleted
- service start and stop order

Somehow I'm not too cozy with all three (or the latter two) mixed in a single list.

docs/how-to/capture-logs-from-files.md
dimaqq: Moved to the paragraph above and expanded, PTAL.

docs/how-to/capture-logs-from-files.md
james-garner-canonical: Ah so the correct long form would be `tail --follow=name --retry /var/log/foo/error.log`, thanks for explaining -- I can see why they added the `-F` flag for that.

docs/how-to/capture-logs-from-files.md
james-garner-canonical: I guess what I'm thinking is that there have been a couple of cases of us saying "oh but what about this edge case", where the answer is "`-F` handles that". That was my motivation for suggesting the long form (but I'm happy with using the short form).

An idea for structuring things, replacing the bullet points below, and the sentences above. Feel free to ignore this idea, it's just what makes sense to me, and it's probably redundant with other information in this document:

The `tail` command will capture content as it is added to the log file, providing a safety net if the log file is already large at startup. The `-F` option means it will follow the log file by name, retrying if the log file doesn't exist yet (for example when the service starts), or if the log file is truncated, deleted, or replaced.

This enables us to require that the helper service `foo-error-log` starts before the main service. Likewise, if the services are disabled, the helper service `foo-error-log` will be stopped after the main service `foo`. This ensures that logs emitted during both startup and termination are captured.

dimaqq: I think I'll leave it in the current, 6a9a0da state.
We can always improve the text based on feedback later.

benhoyt (approved):
Short and sweet -- looks good to me!

dwilding:
@dimaqq I was just reading the latest version, to catch up on changes. I like this a lot:

> The helper service `foo-error-log` starts before the main service. Likewise, if the services are disabled, the helper service `foo-error-log` will be stopped after the main service `foo`. This ensures that logs emitted during both startup and termination are captured.

It feels much better to mention the ordering here, rather than classifying it as a corner case. Nice one :grinning:

PR #599, opened 2025-04-08
feat(https): add https API server [3/4]

flotter:
**Enable the API over HTTPS**

Pebble can now be started with the API enabled on HTTPS:

```
~/> ./pebble run --https=:8888 --http=:4000 --create-dirs
2025-04-08T15:14:40.523Z [pebble] HTTP API server listening on ":4000".
2025-04-08T15:14:40.524Z [pebble] HTTPS API server listening on ":8888".
2025-04-08T15:14:40.524Z [pebble] Started daemon.
2025-04-08T15:14:40.525Z [pebble] POST /v1/services 190.249µs 400
2025-04-08T15:14:40.525Z [pebble] Cannot start default services: no default services
```

The HTTPS (TLS) keypair is signed by the identity certificate (also included in the certificate chain from the server), so you need to connect with the client configured to allow for an untrusted connection.

The typical connection process will require a trust exchange step that allows the client to trust (pin) the identity certificate. This trust exchange processes (or pairing process) will be added in a future patch.

Note that this patch now enables HTTP2 negotiation (using TLS ALPN), with fallback to HTTP1.1 if the client does not explicitly agree.

```
~/>  curl -k https://localhost:8888/v1/health -vv
* Host localhost:8888 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8888...
* Connected to localhost (::1) port 8888
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256 / X25519 / ED25519
* ALPN: server accepted h2
* Server certificate:
*  subject: [NONE]
*  start date: Apr 22 13:30:40 2025 GMT
*  expire date: Apr 22 14:30:40 2025 GMT
*   Certificate level 0: Public key type ED25519 (256/128 Bits/secBits), signed using ED25519
*   Certificate level 1: Public key type ED25519 (256/128 Bits/secBits), signed using ED25519
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://localhost:8888/v1/health
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: localhost:8888]
* [HTTP/2] [1] [:path: /v1/health]
* [HTTP/2] [1] [user-agent: curl/8.5.0]
* [HTTP/2] [1] [accept: */*]
> GET /v1/health HTTP/2
> Host: localhost:8888
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/2 200 
< content-type: application/json
< content-length: 73
< date: Tue, 22 Apr 2025 13:30:41 GMT
< 
* Connection #0 to host localhost left intact
{"type":"sync","status-code":200,"status":"OK","result":{"healthy":true}}
```

### DIFF
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -869,6 +869,9 @@ pebble run --args myservice --port 8080 \; --hold
           --http=        Start HTTP API listening on this address in
                          "<address>:port" format (for example, ":4000",
                          "192.0.2.0:4000", "[2001:db8::1]:4000")
+          --https=       Start HTTPS API listening on this address in
+                         "<address>:port" format (for example, ":8443",
+                         "192.0.2.0:8443", "[2001:db8::1]:8443")
       -v, --verbose      Log all output from services to stdout (also
                          PEBBLE_VERBOSE=1)
           --args=        Provide additional arguments to a service
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -63,7 +63,7 @@ var sharedRunEnterArgsHelp = map[string]string{
 	"--create-dirs": "Create {{.DisplayName}} directory on startup if it doesn't exist",
 	"--hold":        "Do not start default services automatically",
 	"--http":        `Start HTTP API listening on this address in "<address>:port" format (for example, ":4000", "192.0.2.0:4000", "[2001:db8::1]:4000")`,
-	"--https":       `Start HTTPS API listening on this address (e.g., ":8443")`,
+	"--https":       `Start HTTPS API listening on this address in "<address>:port" format (for example, ":8443", "192.0.2.0:8443", "[2001:db8::1]:8443")`,
 	"--verbose":     "Log all output from services to stdout (also PEBBLE_VERBOSE=1)",
 	"--args":        "Provide additional arguments to a service",
 	"--identities":  "Seed identities from file (like update-identities --replace)",
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -199,7 +199,7 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 	dopts := daemon.Options{
 		Dir:          rcmd.pebbleDir,
 		SocketPath:   rcmd.socketPath,
-		IDSigner:   idSigner,
+		IDSigner:     idSigner,
 		HTTPAddress:  rcmd.HTTP,
 		HTTPSAddress: rcmd.HTTPS,
 	}
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -53,6 +53,7 @@ type sharedRunEnterOpts struct {
 	CreateDirs bool       `long:"create-dirs"`
 	Hold       bool       `long:"hold"`
 	HTTP       string     `long:"http"`
+	HTTPS      string     `long:"https"`
 	Verbose    bool       `short:"v" long:"verbose"`
 	Args       [][]string `long:"args" terminator:";"`
 	Identities string     `long:"identities"`
@@ -62,6 +63,7 @@ var sharedRunEnterArgsHelp = map[string]string{
 	"--create-dirs": "Create {{.DisplayName}} directory on startup if it doesn't exist",
 	"--hold":        "Do not start default services automatically",
 	"--http":        `Start HTTP API listening on this address in "<address>:port" format (for example, ":4000", "192.0.2.0:4000", "[2001:db8::1]:4000")`,
+	"--https":       `Start HTTPS API listening on this address (e.g., ":8443")`,
 	"--verbose":     "Log all output from services to stdout (also PEBBLE_VERBOSE=1)",
 	"--args":        "Provide additional arguments to a service",
 	"--identities":  "Seed identities from file (like update-identities --replace)",
@@ -195,14 +197,15 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 	}
 
 	dopts := daemon.Options{
-		Dir:        rcmd.pebbleDir,
-		SocketPath: rcmd.socketPath,
+		Dir:          rcmd.pebbleDir,
+		SocketPath:   rcmd.socketPath,
 		IDSigner:   idSigner,
+		HTTPAddress:  rcmd.HTTP,
+		HTTPSAddress: rcmd.HTTPS,
 	}
 	if os.Getenv("PEBBLE_VERBOSE") == "1" || rcmd.Verbose {
 		dopts.ServiceOutput = os.Stdout
 	}
-	dopts.HTTPAddress = rcmd.HTTP
 
 	d, err := daemon.New(&dopts)
 	if err != nil {
@@ -318,8 +321,8 @@ out:
 	}
 
 	// Close the client idle connection to the server (self connection) before we
-	// start with the HTTP shutdown process. This will speed up the server shutdown,
-	// and allow the Pebble process to exit faster.
+	// start with the HTTP/HTTPS shutdown process. This will speed up the server
+	// shutdown, and allow the Pebble process to exit faster.
 	rcmd.client.CloseIdleConnections()
 
 	return d.Stop(ch)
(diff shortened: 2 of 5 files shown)


docs/reference/cli-commands.md
```diff
@@ -866,8 +866,8 @@ pebble run --args myservice --port 8080 \; --hold
 [run command options]
           --create-dirs  Create Pebble directory on startup if it doesn't exist
           --hold         Do not start default services automatically
-          --http=        Start HTTP API listening on this address (e.g.,
-                         ":4000") and expose open-access endpoints
+          --http=        Start HTTP API listening on this address (e.g. ":4000")
+          --https=       Start HTTPS API listening on this address (e.g. ":8433")
```
anpep, line 870:
```suggestion
          --https=       Start HTTPS API listening on this address (e.g. ":8443")
```

anpep (approved):

docs/reference/cli-commands.md
flotter: Fixed. Resolving.

internals/daemon/daemon.go
```diff
@@ -368,6 +380,22 @@ func (d *Daemon) Init() error {
 		logger.Noticef("HTTP API server listening on %q.", d.httpAddress)
 	}
 
+	if d.httpsAddress != "" {
+		tlsConf := &tls.Config{
+			// We now support HTTP1.1 and HTTP2 when using HTTPS.
```
benhoyt, line 385: What does "now" mean here?

internals/daemon/daemon.go
```diff
@@ -368,6 +380,22 @@ func (d *Daemon) Init() error {
...
+			NextProtos:     []string{"h2", "http/1.1"},
+			MinVersion:     tls.VersionTLS13,
+			MaxVersion:     tls.VersionTLS13,
```
benhoyt, line 388: I'm definitely not a TLS expert, but why do we want a maximum here?

internals/cli/cmd_run.go
```diff
@@ -195,14 +197,15 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
...
 
 	dopts := daemon.Options{
-		Dir:        rcmd.pebbleDir,
-		SocketPath: rcmd.socketPath,
-		IDSigner:   idSigner,
+		Dir:          rcmd.pebbleDir,
+		SocketPath:   rcmd.socketPath,
+		IDSigner:     idSigner,
+		HTTPAddress:  rcmd.HTTP,
+		HTTPSAddress: rcmd.HTTPS,
```
benhoyt, line 204: Do we need to create the `idSigner` if `HTTPAddress` is not specified? Seems a bit strange to be creating the "identity" directory above if it's never going to be used.

internals/daemon/daemon_test.go
```diff
@@ -1305,6 +1313,50 @@ func (s *daemonSuite) TestHTTPAPI(c *C) {
...
+		},
+	})
+
+	request, err = http.NewRequest("GET", fmt.Sprintf("https://localhost:%d/v1/checks", port), nil)
+	c.Assert(err, IsNil)
+	response, err = httpsClient.Do(request)
+	c.Assert(err, IsNil)
+	c.Assert(response.StatusCode, Equals, http.StatusUnauthorized)
+
+	err = d.Stop(nil)
```
benhoyt, line 1354: I see that `TestHTTPAPI` does this the same way, but should this be done in a `defer`, as in `defer d.Stop(nil)` just after the `d.Start()` call?

internals/daemon/daemon.go
flotter: I made the explanation a little bit clearer.

What I meant to say is that now that we have a TLS client / server handshake as part of the HTTPS connection, HTTP protocol negotiation is now supported for the first time. This means that clients that also indicates HTTP2 support ("h2") in the TLS client hello message, will allow the HTTP server to switch to HTTP2.

```
                        // Enable server support for HTTP1.1 and HTTP2 over TLS. The server
                        // will only pick HTTP2 if the client indicated the same support in
                        // the client hello message. The Pebble client will not pick HTTP2
                        // because it requires integrated websockets to work, which does
                        // not support HTTP2 (it does not yet implement RFC8441). However,
                        // external clients, such as curl, will be able to switch to HTTP2
                        // for non-websocket dependant API calls.
```

Any thoughts ?

internals/daemon/daemon.go
flotter: I agree with your feeling. I've removed it :) Resolving.

internals/daemon/daemon_test.go
flotter: Note that there is one test case deliberately done still after the daemon is stopped, so I think that piece has to stay.

However, I've added a conditional defer to handle the case where the test fail early, e.g. one of the ```c.Assert``` cases fail, resulting in a ```runtime.Goexit``` which triggers the defer chain.

I think this should make it reliable, but do let me know if what I have done makes sense for you?

internals/daemon/daemon.go
```diff
@@ -368,6 +380,27 @@ func (d *Daemon) Init() error {
...
+			// will only pick HTTP2 if the client indicated the same support in
+			// the client hello message. The Pebble client will not pick HTTP2
+			// because it requires integrated websockets to work, which does
+			// not support HTTP2 (it does not yet implement RFC8441). However,
+			// external clients, such as curl, will be able to switch to HTTP2
+			// for non-websocket dependant API calls.
+			NextProtos:     []string{"h2", "http/1.1"},
+			MinVersion:     tls.VersionTLS13,
+			GetCertificate: d.overlord.TLSManager().GetCertificate,
+		}
```
hpidcock, line 395: Can the TLSManager have a `ListenConfig() tls.Config` method instead? Or something similar?

hpidcock (approved):
Seems OK. I have no strong opinions.

docs/reference/cli-commands.md
```diff
@@ -868,6 +868,7 @@ pebble run --args myservice --port 8080 \; --hold
           --hold         Do not start default services automatically
           --http=        Start HTTP API listening on this address (e.g.,
                          ":4000") and expose open-access endpoints
+          --https=       Start HTTPS API listening on this address (e.g. ":8443")
```
benhoyt, line 871: See also changes in https://github.com/canonical/pebble/pull/606/files#diff-26d4ccdee2a828c580153cbc020472b62b589e9bec6c89179c6e9f48c4a4e783R62 to give IPv4 and IPv6 examples here -- probably want to follow that.

benhoyt (approved):
I'm happy with this now - one minor comment about potential merge conflict.

Excellent work on all these, Fred.

internals/daemon/daemon_test.go
flotter: Resolving.

internals/daemon/daemon.go
flotter: Added this in the TLS Manager PR. Great idea! Resolving.

docs/reference/cli-commands.md
flotter: Fixed, thank you. Resolving.

ghislainbourgeois (approved):

internals/cli/cmd_run.go
flotter: Discussed with @benhoyt again. We are not used to extra stuff in $PEBBLE, but there is not really a consistent solution around this. Ben is OK for me to proceed.

PR #600, opened 2025-04-08
fix: tighten permissions of --create-dirs directory from 0o755 to 0o700

benhoyt:
Also add an example to the docs of how to tighten up this directly if it already exists.


IronCore864 (approved):

hpidcock (approved):
LGTM. I did some surface level testing on juju 3.6. There may be charms doing stuff here, but we can probably consider that outside of the guaranteed environment that juju/pebble provides.

PR #605, opened 2025-04-18
feat(idkey): add identity key [1/4]

flotter:
Add support for Pebble to have a machine, container or device identity that is based on an Ed25519 private key.

This identity key implements the crypto.Signer interface, allowing it to sign, for example, TLS certificates.

The identity key should eventually be stored in a hardware backed key-store, such as a TPM, OP-TEE or even YubiKey. The first implementation of the package implements disk based storage with restricted permissions, similar to what you would find when looking at your ~/.ssh directory.

The purpose of this package is to allow a machine, container or device to have a lifelong identity, including a standardized fingerprint that represents that identity. 

**Example use-case**

The identity private key can be used to sign short-lived TLS keypairs for the HTTPS Pebble API server, whereby it acts as the signing authority (CA). In this case, an x509 identity certificate with CA signing permissions will be included in the certificate chain, allowing the client to pin the identity certificate (which does not change over time) during a physical trust exchange process (think button press and user assisted identity confirmation).

**Benchmark information**

The identity will only be generated on first boot, and will incur the following startup penalty:

 **3.0 ms** - AMD FX(tm)-8350 Eight-Core Processor
**5.5ms** - Raspberry Pi 5 (Sdcard)
**14.5ms** - Raspberry Pi 3b+ (Sdcard)

### DIFF
--- a/internals/cli/cmd_enter_test.go
+++ b/internals/cli/cmd_enter_test.go
@@ -167,7 +167,7 @@ func (s *PebbleSuite) TestEnterExecListDir(c *C) {
 
 	exitCode := cli.PebbleMain()
 	c.Check(s.Stderr(), Equals, "")
-	c.Check(s.Stdout(), Equals, "bar\nbaz\nfoo\n")
+	c.Check(s.Stdout(), Equals, "bar\nbaz\nfoo\nidentity\n")
 	c.Check(exitCode, Equals, 0)
 }
 
--- a/internals/cli/cmd_enter_test.go
+++ b/internals/cli/cmd_enter_test.go
@@ -162,12 +162,12 @@ func (s *PebbleSuite) TestEnterExecListDir(c *C) {
 		}
 	}
 
-	restore := fakeArgs("pebble", "enter", "exec", "ls", s.pebbleDir)
+	restore := fakeArgs("pebble", "enter", "exec", "ls", "--hide=identity", s.pebbleDir)
 	defer restore()
 
 	exitCode := cli.PebbleMain()
 	c.Check(s.Stderr(), Equals, "")
-	c.Check(s.Stdout(), Equals, "bar\nbaz\nfoo\nidentity\n")
+	c.Check(s.Stdout(), Equals, "bar\nbaz\nfoo\n")
 	c.Check(exitCode, Equals, 0)
 }
 
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -67,8 +67,8 @@ type Options struct {
 	LayersDir string
 
 	// IDSigner is a private key representing the identity of a Pebble
-	// instance (machine, countainer or device), which implements the
-	// crypto.Signer interface (allowing it to sign TLS keypairs).
+	// instance (machine, container or device), which implements the
+	// crypto.Signer interface (for digest signing).
 	IDSigner crypto.Signer
 
 	// SocketPath is an optional path for the unix socket used for the client
--- a/internals/daemon/daemon.go
+++ b/internals/daemon/daemon.go
@@ -17,6 +17,7 @@ package daemon
 import (
 	"bufio"
 	"context"
+	"crypto"
 	"errors"
 	"fmt"
 	"io"
@@ -64,6 +65,11 @@ type Options struct {
 	// LayersDir is an optional path for the layers directory.
 	// Defaults to "layers" inside the pebble directory.
 	LayersDir string
+
+	// IDSigner is a private key representing the identity of a Pebble
+	// instance (machine, countainer or device), which implements the
+	// crypto.Signer interface (allowing it to sign TLS keypairs).
+	IDSigner crypto.Signer
 
 	// SocketPath is an optional path for the unix socket used for the client
 	// to communicate with the daemon. Defaults to a hidden (dotted) name inside
@@ -844,6 +850,7 @@ func New(opts *Options) (*Daemon, error) {
 		RestartHandler: d,
 		ServiceOutput:  opts.ServiceOutput,
 		Extension:      opts.OverlordExtension,
+		IDSigner:       opts.IDSigner,
 	}
 
 	ovld, err := overlord.New(&ovldOptions)
--- a/internals/idkey/idkey.go
+++ b/internals/idkey/idkey.go
@@ -43,53 +43,53 @@ var _ crypto.Signer = (*IDKey)(nil)
 
 type IDKey struct {
 	keyDir string
-	key    any
+	key    ed25519.PrivateKey
 }
 
-// New checks if an existing private identity key exists, and loads the key
+// Get checks if an existing private identity key exists, and loads the key
 // if it does. It creates a new private identity key and persists it to disk
 // if no key was found. Cases where explicit control is desired on when to
-// generate or load can use GenerateKey and LoadKey directly.
-func New(keyDir string) (*IDKey, error) {
+// generate or load can use Generate and Load directly.
+func Get(keyDir string) (*IDKey, error) {
 	keyPath := filepath.Join(keyDir, identityKeyFile)
 	exists, err := pathExists(keyPath)
 	if err != nil {
 		return nil, fmt.Errorf("cannot access key file %q: %w", keyPath, err)
 	}
 	if exists {
-		return LoadKey(keyDir)
+		return Load(keyDir)
 	}
-	return GenerateKey(keyDir)
+	return Generate(keyDir)
 }
 
-// GenerateKey generates a new identity key and persists it to disk. This
+// Generate generates a new identity key and persists it to disk. This
 // function should only ever be called on the first boot otherwise the
 // existing identity will be overwritten.
 //
 // This function is equivalent to running:
 //
 //	openssl genpkey -algorithm Ed25519 -out key.pem
-func GenerateKey(keyDir string) (*IDKey, error) {
-	key := &IDKey{
+func Generate(keyDir string) (*IDKey, error) {
+	k := &IDKey{
 		keyDir: keyDir,
 	}
-	err := key.createDir()
+	err := k.createDir()
 	if err != nil {
 		return nil, fmt.Errorf("cannot create identity directory: %w", err)
 	}
-	err = key.newEd25519()
+	_, k.key, err = ed25519.GenerateKey(rand.Reader)
 	if err != nil {
 		return nil, fmt.Errorf("cannot generate identity key: %w", err)
 	}
-	err = key.save()
+	err = k.save()
 	if err != nil {
 		return nil, fmt.Errorf("cannot save identity key: %w", err)
 	}
-	return key, nil
+	return k, nil
 }
 
-// LoadKey loads an existing identity key from disk.
-func LoadKey(keyDir string) (*IDKey, error) {
+// Load loads an existing identity key from disk.
+func Load(keyDir string) (*IDKey, error) {
 	key := &IDKey{
 		keyDir: keyDir,
 	}
@@ -116,33 +116,24 @@ func (k *IDKey) createDir() error {
 	return os.Mkdir(k.keyDir, 0o700)
 }
 
-// Public implements the crypto.Signer.
+// Public implements part of the crypto.Signer interface.
 func (k *IDKey) Public() crypto.PublicKey {
-	signer := k.key.(crypto.Signer)
-	return signer.Public()
+	return k.key.Public()
 }
 
-// Sign implements the crypto.Signer.
+// Sign implements part of the crypto.Signer interface.
 func (k *IDKey) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
-	signer := k.key.(crypto.Signer)
-	return signer.Sign(rand, digest, opts)
-}
-
-// newEd25519 generates a new ed25519 key.
-func (k *IDKey) newEd25519() error {
-	var err error
-	_, k.key, err = ed25519.GenerateKey(rand.Reader)
-	if err != nil {
-		return err
-	}
-	return nil
+	return k.key.Sign(rand, digest, opts)
 }
 
 // load loads the private identity key from storage.
 func (k *IDKey) load() error {
 	exists, err := pathExists(k.keyDir)
-	if err != nil || !exists {
-		return fmt.Errorf("directory %q is not accessible", k.keyDir)
+	if err != nil {
+		return fmt.Errorf("directory %q is not accessible: %w", k.keyDir, err)
+	}
+	if !exists {
+		return fmt.Errorf("directory %q not found", k.keyDir)
 	}
 	err = expectPermission(k.keyDir, 0o700)
 	if err != nil {
@@ -158,33 +149,27 @@ func (k *IDKey) load() error {
 	if err != nil {
 		return err
 	}
-	for {
-		var block *pem.Block
-		block, pemData = pem.Decode(pemData)
-		if block == nil {
-			break
-		}
-		switch block.Type {
-		case "PRIVATE KEY":
-			k.key, err = x509.ParsePKCS8PrivateKey(block.Bytes)
-			if err != nil {
-				return err
-			}
-		default:
-			return fmt.Errorf("unknown PEM block %q in file %q", block.Type, pemPath)
-		}
+	block, _ := pem.Decode(pemData)
+	if block == nil || block.Type != "PRIVATE KEY" {
+		return fmt.Errorf("missing 'PRIVATE KEY' block in %q", pemPath)
 	}
-	if k.key == nil {
-		return fmt.Errorf("empty PEM file %q", pemPath)
+	key, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		return err
+	}
+	var ok bool
+	k.key, ok = key.(ed25519.PrivateKey)
+	if !ok {
+		return fmt.Errorf("ed25519 type private key expected")
 	}
 	return nil
 }
 
 // save saves the private identity key to storage.
-func (k *IDKey) save() error {
+func (k *IDKey) save() (err error) {
 	exists, err := pathExists(k.keyDir)
 	if err != nil {
-		return fmt.Errorf("directory %q is not accessible", k.keyDir)
+		return fmt.Errorf("directory %q is not accessible: %w", k.keyDir, err)
 	}
 	if exists {
 		err = expectPermission(k.keyDir, 0o700)
@@ -206,7 +191,7 @@ func (k *IDKey) save() error {
 		return err
 	}
 	defer func() {
-		err = errors.Join(err, pemFile.Sync())
+		// err here refers to the named error return.
 		err = errors.Join(err, pemFile.Close())
 	}()
 	keyBytes, err := x509.MarshalPKCS8PrivateKey(k.key)
@@ -218,10 +203,13 @@ func (k *IDKey) save() error {
 		Bytes: keyBytes,
 	}
 	var pemBuffer bytes.Buffer
-	if err := pem.Encode(&pemBuffer, pemPrivateBlock); err != nil {
+	if err = pem.Encode(&pemBuffer, pemPrivateBlock); err != nil {
 		return err
 	}
 	if _, err = pemFile.Write(pemBuffer.Bytes()); err != nil {
+		return err
+	}
+	if err = pemFile.Sync(); err != nil {
 		return err
 	}
 	return nil
--- a/internals/idkey/idkey.go
+++ b/internals/idkey/idkey.go
@@ -12,11 +12,11 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-// idkey supplies an identity key for a machine or device. This package
-// provides an implementation based on an Ed25519 based key, currently
+// idkey supplies an identity key for a machine, container or device. This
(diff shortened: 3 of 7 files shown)


internals/cli/cmd_enter_test.go
```diff
@@ -167,7 +167,7 @@ func (s *PebbleSuite) TestEnterExecListDir(c *C) {
 
 	exitCode := cli.PebbleMain()
 	c.Check(s.Stderr(), Equals, "")
-	c.Check(s.Stdout(), Equals, "bar\nbaz\nfoo\n")
+	c.Check(s.Stdout(), Equals, "bar\nbaz\nfoo\nidentity\n")
```
anpep, line 170: note for other reviewers: this test creates files `bar`, `baz`, `foo` on `$PEBBLE_DIR`; however, `identity` directory is now created by the `idkey` package before creating a daemon instance.

internals/daemon/daemon.go
```diff
@@ -65,6 +66,11 @@ type Options struct {
 	// Defaults to "layers" inside the pebble directory.
 	LayersDir string
 
+	// IDSigner is a private key representing the identity of a Pebble
+	// instance (machine, countainer or device), which implements the
```
anpep, line 70: nit:
```suggestion
	// instance (machine, container or device), which implements the
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+// Sign implements the crypto.Signer.
+func (k *IDKey) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
+	signer := k.key.(crypto.Signer)
+	return signer.Sign(rand, digest, opts)
+}
+
+// newEd25519 generates a new ed25519 key.
+func (k *IDKey) newEd25519() error {
+	var err error
+	_, k.key, err = ed25519.GenerateKey(rand.Reader)
```
anpep, line 134: nit: [according to docs](https://pkg.go.dev/crypto/ed25519#GenerateKey), `crypto/rand.Reader` is the default if `nil` is passed

```suggestion
	_, k.key, err = ed25519.GenerateKey(nil)
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+		return err
+	}
+	return nil
+}
+
+// load loads the private identity key from storage.
+func (k *IDKey) load() error {
+	exists, err := pathExists(k.keyDir)
+	if err != nil || !exists {
+		return fmt.Errorf("directory %q is not accessible", k.keyDir)
```
anpep, line 145: nit:
```suggestion
		return fmt.Errorf("directory %q is not accessible: %w", k.keyDir, err)
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+			break
+		}
+		switch block.Type {
+		case "PRIVATE KEY":
+			k.key, err = x509.ParsePKCS8PrivateKey(block.Bytes)
+			if err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown PEM block %q in file %q", block.Type, pemPath)
```
anpep, line 174: I say let's ignore an unrecognized block type here, since we only care about `PRIVATE KEY`

```suggestion
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+		return fmt.Errorf("empty PEM file %q", pemPath)
+	}
+	return nil
+}
+
+// save saves the private identity key to storage.
+func (k *IDKey) save() error {
+	exists, err := pathExists(k.keyDir)
+	if err != nil {
+		return fmt.Errorf("directory %q is not accessible", k.keyDir)
```
anpep, line 187: nit: as above

```suggestion
		return fmt.Errorf("directory %q is not accessible: %w", k.keyDir, err)
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+	actualPerm := info.Mode().Perm()
+	if actualPerm != perm {
+		return fmt.Errorf("expected permission 0o%o (got 0o%o) for %q", perm, actualPerm, path)
+	}
+	return nil
+}
+
+// pathExists returns true of the path exists, false if it does not. If the
+// operation reports an unrelated error, the error is returned.
+func pathExists(path string) (bool, error) {
```
anpep, line 256: nit: we have `osutil.ExistsIsDir`, if we want to get rid of this

anpep (approved):

internals/idkey/idkey.go
ghislainbourgeois: Because this is crypto code, I would much prefer keeping it explicit.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+		switch block.Type {
+		case "PRIVATE KEY":
+			k.key, err = x509.ParsePKCS8PrivateKey(block.Bytes)
+			if err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown PEM block %q in file %q", block.Type, pemPath)
+		}
+	}
```
ghislainbourgeois, line 176: This whole loop is unnecessary if we expect to have a single private key block in the file. If we want to support multiple blocks, we should definitely ignore the blocks that are not private keys instead of returning an error early.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+
+// Fingerprint returns the identity fingerprint. This is the SHA512/384 hash
+// of the public key, encoded in base32 (without padding). This is a
+// convenient shorthand form of the identity that can be used to identify
+// a specfic machine or device.
+func (k *IDKey) Fingerprint() string {
+	publicBytes := k.Public().(ed25519.PublicKey)
+	hashBytes := sha512.Sum384(publicBytes)
+	return base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(hashBytes[:])
+}
```
ghislainbourgeois, line 238: Currently, most fingerprints are using `sha256` presented as an hexadecimal string. As an example, this is what Chrome will show when inspecting a certificate.

I would recommend at least using an hexadecimal string instead of base32 as it is more standard.

ghislainbourgeois (commented):
This looks very good, I left some small suggestions, but I do not see anything really problematic.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	if exists {
+		return expectPermission(k.keyDir, 0o700)
+	}
+
+	// Create the leaf directory node with 0o700 permissions.
+	return os.Mkdir(k.keyDir, 0o700)
+}
+
+// Public implements the crypto.Signer.
```
benhoyt, line 119: Nit:
```suggestion
// Public implements part of the crypto.Signer interface.
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+func (k *IDKey) Public() crypto.PublicKey {
+	signer := k.key.(crypto.Signer)
+	return signer.Public()
+}
+
+// Sign implements the crypto.Signer.
```
benhoyt, line 125:
```suggestion
// Sign implements part of the crypto.Signer interface.
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+)
+
+const identityKeyFile = "key.pem"
+
+// Identity key must implement a crypto signer.
+var _ crypto.Signer = (*IDKey)(nil)
+
+type IDKey struct {
+	keyDir string
+	key    any
```
benhoyt, line 46: Why is this `any` and not `crypto.Signer`, which is what several methods need anyway. See also comment in `load`.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+//	openssl genpkey -algorithm Ed25519 -out key.pem
+func GenerateKey(keyDir string) (*IDKey, error) {
+	key := &IDKey{
+		keyDir: keyDir,
+	}
+	err := key.createDir()
+	if err != nil {
+		return nil, fmt.Errorf("cannot create identity directory: %w", err)
+	}
+	err = key.newEd25519()
```
benhoyt, line 80: This method has the side effect of setting `key.key`, which given the name, seems odd ("new" functions usually create a new object and return it). Plus, the method is only called once and basically wraps a single function call. What about inlining it here:

```go
_, key.key, err = ed25519.GenerateKey(rand.Reader)
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+		return expectPermission(k.keyDir, 0o700)
+	}
+
+	// Create the leaf directory node with 0o700 permissions.
+	return os.Mkdir(k.keyDir, 0o700)
+}
+
+// Public implements the crypto.Signer.
+func (k *IDKey) Public() crypto.PublicKey {
+	signer := k.key.(crypto.Signer)
```
benhoyt, line 121: If `k.key` was typed as a `crypto.Signer`, we wouldn't need the type conversions at each access.

internals/idkey/idkey.go
benhoyt: I agree with @anpep suggestion here, except we probably want two different cases, because `fmt.Errorf` with `%w` yields the string `%!w(<nil>)` for nil errors. So we probably need to handle the err!=nil and the !exists cases separately.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+		block, pemData = pem.Decode(pemData)
+		if block == nil {
+			break
+		}
+		switch block.Type {
+		case "PRIVATE KEY":
+			k.key, err = x509.ParsePKCS8PrivateKey(block.Bytes)
+			if err != nil {
+				return err
+			}
```
benhoyt, line 172: Other code in this file expects `k.key` to be a `crypto.Signer`, so I think we should type-assert here (and handle the error). Something like:

```go
key, err := x509.ParsePKCS8PrivateKey(block.Bytes)
if err != nil { ... }
if k.key, ok = key.(crypto.Signer); !ok {
    return fmt.Errorf("unknown type %T when parsing private key in file %q", key, pemPath)
}
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+	keyBytes, err := x509.MarshalPKCS8PrivateKey(k.key)
+	if err != nil {
+		return err
+	}
+	pemPrivateBlock := &pem.Block{
+		Type:  "PRIVATE KEY",
+		Bytes: keyBytes,
+	}
+	var pemBuffer bytes.Buffer
+	if err := pem.Encode(&pemBuffer, pemPrivateBlock); err != nil {
```
benhoyt, line 221: I guess this is intentionally creating a new `err` variable? In which case, it's kinda subtle -- please comment why.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+
+// expectPermission return an error if the specified directory or file
+// path is not matching the supplied permissions.
+func expectPermission(path string, perm fs.FileMode) error {
+	info, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+	actualPerm := info.Mode().Perm()
+	if actualPerm != perm {
```
benhoyt, line 248: Do the permissions need to be an exact match, or is it okay if `actualPerm` is a subset of `perm`, for example `r--` instead of `rw-`? But maybe it's best just to be very specific.

internals/idkey/idkey.go
benhoyt: I reckon we should leave it. That `ExistsIsDir` function has a confusing name and odd return values -- it tries to do too much IMO.

internals/idkey/idkey_test.go
```diff
@@ -0,0 +1,154 @@
...
+	// Create a new identity key (first boot)
+	_, err := idkey.GenerateKey(keyDir)
+	c.Assert(err, IsNil)
+
+	err = os.Chmod(filepath.Join(keyDir, "key.pem"), 0o644)
+	c.Assert(err, IsNil)
+
+	// Load the identity key (other boots)
+	_, err = idkey.LoadKey(keyDir)
+	c.Assert(err, ErrorMatches, "cannot load identity key.*")
```
benhoyt, line 101: I think we should have a more specific check here:
```suggestion
	c.Assert(err, ErrorMatches, ".*expected permissions.*")
```

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	err = key.save()
+	if err != nil {
+		return nil, fmt.Errorf("cannot save identity key: %w", err)
+	}
+	return key, nil
+}
+
+// LoadKey loads an existing identity key from disk.
+func LoadKey(keyDir string) (*IDKey, error) {
```
benhoyt, line 92: Naming: similar to above, would `idkey.Load` be enough?

benhoyt (changes requested):
Looks good overall. A few minor changes requested.

internals/daemon/daemon.go
flotter: Fixed, Thanks. Resolving.

internals/idkey/idkey.go
flotter: Let's keep it explicit. Resolving.

internals/idkey/idkey.go
flotter: I now handle both cases separately. Resolving.

internals/idkey/idkey.go
flotter: Simplified it to expect only a single block. Resolving.

internals/idkey/idkey.go
flotter: Resolving.

internals/idkey/idkey.go
flotter: I removed the whole loop and simplified it to expect only the single "PRIVATE KEY" block. Resolving.

internals/idkey/idkey.go
flotter: Discussed this with @ghislainbourgeois.

I think this is a very interesting point.

Throughout the design process with the CTO, for the fingerprint, the focus was on CLI and the look, feel and consistency of the char set for the CLI user.

At attempt at using ed52219 keys on Chrome and Firefox actually highlights this is not yet supported (although I do know X25519 key exchange is supported).

We have a little time to refine this in the next cycle, so I propose we move forward, but I will make a not for discussion.

I see mTLS is also technically supported in browsers, so perhaps we should consider the use-case where a browser / phone / device actually connects to the API.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+type IDKey struct {
+	keyDir string
+	key    any
+}
+
+// New checks if an existing private identity key exists, and loads the key
+// if it does. It creates a new private identity key and persists it to disk
+// if no key was found. Cases where explicit control is desired on when to
+// generate or load can use GenerateKey and LoadKey directly.
+func New(keyDir string) (*IDKey, error) {
```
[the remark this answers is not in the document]
flotter: How about ```Get``` ? I know that often times the GetXXX gets shortened to XXX, but since the package is already ```idkey```, having ```idkey.IDKey()``` I guess is not ideal.

I am just thinking from the perspective of the caller, they do not really care how its obtained when using this call.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// GenerateKey generates a new identity key and persists it to disk. This
+// function should only ever be called on the first boot otherwise the
+// existing identity will be overwritten.
+//
+// This function is equivalent to running:
+//
+//	openssl genpkey -algorithm Ed25519 -out key.pem
+func GenerateKey(keyDir string) (*IDKey, error) {
```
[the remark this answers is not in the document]
flotter: Fixed, Thanks. Resolving.

internals/idkey/idkey.go
flotter: Fixed. Resolving.

internals/idkey/idkey.go
flotter: Fixed. Resolving.

internals/idkey/idkey.go
flotter: Yes. In fact, I think just making this ```ed25519.PrivateKey``` is even more precise ?

internals/idkey/idkey.go
flotter: Fixed. Thanks. Resolving.

internals/idkey/idkey.go
flotter: Yes, after changing to ```ed25519.PrivateKey```, all these extra conversions fall away. Thanks. Resolving.

internals/idkey/idkey.go
flotter: My bad. In fact the defer was broken, it was trying to update the _named_ error return, that was not named. I added a named error return, and added a comment inside defer.

I simplified it a little, only doing the close in the defer, which makes more sense. Let me know what you think.

internals/idkey/idkey.go
flotter: It is an interesting point.

Since the common case for Pebble is to run as root, it would always be able to access the file, and so any other root service (irrespective of the file permissions), right ? I think what I was trying to do here is to make sure this file, which is under our control, will never allow access to anyone else but us, except I do not really check who owns it! I will think on this a bit ...

internals/idkey/idkey_test.go
flotter: Fixed, thanks. Resolving.

internals/idkey/idkey.go
flotter: Yes, nice. Fixed. Resolving.

ghislainbourgeois (approved):
LGTM

internals/idkey/idkey.go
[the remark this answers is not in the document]
benhoyt: Yeah, `Get` works for me, thx

benhoyt (approved):
I'm happy with this now. Please ensure all comments are resolved and then let's merge!

internals/idkey/idkey.go
flotter: Ben indicated he is happy. Resolving.

internals/idkey/idkey.go
flotter: Resolving.

internals/idkey/idkey.go
flotter: @benhoyt 

I started implementing ownership checking here, but after thinking this through, I think this is not a good idea.

I do not want the addition of a Pebble identity (and TLS) restrict the usage of Pebble more in any way, especially since initially this is forced on everyone, whether they use it or not.

If I add ownership checking, and I am really strict about it, if someone runs Pebble first as non-root and then re-runs it as root, the ownership of the identity would not longer match, and Pebble would fail to start. If we argue that Pebble running as root should support any identity user, then there really is no need for ownership checking in the first place, since any other scenario (owned by normal user X, now running as Y) would simply now be possible due to normal Linux permissions being enforced at the OS file access level (it would fail to load).

In summary, I think the current approach of ensuring that only the file owner or root can access it, and only as read or write, is sufficient.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,253 @@
...
+type IDKey struct {
+	keyDir string
+	key    ed25519.PrivateKey
+}
+
+// Get checks if an existing private identity key exists, and loads the key
+// if it does. It creates a new private identity key and persists it to disk
+// if no key was found. Cases where explicit control is desired on when to
+// generate or load can use Generate and Load directly.
+func Get(keyDir string) (*IDKey, error) {
```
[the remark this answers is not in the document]
flotter: Thanks for the review @Ardelean-Calin. I think your proposal is valid, and also what Ben initially proposed.

At the end I settled with ```Get``` from a consumer perspective in mind - the caller using get is really saying that it does not care what happens, just give me the identity (like in the Pebble user case). If the caller does care, then a function saying maybe this or maybe that is not really that helpful, it is basically saying, I cannot tell you how, but I can give it to you. If you really care about the how, then use the ```Load``` or ```Generate``` calls instead. At the end it is probably just my personal preference.

Ardelean-Calin (approved):
Very nice work!

internals/idkey/idkey.go
benhoyt: Sounds reasonable to me, thank you for thinking this through.

internals/idkey/idkey.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	pemPrivateBlock := &pem.Block{
+		Type:  "PRIVATE KEY",
+		Bytes: keyBytes,
+	}
+	var pemBuffer bytes.Buffer
+	if err = pem.Encode(&pemBuffer, pemPrivateBlock); err != nil {
+		return err
+	}
+	if _, err = pemFile.Write(pemBuffer.Bytes()); err != nil {
```
hpidcock, line 212: We should be using io.Copy where we can and checking the written count.

hpidcock (approved):

internals/idkey/idkey.go
flotter: Resolving.

internals/idkey/idkey.go
flotter: Resolving.

PR #606, opened 2025-04-21
feat: add http-address to /v1/system-info

benhoyt:
Per discussion on Matrix > Juju, we'd like to add the ability to get the HTTP port that Pebble is listening on (if any) to the /v1/system-info API response. This PR simply adds that field.

For example, if the Pebble daemon is started with `pebble run --http=:4040`, the API returns this:

```
$ curl --unix-socket /var/lib/pebble/default/.pebble.socket http://_/v1/system-info|jq
{
  "type": "sync",
  "status-code": 200,
  "status": "OK",
  "result": {
    "boot-id": "30435cc9-b428-4e25-a821-c9477fc339eb",
    "http-address": ":4040",
    "version": "v1.21.0-dev"
  }
}
```

### DIFF
--- a/docs/specs/openapi.yaml
+++ b/docs/specs/openapi.yaml
@@ -1226,6 +1226,7 @@ paths:
                   "status": "OK",
                   "result": {
                     "boot-id": "e14ed96e-5a98-4402-80f7-d19dd949eac3",
+                    "http-address": ":4000",
                     "version": "v1.17.0"
                   }
                 }
@@ -1675,12 +1676,18 @@ components:
     systemInfo:
       type: object
       properties:
-        version:
-          type: string
-          description: Version of the Pebble daemon.
         boot-id:
           type: string
           description: A unique string that represents this boot of the server.
+        http-address:
+          type: string
+          description: Address the HTTP server is listening on, for example `:4000`. Only present if the daemon was started with the `--http` argument.
+        version:
+          type: string
+          description: Version of the Pebble daemon.
+      required:
+        - boot-id
+        - version
     checkInfo:
       type: object
       properties:
--- a/docs/specs/openapi.yaml
+++ b/docs/specs/openapi.yaml
@@ -1209,7 +1209,7 @@ paths:
   /v1/system-info:
     get:
       summary: Get system information
-      description: Get the version of the Pebble daemon and the boot ID of the system.
+      description: Get information about the Pebble daemon and the boot ID of the system.
       tags:
         - system info
       responses:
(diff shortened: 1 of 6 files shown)


IronCore864 (approved):

docs/specs/openapi.yaml
```diff
@@ -1675,12 +1676,18 @@ components:
...
       properties:
-        version:
-          type: string
-          description: Version of the Pebble daemon.
         boot-id:
           type: string
           description: A unique string that represents this boot of the server.
+        http-address:
+          type: string
+          description: Address the HTTP server is listening on, for example `:4000`. Only present if the daemon was started with the `--http` argument.
```
dimaqq, line 1684: ### Perhaps off-topic

The documentation for the `--http` flag is hard to find.
It's here: https://documentation.ubuntu.com/pebble/reference/cli-commands/#run

What's missing is the list of formats for the value:
- a single port like `:4000`
- a list of ports?
- this network and port, like `0.0.0.0:4000`?
- named ipv6 address, like `[ffee::1234]:4000`?
- subnets, CIDR?

Now that the value is exposed in the API, maybe it's a good idea to define what acceptable values for the flag are, and therefore what kinds of values the API caller may get.

benhoyt: Fair -- I've spelled this out a bit more in the `--help` docs now.

internals/cli/cmd_run.go
```diff
@@ -59,7 +59,7 @@ type sharedRunEnterOpts struct {
 var sharedRunEnterArgsHelp = map[string]string{
 	"--create-dirs": "Create {{.DisplayName}} directory on startup if it doesn't exist",
 	"--hold":        "Do not start default services automatically",
-	"--http":        `Start HTTP API listening on this address (e.g., ":4000") and expose open-access endpoints`,
+	"--http":        `Start HTTP API listening on this address in "[ipv4_or_v6_address]:port" format (for example, ":4000")`,
```
hpidcock, line 62: Square brackets are a bit confusing, since they are used for v6 addresses.
```suggestion
	"--http":        `Start HTTP API listening on this address in "<address>:port" format (for example, ":4000", "192.0.2.0:4000", "[2001:db8::1]:4000")`,
```

Same for `docs/reference/cli-commands.md`.

hpidcock (approved):
LGTM thank-you. Just one suggestion for the listen help text.

internals/cli/cmd_run.go
benhoyt: Thanks, good call - done

PR #607, opened 2025-04-26
docs: document how to check health based on logs

### LINKED ISSUE #601: Doc or Discourse post on how to assert on pebble service logs from pebble checks
Recent example from @sed-i: a service is up and running, but logs HTTP 500 whenever Prometheus tries to scrape it. Would be useful to have a pebble check surface this for when the user is not relying on full fledged observability.

I suggested you could do that today with an `exec` check that calls `pebble logs` plus a bit of grep.

It would be nice to document this pattern -- maybe in a how-to? Though that's pretty specific for a how-to. @dwilding where would this kind of pattern/recipe content go? Maybe just a Discourse post?

dwilding:
This PR adds a how-to guide called "How to check health based on logs". Closes #601

**[Preview doc](https://canonical-ubuntu-documentation-library--607.com.readthedocs.build/pebble/how-to/check-health-based-on-logs/)**

This PR also adds a couple of "see also" links after the examples on the health checks reference page ([see preview](https://canonical-ubuntu-documentation-library--607.com.readthedocs.build/pebble/reference/health-checks/#examples)).

### DIFF
--- a/docs/how-to/check-health-based-on-logs.md
+++ b/docs/how-to/check-health-based-on-logs.md
@@ -1,1 +1,28 @@
 # How to check health based on logs
+
+Need to be clear that this is not the primary way to check health. It's kind of a hack...
+
+## Define a layer
+
+```yaml
+services:
+  foo:
+    override: replace
+    command: foo
+    startup: enabled
+checks:
+  foo-warning:
+    override: replace
+    threshold: 1
+    exec:
+      command: bash -c '! pebble logs | grep -q WARNING'
+```
+
+- `pebble check foo-warning`
+- `pebble health foo-warning`
+
+## See more
+
+- [`pebble check`](#reference_pebble_check_command) command
+- [`pebble health`](#reference_pebble_health_command) command
+- [](/reference/health-checks)
--- a/docs/how-to/check-health-based-on-logs.md
+++ b/docs/how-to/check-health-based-on-logs.md
@@ -1,0 +1,1 @@
+# How to check health based on logs
--- a/docs/how-to/check-health-based-on-logs.md
+++ b/docs/how-to/check-health-based-on-logs.md
@@ -1,8 +1,8 @@
 # How to check health based on logs
 
-Pebble stores the most recent `stdout` and `stderr` from each service. This guide demonstrates how to set up a health check that fails if the logs contain a particular line, such as a warning message.
+Pebble stores the most recent `stdout` and `stderr` from each service. This guide demonstrates how to set up a check that fails if the logs contain a particular line, such as a warning message.
 
-This is an indirect and potentially unreliable way to catch issues with services. Whenever possible, you should set up health checks that query the service status directly. For more information, see [](./run-services-reliably).
+This method is an alternative to directly querying a service. Directly querying a service is typically a more reliable method of catching issues. For more information, see [](./run-services-reliably).
 
 ## Define a layer
 
@@ -23,7 +23,7 @@ checks:
       # Because of YAML escaping rules, we need to use \\[ to pass \[ to grep.
 ```
 
-The check searches Pebble's service logs for lines such as:
+The check searches Pebble's service logs for lines matching the pattern, such as:
 
 ```text
 2025-04-26T03:22:20.315Z [foo] some WARNING reported by the service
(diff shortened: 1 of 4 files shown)


docs/how-to/check-health-based-on-logs.md
```diff
@@ -0,0 +1,69 @@
+# How to check health based on logs
+
+Pebble stores the most recent `stdout` and `stderr` from each service. This guide demonstrates how to set up a health check that fails if the logs contain a particular line, such as a warning message.
+
+This is an indirect and potentially unreliable way to catch issues with services. Whenever possible, you should set up health checks that query the service status directly. For more information, see [](./run-services-reliably).
```
IronCore864, line 5: Could be more positive:

```suggestion
This method provides an alternative way to detect issues, though direct health checks are preferred. Whenever possible, you should set up health checks that query the service status directly. For more information, see [](./run-services-reliably).
```

docs/how-to/check-health-based-on-logs.md
```diff
@@ -0,0 +1,69 @@
...
+
+```text
+2025-04-26T03:22:20.315Z [foo] some WARNING reported by the service
+```
+
+If a match is found, the `bash -c '...'` command exits with 1 and the check fails.
+
+When the check fails, Pebble doesn't restart `foo`. If `foo` were to restart, the check would still be considered "down" (because the logs would still contain the warning message) and Pebble wouldn't restart `foo` again if another warning occurred.
+
+It's possible to [configure a service to restart when a check fails](#restart-a-service-when-the-health-check-fails). However, we don't recommend that you configure `foo` to restart. Instead, we recommend that you monitor the check and alert a human operator if the check fails.
```
IronCore864, line 36: The above two paragraphs don't feel very logical to me. I tried to rewrite this part. My logic is:

You can configure "restart on check failure" -> but it's not enabled by defaut -> However, in this case, it doesn't help because of X -> It's better to Y.

Following this logic:

```suggestion
By default, Pebble keeps a service running without restarting it even if a health check fails. Although we can configure a service to restart on check failure, this approach isn't helpful here, because even if `foo` restarts, the check remains "down" (since the warning message persists in the logs and restarting won't resolve the logged warning), and subsequent failures wouldn't trigger additional restarts. So, in this particular case, monitor the check and alert a human operator to investigate.
```

docs/how-to/check-health-based-on-logs.md
```diff
@@ -0,0 +1,69 @@
...
+threshold: 1
+change-id: "60"
+logs: |
+    2025-04-26T11:22:27+08:00 ERROR exit status 1
+    2025-04-26T11:22:37+08:00 ERROR exit status 1
+```
+
+## See more
+
+- [`pebble logs`](#reference_pebble_logs_command) command
```
IronCore864, line 66: I don't see a logical order here because one might use any of these three commands as a first debugging step. Maybe use alphanumeric order?

docs/how-to/check-health-based-on-logs.md
```diff
@@ -0,0 +1,69 @@
...
+checks:
+  foo-warning:
+    override: replace
+    threshold: 1
+    exec:
+      command: bash -c '! pebble logs | grep -q "\\[foo\\] .*WARNING"'
+      # Because of YAML escaping rules, we need to use \\[ to pass \[ to grep.
+```
+
+The check searches Pebble's service logs for lines such as:
```
IronCore864, line 26: More precise:

```suggestion
The check searches Pebble's service logs for lines matching the pattern, such as:
```

IronCore864 (changes requested):
One more general comment: The doc uses both "check" and "health check". Maybe we should stick to one for more consistency, preferably "check" since it’s shorter?

docs/how-to/check-health-based-on-logs.md
dwilding: Thanks, I found this awkward to explain. I'm using your suggestion to improve it

dwilding:
Thanks @IronCore864 for the review! I made improvements based on each of your comments - hopefully it looks good now.

IronCore864 (approved):

docs/how-to/check-health-based-on-logs.md
```diff
@@ -0,0 +1,67 @@
...
+
+We'll define service called `foo` and a check called `foo-warning` that inspects the logs from `foo`:
+
+```yaml
+services:
+  foo:
+    override: replace
+    command: /bin/foo
+    startup: enabled
+checks:
```
benhoyt, line 17: Nit: I would typically put a blank line before `checks:` -- slightly more readable.

benhoyt (approved):
This looks reasonable to me, thanks! Have you tested this flow locally, with foo replaced by a fake program that spits out warnings? (Especially the `bash -c` line.)

docs/how-to/check-health-based-on-logs.md
dwilding: Done, thanks!

dwilding (commented):
Thanks for reviewing @benhoyt :slightly_smiling_face: 

> Have you tested this flow locally, with foo replaced by a fake program that spits out warnings? (Especially the bash -c line.)

Yep, I tested using Tiexin's handy [sample service](https://github.com/IronCore864/health-check-sample-service)

PR #611, opened 2025-05-02
fix: when pruning changes prune associated notices

jameinel:
Notices are currently only pruned based on time, which means they can grow excessively long, and they can reference changes that have already been pruned. While this still allows notices to grow without bound within the 1 week before they expire, now they should be pruned back along with their associated changes.

There were also a few changes associated with lint warnings (ignored errors, unused variables, etc).

### DIFF
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -531,6 +531,8 @@ func (s *State) Prune(startOfOperation time.Time, pruneWait, abortWait time.Dura
 	}
 	s.latestWarningTime.Store(&latestWarningTime)
 
+	prunedChangeIds := map[string]struct{}{}
+
 NextChange:
 	for _, chg := range changes {
 		readyTime := chg.ReadyTime()
@@ -559,6 +561,7 @@ NextChange:
 				delete(s.tasks, t.ID())
 			}
 			delete(s.changes, chg.ID())
+			prunedChangeIds[chg.ID()] = struct{}{}
 			readyChangesCount--
 		}
 	}
@@ -568,6 +571,16 @@ NextChange:
 		if t.Change() == nil && t.SpawnTime().Before(pruneLimit) {
 			s.writing()
 			delete(s.tasks, tid)
+		}
+	}
+	// Remove all the notices that refer to changes that have been pruned
+	// Note that we are using the fact that noticeKey redundantly references
+	//  several items from the notice object
+	for k := range s.notices {
+		if k.noticeType == ChangeUpdateNotice {
+			if _, pruned := prunedChangeIds[k.key]; pruned {
+				delete(s.notices, k)
+			}
 		}
 	}
 }
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -531,7 +531,7 @@ func (s *State) Prune(startOfOperation time.Time, pruneWait, abortWait time.Dura
 	}
 	s.latestWarningTime.Store(&latestWarningTime)
 
-	prunedChangeIds := map[string]struct{}{}
+	prunedChangeIDs := make(map[string]struct{})
 
 NextChange:
 	for _, chg := range changes {
@@ -561,7 +561,7 @@ NextChange:
 				delete(s.tasks, t.ID())
 			}
 			delete(s.changes, chg.ID())
-			prunedChangeIds[chg.ID()] = struct{}{}
+			prunedChangeIDs[chg.ID()] = struct{}{}
 			readyChangesCount--
 		}
 	}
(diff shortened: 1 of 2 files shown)


tlm:
Thank you @jameinel, you are the real MVP!

benhoyt (approved):
Thanks, @jameinel! I've made a couple of style tweaks, and also added a style fix (missing error check in a new test), and I'll merge this shortly.

tonyandrewmeyer:
I don't think we should do this. For perform-check and recover-check changes in the charming context in particular, I don't like that the charm may never receive the relevant event based on unpredictable timing (when pruning runs) and hard-to-predict volumes.

I would rather that in Juju instead of ignoring the notice in this situation, we had a `CheckEvent` that had the check name (which is in the notice data) and did not have the change ID, and sent that through. The charm typically needs to get the current state of the check anyway, and only needs the name for that (the change is an implementation detail at that point).

If the goal is to reduce the number of change-updated notices where the change is gone, I think we could consider some other changes:
* Perhaps we could set a different expiry on some change-update notices? Exec, for example, could have a much lower expiry.
* Perhaps we should decrease the likelihood of pruning specific types of changes - so not only sorting by ready time, but also by the change kind.
* This doesn't solve the issue, but if we didn't have notices for change-update changes for exec, then we would have many fewer notices to process and it would also be easier to understand what's happening, and we would have fewer notices to prune in this sort of solution.

I would also like to better understand how the original issue arises - the pruning is done in order of ready time, so if the notice is processed by Juju and the change has already been pruned, that implies a significant amount of changes between the change becoming ready and Juju handling the change-updated notice. I think we need to understand what is happening to better adjust the pruning algorithm.

I'd prefer that we discuss this, at least within Charm-Tech (both Ben and I have individually discussed this with John), before merging.

benhoyt:
Tony and Dima and I discussed this, and Juju ([as of 3.6.4](https://github.com/juju/juju/commit/3f4e5beea97f587a4d6bc7030223fd0e6a4d5470)) actually [doesn't mind](https://github.com/juju/juju/blob/66322cb3a2409f2d59344c88bddfe963c1f5ed94/internal/worker/uniter/pebblenotices.go#L171-L178) if the change is not not found -- it will send the check-failed or check-recovered event in any case. So keeping the notice is actually useful, even if the change object isn't around anymore. So the change in this PR will mean we just miss notice events (in newer Juju).

In that light, we don't think this is the correct change moving forward. I've set up a time to discuss further tomorrow afternoon.

jameinel:
I'd like to understand why we felt the need to prune changes to 500 but feel it is reasonable for notices to grow to whatever context fills 7 days, even when the object they refer to no longer exists.
Surely if the goal for pruning changes before their expiry time was a concern about the size of the state file, having 2000 records in Notices should also be cleaned up.
As for the concern around "don't have notices for exec changes", that might help a tiny bit, but the real issue is having the *change* event for exec changes, since this causes us to prune changes (not notices). Unless we drop change-ids from exec commands, you won't save any pruning steps.

I feel quite firmly that the pebble.state should maintain its referential integrity. If you want to prune the change, you should definitely prune the notice that is associated with it. If you feel that you shouldn't be pruning the notice, then you shouldn't be pruning its change.

jameinel:
If you'd rather a different fix, with something that refused to prune "perform-check" and "recover-check" changes, that would be fine, but we definitely need to do something.

jameinel:
I can see that you are setting the "noPruneAttr"
Here you set it as an attribute that implies you should not prune the change even if the abort limit has been reached:
https://github.com/canonical/pebble/blob/a19c2e05a28ed24c4c9f7e8cd3800a9e7ae090df/internals/overlord/checkstate/manager.go#L65

And here you set it for "perform-check" and "recover-check"
https://github.com/canonical/pebble/blob/a19c2e05a28ed24c4c9f7e8cd3800a9e7ae090df/internals/overlord/checkstate/request.go#L43
https://github.com/canonical/pebble/blob/a19c2e05a28ed24c4c9f7e8cd3800a9e7ae090df/internals/overlord/checkstate/request.go#L63

However, I think that only triggers that "I started this change", once the change actually happens, you are no longer prevent pruning that change.
Do we want to change *that* behavior, and always respect the "check-no-prune"

Since the statement you've made is "we want to keep the notices for perform-check and recover-check" forever, then we shouldn't be pruning the associated Change.

jameinel:
We could prune notices based on whatever criteria, and then use that to decide what changes we want to prune. (eg, prune notices based on the fixed 7 day window, and then only prune changes that are no longer referenced)

benhoyt:
Decisions:

- keep referential integrity (no notices that have dangling change IDs)
- add feature to prune notices over N notices (eg: 500 like changes)
- write a single log message when pruning notices to say "N notices pruned from time X to time Y" (or similar)
- consider bumping up N for pruning changes and notices from 500 to ~2000 (significant bump but not 10x)

benhoyt:
John's going to do this, but closing this PR in favour of his new #613.

PR #612, opened 2025-05-07
docs: create coding style guide

IronCore864:
Style guide for Go code.

Also add a link to the doc style guide in the ops repo.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -53,11 +53,11 @@ Order:
 
 - A special order: For example, if you have a validation function that validates user inputs against a couple of rules and the code uses if/switch and puts them in alphabetical order, when adding new code, you should probably follow that existing pattern.
 - Anywhere? This could work in certain cases. For the same example, if existing validation rules are organized in no particular order and now you are adding a new rule, it's probably ok to add it anywhere in the validation function. At the beginning, at the end, or even in the middle.
-- Some logical order: For the same example, if the new rule you add is a naming convention (e.g., it must start with a certain prefix, it must follow a regex, etc.), then there is a correct place to add it and it's the very beginning because it makes more sense: if the name doesn't even follow the convention, there is no need to check for other rules. Lazy thinking would just add this new rule anywhere or just at the bottom of the function because it's just another rule to be added to a validation function and it's the easiest to do; strategic thinking would try to discover the logic and find the best place.
+- Some logical order: For the same example, if the new rule you add is a naming convention (e.g., it must match a certain pattern), then it's almost certainly correct to put it at the very beginning: If the name doesn't even follow the convention, there's no need to check other rules. Lazy thinking would add this new rule anywhere, because it's just another rule; strategic thinking would try to discover the logic and find the best place.
 
 Ownership:
 
-- When adding a new field/flag to an existing struct, think about if the field really _belongs_ to the thing where you are adding it. Do you add it there just because it's simple to do so, or does it truly belong to it? If not, should a new struct be created specifically for it? Understanding existing code before working on it.
+- When adding a new field/flag to an existing struct, think about if the field really _belongs_ to the thing where you are adding it. Are you adding it there just because it's simple to do so, or does it truly belong to it? If not, should a new struct be created specifically for it? Understand existing code before working on it.
 
 An example of order:
 
@@ -89,9 +89,9 @@ MkdirOptions{
 
 (Pay attention to the place of `Chown` because it relates more to `UserID` and `GroupID`.)
 
-### Refactor
+### Refactoring
 
-Compare the logic before/after, there should be no behaviour change, because it's a "refactor". Make sure they are 100% the same.
+Compare the logic before/after. There should be no behavior changes, because it's a "refactor", so make sure that the behavior is 100% the same.
 
 ### Code style
 
@@ -106,7 +106,7 @@ Use Go style:
 
 ##### Named arguments
 
-Clearer to have named arguments when you're returning multiple things of the same type, and it's not clear which is which. Example:
+It's clearer to have named arguments when you're returning multiple things of the same type, and it's not clear which is which. Example:
 
 - Avoid: `func foo() (<-chan servicelog.Entry, <-chan servicelog.Entry) {}`
 - Prefer: `func foo() (stdoutCh <-chan servicelog.Entry, stderrCh <-chan servicelog.Entry) {}`
@@ -121,11 +121,11 @@ Inside a function, use the more go-idiomatic `:=` short assignment for var decla
 On the other hand, if the zero value you'd write with the first way would never be read, use `var foo type`. For example:
 
 ```go
-	var intName string
+	var numberName string
 	if i == 1 {
-		intName = "one"
+		numberName = "one"
 	} else {
-		intName = "not one"
+		numberName = "not one"
 	}
 ```
 
@@ -202,7 +202,7 @@ The two functions don't provide much value because we can simply use
 	stop := rb.writeIndex
 ```
 
-Another example: `strings.Contains` instead of creating a function `containsSubstring`.
+Another example: Use `strings.Contains` instead of creating a function `containsSubstring`.
 
 Exception: If a very short function is used repeatedly in different places and increases readability, maybe we can keep it.
 
@@ -311,10 +311,11 @@ Use pointer when:
 
 - The embedded struct is large: using a pointer avoids copying the entire struct when passing it around.
 - Optional field: If the field can be nil/null (truly optional), a pointer makes this explicit.
-- Mutable shared state: If you need to modify the original CheckInfo from different places.
+- Mutable shared state: If you need to modify the original data from different places.
 - JSON null values: If you need to distinguish between a zero-valued struct and an absent/missing value in JSON (pointer can be nil).
 
 Use direct value when:
+
 - Small struct
 - Required field: If the field should never be nil in a valid struct.
 - Value semantics: If you want each struct to have its own copy of the data.
@@ -328,7 +329,7 @@ Common Practice: Use pointers for struct fields when:
 
 ##### Trailing comma
 
-Trailing comma after the last field, the closing brace is on its own line. Example:
+Trailing comma after the last field, the closing brace is on its own line.
 
 Avoid:
 
@@ -353,7 +354,7 @@ Prefer:
 
 ##### Locks
 
-If func A calls B, B locks/unlocks a lock, then A also locks/unlocks it, it's messy and inefficient. Maybe only do one lock in A.
+Suppose that func A calls B, where B locks/unlocks a lock, and then A also locks/unlocks the same lock. This is messy and inefficient. Put everything lock-related in A.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


dwilding (approved):
Thank you for putting this together! I added a lot of comments, but almost all very minor.

STYLE.md
```diff
@@ -0,0 +1,925 @@
...
+var detailsErr *detailsError
+
+if errors.As(err, &detailsErr) && detailsErr.Details() != "" {
+	message += "; " + detailsErr.Details()
+}
+```
+
+In general, a custom error type should be considered as a marker indicating that an error is somehow "recoverable".
+
+Use a custom type if the consumer of your API can be expected to perform some special action in response to an error. If necessary, add public fields to expose data which allows a user to make good decisions about what just went wrong.
```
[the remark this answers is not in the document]
IronCore864: This is also added merging with our internal Go best practice, I didn't have this paragraph in my original draft, I think it's nice to add it here.

STYLE.md
```diff
@@ -0,0 +1,925 @@
...
+```go
+	logs, err := cmd.taskLogs(info.ChangeID)
+	if err != nil {
+		return err
+	}
+```
+
+Prefer: `fmt.Errorf("cannot get task logs for change %s: %w", info.ChangeID, err)`
+
+Using `%w` implies that the wrapped error may later be inspected in order to perform some specific action in response. Use this to allow for error recovery. Using `%v` implies that the pasted error is unrecoverable. Given that the only way to extract the underlying error would be to use very fragile string matching, this method very clearly discourages such attempts.
```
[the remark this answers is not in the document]
IronCore864: Thanks for pointing it out. This is also a new paragraph added from our Go best practice doc.

STYLE.md
```diff
@@ -0,0 +1,925 @@
...
+
+#### Avoid cached results with `-count=1`
+
+Test outputs are cached to speed up tests. If the code doesn't change, the test output shouldn't change either. Of course, this is not necessarily true, tests may read info from external sources or may use time and random related data which could change from run to run.
+
+When you request multiple test runs using the `-count` flag the intention is to run the tests multiple times, there's no logic in running them just once and showing n-1 times the same result. So `-count` triggers omitting the cached results. Setting `-count=1` explicitly (although the default value is also 1, but the behaviour is different) will simply cause running the tests once, omitting previously cached outputs.
+
+#### `t.Fatalf` or `t.Errorf`?
+
+Be careful, if the test should continue. This makes a difference.
```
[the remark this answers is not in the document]
IronCore864: Thanks, I will rewrite this.

STYLE.md
```diff
@@ -0,0 +1,740 @@
...
+
+### Housekeeping rule
+
+Be careful when the code you change has some comments/notes or even links to issues. Read them carefully. If your code change solves the issue, remove the link - keep the comment clean and true.
+
+### Spell-check and grammar-check comments
+
+Do this.
+
+### Review newly added/updated comments before committing
```
benhoyt, line 738: I think this goes without saying too, at least not here; all committed code and comments should be reviewed.

benhoyt (changes requested):
Thanks for this; it's a good start, and good distillation of the code review comments over the last months. I've added dozens of comments, mostly minor simplifications, cleanups, and clarifications.

STYLE.md
```diff
@@ -0,0 +1,740 @@
...
+
+Because in the context, it can be inferred the username and password are for basic auth.
+
+### Be consistent
+
+If `func foo()` returns a `code`, then `returnCode := foo()` is more accurate than `returnVal := foo()` - the result variable naming should be consistent with the function called.
+
+If existing code uses `json:"error,omitempty"` instead of `json:"err,omitempty"`, stick to the existing convention.
+
+If the rest of the API uses `verbNoun` then unless there is a very good reason not to, the next function should be of the form `verbNoun`.
```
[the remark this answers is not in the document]
IronCore864: An example is added:

```
For example, if the API has `createUser()`, `updateUser()`, `deleteUser()`, then adding `getUser()` fits perfectly, but if you add `userFetch()` instead of `fetchUser()`, it breaks the consistent pattern, making the API harder to learn and use.
```

STYLE.md
```diff
@@ -0,0 +1,740 @@
...
+
+### Adding code in existing functions and structs
+
+Think stratigically about _where to add the new code._ Study existing code and try to discover a logic or pattern.
+
+Order:
+
+- A special order: For example, if you have a validation function that validates user inputs against a couple of rules and the code uses if/switch and puts them in alphabetical order, when adding new code, you should probably follow that existing pattern.
+- Anywhere? This could work in certain cases. For the same example, if existing validation rules are organized in no particular order and now you are adding a new rule, it's probably ok to add it anywhere in the validation function. At the beginning, at the end, or even in the middle.
+- Some logical order: For the same example, if the new rule you add is a naming convention (e.g., it must match a certain pattern), then it's almost certainly correct to put it at the very beginning: If the name doesn't even follow the convention, there's no need to check other rules. Lazy thinking would add this new rule anywhere, because it's just another rule; strategic thinking would try to discover the logic and find the best place.
```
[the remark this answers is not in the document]
IronCore864: The example came from a PR where, in a validation function, we first validate the naming before validating other rules. It's too specific; I will replace it with another example, such as frequency of usage.

tonyandrewmeyer:
> Apparently, closing and re-opening the PR will get the cla-check workflow to pass, so testing that out.

It worked - it's failing because it seems like @IronCore864 hasn't signed the CLA with the address associated with GitHub and it's not @canonical.com so not auto-approved. But the workflow is doing what it should, at least.

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+
+- Avoid `func foo(a string, b string, c string)`
+- Prefer `func foo(a, b, c string)`
+
+### Named return values
+
+It's usually clearer to use named arguments when you're returning multiple values of the same type, for example:
+
+- Avoid: `func run() (<-chan servicelog.Entry, <-chan servicelog.Entry) {}`
+- Prefer: `func run() (stdoutCh <-chan servicelog.Entry, stderrCh <-chan servicelog.Entry) {}`
```
benhoyt: This should follow the rule above, merging arguments/return values of the same type:

```suggestion
- Prefer: `func run() (stdoutCh, stderrCh <-chan servicelog.Entry) {}`
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+
+```go
+err := c.d.overlord.CheckManager().RunCheck(r.Context(), check)
+```
+
+- Prefer:
+
+```go
+checkMgr := c.d.overlord.CheckManager()
+checks, err := checkMgr.Checks()
```
benhoyt: This isn't the same as what's above. Let's make it match:
```suggestion
err := checkMgr.RunCheck(r.Context(), check)
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+	// ...
+}
+```
+
+### Cancellation channels
+
+Cancellation channels should be unbuffered channels that are closed. For example:
+
+```go
+	stopStdout := make(chan struct{})
```
benhoyt: Please remove the unnecessary indent in this block.

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+
+### Octal number literals
+
+For new code, prefer the `0o755` format over `0755`, to make it very clear the number is octal.
+
+## Error handling
+
+### Start error messages with "cannot"
+
+Where possible, start error messages with "cannot X" for consistency."
```
benhoyt:
```suggestion
Where possible, start error messages with "cannot X" for consistency.
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+
+- Avoid: `fmt.Println("Setup failed with error:", err)`
+- Prefer: `fmt.Println("Cannot build pebble binary:", err)`
+
+### Be consistent
+
+The form of the phrases used in error messages should be consistent with other code in the same function or even the same module. For example, if existing code uses:
+
+```go
+	if !osutil.IsDir(o.pebbleDir) {
```
benhoyt: Please remove unnecessary indent.

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+```
+
+When adding a new error for no write permissions:
+
+- Avoid: `fmt.Errorf("no write permission in directory %q", o.pebbleDir)`
+- Prefer: `fmt.Errorf("directory %q not writeable, o.pebbleDir)` (Follow existing convention)
+
+### Use `errors.Is`
+
+Use `errors.Is()` to check error types, for example:
```
benhoyt:
```suggestion
Use `errors.Is()` instead of `==` to check if an error is a specific value (this handles wrapped errors correctly). For example:
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+`Setenv` calls `os.Setenv(key, value)` and uses Cleanup to restore the environment variable to its original value after the test. So, instead of doing:
+
+Avoid:
+
+```go
+func TestSomething(t *testing.T) {
+    // Manually set and cleanup env var
+    originalValue := os.Getenv("FOO")
+    os.Setenv("FOO", "1")
+    defer os.Setenv("FOO", originalValue)  // Risky - might not restore properly if test fails
```
benhoyt: I don't think this comment is true -- the defer will be run even if the test panics or fails.

```suggestion
    defer os.Setenv("FOO", originalValue)
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+	tmpDir := t.TempDir()
+	pebbleDir := filepath.Join(tmpDir, "pebble")
+	_, stderrCh := pebbleDaemon(t, pebbleDir, "run", "--create-dirs")
+	// ...
+}
+```
+
+### Check behaviour, not logs
+
+Do not bother checking the logs in the tests, because stable log formatting isn't part of the contract. Check the expected behaviour or output.
```
benhoyt:
```suggestion
Stable log formatting isn't usually part of the contract, so you shouldn't usually check logs in a test. Check the expected behaviour or output instead.
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+
+### Time-dependent tests
+
+When a test needs to wait (in a `for` loop or a sleep), make sure the time is long enough to ensure that the test passes even when the CPU is loaded. Also make sure the time is not excessively long, which would slow the tests drastically. Use a reasonable value; refer to existing tests, follow existing convention, and maybe run the test multiple times to get a reasonable value.
+
+## Comments
+
+### Be precise
+
+Think about choice of words, especially verbs. Are errors "returned" or "thrown"?
```
benhoyt:
```suggestion
Think about choice of words, especially verbs. For example, errors in Go are "returned", not "thrown".
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+
+### Helper/utility functions
+
+Add comments for complex helper and utility functions to describe their use.
+
```
benhoyt: This is true for all comments. I think we should just delete this section.
```suggestion
```

STYLE.md
```diff
@@ -0,0 +1,664 @@
...
+For example:
+
+- Avoid: "MkdirOptions is a struct of options used for Mkdir()."
+- Prefer: Either remove the comment or: "MkdirOptions holds the options for a call to Mkdir."
+
+For another example, the comment below isn't necessary since the following line is straightforward to understand:
+
+```go
+	// if path already exists
```
benhoyt: I think we should use a better example here, as `os.Stat` can return other errors than `ErrNotExist`, so the comment isn't strictly true/obvious. Also, please remove unnecessary indent. Perhaps:

```suggestion
// If string has prefix "foo"
if strings.HasPrefix(s, "foo") {
    // ...
}
```

benhoyt (approved):
Some very minor comments. Go ahead and merge after fixing these!

PR #613, opened 2025-05-09
fix: when pruning changes preserve referential integrity for notices

jameinel:
When pruning changes, also prune notices under a few rules:

1) Remove Changes down to 1000 (changed from 500) according to the existing logic.
2) Remove "change-update" Notices that refer to changes that do not exist.
3) Further remove Notices down to 1000 Notices.
4) Log an aggregate result of what changes and notices have been pruned.

This is an update to https://github.com/canonical/pebble/pull/611, built off of 1.19.0. Some of the key differences to 611 are:
a) We prune all notices that refer to an unknown change, not just the ones that were recently pruned. This is better for consistency, and it means we don't need to track an extra set of what has changed.
b) Notices are also kept to an absolute maximum count.

I ran this with a lightly customized overload.go and state/notices.go to run prune every minute, and to not prune expired notices on load. The new messages look like (before review changes):

```
2025-05-09T08:48:59.675Z [pebble] pruned 500 changes from 2025-04-29 23:00:12.035231144 +0000 UTC to 2025-05-01 13:24:18.814988887 +0000 UTC
2025-05-09T08:48:59.675Z [pebble] pruned 2003 notices from 2025-04-24 21:40:52.669570839 +0000 UTC to 2025-05-01 13:24:18.814994277 +0000 UTC
2025-05-09T08:49:59.693Z [pebble] pruned 0 changes and 0 notices
2025-05-09T08:42:31.452Z [pebble] pruned 500 changes from 2025-04-29 23:00:12.035231144 +0000 UTC to 2025-05-01 13:24:18.814988887 +0000 UTC
2025-05-09T08:42:31.452Z [pebble] pruned 0 notices
```

### DIFF
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -51,7 +51,11 @@ var (
 	// In snapd this is 7d, but also increase that in the context of Pebble.
 	abortWait = 24 * time.Hour * 14
 
+	// Hold no more than 500 completed changes, even if they have not yet expired
 	pruneMaxChanges = 500
+
+	// Hold no more than 1000 completed notices, even if they have not yet expired
+	pruneMaxNotices = 1000
 )
 
 var pruneTickerC = func(t *time.Ticker) <-chan time.Time {
@@ -404,7 +408,7 @@ func (o *Overlord) Loop() {
 			case <-pruneC:
 				st := o.State()
 				st.Lock()
-				st.Prune(o.startOfOperationTime, pruneWait, abortWait, pruneMaxChanges)
+				st.Prune(o.startOfOperationTime, pruneWait, abortWait, pruneMaxChanges, pruneMaxNotices)
 				st.Unlock()
 			}
 		}
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -51,8 +51,8 @@ var (
 	// In snapd this is 7d, but also increase that in the context of Pebble.
 	abortWait = 24 * time.Hour * 14
 
-	// Hold no more than 500 completed changes, even if they have not yet expired
-	pruneMaxChanges = 500
+	// Hold no more than 1000 completed changes, even if they have not yet expired
+	pruneMaxChanges = 1000
 
 	// Hold no more than 1000 completed notices, even if they have not yet expired
 	pruneMaxNotices = 1000
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -632,7 +632,7 @@ func (p *pruneStats) IncludeNotice(n *Notice) {
 func (p *pruneStats) Log() {
 	if p.numChangesPruned == 0 && p.numNoticesPruned == 0 {
 		// For this common case, just log a single line, we might want to move this to Debug
-		logger.Noticef("Prune() removed no Changes or Notices")
+		logger.Noticef("pruned 0 changes and 0 notices")
 	} else {
 		if p.numChangesPruned == 0 {
 			logger.Noticef("pruned 0 changes")
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -589,8 +589,23 @@ NextChange:
 	if len(s.notices) > maxNotices {
 		s.pruneMaxNotices(maxNotices, stats)
 	}
-	logger.Noticef("pruned %d changes from %s to %s", stats.numChangesPruned, stats.oldestChangePruned, stats.newestChangePruned)
-	logger.Noticef("pruned %d notices from %s to %s", stats.numNoticesPruned, stats.oldestNoticePruned, stats.newestNoticePruned)
+	if stats.numChangesPruned == 0 && stats.numNoticesPruned == 0 {
+		// For this common case, just log a single line, we might want to move this to Debug
+		logger.Noticef("Prune() removed no Changes or Notices")
+	} else {
+		if stats.numChangesPruned == 0 {
+			logger.Noticef("pruned 0 changes")
+		} else {
+			logger.Noticef("pruned %d changes from %s to %s",
+				stats.numChangesPruned, stats.oldestChangePruned, stats.newestChangePruned)
+		}
+		if stats.numNoticesPruned == 0 {
+			logger.Noticef("pruned 0 notices")
+		} else {
+			logger.Noticef("pruned %d notices from %s to %s",
+				stats.numNoticesPruned, stats.oldestNoticePruned, stats.newestNoticePruned)
+		}
+	}
 }
 
 type pruneStats struct {
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -501,9 +501,9 @@ func (s *State) RegisterPendingChangeByAttr(attr string, f func(*Change) bool) {
 //
 //   - it removes expired warnings and notices. If the notice refers to a
 //     change that was removed, then the notice is removed, if there are still
-//     more than maxNotices they are removed until we reach maxNotices. The
-//     order of pruning is based on which notices would expire first, and then
-//     by oldest lastOccurred
+//     more than maxNotices they are removed until we reach maxNotices. When pruning
+//     down to maxNotices, the oldest lastOccurred are removed (iow, we keep
+//     recent notices in favor of older notices.)
 func (s *State) Prune(startOfOperation time.Time, pruneWait, abortWait time.Duration, maxReadyChanges, maxNotices int) {
 	now := time.Now()
 	pruneLimit := now.Add(-pruneWait)
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -16,11 +16,11 @@
 package state
 
 import (
-	"container/heap"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"slices"
 	"sort"
 	"strconv"
 	"sync"
@@ -579,7 +579,7 @@ NextChange:
 				latestWarningTime = n.lastRepeated
 			}
 		} else if n.noticeType == ChangeUpdateNotice {
-			if _, exists := s.changes[n.key]; !exists {
+			if _, changeExists := s.changes[n.key]; !changeExists {
 				stats.IncludeNotice(n)
 				delete(s.notices, k)
 			}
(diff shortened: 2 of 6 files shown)


jameinel:
@benhoyt this is also built against v1.19.0, but there doesn't seem to be a 1.19-maintenance branch. I talked with Harry, and we do think it is reasonable to be pretty conservative with what we land into the 3.6 LTS

jameinel:
If I set my computer's date to 2025-05-02 I can then run against the .pebble.state file that I've shared out of band. That results in:
```
2025-05-02T08:57:22.011Z [pebble] POST /v1/notices 47.387888ms 200
2025-05-02T08:58:15.049Z [pebble] pruned 0 changes
2025-05-02T08:58:15.049Z [pebble] pruned 1367 notices from 2025-04-25 09:01:02.18641593 +0000 UTC to 2025-04-29 23:00:08.526685038 +0000 UTC
```

(I tried to use `faketime` but it seems whatever it does to alter time requests doesn't work with a go process)

jameinel:
Note that pebble will be logging a line every 10 minutes of:
```
2025-05-02T09:03:15.068Z [pebble] pruned 0 changes and 0 notices
```

Should we make the "nothing to do" log at Debug level instead of Notice?

benhoyt:
Thanks @jameinel. I'll take a closer look at the code soon (hopefully today or next week). Regarding `1.19-maintenance`, I've created that branch now, but what I've usually done is merge the fix PR into `master` (like you've got here already) and then cherry-pick it onto the maintenance branch.

jameinel:
Sure. But at least we know it is a clean patch for 1.19-maintenance :).

internals/overlord/state/state.go
```diff
@@ -498,12 +499,18 @@ func (s *State) RegisterPendingChangeByAttr(attr string, f func(*Change) bool) {
 //     changes than the limit set via "maxReadyChanges" those changes in ready
 //     state will also removed even if they are below the pruneWait duration.
 //
-//   - it removes expired warnings and notices.
-func (s *State) Prune(startOfOperation time.Time, pruneWait, abortWait time.Duration, maxReadyChanges int) {
+//   - it removes expired warnings and notices. If the notice refers to a
+//     change that was removed, then the notice is removed, if there are still
+//     more than maxNotices they are removed until we reach maxNotices. When pruning
+//     down to maxNotices, the oldest lastOccurred are removed (iow, we keep
+//     recent notices in favor of older notices.)
```
hpidcock, line 506: Please expand `iow`.

hpidcock (commented):
Seems OK, will leave it for @benhoyt to approve.

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+
+func (p *pruneStats) IncludeChange(chg *Change) {
+	p.numChangesPruned++
+	if chg == nil || chg.readyTime.IsZero() {
+		return
+	}
+	if p.oldestChangePruned.IsZero() || p.oldestChangePruned.After(chg.readyTime) {
+		p.oldestChangePruned = chg.readyTime
+	}
+	if p.newestChangePruned.IsZero() || p.newestChangePruned.Before(chg.readyTime) {
```
benhoyt, line 614: To me this logic seemed slightly confusing/backward. I think we should say "if *this* change's time is before/after" rather than the other way.

Also, we don't need to check `p.newestChange.IsZero()` because the `readyTime` will always be after zero.

Similar in `IncludeNotice`.

```suggestion
	if p.oldestChange.IsZero() || chg.readyTime.Before(p.oldestChange) {
		p.oldestChange = chg.readyTime
	}
	if chg.readyTime.After(p.newestChange) {
```

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+			logger.Noticef("pruned %d changes from %s to %s",
+				p.numChangesPruned, p.oldestChangePruned, p.newestChangePruned)
+		}
+		if p.numNoticesPruned == 0 {
+			logger.Noticef("pruned 0 notices")
+		} else {
+			logger.Noticef("pruned %d notices from %s to %s",
+				p.numNoticesPruned, p.oldestNoticePruned, p.newestNoticePruned)
+		}
+	}
```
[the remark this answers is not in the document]
jameinel: Your proposed change stops reporting on what notices have been pruned, and only reports changes that are removed. Was it intentional? (eg, you don't want to see how many notices are pruned, and what time range was removed?)

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+		p.oldestChangePruned = chg.readyTime
+	}
+	if p.newestChangePruned.IsZero() || p.newestChangePruned.Before(chg.readyTime) {
+		p.newestChangePruned = chg.readyTime
+	}
+}
+
+func (p *pruneStats) IncludeNotice(n *Notice) {
+	p.numNoticesPruned++
+	if n == nil || n.lastOccurred.IsZero() {
```
benhoyt, line 621: Similar to above, let's remove `n == nil`. Also, `lastOccured` will always be set, so I think we should either remove this or panic if it's zero with "internal error: lastOccurred unexpectedly zero".

internals/overlord/state/state.go
jameinel: :+1:

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+	}
+	if p.oldestNoticePruned.IsZero() || p.oldestNoticePruned.After(n.lastOccurred) {
+		p.oldestNoticePruned = n.lastOccurred
+	}
+	if p.newestNoticePruned.IsZero() || p.newestNoticePruned.Before(n.lastOccurred) {
+		p.newestNoticePruned = n.lastOccurred
+	}
+}
+
+func (p *pruneStats) Log() {
```
benhoyt, line 632: I think we should significantly simplify this (and don't log anything if they're both 0). Also nit: let's change the receiver name `p` to `s` ("stats" makes more sense when its used). How about:

```go
if s.numChanges > 0 {
    logger.Noticef("pruned %d changes from %s to %s", ...)
}
if s.numNotices > 0 {
    logger.Noticef("pruned %d notices from %s to %s", ...)
}
```

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+		if p.numNoticesPruned == 0 {
+			logger.Noticef("pruned 0 notices")
+		} else {
+			logger.Noticef("pruned %d notices from %s to %s",
+				p.numNoticesPruned, p.oldestNoticePruned, p.newestNoticePruned)
+		}
+	}
+}
+
+// noticeHeap tracks the lastOccurred time in a heap. It will maintain the property that
```
benhoyt, line 652: I would like to simplify this to `slices.SortFunc`, as I think this is unnecessarily complicated. We're only calling this when we've reached the 1000, and the sort will be N*log(N), but N is going to be relatively small (likely not more than a couple of thousand), so log(N) is basically a constant. In addition, slices.SortFunc is heavily optimised, doesn't go through an interface, so is likely more "machine-sympathetic" than the heap+interface, so it's likely to be as fast in any case (though I haven't tested). Even if a tiny bit slower I'd prefer the simplicity.

benhoyt (changes requested):
Thanks for this, John. I'd like to simplify in a couple of places. Happy to chat over lunch if you strongly disagree. :-)

internals/overlord/overlord.go
```diff
@@ -51,7 +51,11 @@ var (
 	// In snapd this is 7d, but also increase that in the context of Pebble.
 	abortWait = 24 * time.Hour * 14
 
+	// Hold no more than 500 completed changes, even if they have not yet expired
```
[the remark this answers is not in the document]
jameinel: :+1:

internals/overlord/state/state.go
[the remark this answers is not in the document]
hpidcock: > Your proposed change stops reporting on what notices have been pruned, and only reports changes that are removed. Was it intentional? (eg, you don't want to see how many notices are pruned, and what time range was removed?)

Ahhh it looked like duplicate code.

```suggestion
	if p.numChangesPruned == 0 && p.numNoticesPruned == 0 {
		// For this common case, just log a single line, we might want to move this to Debug
		logger.Noticef("pruned 0 changes and 0 notices")
		return
	}
	
	if p.numChangesPruned == 0 {
		logger.Noticef("pruned 0 changes")
	} else {
		logger.Noticef("pruned %d changes from %s to %s",
			p.numChangesPruned, p.oldestChangePruned, p.newestChangePruned)
	}
	
	if p.numNoticesPruned == 0 {
		logger.Noticef("pruned 0 notices")
	} else {
		logger.Noticef("pruned %d notices from %s to %s",
			p.numNoticesPruned, p.oldestNoticePruned, p.newestNoticePruned)
	}
```

This is better.

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+	if len(s.notices) > maxNotices {
+		s.pruneMaxNotices(maxNotices, stats)
+	}
+	stats.Log()
+}
+
+// pruneStats tracks how many changes and notices have been pruned, and what
+// time range has been affected.
+type pruneStats struct {
+	numChangesPruned   int
```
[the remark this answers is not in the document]
jameinel: I'm ok with this, but I will note it is used as:
```
p.numNotices
```

But this has been done

internals/overlord/state/state.go
jameinel: done

internals/overlord/state/state.go
```diff
@@ -570,6 +568,131 @@ NextChange:
...
+	numNoticesPruned   int
+	oldestChangePruned time.Time
+	newestChangePruned time.Time
+	oldestNoticePruned time.Time
+	newestNoticePruned time.Time
+}
+
+func (p *pruneStats) IncludeChange(chg *Change) {
+	p.numChangesPruned++
+	if chg == nil || chg.readyTime.IsZero() {
```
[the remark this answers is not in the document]
jameinel: This one I'm more likely to push back on. If you want a log message/something more visible, I'd be ok with it, but I'm really pretty strict about never *panic* in the field. The customer can't do anything about a programming bug but have their service be dying until you get a new release.

internals/overlord/state/state.go
jameinel: I updated the receiver, but I'm not sure I agree about not reporting when they are both 0. I'm happy to drop it to Debug, but not knowing whether it is actually occurring or not is very relevant in logs.
I tweaked it a bit. Let me know what you think.

internals/overlord/state/state.go
jameinel: So part of this is because I already had to pass over the map to turn it into a slice, so I might as well figure out which items to keep while I was doing so.
But I did move to slices.SortFunc.

internals/overlord/state/state.go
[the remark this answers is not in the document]
benhoyt: I won't fight you on this. :-) I'd prefer logic bugs like this be caught loudly when the tests fail with a panic, but you have a point.

jameinel:
@benhoyt I think I've done the changes you suggested, with the caveats of
1) nil pointers are logged as Noticef, but not a panic. If we have a pattern to panic during the test suite, I'd be happy to have that. But I really don't ever want a panic in the field.
2) I still log at Debug level when the pruning operation runs, but by default it won't report "no changes" to the log.

internals/overlord/state/state.go
```diff
@@ -570,6 +568,110 @@ NextChange:
...
+		s.oldestNotice = n.lastOccurred
+	}
+	if n.lastOccurred.After(s.newestNotice) {
+		s.newestNotice = n.lastOccurred
+	}
+}
+
+func (s *pruneStats) Log() {
+	if s.numChanges == 0 && s.numNotices == 0 {
+		logger.Debugf("pruned 0 changes and 0 notices")
```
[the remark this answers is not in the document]
jameinel: :+1:

internals/overlord/state/state.go
```diff
@@ -570,6 +568,110 @@ NextChange:
...
+		notices = append(notices, n)
+	}
+	slices.SortFunc(notices, func(a, b *Notice) int {
+		if a.lastOccurred.Before(b.lastOccurred) {
+			return -1
+		}
+		if a.lastOccurred.Equal(b.lastOccurred) {
+			return 0
+		}
+		return 1
```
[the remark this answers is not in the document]
jameinel: :+1: 
I had tried cmp.Compare, but I didn't realize there was a time.Compare

benhoyt (approved):

internals/overlord/state/state.go
benhoyt: FWIW, I did a [quick benchmark](https://gist.github.com/benhoyt/fbde2a4195aed9a2f80e86aa0ebe34e5) of heap vs sort. Heap is faster, but they're in the same ballpark, and neither blows up even with large N. I'm happy with the simplification.

```
BenchmarkHeap1500-12      	    4879	    254608 ns/op
BenchmarkHeap10000-12     	     528	   2148942 ns/op
BenchmarkHeap100000-12    	      37	  42236854 ns/op
BenchmarkSort1500-12      	    3307	    343588 ns/op
BenchmarkSort10000-12     	     339	   3687242 ns/op
BenchmarkSort100000-12    	      16	  77175471 ns/op
```

jameinel: The Heap version was designed around the idea that you might have a large N of notices, but are only pruning a small M of them. (eg, 1000 notices, but pruning 10-100). If I change your benchmark to:
```
-		st.pruneMaxNoticesHeap(maxNotices, stats)
+		st.pruneMaxNoticesHeap(numNotices-1000, stats)
```

Then you can see how heap starts to pull away. In my math, it is O(N log M) where M is the number of notices being pruned. In your benchmark M ~= N, so that becomes the same O(N log N) as the sort version.
However, with M=1000, I get:
```
BenchmarkHeap1500-14              2768        424352 ns/op
BenchmarkHeap10000-14              742       1643452 ns/op
BenchmarkHeap100000-14             100      11774590 ns/op
BenchmarkHeap1000000-14              13      91848429 ns/op
BenchmarkSort1500-14              2215        517414 ns/op
BenchmarkSort10000-14              354       3415482 ns/op
BenchmarkSort100000-14              20      56381044 ns/op
BenchmarkSort1000000-14               2     771498580 ns/op
```

So at N=1500, it is 1.29x, at 10k it is 2.07x, 100k  it is 4.7x, and 1M is 8.3x.

These are rather significant, but only if we were at 100k+, so I'm fine with a more understandable method. But it *is* a bigger advantage.

PR #615, opened 2025-05-23
docs: update doc starter pack, including page styles and build scripts

dwilding:
This PR updates Pebble docs to the latest version of [Canonical Sphinx starter pack](https://github.com/canonical/sphinx-docs-starter-pack).

**[Preview docs build](https://canonical-ubuntu-documentation-library--615.com.readthedocs.build/pebble/)**

There are some minor page style changes. For example, the width of the right-hand TOC, links are no longer underlined, and the footer is simpler (related links are available from "More resources" in the top banner).

The build scripts have also changed. There's now a makefile for checking and building the docs. I removed `tox.ini` and everything related to tox. We now need to use the following commands in the `docs` directory:

- `make run` to build the docs and start a local server
- `make cli-help` to regenerate help output in CLI the reference

The starter pack recommends having a GitHub workflow for doc checks. I've added a cut-down `.github/workflows/docs.yaml` file, as we did for Jubilant. I haven't enabled the spelling check (see [discussion for Jubilant](https://github.com/canonical/jubilant/pull/65#issuecomment-2750224585)) - but I did do a one-time scan for any misspelled words.

---

After merging this PR:

- Move `.readthedocs.yaml` to the root of the repo. It's supposed to live there, but I didn't do that in this PR because it would have broken the Read the Docs build. I'll do that separately (and reconfigure RTD at the same time).
- Find external links that should be converted to intersphinx links.


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Links
+        id: linkcheck-step
+        if: success() || failure()
+        uses: canonical/documentation-workflows/linkcheck@main
+        with:
+          working-directory: docs
```
dwilding, line 36: I'm not massively confident in the robustness of the link checker. The gnu.org link on the docs homepage was failing, and I'm not sure why, so I excluded it from the checker. If we have other problems with the link checker, we could remove it from this workflow.

dwilding:
This is ready for review. Note that checks will fail until we resolve #618.

IronCore864 (approved):

docs/tox.ini
benhoyt, line 1: This file was our own thing, right? I think we can probably drop this now, and just rely on the Starter Pack's Makefile (we'll need to update the commands in HACKING.md).

benhoyt (commented):
Thanks for this, output looks good to me! I presume you meant "linked are no longer underlined" in the PR description, rather than "links are now underlined"?

Just a couple of minor comments.

dwilding:
As we discussed today, I've removed `tox.ini` and everything related to tox. For building the docs, we now have `make run` in the `docs` directory. For updating the CLI reference, we now have `make cli-help` in the docs directory (I tried to pick a name that might be somewhat generic across Canonical). I've also updated [HACKING.md](https://github.com/canonical/pebble/pull/615/files#diff-9769a3f5504af3cd735e08f078b94f827bdd51eacac2e9d80ac9761bc485171f) accordingly.

Issue #614, opened 2025-05-16
API/swagger looks suboptimal in dark mode


dimaqq:
<img width="1247" alt="Image" src="https://github.com/user-attachments/assets/c0014a33-d555-4709-891a-e1f742b00bf3" />

IronCore864:
After playing with it a bit and getting some help from @dwilding, we figured out that this only occurs when the page is set to auto and the system is set to dark, on both macOS and Ubuntu.

PR #617, opened 2025-05-23
fix: page auto and system dark mode

IronCore864:
Fix an issue when the page theme is set to "auto", but the system appearance is set to "dark", the Sphinx docs titles and background colour inversion (done in a previous PR to improve readability in dark mode) don't work properly.

After playing with it a bit and getting some help from @dwilding, we figured out that this only occurs when the page is set to auto and the system is set to dark, on both macOS and Ubuntu. If the user explicitly sets the colour theme to dark on the page, it works properly.

See more details [here in this issue](https://github.com/canonical/pebble/issues/614)

Closes https://github.com/canonical/pebble/issues/614.

Preview: https://canonical-ubuntu-documentation-library--617.com.readthedocs.build/pebble/reference/api/#/changes-and-tasks (Remember to set your system colour mode to "dark").

### DIFF
--- a/docs/.sphinx/_static/swagger-override.css
+++ b/docs/.sphinx/_static/swagger-override.css
@@ -100,4 +100,16 @@
             filter: invert(100%) hue-rotate(180deg);
         }
     }
+
+    @media (prefers-color-scheme: dark) {
+        body[data-theme="auto"] {
+            .swagger-ui {
+                filter: invert(88%) hue-rotate(180deg);
+            }
+
+            .swagger-ui .microlight {
+                filter: invert(100%) hue-rotate(180deg);
+            }
+        }   
+    }
 }


IronCore864:
I had another draft that uses both CSS and JS to achieve the same goal: Using JavaScript to detect settings and apply a forced class. At the beginning, I thought it was more reliable since it detects things on the fly and listens to button-clicking events. But after some comparisons and tests, the version in this PR is much simpler and works reliably too, so I prefer the version in this PR.

Anyway, for possible future references, I put the other, more complicated version here:

CSS:

```css
@media not print {
  .force-invert .swagger-ui {
    filter: invert(88%) hue-rotate(180deg);
  }
  .force-invert .swagger-ui .microlight {
    filter: invert(100%) hue-rotate(180deg);
  }
}
```

JS:

```javascript
  function shouldInvert() {
    const userSelectedTheme = localStorage.getItem('theme') || 'auto'; // Check saved preference
    if (userSelectedTheme === 'dark') {
      return true;
    }

    const isSystemDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
    if (userSelectedTheme === 'auto' && isSystemDark) {
      return true;
    }

    return false;
  }

  function updateInversion() {
    document.documentElement.classList.toggle('force-invert', shouldInvert());
  }

  // --- Initialize ---
  updateInversion();

  // --- Listen for changes ---
  // If theme is toggled via localStorage
  window.addEventListener('storage', (e) => {
    if (e.key === 'theme') updateInversion();
  });

  // If the theme is toggled via HTML attributes
  new MutationObserver(updateInversion).observe(document.body, { 
    attributes: true, 
    attributeFilter: ['data-theme'] 
  });

  // If the theme is toggled via a button
  document.addEventListener('click', (e) => {
    if (e.target.closest('.theme-toggle')) {
      setTimeout(updateInversion, 50);
    }
  });
```

benhoyt:
Is it possible we could push this fix upstream into the starter pack? Or will the need for this go away when David updates to the latest starter pack?

IronCore864:
> Is it possible we could push this fix upstream into the starter pack? Or will the need for this go away when David updates to the latest starter pack?

This is some override on the Swagger UI, not on the starter pack, so it won't be pushed to the upstream.

I have also done some research regarding Swagger UI, and it seems there aren't many people who have issues with dark/auto mode, so we probably can't push this to the Swagger UI upstream either.

I think it has something to do with the furo CSS in our starter pack, which does something for `prefers-color-scheme: dark`, and it interferes with Swagger UI. Probably this is why there aren't many people who have met this issue. Not sure if the latest starter pack would render this override unnecessary. @dwilding do you have more information on this?

I also tried the latest Swagger UI version 5.22 (we have 5.11 at the moment), and it behaves the same. Since the Swagger docs still suggest 5.11, I did not update the Swagger UI version in this PR because it doesn't make any difference anyway.

benhoyt:
> This is some override on the Swagger UI, not on the starter pack, so it won't be pushed to the upstream.

Ah, of course -- thanks.

> I have also done some research regarding Swagger UI, and it seems there aren't many people who have issues with dark/auto mode, so we probably can't push this to the Swagger UI upstream either.

Got it, thanks for the research.

PR #618, opened 2025-05-23
chore: fix markdown issues in HACKING.md

dwilding:
This PR adjusts the formatting of HACKING.md to resolve linting issues that will be caught by the latest doc starter pack (see #615).

@benhoyt I don't have a particularly strong opinion here, so I can find an alternative approach if you'd rather not do this.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -109,6 +109,7 @@ Recommended prefixes are: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `
 ### Imports
 
 Pebble imports should be arranged in three groups:
+
 - standard library imports
 - third-party / non-Pebble imports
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -24,7 +24,7 @@ Usage: pebble <command> [<options>...]
 If you want to build and install the executable to your `~/go/bin` directory (which you may want to add to your path), use `go install`:
 
 ```
-$ go install ./cmd/pebble
+go install ./cmd/pebble
```
dwilding, line 27: The rationale here is: When a command has no output, we might as well make copy-pasting easier by dropping the $.

benhoyt (approved):
Yep, fine with me. I'm fine with the `$`-removal change, and the blank lines seem better to me as well.

Issue #561, opened 2025-01-21
Have client/client include a Content-Type: application/json header by default


tonyandrewmeyer:
Most requests should have a Content-Type: application/json header, but only client/services currently does this. It would be nice if we set the header by default, and then anything that wants something else (like for files) can override that.

As suggested by @benhoyt [in another PR](https://github.com/canonical/pebble/pull/560#discussion_r1923004477).

PR #619, opened 2025-05-24
feat: add default Content-Type header to client requests

IronCore864:
Add a default `Content-Type` header for the client requester. If not specified, use "application/json". It will be overridden when calling `client.Requester().Do` with headers set in `RequestOptions`. Also, remove the explicit Content-Type header with "application/json" in specific client calls.

Closes https://github.com/canonical/pebble/issues/561.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -155,6 +155,9 @@ type Config struct {
 
 	// UserAgent is the User-Agent header sent to the Pebble daemon.
 	UserAgent string
+
+	// ContentType is the Content-Type header sent to the Pebble daemon.
+	ContentType string
 }
 
 // defaultTLSVerifier blocks all TLS (HTTPS) access by always rejecting the
@@ -196,6 +199,11 @@ func New(config *Config) (*Client, error) {
 	// The default verifier never trusts any server TLS certificates.
 	if localConfig.VerifyTLSConnection == nil {
 		localConfig.VerifyTLSConnection = defaultTLSVerifier
+	}
+
+	// Set default content type if not specified.
+	if localConfig.ContentType == "" {
+		localConfig.ContentType = "application/json"
 	}
 
 	client := &Client{}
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -155,6 +155,9 @@ type Config struct {
 
 	// UserAgent is the User-Agent header sent to the Pebble daemon.
 	UserAgent string
+
+	// ContentType is the Content-Type header sent to the Pebble daemon.
+	ContentType string
```
benhoyt, line 160: I don't think we need to make this configurable: for Pebble API requests, it should always be `application/json`.

client/client.go
```diff
@@ -269,6 +277,9 @@ func (rq *defaultRequester) dispatch(ctx context.Context, method, urlpath string
 	if rq.userAgent != "" {
 		req.Header.Set("User-Agent", rq.userAgent)
 	}
+	if rq.contentType != "" {
+		req.Header.Set("Content-Type", rq.contentType)
+	}
```
benhoyt, line 282: So per above, I think we can just set this always:
```suggestion
	req.Header.Set("Content-Type", "application/json")
```

client/client_test.go
```diff
@@ -292,6 +292,23 @@ func (cs *clientSuite) TestUserAgent(c *C) {
 	c.Check(cs.req.Header.Get("User-Agent"), Equals, "some-agent/9.87")
 }
 
+func (cs *clientSuite) TestContenType(c *C) {
```
benhoyt, line 295:
```suggestion
func (cs *clientSuite) TestContentType(c *C) {
```

client/services.go
```diff
@@ -75,15 +75,12 @@ func (client *Client) doMultiServiceAction(actionName string, services []string)
...
-	headers := map[string]string{
-		"Content-Type": "application/json",
-	}
 
 	resp, err := client.Requester().Do(context.Background(), &RequestOptions{
 		Type:    AsyncRequest,
 		Method:  "POST",
 		Path:    "/v1/services",
-		Headers: headers,
+		Headers: nil,
```
benhoyt, line 83: Let's make this consistent with the others and drop the line altogether:

```suggestion
```

benhoyt (approved):
Looks good -- a couple of very minor suggestions before merging.

PR #620, opened 2025-05-27
feat: make checks report "unknown" before they've been run

### LINKED ISSUE #164: Reconsider checks reporting "up" status before being executed
### Problem

If we list the checks very early in the lifetime of the Pebble plan, the output will return a default misleading "up" status, even though we aren't yet capable of making such an assessment.

### Reproduce

Use an intentionally faulty check, like:

```yaml
checks:
    chk1:
        override: replace
        level: ready
        http:
            url: http://23234sdffsdfsadf.com/
```

Start the server: `pebble run`

Immediately list the checks with `pebble checks`, and the output will be

```
Check  Level  Status  Failures
chk1   ready  up      0/3
```

Notice the `Status = up`, which is misleading, and will eventually switch to "down".

benhoyt:
Currently `/v1/checks` returns a status of "up" before a check has been run the first time. This is misleading, as it's not really up, but "unknown". This changes the API to return "unknown" for this state.

However, the `/v1/health` endpoint still returns healthy (true) for this period, to avoid too a breaking change with health probes (plus, it's not clear that that's incorrect -- a boolean is all we have there, and it's not "down" either).

Fixes #164

### DIFF
--- a/internals/overlord/checkstate/handlers.go
+++ b/internals/overlord/checkstate/handlers.go
@@ -64,7 +64,7 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 				// Update number of failures in check info. In threshold
 				// case, check data will be updated with new change ID by
 				// changeStatusChanged.
-				m.updateCheckData(config, changeID, details.Failures)
+				m.updateCheckData(config, changeID, CheckStatusUp, details.Failures)
 			}
 
 			m.state.Lock()
@@ -91,9 +91,8 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 		}
 
 		m.incSuccessCount(config)
+		m.updateCheckData(config, changeID, CheckStatusUp, 0)
 		if details.Failures > 0 {
-			m.updateCheckData(config, changeID, 0)
-
 			m.state.Lock()
 			task.Logf("succeeded after %s", pluralise(details.Failures, "failure", "failures"))
 			details.Failures = 0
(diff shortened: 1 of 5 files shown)


internals/overlord/checkstate/handlers.go
```diff
@@ -91,9 +91,8 @@ func (m *CheckManager) doPerformCheck(task *state.Task, tomb *tombpkg.Tomb) erro
 		}
 
 		m.incSuccessCount(config)
+		m.updateCheckData(config, changeID, CheckStatusUp, 0)
```
benhoyt, line 94: This had to be moved out of the if statement. That was okay before as the status was already up, so it wouldn't change in this case, but now it's changing from "unknown" to "up" so needs to be updated.

benhoyt:
@hpidcock Would love your review here also. This is definitely a change in the API response for checks that haven't yet been run, but I very much doubt it'll break anything, as we're not changing the health probe endpoint (`/v1/health`). Please push back if you disagree.

IronCore864 (approved):

hpidcock (approved):
Overall I am happy with this change as it does not affect the health endpoint.
That being said, I think, in a follow up PR, it should affect the health endpoint.

Specifically, when the checks that contribute to a health status are in an unknown
state, they should prevent a response being sent until they are in another state.

benhoyt:
> That being said, I think, in a follow up PR, it should affect the health endpoint.
> Specifically, when the checks that contribute to a health status are in an unknown
> state, they should prevent a response being sent until they are in another state.

I'm pretty uncomfortable with that. That would be a significant/breaking change, as the health endpoint wouldn't respond in a timely fashion the first time (or when checks are restarted). The default check period is 10s, so it could be up to 10 seconds + check time.

In any case, I'll go ahead and merge this PR.

PR #623, opened 2025-05-29
chore: move Read the Docs config to repo root

dwilding:
This PR is a follow-on to #615. For consistency with the [Canonical Sphinx starter pack](https://github.com/canonical/sphinx-docs-starter-pack), `.readthedocs.yaml` should be at the repo root.


dwilding:
I'd be totally good with staying as-is for Pebble. As long as we're consistent with our own projects - how about I move the file into `docs` for Jubilant?

benhoyt:
Sounds good. And same for canonical/operator?

dwilding:
For Ops, I was wondering whether we should wait until updating the starter pack. But they don't really depend on each other, so might as well do it right away!

PR #628, opened 2025-06-03
fix: revert making checks report "unknown" before they've been run

benhoyt:
This reverts PR #620 (commit d7127c9748f2cb56756242f186b3ab5589925bba) as it was too invasive. I had incorrectly assumed that charms weren't using the check status at this level, and it'd be okay to change. However, many charms use `get_checks()` and `CheckStatus.UP` or `.DOWN`, for example:

  + ranger-k8s-operator does get_check().status != CheckStatus.UP in update-status hook, sets MaintenanceStatus if so
  + temporal-ui-k8s-operator (similar)
  + trino-k8s-operator (similar)
  + sunbeam-charms (similar, but set BlockedStatus)
  + katib-operators (similar)
  + superset-k8s-operator (similar)
  + notebook-operators (similar)

The fact that they're comparing `!=UP` means these charms will now report a blocked or maintenance status in that initial "unknown" period before the check is run. And other charms could be doing something more invasive, like restarting a workload based on a health check status `!=UP`.

So we'd like to revert this approach, and then step back and consider how to add this in a backwards-compatible way (or whether the feature is needed by charmers at all).

### DIFF
--- a/client/checks.go
+++ b/client/checks.go
@@ -57,7 +57,6 @@ const (
 type CheckStatus string
 
 const (
-	CheckStatusUnknown  CheckStatus = "unknown"
 	CheckStatusUp       CheckStatus = "up"
 	CheckStatusDown     CheckStatus = "down"
 	CheckStatusInactive CheckStatus = "inactive"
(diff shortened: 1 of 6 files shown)


benhoyt:
Just a simple revert, merging without further review.

PR #633, opened 2025-06-11
feat: add "successes" field to checks API and CLI output

benhoyt:
This adds a "Successes" column to the output of `pebble checks` (and field to the `pebble check` YAML), as well as a "successes" field in the `/v1/checks` API response.

"Successes" is the number of times this check has succeeded. It is reset when the check succeeds again after the check's failure threshold was reached. It will be zero if the check has never run, or has never run successfully. If successes==0 and failures==0, the caller knows the check has not yet been run.

The client field `CheckInfo.Successes` is typed as a pointer (`*int`) as it's being added in a much later version of Pebble -- this means clients will be able to tell the difference between `*Successes==0` and `Successes==nil` (nil meaning the field is not present in the API response because they're querying on older version of Pebble). In the `pebble checks` CLI output the column will show `?` if running against an older Pebble daemon. I figure we can do something similar on the Python side (`successes: int | None`). This does make it a bit more awkward to use, and I'm open to push-back on this choice if you have a better idea.

Example `pebble checks` output:

```
$ pebble checks
Check   Level  Startup   Status    Successes  Failures  Change
up      alive  enabled   up        5          0/1       10
```

Note that this PR is an alternative to the new status "unknown" reverted in https://github.com/canonical/pebble/pull/628.

### DIFF
--- a/client/checks.go
+++ b/client/checks.go
@@ -90,8 +90,8 @@ type CheckInfo struct {
 
 	// Successes is the number of times this check has succeeded. It is reset
 	// when the check succeeds again after the check's failure threshold was
-	// reached. This will be zero if the check has never run, or has never run
-	// successfully.
+	// reached, or if the check is stopped and started again. This will be
+	// zero if the check has never run, or has never run successfully.
 	//
 	// This field will be nil if running against a version of the daemon
 	// before this field was added to the API.
--- a/client/checks.go
+++ b/client/checks.go
@@ -88,6 +88,15 @@ type CheckInfo struct {
 	// the check is inactive.
 	Status CheckStatus `json:"status"`
 
+	// Successes is the number of times this check has succeeded. It is reset
+	// when the check succeeds again after the check's failure threshold was
+	// reached. This will be zero if the check has never run, or has never run
+	// successfully.
+	//
+	// This field will be nil if running against a version of the daemon
+	// before this field was added to the API.
+	Successes *int `json:"successes"`
+
 	// Failures is the number of times in a row this check has failed. It is
 	// reset to zero as soon as the check succeeds.
 	Failures int `json:"failures"`
--- a/docs/reference/health-checks.md
+++ b/docs/reference/health-checks.md
@@ -127,9 +127,9 @@ You can view check status using the `pebble checks` command. This reports the ch
    :input: pebble checks
 Check   Level  Startup   Status    Successes  Failures  Change
 up      alive  enabled   up        5          0/1       10
-online  ready  enabled   down      2          1/3       13 (dial tcp 127.0.0.1:8000: connect: connection refused)
+online  ready  enabled   up        2          1/3       13 (dial tcp 127.0.0.1:8000: connect: connection refused)
 test    -      disabled  down      0          42/3      14 (Get "http://localhost:8080/": dial t... run "pebble tasks 14" for more)
-extra   -      disabled  inactive  0          -         -
+extra   -      disabled  inactive  -          -         -
 ```
 
 The "Successes" column shows the number of times the check has succeeded. It is reset when the check succeeds again after the check's failure threshold was reached. This will be zero if the check has never run, or has never run successfully.
--- a/docs/reference/health-checks.md
+++ b/docs/reference/health-checks.md
@@ -154,6 +154,7 @@ You can run a check immediately using the `pebble check <name> --refresh` comman
 name: chk1
 startup: enabled
 status: up
+successes: 1
 failures: 0
 threshold: 3
 change-id: "1"
@@ -166,6 +167,7 @@ If running the check fails, the result will also show the error and logs:
 name: chk1
 startup: enabled
 status: up
+successes: 0
 failures: 1
 threshold: 3
 change-id: "1"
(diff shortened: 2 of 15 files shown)


internals/overlord/checkstate/manager.go
```diff
@@ -418,16 +411,17 @@ type refreshInfo struct {
...
+	name          string
+	level         plan.CheckLevel
+	startup       plan.CheckStartup
+	status        CheckStatus
+	successes     int
+	failures      int
+	threshold     int
+	changeID      string
+	successMetric int64
+	failureMetric int64
```
benhoyt, line 423: I renamed these fields to avoid confusion with `successes` and `failures`. These are always-increasing numbers only used for the Prometheus metrics.

internals/overlord/servstate/handlers.go
```diff
@@ -588,6 +588,9 @@ func (s *serviceData) exited(exitCode int) error {
 			s.transition(stateStopped)
 		}
 
+	case stateStopped:
```
benhoyt, line 591: This is a drive-by fix. It doesn't actually change any behaviour, but avoids an unnecessary info log in this situation (presumably primarily during test runs).

docs/reference/health-checks.md
```diff
@@ -125,13 +125,15 @@ You can view check status using the `pebble checks` command. This reports the ch
...
 ```{terminal}
    :input: pebble checks
-Check   Level  Startup   Status    Failures  Change
-up      alive  enabled   up        0/1       10
-online  ready  enabled   down      1/3       13 (dial tcp 127.0.0.1:8000: connect: connection refused)
-test    -      disabled  down      42/3      14 (Get "http://localhost:8080/": dial t... run "pebble tasks 14" for more)
-extra   -      disabled  inactive  -         -
+Check   Level  Startup   Status    Successes  Failures  Change
+up      alive  enabled   up        5          0/1       10
+online  ready  enabled   down      2          1/3       13 (dial tcp 127.0.0.1:8000: connect: connection refused)
```
tonyandrewmeyer: This is not introduced by your PR, but why is this check `down`? It's failed once and the threshold is three (and has now succeeded twice). Unless I'm missing something, we should either bump the failures to >=3 (succeeded twice but failed enough times to hit the threshold) or change the status to `up` (ran three times, two success one failure, not enough to hit the threshold).

docs/specs/openapi.yaml
```diff
@@ -1702,7 +1703,10 @@ components:
         status:
           type: string
           description: Status of the check.
-          enum: [down, up]
+          enum: [inactive, down, up]
```
tonyandrewmeyer, line 1706: Nice catch, thanks!

docs/reference/health-checks.md
```diff
@@ -125,13 +125,15 @@ You can view check status using the `pebble checks` command. This reports the ch
...
+test    -      disabled  down      0          42/3      14 (Get "http://localhost:8080/": dial t... run "pebble tasks 14" for more)
+extra   -      disabled  inactive  0          -         -
```
tonyandrewmeyer: It looks like the code has "Successes" as `-` when the check is `inactive` (the same as "Failures"), not `0`.

internals/overlord/checkstate/manager.go
```diff
@@ -353,7 +344,7 @@ func (m *CheckManager) ensureCheck(name string) *checkData {
 	return check
 }
 
-func (m *CheckManager) updateCheckData(config *plan.Check, changeID string, failures int) {
+func (m *CheckManager) updateCheckData(config *plan.Check, changeID string, successes, failures int) {
```
tonyandrewmeyer, line 347: Interesting, this thing where you can leave out the type when it's the same as the next item is a Go-ism new to me (the pretty new to Go person). I can't say that I like it :).

client/checks.go
```diff
@@ -88,6 +88,15 @@ type CheckInfo struct {
 	// the check is inactive.
 	Status CheckStatus `json:"status"`
 
+	// Successes is the number of times this check has succeeded. It is reset
+	// when the check succeeds again after the check's failure threshold was
+	// reached. This will be zero if the check has never run, or has never run
+	// successfully.
```
tonyandrewmeyer: Also zero if it has been stopped and then started again (as is "failures"). So stopping and then starting again looks the same as "never run". I think it's probably fine to consider a stop/start sequence as resetting, but I do think we should document that somewhere.

tonyandrewmeyer (changes requested):
If I do:
* Add a layer with a check.
* Have the check fail >= threshold times.
* Have the check succeed.
* Run `pebble checks`

I see a "Successes" value (and "Failures" value) of `0` (until it succeeds another time). It seems like there's an off-by-one error somewhere?

I'm good with the `*int` - as you say, it's a bit annoying, but I agree that it's worth doing.

I like the `?` in the `checks` output when the server is too old (and verified that it works for me with master as the server and this branch as the client).

internals/overlord/checkstate/manager.go
benhoyt: Heh, yeah. It's quite nice in terms of terseness, though I wonder if it could do with a bit more punctuation. I think it's somewhat [derived from Pascal](https://en.wikibooks.org/wiki/Pascal_Programming/Variables_and_Constants#More_variables) (which Go has surprising connections to), but there you have a `:` separator between names and type:

```
var foo, bar, x: integer;
```

It's also not so different from C's declarations, but in reverse:

```
int foo, bar, x;
```

tonyandrewmeyer: True, but I'm pretty sure you can't do:

```c
int foo, bar, float x;
```

in C. I have never really written any Pascal other than as a kid, but I don't think you can do this either, right?

```pascal
var foo, bar: integer, x: real;
```

Anyway, I should probably have commented this in our watercooler not the PR, sorry!

docs/reference/health-checks.md
benhoyt: Good point - fixed!

docs/reference/health-checks.md
benhoyt: Good catch, fixed.

client/checks.go
benhoyt: Good point - updated the comment.

tonyandrewmeyer (approved):
Looks good, and all seems to work for me locally, thanks!

Perhaps add a "successes" note [here in the docs](https://github.com/canonical/pebble/blob/6905490894ae2077dfc527210f278341f5d64161/docs/reference/health-checks.md?plain=1#L175)?

benhoyt:
> Perhaps add a "successes" note [here in the docs](https://github.com/canonical/pebble/blob/6905490894ae2077dfc527210f278341f5d64161/docs/reference/health-checks.md?plain=1#L175)?

Good call, done.

PR #642, opened 2025-06-22
feat: add max start retries to service spec

damianj:
Hi, I have been using pebble but sometimes would like to specify a maximum number of restart retries. Wondering if such a change would be considered -- opening this draft to just inquire on the possibility of adding this functionality.

# Description

Currently the behavior of pebble is to keep trying to restart a service while employing a back-off. I have a use case, and I think others may as well, where I'd like to let the service stop if it can't be started after a certain number of retries. At this point I may have some other action based on the service being inactive or a health-check failing. You can sort of mimic this by setting a really big backoff value, but that feels clunky as opposed to just being able to specify a limit when you want this behavior. This means adding `max-start-retries` to the service configuration. The expected behavior is that if the max retries is set to 3 then a failing service will be started the first time and then attempted to be started 3 additional times. This should not affect previous functionality as only a value greater than 0 for `max-start-retries` will enable this behavior, so previous service configurations should behave the same.

# New Usage in Service Configuration

```yaml
summary: Simple layer

description: |
    A simple layer.

services:
    http-server:
        override: replace
        summary: demo http server
        command: python3 -m http.server 8080
        startup: enabled
        max-start-retries: 3
```

### DIFF
--- a/internals/overlord/servstate/handlers.go
+++ b/internals/overlord/servstate/handlers.go
@@ -96,20 +96,21 @@ const (
 
 // serviceData holds the state and other data for a service under our control.
 type serviceData struct {
-	manager      *ServiceManager
-	state        serviceState
-	config       *plan.Service
-	workload     *workloads.Workload
-	logs         *servicelog.RingBuffer
-	started      chan error
-	stopped      chan error
-	cmd          *exec.Cmd
-	backoffNum   int
-	backoffTime  time.Duration
-	resetTimer   *time.Timer
-	restarting   bool
-	currentSince time.Time
-	startCount   atomic.Int64
+	manager           *ServiceManager
+	state             serviceState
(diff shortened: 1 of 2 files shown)


benhoyt:
I think if we did add this feature it should be called `backoff-foo`, for example `backoff-attempts`, similar to the other related fields.

Regarding your use case, what harm does it do just to let Pebble keep retrying (`backoff-limit` time apart)? Wouldn't the (Fargate or other) health checks fail if the service isn't starting up?

benhoyt:
I'm going to close this PR, as if we did add this feature it'd be something like `backoff-attempts`. However, I think we'd need a bit more weight/justification to add it, and a bit more design discussion. If you still feel strongly about it, open an issue and we can spec it out a bit more and prioritize it.

PR #649, opened 2025-07-04
feat: use libcrypt when built with cgo

hpidcock:
This is an attempt to use the system provided libcrypt (or libxcrypt) to
verify passwords instead of a pure-go implementation. This enables a
FIPS aware libcrypt to ensure FIPS compliance.

### DIFF
--- a/internals/cryptutil/crypt_cgo.go
+++ b/internals/cryptutil/crypt_cgo.go
@@ -31,6 +31,7 @@ package cryptutil
 import "C"
 
 import (
+	"crypto/subtle"
 	"errors"
 	"fmt"
 	"runtime"
@@ -49,7 +50,7 @@ func crypt6_verify(hashedKey string, key string) error {
 	defer runtime.UnlockOSThread()
 	lastSep := strings.LastIndex(hashedKey, "$")
 	if lastSep < 0 {
-		return errors.New("illformed key hash")
+		return errors.New("ill-formed key hash")
 	}
 	salt := hashedKey[:lastSep]
 	csalt := C.CString(salt)
@@ -65,8 +66,8 @@ func crypt6_verify(hashedKey string, key string) error {
 	if err != nil {
 		return fmt.Errorf("unable to verify passwd: %w", err)
 	}
-	resultHashedKey := C.GoString(chashedkey)
-	if resultHashedKey == hashedKey {
+	resHashedKey := C.GoString(chashedkey)
+	if subtle.ConstantTimeCompare([]byte(resHashedKey), []byte(hashedKey)) == 1 {
 		return nil
 	}
 	return errors.New("unable to verify passwd")
--- a/internals/cryptutil/crypt_cgo.go
+++ b/internals/cryptutil/crypt_cgo.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2025 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//go:build cgo && linux
+
+package cryptutil
+
+//#cgo pkg-config: libcrypt
+//#include <crypt.h>
+//#include <string.h>
+//#include <malloc.h>
+//struct crypt_data *alloc_cryptdata() {
+//  struct crypt_data *d = malloc(sizeof(struct crypt_data));
+//  if(d == 0) {
+//    return 0;
+//  }
+//  memset(d, 0, sizeof(struct crypt_data));
+//  return d;
+//}
+import "C"
+
+import (
+	"errors"
+	"fmt"
+	"runtime"
+	"strings"
+	"sync/atomic"
+	"unsafe"
+)
+
+var (
+	cryptNotCalled atomic.Bool
+)
+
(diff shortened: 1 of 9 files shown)


dimaqq:
Thinking outside the box: dlopen+dlsym and some register conventions to avoid cgo.

There are projects like https://github.com/ebitengine/purego (apache license)

internals/cryptutil/crypt_other.go
```diff
@@ -0,0 +1,24 @@
...
+
+//go:build !(cgo && linux)
+
+package cryptutil
+
+import "github.com/GehirnInc/crypt/sha512_crypt"
+
+func crypt6_verify(hashedKey string, key string) error {
+	crypt := sha512_crypt.New()
+	return crypt.Verify(hashedKey, []byte(key))
```
dimaqq, line 23: Go supports BoringCrypto specifically for FIPS compliance.
Would that be better than someone else's package?

Refs:
https://go.dev/doc/security/fips140
https://boringssl.googlesource.com/boringssl/+/master/crypto/fipsmodule/FIPS.md

internals/cryptutil/crypt_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+	cryptNotCalled atomic.Bool
+)
+
+func crypt6_verify(hashedKey string, key string) error {
+	cryptNotCalled.Store(false)
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+	lastSep := strings.LastIndex(hashedKey, "$")
+	if lastSep < 0 {
+		return errors.New("illformed key hash")
```
dimaqq: **total nit**

```suggestion
		return errors.New("ill-formed key hash")
```

I guess it's a real term, though maybe malformed would be understood wider.

internals/cryptutil/crypt_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	salt := hashedKey[:lastSep]
+	csalt := C.CString(salt)
```
dimaqq, line 56: Wdyt about validating that salt starts with `$6$` to ensure that the password was hashed  with sha-512?

internals/cryptutil/crypt_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+	defer C.free(unsafe.Pointer(csalt))
+	ckey := C.CString(key)
```
dimaqq, line 58: **scratching my head really hard**
Go strings may contain NULL bytes, but C strings cannot.
This means someone could trick the algorithm to use a two-letter salt, thus forcing 56-bit security.
I'm not sure if this can actually be exploited, but maybe it's better to validate the entire input or every single string converted to a C string.

internals/cryptutil/crypt_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+//#include <string.h>
+//#include <malloc.h>
+//struct crypt_data *alloc_cryptdata() {
+//  struct crypt_data *d = malloc(sizeof(struct crypt_data));
+//  if(d == 0) {
+//    return 0;
+//  }
+//  memset(d, 0, sizeof(struct crypt_data));
+//  return d;
+//}
```
dimaqq, line 30: There's `crypt_ra()` to do this, specifically so that applications don't `malloc()` for sensitive data:

https://manpages.debian.org/experimental/libcrypt1-dev/crypt_r.3.en.html

dimaqq (commented):
I wonder if go+BoringCrypto can do the trick.

internals/cryptutil/crypt_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+	if ccrypt_data == nil {
+		panic("cannot alloc crypt_data")
+	}
+	defer C.free(unsafe.Pointer(ccrypt_data))
+	chashedkey, err := C.crypt_r(ckey, csalt, ccrypt_data)
+	if err != nil {
+		return fmt.Errorf("unable to verify passwd: %w", err)
+	}
+	resultHashedKey := C.GoString(chashedkey)
+	if resultHashedKey == hashedKey {
```
dimaqq: **totally theoretical**
Some folks claim that constant-time comparison must be used even when comparing hashes.

The idea is a mixed online and offline attack:
- attacker doesn't know the password, or the hash
- attacker learns a few bytes of the hash from a timing attack on this line
- then they crunch passwords offline, discarding those whose hashes don't match the first bytes
- then they try the remaining passwords against the live system

If the passwords are made by humans, this reduces the dictionary size by a lot.

https://crypto.stackexchange.com/questions/25607/practical-uses-for-timing-attacks-on-hash-comparisons-e-g-md5/25609#25609

internals/cryptutil/crypt_cgo.go
hpidcock: Illformed is an ill-formed form of ill-formed :laughing:

internals/cryptutil/crypt_cgo.go
hpidcock: Yeah I thought about this. But, this is just hash verification, someone still had to produce the hash outside of pebble (with `openssl passwd -6` for example). A key with a NUL character will just always fail with pebble.

internals/cryptutil/crypt_cgo.go
hpidcock: Ack. Good spot.

internals/cryptutil/crypt_other.go
hpidcock: The concern here is not the SHA-512 implementation, we need a FIPS certified version anyway.

I think the concern is raised about the crypt implementation itself. Since libcrypt/libxcrypt is broadly used and included in Ubuntu (libxcrypt is main since focal, libcrypt prior was from glibc but was removed). 

I'm still not certain.

hpidcock:
Closing this as the approach needs to change. Pebble should not rely on any external
packages to provide cryptographic functionality. It should rely entirely on standard
library packages or x/crypto packages.

Moving to "crypto/pbkdf2" for key generation for password based authentication would
be ideal.

Issue #499, opened 2024-09-16
Question regarding plan deep copy on combine


flotter:
Today, a basic merge strategy pattern exist in the plan library that looks something like this:

```
    combined := &Layer{
        Services:   make(map[string]*Service),
        Checks:     make(map[string]*Check),
        LogTargets: make(map[string]*LogTarget),
        Sections:   make(map[string]Section),
    }
    :
    combined.Summary = last.Summary
    combined.Description = last.Description
    for _, layer := range layers {
        for name, service := range layer.Services {
            switch service.Override {
            case MergeOverride:
                if old, ok := combined.Services[name]; ok {
                    copied := old.Copy()    <<<============================== ?
                    copied.Merge(service)
                    combined.Services[name] = copied
                    break
                }
                fallthrough
            case ReplaceOverride:
                combined.Services[name] = service.Copy()
            case UnknownOverride:
                return nil, &FormatError{
                    Message: fmt.Sprintf(`layer %q must define "override" for service %q`,
                        layer.Label, service.Name),
                }
            default:
                return nil, &FormatError{
                    Message: fmt.Sprintf(`layer %q has invalid "override" value for service %q`,
                        layer.Label, service.Name),
                }
            }
        }
        :    
```

flotter:
@benhoyt @hpidcock If you could share some wisdom here that would be very much appreciated! This is not urgnet.

benhoyt:
Yeah, I think you're right. Feel free to put up a PR to fix (there's probably already test coverage of these cases -- but worth double-checking). I've also never liked that `fallthrough`, as I just find it confusing. What about this:

```go
			switch service.Override {
			case MergeOverride:
				if previous, ok := combined.Services[name]; ok {
					previous.Merge(service)
				} else {
					combined.Services[name] = service.Copy()
				}
			case ReplaceOverride:
				combined.Services[name] = service.Copy()
			...
			}
```

benhoyt:
Hiya @flotter, any update on this?

PR #650, opened 2025-07-04
fix: avoid second copy when combining layers

IronCore864:
Avoid a second copy of services when combining layers, also add a unit test to cover `override: merge`.

Closes https://github.com/canonical/pebble/issues/499.

See more discussion [here](https://github.com/canonical/pebble/issues/499#issuecomment-2354083170).

### DIFF
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -740,12 +740,10 @@ func CombineLayers(layers ...*Layer) (*Layer, error) {
 			switch service.Override {
 			case MergeOverride:
 				if old, ok := combined.Services[name]; ok {
-					copied := old.Copy()
-					copied.Merge(service)
-					combined.Services[name] = copied
-					break
+					old.Merge(service)
+				} else {
+					combined.Services[name] = service.Copy()
 				}
-				fallthrough
 			case ReplaceOverride:
 				combined.Services[name] = service.Copy()
 			case UnknownOverride:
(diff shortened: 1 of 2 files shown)


internals/plan/plan.go
flotter, line 1: Hi @IronCore864 this looks great, and thanks for driving this (I never got to starting this).

I think this pattern occurs in multiple places, both inside ```plan.go```, but also under ```internals/worksloads```. Would it make sense to align all the cases?

I will apply your changes also in our derived use of Pebble, where we add additional plan extensions, which again follows a similar pattern.

internals/plan/plan_test.go
```diff
@@ -1540,6 +1540,69 @@ services:
 	c.Assert(ok, Equals, true, Commentf("error must be *plan.FormatError, not %T", err))
 }
 
+func (s *S) TestCombineLayersOverrideMerge(c *C) {
```
flotter: Maybe I am missing something, but  do some of the ```TestParseLayer``` test cases not cover override merge ?

flotter (commented):
Looks good. Just two questions.

internals/plan/plan_test.go
IronCore864: Thanks for pointing this out, I had another look, the first test case in `TestParseLayer` did cover most of it (at least `before`/`after` for `override: merge`). I will remove this redundant test case.

internals/plan/plan.go
IronCore864: Thanks for pointing this out. Reading the original issue, I thought we only needed to handle the copy of services, but since the copy of checks and log targets are of the same logic, I will update them all, and also the `combine` method in `internal/workloads/workload.go`. I did a search on the whole code base, and found similar patterns in tests, which I will also fix.

Issue #639, opened 2025-06-19
Document how a service can wait for another service to have exited


dwilding:
We have this note in the docs:

> Currently, `before` and `after` are of limited usefulness, because Pebble only waits 1 second before moving on to start the next service, with no additional checks that the previous service is operating correctly.

in [How to manage service dependencies](https://documentation.ubuntu.com/pebble/how-to/service-dependencies/#specify-start-order-for-dependent-services) and [Service start order](https://documentation.ubuntu.com/pebble/explanation/service-start-order/).

People sometimes ask how to have a service B that starts only after the process of service A has exited, regardless of how long that takes. It would be helpful if we could document a workaround for this: Service A and B need to be combined into a single service, using something like

```yaml
command: bash -c 'run-a && run-b'
```

PR #651, opened 2025-07-04
docs: update service start order

IronCore864:
The network team asks how to have a service B that starts only after the process of service A has exited, regardless of how long that takes. It would be helpful if we could document a workaround for this.

Closes https://github.com/canonical/pebble/issues/639.

### DIFF
--- a/docs/explanation/service-start-order.md
+++ b/docs/explanation/service-start-order.md
@@ -8,3 +8,5 @@ The `before` option is a list of services that this service must start before (i
    :start-after: Start: Service start order note
    :end-before: End: Service start order note
 ```
+
+Note that the `before` and `after` options are not designed for scenarios where you need to start service B only after service A has fully exited. For example, in systemd, this can be achieved by running service B from service A's `ExecStopPost=` directive, and in supervisord, by using event listeners. In Pebble, a common workaround is to combine both services into a single service definition, using a command such as `bash -c 'run-service-a && run-service-b'` to ensure service B starts only after service A completes.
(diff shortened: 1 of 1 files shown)


docs/explanation/service-start-order.md
```diff
@@ -8,3 +8,5 @@ The `before` option is a list of services that this service must start before (i
    :start-after: Start: Service start order note
    :end-before: End: Service start order note
 ```
+
+Note that the `before` and `after` options are not designed for scenarios where you need to start service B only after service A has fully exited. For example, in systemd, this can be achieved by running service B from service A's `ExecStopPost=` directive, and in supervisord, by using event listeners. In Pebble, a common workaround is to combine both services into a single service definition, using a command such as `bash -c 'run-service-a && run-service-b'` to ensure service B starts only after service A completes.
```
dwilding: I suggest switching the order round to put the Pebble workaround first. Plus a couple of minor suggestions for consistency:

```suggestion
The `before` and `after` options are not designed for scenarios where you need to start service B only after service A has exited. A common workaround is to combine both services into a single service definition, using a command such as `bash -c 'run-service-a && run-service-b'` to ensure that service B starts only after service A exits successfully.

For comparison, in systemd this can be achieved by running service B from service A's `ExecStopPost=` directive. In supervisord, this can be achieved using event listeners.
```

IronCore864:
Considering this is a relatively small change, I'm going to merge it now and ask Ben to have a look when he's back.

PR #653, opened 2025-07-07
feat: prune service logs and serviceData in service manager

### LINKED ISSUE #539: Prune service log buffers
If you've started lots of services and then stopped them, the memory used by the service information (particularly the log ring buffer) stays around and is never pruned. Pebble is used in several contexts including memory-constrained containers and small devices, so we should design and implement a way to prune these when the services has been stopped for a while or there are too many services. Obviously we can't prune running services; we may be able to shrink/deallocate the log buffers for services on backoff state if they've been in that state for too long.

For comparison, old (done) changes are pruned when there are more than 500 of them or they've been done for more than 7 days (the interval for checking whether to prune is 10m).

IronCore864:
Prune service logs and `serviceData` in the service manager.

- If a service is inactive, and its `currentSince` is older than `pruneWait`, it will be pruned.
- If the total number of entries of serviceData is still greater than `maxServiceData`, inactive services will be pruned even if they are _not_ older than `pruneWait` to keep the total number of serviceData entries under control. In this case, it's sorted by `currentSince` to ensure pruning older inactive services first.

Also added a few unit test cases to cover this feature. Test cases for the overlord are not updated because the outer prune loop is already tested; no integration test is added because the unit tests already cover all cases.

Fixes #539.

### DIFF
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -58,8 +58,8 @@ var (
 	// Hold no more than 1000 completed notices, even if they have not yet expired
 	pruneMaxNotices = 1000
 
-	// Hold	no more than 100 serviceData.
-	pruneMaxServiceData = 100
+	// Hold no more than 100 inactive services.
+	pruneMaxInactiveServices = 100
 )
 
 var pruneTickerC = func(t *time.Ticker) <-chan time.Time {
@@ -429,7 +429,7 @@ func (o *Overlord) Loop() {
 				st.Prune(o.startOfOperationTime, pruneWait, abortWait, pruneMaxChanges, pruneMaxNotices)
 				st.Unlock()
 				serviceMgr := o.ServiceManager()
-				serviceMgr.Prune(pruneWait, pruneMaxServiceData)
+				serviceMgr.Prune(pruneWait, pruneMaxInactiveServices)
 			}
 		}
 	})
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -58,8 +58,8 @@ var (
 	// Hold no more than 1000 completed notices, even if they have not yet expired
 	pruneMaxNotices = 1000
 
-	// Hold	no more than 1000 serviceData.
-	pruneMaxServiceData = 1000
+	// Hold	no more than 100 serviceData.
+	pruneMaxServiceData = 100
 )
 
 var pruneTickerC = func(t *time.Ticker) <-chan time.Time {
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -57,6 +57,10 @@ var (
 
 	// Hold no more than 1000 completed notices, even if they have not yet expired
 	pruneMaxNotices = 1000
+
+	// Hold	no more than 1000 serviceData,
+	// even if it means pruning in active services less than pruneWait.
+	pruneMaxServiceData = 1000
 )
 
 var pruneTickerC = func(t *time.Ticker) <-chan time.Time {
@@ -425,6 +429,8 @@ func (o *Overlord) Loop() {
 				st.Lock()
 				st.Prune(o.startOfOperationTime, pruneWait, abortWait, pruneMaxChanges, pruneMaxNotices)
 				st.Unlock()
+				serviceMgr := o.ServiceManager()
+				serviceMgr.Prune(pruneWait, pruneMaxServiceData)
 			}
 		}
 	})
@@ -606,6 +612,7 @@ func FakeWithState(handleRestart func(restart.RestartType)) *Overlord {
 	s := state.New(fakeBackend{o: o})
 	o.stateEng = NewStateEngine(s)
 	o.runner = state.NewTaskRunner(s)
+	o.serviceMgr, _ = servstate.NewManager(s, o.runner, nil, nil, nil)
 	return o
 }
 
--- a/internals/overlord/overlord.go
+++ b/internals/overlord/overlord.go
@@ -58,8 +58,7 @@ var (
 	// Hold no more than 1000 completed notices, even if they have not yet expired
 	pruneMaxNotices = 1000
 
-	// Hold	no more than 1000 serviceData,
-	// even if it means pruning in active services less than pruneWait.
+	// Hold	no more than 1000 serviceData.
 	pruneMaxServiceData = 1000
 )
 
--- a/internals/overlord/servstate/manager.go
+++ b/internals/overlord/servstate/manager.go
@@ -17,6 +17,9 @@ import (
 	"github.com/canonical/pebble/internals/workloads"
 )
 
+// timeNow can be faked during testing.
+var timeNow = time.Now
+
 type ServiceManager struct {
 	state *state.State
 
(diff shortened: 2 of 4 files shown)


internals/overlord/servstate/manager.go
```diff
@@ -368,6 +371,41 @@ func (m *ServiceManager) WriteMetrics(writer metrics.Writer) error {
...
+//     serviceData even if they are not older than pruneWait. Inactive services are sorted by currentSince,
+//     and remove the older ones first.
+func (m *ServiceManager) Prune(pruneWait time.Duration, maxServiceData int) {
+	m.servicesLock.Lock()
+	defer m.servicesLock.Unlock()
+
+	now := timeNow()
+	pruneLimit := now.Add(-pruneWait)
+	for name, s := range m.services {
+		if s != nil && stateToStatus(s.state) == StatusInactive && s.currentSince.Before(pruneLimit) {
```
benhoyt: Entries in `m.services` can never be nil, so we can remove the `s != nil` check (other similar code in manager.go doesn't).
```suggestion
		if stateToStatus(s.state) == StatusInactive && s.currentSince.Before(pruneLimit) {
```

internals/overlord/servstate/manager.go
```diff
@@ -368,6 +371,41 @@ func (m *ServiceManager) WriteMetrics(writer metrics.Writer) error {
...
+			delete(m.services, name)
+		}
+	}
+
+	if len(m.services) > maxServiceData {
+		var inactive []*serviceData
+		for _, s := range m.services {
+			if s != nil && stateToStatus(s.state) == StatusInactive {
```
benhoyt:
```suggestion
			if stateToStatus(s.state) == StatusInactive {
```

internals/overlord/servstate/manager.go
```diff
@@ -368,6 +371,41 @@ func (m *ServiceManager) WriteMetrics(writer metrics.Writer) error {
...
+				inactive = append(inactive, s)
+			}
+		}
+		sort.Slice(inactive, func(i, j int) bool {
+			return inactive[i].currentSince.Before(inactive[j].currentSince)
+		})
```
benhoyt: The slightly more modern, and slightly more efficient approach would use `slices.SortFunc`:

```suggestion
		slices.SortFunc(inactive, func(a, b *serviceData) bool {
			return a.currentSince.Compare(b.currentSince)
		})
```

internals/overlord/servstate/manager.go
```diff
@@ -368,6 +371,41 @@ func (m *ServiceManager) WriteMetrics(writer metrics.Writer) error {
...
+		excess := len(m.services) - maxServiceData
+		toDelete := min(len(inactive), excess)
```
benhoyt: Why is this line needed? If `excess` is negative (because `len(m.services)` is small), then the for loop would never run anyway.

internals/overlord/servstate/manager_test.go
```diff
@@ -2396,3 +2396,177 @@ pebble_service_start_count{service="test2"} 2
...
+	chg = s.stopServices(c, [][]string{{"test3", "test4", "test5"}})
+	s.st.Lock()
+	c.Check(chg.Status(), Equals, state.DoneStatus, Commentf("Error: %v", chg.Err()))
+	s.st.Unlock()
+
+	// Immediately prune; all inactive services are not older than pruneWait, so nothing will be pruned.
+	s.manager.Prune(7*24*time.Hour, 1000)
+	services, err := s.manager.Services(nil)
+	c.Assert(err, IsNil)
+	for i := range len(services) {
```
benhoyt: Could you use a regular range loop here? Similar throughout.
```suggestion
	for _, service := range services {
```

internals/overlord/servstate/manager_test.go
```diff
@@ -2396,3 +2396,177 @@ pebble_service_start_count{service="test2"} 2
...
+	s.st.Lock()
+	c.Check(chg.Status(), Equals, state.DoneStatus, Commentf("Error: %v", chg.Err()))
+	s.st.Unlock()
+
+	// Mock time so that the inactive test3 is older than pruneWait and will be pruned.
+	testBaseTime := getTestTime(2100, 1, 1)
+	restoreTime := servstate.FakeTimeNow(testBaseTime)
+	defer restoreTime()
+
+	// test4 and test5 become inactive but not older than pruneWait.
```
benhoyt: I don't think this test is doing what it's intended to do, because `servstate.FakeTimeNow` only affects `Prune`, not starting or stopping services. So these will all be pruned because of pruneWait, not maxServiceData (I confirmed this by adding a panic to that second `delete()` and running this test -- the panic was never hit).

internals/overlord/servstate/manager_test.go
```diff
@@ -2396,3 +2396,177 @@ pebble_service_start_count{service="test2"} 2
...
+	s.st.Lock()
+	c.Check(chg.Status(), Equals, state.DoneStatus, Commentf("Error: %v", chg.Err()))
+	s.st.Unlock()
+
+	chg = s.stopServices(c, [][]string{{"test3"}})
+	s.st.Lock()
+	c.Check(chg.Status(), Equals, state.DoneStatus, Commentf("Error: %v", chg.Err()))
+	s.st.Unlock()
+
+	// Mock time, and stop test4 and test5, so that they have a newer currentSince.
```
benhoyt: The mocked time only applies to `Prune` itself, so this comment is wrong. You should just be able to to stop them separately, one after the other, to get different times.

internals/overlord/servstate/manager_test.go
```diff
@@ -2396,3 +2396,177 @@ pebble_service_start_count{service="test2"} 2
...
+	s.manager.Prune(7*24*time.Hour, 1000)
+	services, err := s.manager.Services(nil)
+	c.Assert(err, IsNil)
+	for i := range len(services) {
+		// CurrentSince is not nil, meaning the serviceData is not pruned.
+		c.Assert(services[i].CurrentSince, Not(Equals), time.Time{})
+	}
+
+	// Mock time so that all inactive services' serviceData will be pruned.
+	testBaseTime := getTestTime(2100, 1, 1)
```
benhoyt: I know it's silly, as I wouldn't mind if these tests started failing in 75 years, but maybe we should be safer and use 2999 or 9999 for the test date. :-)

internals/overlord/servstate/manager.go
IronCore864: If `len(m.services)` is 1000, `maxServiceData` is 100, `excess` is 900. But if 990 services are active, we can't delete that, `len(inactive)` is only 10, so we can only delete `min(len(inactive), excess)`.

internals/overlord/overlord.go
```diff
@@ -57,6 +57,9 @@ var (
 
 	// Hold no more than 1000 completed notices, even if they have not yet expired
 	pruneMaxNotices = 1000
+
+	// Hold	no more than 100 serviceData.
```
benhoyt:
```suggestion
	// Hold no more than 100 inactive services.
```

internals/overlord/overlord.go
```diff
@@ -57,6 +57,9 @@ var (
...
+	pruneMaxServiceData = 100
```
benhoyt: Perhaps we should include the word "inactive" in the name, as that's what it actually is:
```suggestion
	pruneMaxInactiveServices = 100
```

internals/overlord/servstate/manager.go
```diff
@@ -368,6 +372,41 @@ func (m *ServiceManager) WriteMetrics(writer metrics.Writer) error {
...
+			}
+		}
+		slices.SortFunc(inactive, func(a *serviceData, b *serviceData) int {
+			return a.currentSince.Compare(b.currentSince)
+		})
+		excess := len(m.services) - maxServiceData
+		toDelete := min(len(inactive), excess)
+		for i := 0; i < toDelete; i++ {
+			delete(m.services, inactive[i].config.Name)
+		}
```
benhoyt: Now that I think about it, "excess" is not really `len(m.services) - maxServiceData` but `len(inactive) - maxServiceData` or specifically `max(len(inactive) - maxServiceData, 0)`. So this seems clearer to me:

```suggestion
		excess := max(len(inactive) - maxServiceData, 0)
		for i := range excess {
			delete(m.services, inactive[i].config.Name)
		}
```

Is that right?

internals/overlord/overlord.go
IronCore864: I see the issue now (which explains the discussion in the other comment), it was because I thought we would only keep 100 entries of serviceData, both active or not, that was why I was calculating how many inactive serviceData entries need to be pruned to reach this target.

This comment cleared things up. If it is the number of inactive services, rather than the total entries of serviceData, then it's more straightforward. I will make the changes accordingly, and probably need to update tests as well.

internals/overlord/servstate/manager.go
IronCore864: Yes, I was thinking maxServiceData is for the total number of serviceData entries, active or not. If maxServiceData means inactive services only, your logic is correct, see my reply in the other comment, and I have made the changes accordingly.

Issue #590, opened 2025-03-14
Add in-memory state option, PEBBLE_PERSIST=never


benhoyt:
Per spec [OP059](https://docs.google.com/document/d/1tgVvYzswCT4nTXsefNq2CU_mtEJEA-51TXsJNUFB0E8/edit?tab=t.0#heading=h.a291qhvyhsw3), we'd like to add a way to tell the Pebble daemon (`pebble run`) not to persist the state to disk, for environments where that makes sense.

```
PEBBLE_PERSIST=always    # the default
PEBBLE_PERSIST=never     # new feature
```

- always: always persist state to disk, and fail loudly if we can’t (the current behaviour). This would be the default if the environment variable is unset.
- never:  never persist state to disk, use in-memory state only.

If the new “never” in-memory mode was specified, the Pebble daemon would use a different [Backend](https://github.com/canonical/pebble/blob/e3b81e139fc188e8ef01311e9fb13fe26402f816/internals/overlord/state/state.go#L34) implementation which did nothing in its Checkpoint method; the default writes the state JSON to a file on disk.

For efficiency, we should try to entirely avoid serialising the state to JSON in this case, perhaps by adding a new NeedsCheckpoint() bool method on the interface.

PR #658, opened 2025-07-15
feat: in-memory state

IronCore864:
Add a feature that if the environment variable `PEBBLE_PERSIST` is set to `never`, Pebble daemon will not persist the state to the disk.

Closes https://github.com/canonical/pebble/issues/590.

By default, if `PEBBLE_PERSIST` is not set, or set to any value other than `never` (I'm not sure if this is the desired behaviour), Pebble will still write its state to the disk, and for a read-only file system, it will fail.

For more information, see the [spec here](https://docs.google.com/document/d/1tgVvYzswCT4nTXsefNq2CU_mtEJEA-51TXsJNUFB0E8/edit?tab=t.0#heading=h.a291qhvyhsw3) (internal only).


internals/overlord/overlord_test.go
```diff
@@ -77,7 +78,32 @@ func (ovs *overlordSuite) TestNew(c *C) {
 	restore := patch.Fake(42, 2, nil)
 	defer restore()
 
-	o, err := overlord.New(&overlord.Options{PebbleDir: ovs.dir})
+	o, err := overlord.New(&overlord.Options{PebbleDir: ovs.dir, Persist: true})
```
IronCore864: The diff here does not display properly. The change is that I added `Persist: true` in `TestNew` (the only change), then created a new test case `TestNewInMemoryBackend` with `Persist` set to false.

internals/overlord/overlord_test.go
```diff
@@ -111,7 +140,7 @@ func (ovs *overlordSuite) TestNewWithGoodState(c *C) {
 	err := os.WriteFile(ovs.statePath, fakeState, 0600)
 	c.Assert(err, IsNil)
 
-	o, err := overlord.New(&overlord.Options{PebbleDir: ovs.dir})
+	o, err := overlord.New(&overlord.Options{PebbleDir: ovs.dir, Persist: true})
```
IronCore864: This change (including the following in the same file) is to make the existing test pass.

internals/overlord/state/state_test.go
```diff
@@ -226,6 +226,10 @@ func (b *fakeStateBackend) EnsureBefore(d time.Duration) {
 	b.ensureBefore = d
 }
 
+func (b *fakeStateBackend) NeedsCheckpoint() bool {
+	return true
```
IronCore864, line 230: NeedsCheckpoint for `fakeStateBackend` must be true because stateSuite has tests testing the checkpoint behaviour.

internals/daemon/daemon.go
```diff
@@ -158,6 +158,9 @@ type Options struct {
 	// OverlordExtension is an optional interface used to extend the capabilities
 	// of the Overlord.
 	OverlordExtension overlord.Extension
+
+	// Persist specifies whether the state should be persisted to disk.
+	Persist bool
```
benhoyt: I think it'd be better if we made the zero value the default (to persist). Let's change this to `Persist PersistMode` and then have:

```go
type PersistMode int

const (
    PersistDefault PersistMode = 0
    PersistNever PersistMode = 1
)
```

Similar for the `Overlord` field.

This would also mean you don't have to specify the default `Persist: true` in various places -- you'd just leave the field as the zero value (default).

internals/overlord/overlord.go
```diff
@@ -143,12 +145,21 @@ func New(opts *Options) (*Overlord, error) {
...
+			ensureBefore: o.ensureBefore,
+		}
+	} else {
+		// In-memory backend.
+		backend = &inMemoryBackend{
+			ensureBefore: o.ensureBefore,
+		}
 	}
+
 	s, restartMgr, err := loadState(statePath, opts.RestartHandler, backend)
```
benhoyt: It's weird that a function called `loadState` is being called even when we're not loading state. And now `loadState` also has two calls to `backend.NeedsCheckpoint` in it, and most of the code is skipped if it's false. I think it'd be better to create a new function `setupState` or `newState` that handles the persist=never case instead.

And perhaps factor out the "get current boot ID" code into a separate function, called from both (or, probably better, called once beforehand, and the boot ID passed in to both).

internals/overlord/overlord_test.go
```diff
@@ -77,7 +78,32 @@ func (ovs *overlordSuite) TestNew(c *C) {
 	restore := patch.Fake(42, 2, nil)
 	defer restore()
 
-	o, err := overlord.New(&overlord.Options{PebbleDir: ovs.dir})
+	o, err := overlord.New(&overlord.Options{PebbleDir: ovs.dir, Persist: true})
+	c.Assert(err, IsNil)
+	c.Check(o, NotNil)
```
benhoyt: Probably don't need this check. Or if anything, it should be an Assert (we can't execute the following `o.StateEngine()` and similar if it's nil).

internals/overlord/overlord_test.go
```diff
@@ -96,6 +122,9 @@ func (ovs *overlordSuite) TestNew(c *C) {
 	c.Check(patchLevel, Equals, 42)
 	s.Get("patch-sublevel", &patchSublevel)
 	c.Check(patchSublevel, Equals, 2)
+
+	_, err = os.Stat(ovs.statePath)
+	c.Check(errors.Is(err, fs.ErrNotExist), Equals, true)
```
benhoyt, line 128: Should we add the opposite check to `TestNew` -- that the state file does exist?

internals/overlord/state/state.go
```diff
@@ -296,20 +297,25 @@ func (s *State) Unlock() {
...
 
-	data := s.checkpointData()
-	var err error
-	start := time.Now()
-	for time.Since(start) <= unlockCheckpointRetryMaxTime {
-		if err = s.backend.Checkpoint(data); err == nil {
-			s.modified = false
-			return
-		}
+	if s.backend.NeedsCheckpoint() {
```
benhoyt: I think it'd be clearer/simpler with an early return:

```suggestion
	if !s.backend.NeedsCheckpoint() {
		s.modified = false
		return
	}
	// existing code at top level
```

tests/enter_test.go
```diff
@@ -62,3 +63,32 @@ sleep 1.1
...
+// TestEnterPersistNever tests that when the environment variable `PEBBLE_PERSIST` is set to "never",
+// Pebble does not persist its state to the disk.
+func TestEnterPersistNever(t *testing.T) {
+	t.Setenv("PEBBLE_PERSIST", "never")
+	pebbleDir := t.TempDir()
+
+	script := `
+#!/bin/sh
+echo "hello world"
+sleep 1.1
```
benhoyt: Is there any reason this needs to be over 1 second, or could it just be `sleep 0.1` to make the tests run faster? (Service startup needs to be longer than 1 second, but execs shouldn't.)

tests/enter_test.go
```diff
@@ -62,3 +63,32 @@ sleep 1.1
...
+	`
+	path := filepath.Join(pebbleDir, "test.sh")
+	err := os.WriteFile(path, []byte(script), 0755)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	stdoutCh, _ := pebbleDaemon(t, pebbleDir, "enter", "exec", "/bin/sh", "-c", path)
```
benhoyt: I don't think you actually need to write a script at all. `sh -c` means "read commands from command line". So you should be able to skip the script and just do:

```suggestion
	stdoutCh, _ := pebbleDaemon(t, pebbleDir, "enter", "exec", "/bin/sh", "-c", "echo 'hello world'; sleep 0.1")
```

If so, we can make similar changes above.

tests/run_test.go
```diff
@@ -243,3 +243,113 @@ local:
...
+	_, err := os.Stat(filepath.Join(pebbleDir, ".pebble.state"))
+	if err == nil {
+		t.Fatalf("pebble run with PEBBLE_PERSIST set to 'never' still created the state file")
+	}
+	if !errors.Is(err, os.ErrNotExist) {
+		t.Fatalf("pebble run with PEBBLE_PERSIST set to 'never' got error other than ErrNotExist: %v", err)
+	}
+}
+
+// TestPersistAlways tests that when the environment variable `PEBBLE_PERSIST` is set to "always",
```
benhoyt: I think we can drop the "always" and "other/foo" tests, as these aren't values we test for.

internals/overlord/state/taskrunner_test.go
```diff
@@ -56,6 +56,8 @@ func (b *stateBackend) EnsureBefore(d time.Duration) {
 	}
 }
 
+func (b *stateBackend) NeedsCheckpoint() bool { return false }
```
benhoyt: Nit: let's be consistent with the above and split this onto multiple lines.

internals/overlord/backend.go
```diff
@@ -32,3 +32,25 @@ func (osb *overlordStateBackend) Checkpoint(data []byte) error {
...
+}
+
+type inMemoryBackend struct {
+	ensureBefore func(d time.Duration)
+}
+
+// Checkpoint of the in-memory backend does nothing because
+// we're keeping the state in memory only.
+func (imb *inMemoryBackend) Checkpoint(data []byte) error {
+	return nil
```
benhoyt: I guess we should probably make this a `panic("internal error: inMemoryBackend.Checkpoint should never be called")` as `NeedsCheckpoint()` returns false.

internals/overlord/overlord.go
```diff
@@ -249,12 +275,10 @@ func (o *Overlord) Extension() Extension {
...
 }
 
-func loadState(statePath string, restartHandler restart.Handler, backend state.Backend) (*state.State, *restart.RestartManager, error) {
-	timings := timing.Start("", "", map[string]string{"startup": "load-state"})
-
+func getCurrentBootID() (string, error) {
 	curBootID, err := osutil.BootID()
 	if err != nil {
-		return nil, nil, fmt.Errorf("fatal: cannot find current boot ID: %v", err)
+		return curBootID, fmt.Errorf("fatal: cannot find current boot ID: %v", err)
```
benhoyt: Nit: better to explicitly return the zero value `""` in the error case (also below).

docs/explanation/security.md
```diff
@@ -24,6 +24,7 @@ The `$PEBBLE` directory must be readable and writable by the UID of the pebble p
 
 The file `$PEBBLE/.pebble.state` contains the internal state of the Pebble daemon. You shouldn't try to edit this file or change its permissions.
 
+If `$PEBBLE_PERSIST` is set to "never", then Pebble will only keep the state in memory without persisting it to the file `$PEBBLE/.pebble.state`.
```
benhoyt: Nit:
```suggestion
If `$PEBBLE_PERSIST` is set to "never", then Pebble will only keep the state in memory without persisting it to the state file.
```

docs/how-to/manage-a-remote-system.md
```diff
@@ -124,6 +124,12 @@ For more information about related Pebble commands, see:
 - {ref}`reference_pebble_push_command`
 - {ref}`reference_pebble_rm_command`
 
+## Run Pebble with a read-only file system
+
+Sometimes Pebble needs to be run in a remote system whose filesystem is read-only. For example, when Pebble is used in a Docker context with `docker run --read-only`, or in a Kubernetes context with `readOnlyRootFilesystem` set to true, mounting the container's root filesystem as read-only. In such cases, Pebble can not persist its state to the file `$PEBBLE/.pebble.state`.
```
benhoyt: Readability nit:
```suggestion
Sometimes Pebble needs to be run in a remote system whose filesystem is read-only. For example, when Pebble is used in a Docker context with `docker run --read-only`, or in a Kubernetes context with `readOnlyRootFilesystem` set to true, the container's root filesystem is mounted as read-only. In such cases, Pebble cannot persist its state to the state file.
```

benhoyt (approved):
Great, looks good to me. Two nit comments on the docs, but approving.

PR #661, opened 2025-07-20
ci: bump the setup-go version and explicitly request Go 1.24 in the TIOBE workflow

tonyandrewmeyer:
#660 was meant to adjust the TIOBE workflow to run on self-hosted runners with no other changes, but that promise was unfortunately not fulfilled. The team suggests that bumping the setup-go action to v5 and explicitly specifying the version of Go to use may fix the failing job.

While making the adjustments, a drive-by of minor security improvements:
* Default permissions to none.
* Don't persist the git credentials when cloning.
* Pin the TIOBE action to a specific revision.
* Allow actions/ and github/ actions to be pinned to versions rather than revisions.

### DIFF
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -5,14 +5,20 @@ on:
   schedule:
     - cron:  '0 7 1 * *'
 
+permissions: {}
+
 jobs:
   TICS:
     runs-on: [self-hosted, reactive, amd64, tiobe, noble]
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/tiobe.yaml
```diff
@@ -5,14 +5,20 @@ on:
...
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
```
benhoyt: Can we use `go-version-file` (to get the version from go.mod) instead? https://github.com/actions/setup-go/?tab=readme-ov-file#getting-go-version-from-the-gomod-file

tonyandrewmeyer: TIL :)

benhoyt (approved):

PR #663, opened 2025-07-22
feat: send logs in opentelemetry format

IronCore864:


IronCore864:
Some notes:

`serviceLog.Entry` has no "level", but for otelcol it's better to set severity number/severity text. Although optional, without, them, it's not extremely useful in the centralized logging system's UI because most tools can filter logs by levels:

Screenshot 2025-07-17 at 16 03 04

Possible choices:

- Parse log text, if it contains "error", "info" at the beginning, use it as level; if it does not exist, defaults to info.
- Set all as info.
- Don't set (in this PR).

internals/overlord/logstate/gatherer.go
```diff
@@ -368,6 +369,8 @@ func newLogClient(target *plan.LogTarget) (logClient, error) {
 	switch target.Type {
 	case plan.LokiTarget:
 		return loki.NewClient(target), nil
+	case plan.OpenTelemetryTarget:
+		return opentelemetry.NewClient(target), nil
 	//case plan.SyslogTarget: TODO
```
benhoyt: Unrelated, but I've opened https://github.com/canonical/pebble/pull/665 to remove the `plan.SyslogTarget` constant (until such a time as syslog is supported) so that the type is validated earlier, at plan validation time rather than later when the log gatherer is created.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// Resource information, partially from 'type Resource struct' in
+// opentelemetry-collector/pdata/internal/data/protogen/resource/v1/resource.pb.go
+type Resource struct {
+	Attributes []KeyValue `json:"attributes"`
+}
+
+type KeyValue struct {
+	Key   string         `json:"key"`
```
benhoyt: Possibly doesn't matter, but [in the original](https://github.com/open-telemetry/opentelemetry-collector/blob/3c0fd3946f70a0b1fa97813c39dbc4d91d95afa6/pdata/internal/data/protogen/common/v1/common.pb.go#L286) this is marked as "omitempty", so we should probably match:
```suggestion
	Key   string         `json:"key,omitempty"`
```

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	Value AttributeValue `json:"value"`
```
benhoyt: In the original, this is called `AnyValue`. Should we match that name?

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+const (
+	requestTimeout    = 10 * time.Second
+	maxRequestEntries = 100
+	batchSize         = 10
+)
+
+// A collection of ScopeLogs from a Resource.
+// Refer to `type ResourceLogs struct` in
+// opentelemetry-collector/pdata/internal/data/protogen/logs/v1/logs.pb.go
+type ResourceLogs struct {
```
benhoyt: We should probably make all of these type names start with a lowercase letter, that is, non-exported. But the fields within need to stay uppercased.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+type KeyValue struct {
+	Key   string         `json:"key"`
+	Value AttributeValue `json:"value"`
+}
+
+// AttributeValue represents the OTLP attribute value format.
+// Refer to `type AnyValue struct` in
+// opentelemetry-collector/pdata/internal/data/protogen/common/v1/common.pb.go
+type AttributeValue struct {
+	StringValue *string      `json:"stringValue,omitempty"`
```
benhoyt: It seems like we only use the `StringValue` type. Can we drop the rest of these fields? In other places we've only included the fields we need, which seems good to me.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+
+const (
+	requestTimeout    = 10 * time.Second
+	maxRequestEntries = 100
+	batchSize         = 10
+)
+
+// A collection of ScopeLogs from a Resource.
+// Refer to `type ResourceLogs struct` in
+// opentelemetry-collector/pdata/internal/data/protogen/logs/v1/logs.pb.go
```
benhoyt: Can we change these to full permanent links, such as https://github.com/open-telemetry/opentelemetry-collector/blob/3c0fd3946f70a0b1fa97813c39dbc4d91d95afa6/pdata/internal/data/protogen/logs/v1/logs.pb.go#L223

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+			})
+		}
+	}
+
+	if len(resourceLogs) == 0 {
+		return nil
+	}
+
+	payload := map[string]any{
+		"resourceLogs": resourceLogs,
```
benhoyt: Why don't we make a top-level struct with `json` tags for this one too, instead of a map? For example:

```go
type payload struct {
    ResourceLogs []resourceLogs
}
```

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+
+// A collection of ScopeLogs from a Resource.
+// Refer to `type ResourceLogs struct` in
+// opentelemetry-collector/pdata/internal/data/protogen/logs/v1/logs.pb.go
+type ResourceLogs struct {
+	// The resource for the logs in this message.
+	// If this field is not set then resource info is unknown.
+	Resource Resource `json:"resource"`
+	// A list of ScopeLogs that originate from a resource.
+	ScopeLogs []*ScopeLogs `json:"scope_logs,omitempty"`
```
benhoyt: I notice that the official uses camelCase like `scopeLogs`, [for example](https://github.com/open-telemetry/opentelemetry-collector/blob/3c0fd3946f70a0b1fa97813c39dbc4d91d95afa6/pdata/plog/json_test.go#L61). You've used camelCase sometimes and snake_case others. So let's be consistent and use what they do, camelCase (the server seems to support both).

benhoyt: Also, in some places here you've used `[]*someStruct`, in others `[]someStruct`. Seeing these structs aren't too large, it's probably more efficient to use just `[]someStruct` for them all.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+// A collection of Logs produced by a Scope.
+// Refer to `type ScopeLogs struct` in
+// opentelemetry-collector/pdata/internal/data/protogen/logs/v1/logs.pb.go
+type ScopeLogs struct {
+	// The instrumentation scope information for the logs in this message.
+	// Semantically when InstrumentationScope isn't set, it is equivalent with
+	// an empty instrumentation scope name (unknown).
+	Scope Scope `json:"scope"`
+	// A list of log records.
+	LogRecords []*LogRecord `json:"log_records,omitempty"`
```
benhoyt: Similar here:
```suggestion
	LogRecords []LogRecord `json:"logRecords,omitempty"`
```

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+
+// A log record according to OpenTelemetry Log Data Model:
+// https://github.com/open-telemetry/oteps/blob/main/text/logs/0097-log-data-model.md
+// Refer to `type LogRecord struct` in
+// opentelemetry-collector/pdata/internal/data/protogen/logs/v1/logs.pb.go
+type LogRecord struct {
+	// time_unix_nano is the time when the event occurred.
+	// Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+	// Value of 0 indicates unknown or missing timestamp.
+	TimeUnixNano uint64 `json:"timeUnixNano"`
```
benhoyt: Interestingly, these are actually serialised as strings in the [source repo](https://github.com/open-telemetry/opentelemetry-collector/blob/3c0fd3946f70a0b1fa97813c39dbc4d91d95afa6/pdata/plog/json_test.go#L61):

```
var logsJSON = `..., "logRecords":[{"eventName":"someEvent","timeUnixNano":"1684617382541971000", ...`
```

I guess this is due to JSON issues with large numbers on some implementations (that use float64 for everything and lose precision). Should we follow that and output a string?

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+// resetBuffer drops all buffered logs (in the case of a successful send, or an unrecoverable error).
+func (c *Client) resetBuffer() {
+	// Zero removed elements to allow garbage collection.
+	for i := 0; i < len(c.entries); i++ {
+		c.entries[i] = otelEntryWithService{}
+	}
+	c.entries = c.buffer[:0]
+}
+
+type otelEntryWithService struct {
```
benhoyt: This package is "otel", so we can shorten this name:
```suggestion
type entryWithService struct {
```

benhoyt: This package is "otel", so we can shorten this name:
```suggestion
type entryWithService struct {
```

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	return nil
+}
+
+func encodeEntry(entry servicelog.Entry) LogRecord {
+	message := strings.TrimSuffix(entry.Message, "\n")
+
+	return LogRecord{
+		TimeUnixNano: uint64(entry.Time.UnixNano()),
+		Body:         AttributeValue{StringValue: &message},
+		Attributes:   []KeyValue{},
```
benhoyt: Presumably this can just be set to a nil slice (or better, omitted), instead of a zero-length slice? In fact, we can probably just drop the `Attributes` field from `LogRecord` unless/until we use it.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	for _, serviceName := range serviceNames {
+		batch := serviceBatches[serviceName]
+		if len(batch) > 0 {
+			resourceAttributes := c.resourceAttributes[serviceName]
+			resource := Resource{
+				Attributes: resourceAttributes,
+			}
+			scope := Scope{
+				Name:    "pebble",
+				Version: cmd.Version,
```
benhoyt: It seems a weird dependency loop for something in internal to depend on cmd. Plus, this will hardcode the current version (1.23.0-dev) in the tests, which is not good. How about having a `UserAgent` setting in the client options that we pass in (and set in tests)?

(I see that the loki package imports cmd. I think that's not great either, but we can change that in another PR if we want.)

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", fmt.Sprintf("pebble/%s", cmd.Version))
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return resp, fmt.Errorf("error sending logs: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 400 {
```
benhoyt: Isn't this already handled in `handleServerResponse`?

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	Body AttributeValue `json:"body"`
+	// Additional attributes that describe the specific event occurrence. [Optional].
+	// Attribute keys MUST be unique (it is not allowed to have more than one
+	// attribute with the same key).
+	Attributes []KeyValue `json:"attributes,omitempty"`
+}
+
+type Client struct {
+	options    *ClientOptions
+	target     *plan.LogTarget
```
benhoyt: I think we only use the Name and Location from here. I would prefer to have more explicit dependencies. How about we add `TargetName` and `Location` to the `ClientOptions` struct and get rid of this?

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	// To store log entries, keep a buffer of size 2*MaxRequestEntries with a
+	// sliding window "entries" of size MaxRequestEntries.
+	buffer  []otelEntryWithService
+	entries []otelEntryWithService
+
+	// Store the custom labels for each service (resource attributes in OTEL).
+	resourceAttributes map[string][]KeyValue
+}
+
+func NewClient(target *plan.LogTarget) *Client {
```
benhoyt: Especially now that we'll always need the options (given the comment above), let's drop `NewClientWithOptions` and just have `func NewClient(options *ClientOptions)`.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	return NewClientWithOptions(target, &ClientOptions{})
+}
+
+// ClientOptions allows overriding default parameters (e.g. for testing).
+type ClientOptions struct {
+	RequestTimeout    time.Duration
+	MaxRequestEntries int
+	BatchSize         int
```
benhoyt: Are we ever actually going to override these? If not, we could just drop them from the options and use the constants. Or do we override them for tests? Any we don't actually override let's drop from here and just use a const.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+func NewClientWithOptions(target *plan.LogTarget, options *ClientOptions) *Client {
+	options = fillDefaultOptions(options)
```
benhoyt: I would prefer we didn't modify the caller's options struct. Perhaps have the Client field be `options ClientOptions` and then:

```go
if options.RequestTimeout != 0 {
    c.options.RequestTimeout = options.RequestTimeout
}
```

And so on.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+		delete(c.resourceAttributes, serviceName)
+		return
+	}
+
+	// Convert attributes to KeyValue format.
+	keyValuePairs := make([]KeyValue, 0, len(attributes)+1)
+
+	// Add service.name attribute.
+	keyValuePairs = append(keyValuePairs, KeyValue{
+		Key:   "service.name",
```
benhoyt, line 174: This is called `pebble_service` in the Loki labels. Would it be good to be consistent with that, or is `service.name` a standard OpenTelemetry thing?

benhoyt, line 174: This is called `pebble_service` in the Loki labels. Would it be good to be consistent with that, or is `service.name` a standard OpenTelemetry thing?

benhoyt, line 174: This is called `pebble_service` in the Loki labels. Would it be good to be consistent with that, or is `service.name` a standard OpenTelemetry thing?

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,405 @@
...
+	// time_unix_nano is the time when the event occurred.
+	// Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
+	// Value of 0 indicates unknown or missing timestamp.
+	TimeUnixNano uint64 `json:"timeUnixNano"`
+	// The severity text (also known as log level). The original string representation as
+	// it is known at the source. [Optional].
+	SeverityText string `json:"severityText"`
+	// Numerical value of the severity, normalized to values described in Log Data Model.
+	// [Optional].
+	SeverityNumber int `json:"severityNumber"`
```
benhoyt: I think these two should be `,omitempty` so they don't appear in the JSON (unless they're set, which we don't now).

internals/overlord/logstate/opentelemetry/opentelemetry_test.go
```diff
@@ -0,0 +1,385 @@
...
+				"severityNumber": 0,
+				"body": {"stringValue": "log line #9"}
+			}]
+		}]
+	}
+]}`)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		c.Assert(r.Method, Equals, http.MethodPost)
+		c.Assert(r.Header.Get("Content-Type"), Equals, "application/json")
```
benhoyt, line 158: We should probably assert on the custom `User-Agent` too.

internals/overlord/logstate/opentelemetry/opentelemetry_test.go
```diff
@@ -0,0 +1,385 @@
...
+		c.Assert(err, IsNil)
+		c.Assert(string(reqBody), DeepEquals, string(expected))
+	}))
+	defer server.Close()
+
+	client := opentelemetry.NewClient(&plan.LogTarget{Location: server.URL})
+	client.SetLabels("svc1", map[string]string{})
+	client.SetLabels("svc2", map[string]string{})
+	client.SetLabels("svc3", map[string]string{})
+	client.SetLabels("svc4", map[string]string{})
```
benhoyt: I think if we're going to have a separate `TestLabels` we should drop these calls. But we could just call `SetLabels` with a couple of labels on (say) svc4 and combine the two tests. Most of it will be the same.

internals/overlord/logstate/opentelemetry/opentelemetry_test.go
```diff
@@ -0,0 +1,385 @@
...
+	for _, entry := range input {
+		err := client.Add(entry)
+		c.Assert(err, IsNil)
+	}
+
+	err := client.Flush(context.Background())
+	c.Assert(err, IsNil)
```
benhoyt, line 178: If the request goes through, we're asserting in the handler that it's the expected string, but we're not actually asserting that 1 request did go through. We should have a numRequests variable that's incremented in the handler and asserted on here.

benhoyt (changes requested):
Looks like a good start.

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: [Here](https://github.com/open-telemetry/opentelemetry-collector/blob/3c0fd3946f70a0b1fa97813c39dbc4d91d95afa6/pdata/internal/data/protogen/logs/v1/logs.pb.go#L229) it seems with protobuf, it's `json=scopeLogs` (camelCase) and with pure json, it's `json:"scope_logs"` (snake_case). It seems in the test it is different (same for other places like TimeUnixNano mentioned in another comment). As discussed, I also tried to figure it out using the official go SDK, but no matter how I configure it, it sends protobuf instead of JSON, so I could not figure it out. I think I will stick to the tests, meaning `camelCase`.

IronCore864: `[]*someStruct` v.s. `[]someStruct`: I tried to respect the original code, for example, in the original code it is `[]*ScopeLogs` and `[]KeyValue`, so I kept that. These structs won't be too large indeed, but I'm not sure if I should be consistent within our code, or consistent with the original code. Since for all the namings and json keys we stick with the original code, I think here we should also be consistent with the original code?

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: `otelEntryWithService` v.s. `entryWithService`: It is `lokiEntryWithService` in the loki package, so I think we should stick with existing naming convention. Or shall I rename the loki one to `entryWithService` as well? It is shorter.

benhoyt: Yeah, sure, we can rename the one in the Loki package too. Probably best in a separate PR though (and any other Loki changes you think we need to make there.)

internals/overlord/logstate/opentelemetry/opentelemetry.go
benhoyt: For this one I think it should be internally consistent, so let's keep the naming consistent with the original, but use `[]slice` everywhere here (the fields are only a subset too).

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: user-agent is already set in `req.Header`.

scope_name and scope_version are set by otel SDK, so here I was trying to set the same. After more investigation, it seems the otel Python SDK would set scope_name to "root" and "scope_version" to empty by default, so I think it's OK not to set the version here. I will remove the version, only leaving the name for the scope.

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: `service.name` is a standard OpenTelemetry thing. For example:

```json
{
  "body": "hello world log message",
  "id": "0hLbG3uuXKargNtNbgvROReXpjK",
  "timestamp": "2025-07-21T09:31:10.501084928Z",
  "attributes": {
    "otelTraceSampled": true,
    "code.line.number": 8,
    "code.file.path": "/Users/tiexin/work/test/main.py",
    "code.function.name": "hello_world",
    "otelServiceName": "unknown_service",
    "otelSpanID": "d92b6a1ba1d1b6d9",
    "otelTraceID": "eb37f86f263ee2f74de147770b76dd28"
  },
  "resources": {
    "service.name": "unknown_service",
    "telemetry.auto.version": "0.56b0",
    "telemetry.sdk.language": "python",
    "telemetry.sdk.name": "opentelemetry",
    "telemetry.sdk.version": "1.35.0"
  },
  "scope": {},
  "severity_text": "WARN",
  "severity_number": 13,
  "scope_name": "root",
  "scope_version": "",
  "span_id": "d92b6a1ba1d1b6d9",
  "trace_flags": 1,
  "trace_id": "eb37f86f263ee2f74de147770b76dd28"
}
```

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: fill/modify/return is preferred for two reasons:
1. `options.RequestTimeout` is also used in `httpClient`, not only in `Client.options`
2. It's the same logic as how the loki package does it.

internals/overlord/logstate/opentelemetry/opentelemetry_test.go
IronCore864: I will remove these calls and keep `TestLabels` because I found that one is flaky due to the order of the labels. I will add sorting on the labels to make it deterministic.

IronCore864:
Finished a round of refactoring according to review, a few comments left open so that it's easier to read the reasons why I didn't make those changes.

- [ ] TODO: manual test with otelcol/signoz to make sure the logs are still arriving at the target correctly.

docs/reference/layer-specification.md
```diff
@@ -253,12 +253,18 @@ log-targets:
...
+    # - opentelemetry: Use the OpenTelemetry protocol (OTLP). A "service.name"
+    # label is added automatically, with the name of the Pebble service as its
+    # value.
     type: loki
 
     # (Required) The URL of the remote log target.
     # For Loki, this needs to be the fully-qualified URL of the push API,
     # including the API endpoint, e.g.
     #     http://<ip-address>:3100/loki/api/v1/push
+    # For OpenTelemetry, this needs to use the HTTP port (the default value is
```
benhoyt: Nit:
```suggestion
    # For OpenTelemetry, this needs to include the TCP port (normally
```

docs/reference/layer-specification.md
```diff
@@ -253,12 +253,18 @@ log-targets:
...
+    # 4318) without the API endpoint, e.g.
```
benhoyt:
```suggestion
    # 4318) without the API endpoint, for example:
```

docs/reference/log-forwarding.md
```diff
@@ -21,8 +21,8 @@ log-targets:
 Required configuration:
 
 - `override`: How this log target definition is combined with other pre-existing definitions with the same name in the plan. Supported values are `merge` and `replace`.
-- `type`: The type of log target. The only supported type is `loki`.
-- `location`: The URL of the remote log target. For Loki, this needs to be the fully-qualified URL of the push API, including the API endpoint. Use the format `http://<ip-address>:3100/loki/api/v1/push`.
+- `type`: The type of log target. Supported types are `loki` and `opentelemetry`.
+- `location`: The URL of the remote log target. For Loki, this needs to be the fully-qualified URL of the push API, including the API endpoint; use the format `http://<ip-address>:3100/loki/api/v1/push`. For OpenTelemetry, use the HTTP port (default 4318) without the API endpoint, for example: `http://<ip-address>:4318`.
```
benhoyt:
```suggestion
- `location`: The URL of the remote log target. For Loki, this needs to be the fully-qualified URL of the push API, including the API endpoint; use the format `http://<ip-address>:3100/loki/api/v1/push`. For OpenTelemetry, include the TCP port (normally 4318) without the API endpoint, for example: `http://<ip-address>:4318`.
```

internals/overlord/logstate/opentelemetry/opentelemetry_test.go
```diff
@@ -83,117 +83,97 @@ func (*suite) TestRequest(c *C) {
 {"resourceLogs": [
 	{
 		"resource": {
-			"attributes": [{"key": "service.name", "value": {"stringValue": "svc1"}}]
+			"attributes": null
```
benhoyt, line 85: Why did this change to null? Shouldn't service.name always be there?

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -20,162 +20,140 @@ import (
...
-	Attributes []KeyValue `json:"attributes"`
+// https://github.com/open-telemetry/opentelemetry-collector/blob/3c0fd3946f70a0b1fa97813c39dbc4d91d95afa6/pdata/internal/data/protogen/resource/v1/resource.pb.go#L30
+type resource struct {
+	Attributes []keyValue `json:"attributes"`
 }
 
-type KeyValue struct {
-	Key   string         `json:"key"`
-	Value AttributeValue `json:"value"`
+// keyValue is a key-value pair that stores	attributes.
```
benhoyt: Nit:
```suggestion
// keyValue is a key-value pair that stores attributes.
```

internals/overlord/logstate/opentelemetry/opentelemetry.go
benhoyt: I don't mind the existence of `fillDefaultOptions`, but it seems quite bad practice to modify the caller's struct (it's also bad that the `loki` package does this -- we should change this).

So how about we make a copy instead and fill the copy? Something like this:

```go
opts := *options // make a copy of caller's options
fillDefaultOptions(&opts) // note: no need for this to return a value
...
```

Same effect, but we haven't modified the caller's struct.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -264,37 +253,34 @@ func (c *Client) Flush(ctx context.Context) error {
 	// Sort service names to ensure deterministic order.
 	sort.Strings(serviceNames)
 
-	resourceLogs := make([]ResourceLogs, 0, len(serviceNames))
+	logs := make([]resourceLogs, 0, len(serviceNames))
 	for _, serviceName := range serviceNames {
 		batch := serviceBatches[serviceName]
 		if len(batch) > 0 {
```
benhoyt: Nit: it would avoid an extra nesting level to use an early-continue here: `if len(batch) == 0 { continue }`

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -264,37 +253,34 @@ func (c *Client) Flush(ctx context.Context) error {
...
+			resource := resource{
 				Attributes: resourceAttributes,
 			}
-			scope := Scope{
-				Name:    "pebble",
-				Version: cmd.Version,
-			}
-			scopeLogs := []*ScopeLogs{
+			scope := instrumentationScope{Name: "pebble"}
+			scopeLogs := []scopeLogs{
```
benhoyt: Nit: let's use cuddled braces here.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -305,13 +291,13 @@ func (c *Client) Flush(ctx context.Context) error {
...
 
-func (c *Client) sendBatch(ctx context.Context, payload map[string]any) (*http.Response, error) {
+func (c *Client) sendBatch(ctx context.Context, payload payload) (*http.Response, error) {
 	jsonData, err := json.Marshal(payload)
 	if err != nil {
 		return nil, fmt.Errorf("error marshaling log batch: %v", err)
 	}
 
-	req, err := http.NewRequestWithContext(ctx, "POST", c.target.Location+"/v1/logs", bytes.NewBuffer(jsonData))
+	req, err := http.NewRequestWithContext(ctx, "POST", c.options.Location+"/v1/logs", bytes.NewBuffer(jsonData))
```
benhoyt, line 290: We should probably make this match the loki config and have the user include `/v1/logs` in the plan/config location. Slightly annoying, but gives them more flexibility if it's somehow at a different path, and matches the Loki "location" option.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling log batch: %v", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", c.options.Location+"/v1/logs", bytes.NewBuffer(jsonData))
+	if err != nil {
+		return nil, fmt.Errorf("error creating request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", fmt.Sprintf("pebble/%s", cmd.Version))
```
benhoyt: Per earlier discussion, let's avoid the weird dependency from internals to cmd (and avoid hardcoding "pebble" here). Instead, let's add a `UserAgent string` field to the `ClientOptions` struct and have the caller set that to `cmd.ProgramName + "/" + cmd.Version`. An empty string could either mean "don't set User-Agent" or we could default to "pebble".

benhoyt (commented):
Thanks for the updates. Pretty close now I think. A few more minor comments and a couple of suggestions.

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: I kindly disagree, because Loki and OpenTelemetry are two different things, and we should follow their conventions respectively, not matching them.

In Loki, the official documentation suggests that the URL includes the endpoint "/loki/api/v1/push", as seen in [here](https://grafana.com/docs/loki/latest/send-data/promtail/configuration/#clients) and [here](https://grafana.com/docs/alloy/latest/tutorials/send-logs-to-loki/). Whereas in OpenTelemetry, the convention is without, see [here](https://opentelemetry.io/docs/collector/configuration/).

The users would be confused to have different URLs between their Pebble config and other manual/local/automation. On the other hand, it's less likely to be confused when they switch from Loki to OpenTelemetry and have to update the config, because that is expected, since they switched to a new tool.

internals/overlord/logstate/opentelemetry/opentelemetry.go
IronCore864: Got it, thanks, I did not fully get the previous comment on the user agent...

internals/overlord/logstate/opentelemetry/opentelemetry_test.go
IronCore864: Because in the previous comment, you mentioned that since we already have a dedicated test that tests `SetLabels`, I removed it from here, and the default label `service.name` is set in `SetLabels`, same as the Loki package. In `gatherer`, this won't be a problem, since when the plan is changed, it always sets labels.

internals/overlord/logstate/opentelemetry/opentelemetry.go
benhoyt: I think that's fair, thanks.

dimaqq:
Please take a look at https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding if you haven't already, there are a few gotchas in the OTLP format:
- [ ] traceid/spanid (the only randomly generated id): hex strings without `0x`, typically padded with zeros to exact size by implementations.
- [ ] enum values: integers
- [ ] JSON object keys: `lowerCamelCase`
- [ ] large integers: decimal strings preferred, ints accepted
- [ ] Content-Type: application/json

Examples are here: https://github.com/open-telemetry/opentelemetry-proto/blob/v1.7.0/examples/README.md

IIRC there's not much difference between OTLP 1.5.0 and 1.7.0, but please check, as 1.7.0 is relatively new.

internals/overlord/logstate/opentelemetry/opentelemetry.go
benhoyt: This still hasn't been done. Please do a similar copy-then-fill change here to what you did in `gatherer.go`, to avoid modifying the caller's struct.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,381 @@
...
+		service: entry.Service,
+	})
+	return nil
+}
+
+func encodeEntry(entry servicelog.Entry) logRecord {
+	message := strings.TrimSuffix(entry.Message, "\n")
+
+	return logRecord{
+		TimeUnixNano: fmt.Sprintf("%d", entry.Time.UnixNano()),
```
benhoyt: It's a bit more efficient to use the more specific `strconv.FormatInt(entry.Time.UnixNano(), 10)` function that the generalized `fmt.Sprintf`.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,381 @@
...
+		batch := serviceBatches[serviceName]
+		if len(batch) == 0 {
+			continue
+		}
+
+		resourceAttributes := c.resourceAttributes[serviceName]
+		resource := resource{
+			Attributes: resourceAttributes,
+		}
+		scope := instrumentationScope{Name: "pebble"}
```
benhoyt: Similar as with user agent -- let's add `ScopeName string` to `ClientOptions` and pass this in from above and `cmd.ProgramName`, rather than hardcode it here.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,381 @@
...
+		return err
+	}
+
+	return c.handleServerResponse(resp)
+}
+
+func (c *Client) sendBatch(ctx context.Context, payload payload) (*http.Response, error) {
+	jsonData, err := json.Marshal(payload)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling log batch: %v", err)
```
benhoyt: Let's change to the canonical error format with "cannot", eg: `cannot marshal log batch: %v`. Similar elsewhere in this function.

IronCore864:
On the `fillDefaultOptions` issue, I made an error and changed the wrong place. Reverted the unnecessary change and pushed a new commit to refactor according to the latest review. Now it's ready.

benhoyt (approved):
Cool, thank you! Let's get Harry's external eyes on it before merging. I'll ask him.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,383 @@
...
+			Key:   k,
+			Value: anyValue{StringValue: &v},
+		})
+	}
+
+	c.resourceAttributes[serviceName] = keyValuePairs
+}
+
+func (c *Client) Add(entry servicelog.Entry) error {
+	if n := len(c.entries); n >= c.options.MaxRequestEntries {
```
hpidcock: n declaration and assignment is superfluous

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,383 @@
...
+		}}
+		logs = append(logs, resourceLogs{
+			Resource:  resource,
+			ScopeLogs: scopeLogs,
+		})
+	}
+
+	if len(logs) == 0 {
+		return nil
+	}
```
hpidcock: This is not possible.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,383 @@
...
+	// Sort service names to ensure deterministic order.
+	sort.Strings(serviceNames)
+
+	logs := make([]resourceLogs, 0, len(serviceNames))
+	for _, serviceName := range serviceNames {
+		batch := serviceBatches[serviceName]
+		if len(batch) == 0 {
+			continue
+		}
```
hpidcock: This is not possible.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,383 @@
...
+		return nil, fmt.Errorf("cannot create request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", c.options.UserAgent)
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return resp, fmt.Errorf("cannot send logs: %v", err)
+	}
+	defer resp.Body.Close()
```
hpidcock: Why are you closing the body here? `handleServerResponse` does it.

internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,383 @@
...
+
+	resp, err := c.sendBatch(ctx, payload)
+	if err != nil {
+		return err
+	}
+
+	return c.handleServerResponse(resp)
+}
+
+func (c *Client) sendBatch(ctx context.Context, payload payload) (*http.Response, error) {
```
hpidcock: Don't return http response. Handle everything here.

IronCore864: I referred to the existing Loki package to follow the same logic, but after thinking about it, I agree that a better design is that `sendBatch` should call `handleServerResponse` internally and return only an error, because this would:

- eliminate the body close defer issue
- keep response handling encapsulated within this function
- simplify the API/signature

I will refactor accordingly, thanks for pointing this out :)

PR #666, opened 2025-07-30
Security event logging

benhoyt:
DRAFT

### DIFF
--- a/internals/logger/logger_test.go
+++ b/internals/logger/logger_test.go
@@ -65,17 +65,33 @@ func (s *LogSuite) TestDebugfEnv(c *C) {
 
 func (s *LogSuite) TestNoticef(c *C) {
 	logger.Noticef("xyzzy")
-	c.Check(s.logbuf.String(), Matches, `20\d\d-\d\d-\d\dT\d\d:\d\d:\d\d.\d\d\dZ PREFIX: xyzzy\n`)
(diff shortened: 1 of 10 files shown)


internals/logger/logger_test.go
```diff
@@ -65,17 +65,33 @@ func (s *LogSuite) TestDebugfEnv(c *C) {
 
 func (s *LogSuite) TestNoticef(c *C) {
 	logger.Noticef("xyzzy")
-	c.Check(s.logbuf.String(), Matches, `20\d\d-\d\d-\d\dT\d\d:\d\d:\d\d.\d\d\dZ PREFIX: xyzzy\n`)
+	c.Check(s.logbuf.String(), Matches, `2\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d\.\d\d\dZ PREFIX: xyzzy\n`)
```
benhoyt, line 68: Driveby to fix y2k1 bugs. :-)

IronCore864 (approved):

tonyandrewmeyer (approved):
I noticed that you don't get an add-layer event when Pebble adds layers on startup. I think this is probably fine: it's implied by the startup event, so if I was trying to figure out how a layer was added if I didn't see a specific event for it then I'd assume it was on startup. I mention it just in case you hadn't considered it and wanted a different behaviour.

No other comments: everything looks good and works locally for me.

Issue #634, opened 2025-06-13
Check failure details don't show in 'pebble checks' when failures==threshold


tonyandrewmeyer:
When the failure count is exactly the threshold, `pebble checks` doesn't provide information about the failure.

```
$ PEBBLE=/tmp go run ./cmd/pebble checks
Check  Level  Startup  Status  Failures  Change
chk1   -      enabled  up      0/3       12
$ PEBBLE=/tmp go run ./cmd/pebble checks
Check  Level  Startup  Status  Failures  Change
chk1   -      enabled  up      1/3       12 (exit status 2; ls: cannot access '/t... run "pebble tasks 12" for more)
$ PEBBLE=/tmp go run ./cmd/pebble checks
Check  Level  Startup  Status  Failures  Change
chk1   -      enabled  up      2/3       12 (exit status 2; ls: cannot access '/t... run "pebble tasks 12" for more)
$ PEBBLE=/tmp go run ./cmd/pebble checks
Check  Level  Startup  Status  Failures  Change
chk1   -      enabled  down    3/3       13
$ PEBBLE=/tmp go run ./cmd/pebble checks
Check  Level  Startup  Status  Failures  Change
chk1   -      enabled  down    4/3       13 (exit status 2; ls: cannot access '/t... run "pebble tasks 13" for more)
```

@benhoyt pointed out that it's happening because it's switched to the new change ID (13) but there are no failures on that change yet (they're still on 12).

It would be nice if we could improve this somehow.

IronCore864:
I did some initial research, it seems [here](https://github.com/canonical/pebble/blob/master/internals/overlord/checkstate/request.go#L56) a new task and a new change are created.

How about:

- We pass in the previous change.
- Get the final task of the previous change.
- Get the final log line of the final task from the previous change.
- Set it to the new task's logs?

For example:

```go
func recoverCheckChange(st *state.State, config *plan.Check, successes, failures int, previousChange *state.Change) (changeID string) {
	summary := fmt.Sprintf("Recover %s check %q", checkType(config), config.Name)
	task := st.NewTask(recoverCheckKind, summary)
	task.Set(checkDetailsAttr, &checkDetails{
		Name:      config.Name,
		Successes: successes,
		Failures:  failures,
	})

	tasks := previousChange.Tasks()
	previousTaskLogs := tasks[len(tasks)-1].Log()
	if len(previousTaskLogs) > 0 {
		lastLogLine := logs[len(previousTaskLogs)-1]
[...]

PR #668, opened 2025-08-01
fix: check error log when failure is at threshold

IronCore864:
Fix an issue where the error logs of a check can't be displayed properly when the check failures count is at the threshold.

Closes https://github.com/canonical/pebble/issues/634.

### DIFF
--- a/client/changes.go
+++ b/client/changes.go
@@ -26,13 +26,14 @@ import (
 
 // Change is a modification to the system state.
 type Change struct {
-	ID      string  `json:"id"`
-	Kind    string  `json:"kind"`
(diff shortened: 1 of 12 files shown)


benhoyt:
I've had a look over this, and I don't hate it, but on reflection, I'm not sure it's much cleaner than the "just copy over the last log" approach you mentioned [here](https://github.com/canonical/pebble/issues/634#issuecomment-3116191335). I'm a bit unsure what's best here -- I'm going to ask Gustavo what he thinks of the two approaches and if he has any better ideas. I'll report back on #634.

PR #669, opened 2025-08-02
chore: update TIOBE dependencies to support Go 1.24

tonyandrewmeyer:
Our TIOBE contact let us know that we need to update the staticcheck tool that the workflow uses, to support Go 1.24.

In order to have this work in CI, also bump the version of Go used in the unit and integration tests to 1.24, by pulling the version from `go.mod`. This removes the Go version matrix testing in the tests, but we only had a single version being tested, so the matrix wasn't really used.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    name: Install Go
+    name: Unit tests
     steps:
     - uses: actions/checkout@v3
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,19 +10,14 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-      matrix:
-        go: ['1.22']
-
-    name: Go ${{ matrix.go }}
+    name: Install Go
     steps:
     - uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer:
Our go.mod says 1.24, but our test matrix installs 1.22. @benhoyt should they be the same, or am I missing something?

benhoyt:
> Our go.mod says 1.24, but our test matrix installs 1.22. @benhoyt should they be the same, or am I missing something?

I think they should be the same. We might be able to use `go-version-file: go.mod` (or whatever the syntax is) to avoid hardcoding a version there.

benhoyt (approved):
Thanks!

tonyandrewmeyer:
@benhoyt I've changed the CI to use go.mod - are you still ok with this? And do you know what the story is with the "Go 1.22" check being expected? Is that listed in the repo settings as a required check under that exact name? Or just expected because it's there in master?

.github/workflows/tests.yml
```diff
@@ -10,19 +10,14 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-      matrix:
-        go: ['1.22']
-
-    name: Go ${{ matrix.go }}
+    name: Install Go
```
benhoyt: I think this is incorrect -- it should be "Unit tests" or similar (it includes the main "Test and build" step).

Issue #586, opened 2025-03-12
Suspicious errors reported during pebble exec termination


flotter:
## 1. Run the pebble server
```sh
export PEBBLE_DEBUG=1
sudo -E ./pebble run --create-dirs
```

## 2. Start an exec session
```sh
sudo ./pebble exec busybox ls /
```

## 3. Inspect the pebble daemon output

```
2025-03-12T06:27:37.428Z [pebble] DEBUG Exec 6: started mirroring websocket
2025-03-12T06:27:37.429Z [pebble] DEBUG Exec 6: control handler started for PID 260895
2025-03-12T06:27:37.431Z [pebble] DEBUG Reaper received SIGCHLD.
2025-03-12T06:27:37.431Z [pebble] DEBUG Reaped PID 260895 which exited with code 0.
2025-03-12T06:27:37.431Z [pebble] DEBUG Detected poll(POLLHUP) event: exiting.
2025-03-12T06:27:37.431Z [pebble] DEBUG Exec 6: cannot get next websocket reader for PID 260895: read unix /var/lib/pebble/default/.pebble.socket->@: use of closed network connection
2025-03-12T06:27:37.431Z [pebble] DEBUG Exec 6: control handler finished
2025-03-12T06:27:37.431Z [pebble] DEBUG Sending write barrier
2025-03-12T06:27:37.431Z [pebble] DEBUG Exec 6: finished mirroring websocket
2025-03-12T06:27:37.432Z [pebble] DEBUG Detected poll(POLLNVAL) event.
2025-03-12T06:27:37.432Z [pebble] DEBUG Cannot get next reader: websocket: close 1006 (abnormal closure): unexpected EOF
2025-03-12T06:27:37.440Z [pebble] DEBUG pid=260889;uid=0;socket=/var/lib/pebble/default/.pebble.socket; GET /v1/changes/6/wait 10.822537ms 200
2025-03-12T06:27:37.440Z [pebble] GET /v1/changes/6/wait 10.822537ms 200
```

There are two debug messages that worries me:

1. ```cannot get next websocket reader for PID 260895: read unix /var/lib/pebble/default/.pebble.socket->@: use of closed network connection```

2. ```Cannot get next reader: websocket: close 1006 (abnormal closure): unexpected EOF```

Everything still appears to work as expected, but I am observing this with every websocket / exec use case and I do not know when this is going to bite us.

benhoyt:
[...]

PR #670, opened 2025-08-04
fix(cmdstate,wsutil): don't report websocket close errors

anpep:
This issue is related to https://github.com/canonical/pebble/issues/586

When the websocket connection is closed from the client side, websocket close errors are sometimes incorrectly reported:

```
2025-03-12T06:27:37.428Z [pebble] DEBUG Exec 6: started mirroring websocket
2025-03-12T06:27:37.429Z [pebble] DEBUG Exec 6: control handler started for PID 260895
2025-03-12T06:27:37.431Z [pebble] DEBUG Reaper received SIGCHLD.
2025-03-12T06:27:37.431Z [pebble] DEBUG Reaped PID 260895 which exited with code 0.
2025-03-12T06:27:37.431Z [pebble] DEBUG Detected poll(POLLHUP) event: exiting.
>>>>> 2025-03-12T06:27:37.431Z [pebble] DEBUG Exec 6: cannot get next websocket reader for PID 260895: read unix /var/lib/pebble/default/.pebble.socket->@: use of closed network connection
2025-03-12T06:27:37.431Z [pebble] DEBUG Exec 6: control handler finished
2025-03-12T06:27:37.431Z [pebble] DEBUG Sending write barrier
2025-03-12T06:27:37.431Z [pebble] DEBUG Exec 6: finished mirroring websocket
2025-03-12T06:27:37.432Z [pebble] DEBUG Detected poll(POLLNVAL) event.
>>>>> 2025-03-12T06:27:37.432Z [pebble] DEBUG Cannot get next reader: websocket: close 1006 (abnormal closure): unexpected EOF
2025-03-12T06:27:37.440Z [pebble] DEBUG pid=260889;uid=0;socket=/var/lib/pebble/default/.pebble.socket; GET /v1/changes/6/wait 10.822537ms 200
2025-03-12T06:27:37.440Z [pebble] GET /v1/changes/6/wait 10.822537ms 200
```

This patch applies better error handling around connection closure so no bogus errors end up in the logs.

### DIFF
--- a/internals/overlord/cmdstate/handlers.go
+++ b/internals/overlord/cmdstate/handlers.go
@@ -475,10 +475,11 @@ func (e *execution) controlLoop(execID string, pidCh <-chan int, stop <-chan str
 		}
 
 		if err != nil {
-			if websocket.IsUnexpectedCloseError(err, websocket.CloseAbnormalClosure, websocket.CloseGoingAway) {
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway) {
 				logger.Debugf("Exec %s: cannot get next websocket reader for PID %d: %v", execID, pid, err)
+			}
 
-				// If an abnormal closure occurred, kill the attached process.
+			if websocket.IsCloseError(err, websocket.CloseAbnormalClosure) {
 				err := unix.Kill(pid, unix.SIGKILL)
 				if err != nil {
 					logger.Noticef("Exec %s: cannot send SIGKILL to pid %d: %v", execID, pid, err)
@@ -486,6 +487,7 @@ func (e *execution) controlLoop(execID string, pidCh <-chan int, stop <-chan str
 					logger.Debugf("Exec %s: sent SIGKILL to pid %d", execID, pid)
 				}
 			}
+
 			break
 		}
 
--- a/internals/overlord/cmdstate/handlers.go
+++ b/internals/overlord/cmdstate/handlers.go
@@ -469,27 +469,17 @@ func (e *execution) controlLoop(execID string, pidCh <-chan int, stop <-chan str
 	logger.Debugf("Exec %s: control handler started for PID %d", execID, pid)
 	for {
 		controlConn := e.getWebsocket(wsControl)
-		mt, r, err := controlConn.NextReader()
-		if mt == websocket.CloseMessage {
-			break
-		}
-
+		_, r, err := controlConn.NextReader()
 		if err != nil {
-			logger.Debugf("Exec %s: cannot get next websocket reader for PID %d: %v", execID, pid, err)
-			er, ok := err.(*websocket.CloseError)
-			if !ok {
-				break
-			}
-			if er.Code != websocket.CloseAbnormalClosure {
-				break
-			}
-
-			// If an abnormal closure occurred, kill the attached process.
-			err := unix.Kill(pid, unix.SIGKILL)
-			if err != nil {
-				logger.Noticef("Exec %s: cannot send SIGKILL to pid %d: %v", execID, pid, err)
-			} else {
-				logger.Debugf("Exec %s: sent SIGKILL to pid %d", execID, pid)
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseAbnormalClosure, websocket.CloseGoingAway) {
+				logger.Debugf("Exec %s: cannot get next websocket reader for PID %d: %v", execID, pid, err)
+				// If an abnormal closure occurred, kill the attached process.
+				err := unix.Kill(pid, unix.SIGKILL)
+				if err != nil {
+					logger.Noticef("Exec %s: cannot send SIGKILL to pid %d: %v", execID, pid, err)
+				} else {
+					logger.Debugf("Exec %s: sent SIGKILL to pid %d", execID, pid)
+				}
 			}
 			break
 		}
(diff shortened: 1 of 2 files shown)


internals/wsutil/wsutil.go
```diff
@@ -88,7 +88,9 @@ func recvLoop(w io.Writer, conn MessageReader) {
 	for {
 		mt, r, err := conn.NextReader()
 		if err != nil {
-			logger.Debugf("Cannot get next reader: %v", err)
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseAbnormalClosure, websocket.CloseGoingAway) {
```
benhoyt: In practice this error seems to always be `AbnormalClosure`, not `CloseGoingAway`. What was your motivation for allowing both of those here?

internals/overlord/cmdstate/handlers.go
```diff
@@ -475,21 +475,16 @@ func (e *execution) controlLoop(execID string, pidCh <-chan int, stop <-chan str
...
-			}
-
-			// If an abnormal closure occurred, kill the attached process.
-			err := unix.Kill(pid, unix.SIGKILL)
-			if err != nil {
-				logger.Noticef("Exec %s: cannot send SIGKILL to pid %d: %v", execID, pid, err)
-			} else {
-				logger.Debugf("Exec %s: sent SIGKILL to pid %d", execID, pid)
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseAbnormalClosure, websocket.CloseGoingAway) {
+				logger.Debugf("Exec %s: cannot get next websocket reader for PID %d: %v", execID, pid, err)
```
benhoyt, line 479: In my testing, when this error is printed out it's actually a `*websocket.netError`, not a `*websocket.CloseError` ("read unix /var/lib/pebble/default/.pebble.socket->@: use of closed network connection").

So is this actually the logic we want?

internals/overlord/cmdstate/handlers.go
```diff
@@ -475,21 +475,16 @@ func (e *execution) controlLoop(execID string, pidCh <-chan int, stop <-chan str
...
+
+				// If an abnormal closure occurred, kill the attached process.
```
benhoyt: I believe this logic is now incorrect/changed: now the logic is saying "if it's a CloseError other than AbnormalClosure or CloseGoingAway, kill the process". Whereas previously it was saying "if it's an AbnormalClosure, kill the process".

internals/wsutil/wsutil.go
anpep: I've researched and `AbnormalClosure` is set by the library when the connection is closed abruptly without sending a "close" frame, and thought `GoingAway` was sent by both the client and a server that is exiting, however the latter does not apply here, and I left it while testing. Thanks!

IronCore864:
I did some research and tested this PR, while the provided solution does depress the two warning logs ("cannot get next websocket reader for PID" and "Cannot get next reader"), I'm not certain this is the right way, because according to the [official doc here](https://pkg.go.dev/github.com/gorilla/websocket#IsUnexpectedCloseError), it seems we should not exclude "CloseAbnormalClosure":

> The only expected close message from the client has the code websocket.CloseGoingAway. All other close messages are likely the result of an application or protocol error and are logged to aid debugging.

There is also an example in the doc where it only excludes `CloseGoingAway`.

I believe the logic in the handler should be something like:

- If the error is not `CloseGoingAway`, log "cannot get next websocket reader for PID".
- If the error is `CloseAbnormalClosure`, kill the attached process.
- Else: break.

For example:

```go
if err != nil {
	// The only expected close message from the client has the code websocket.CloseGoingAway.
	// All other close messages are likely the result of an application or protocol error and are logged to aid debugging.
	if websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway) {
		logger.Debugf("Exec %s: cannot get next websocket reader for PID %d: %v", execID, pid, err)
	}

	if websocket.IsCloseError(err, websocket.CloseAbnormalClosure) {
		err := unix.Kill(pid, unix.SIGKILL)
		if err != nil {
			logger.Noticef("Exec %s: cannot send SIGKILL to pid %d: %v", execID, pid, err)
		} else {
			logger.Debugf("Exec %s: sent SIGKILL to pid %d", execID, pid)
		}
	}

	break
}
```

This suppresses the error log "cannot get next websocket reader for PID".

But if we do the same, only whitelisting `CloseGoingAway` in `wsutil.recvLoop`:

```go
for {
	mt, r, err := conn.NextReader()
	if err != nil && websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway) {
		logger.Debugf("Cannot get next reader: %v", err)
		return
	}

	// ...
}
```

The "Cannot get next reader" error log is still there, and this might suggest there is something wrong in the code. I tried to debug this, but without success. The client-side connection closes without error, but the server-side still reports `CloseAbnormalClosure`. Maybe it's OK to whitelist `CloseAbnormalClosure` here? As I find in a few examples which do so, see [here](https://superdev.school/en/blogs/go-error-handling-logging-websocket) and [here](https://webreference.com/go/web/websockets/).

internals/wsutil/wsutil.go
benhoyt: @anpep Did you intend to update this?

anpep: Yes, fixed this now

benhoyt (approved):
Thanks for this fix!

PR #672, opened 2025-08-06
servstate: Shutdown logs after service shutdown

brancz:
Shutting down loggers is required because otherwise service loggers and the log gathering infrastructure do not know to stop pulling, and will only shutdown after a hard-shutdown timeout.


benhoyt:
Thanks! I think I'll get a chance to review this issue and the fix tomorrow.

brancz:
Good that this is so well exercised in tests. I won't be able to find a fix for this today, but the fundamental problem is clear, the ringbuffer never closes on shutdown. Where/how the right way to go about resolving that has yet to be figured out, looks like it's more intricate than initially thought.

PR #673, opened 2025-08-08
Improve `servicelog.Parse` performance

brancz:
This PR contains two improvements to parsing servicelogs:

1. 1436453333ea8e6ff3e7def8466c8eed3b6d8ab3 Use a parser approach instead of `bytes.SplitN`, which allocates a slice of length N for every log line parsed. This new approach is allocation-free.
2. 6a3258bf4095e63d6b1c5b0b42a15dc70907c1f6 Use a specialized parser for parsing dates. Because the timestamp format being used is non-standard (eg. not RFC3339) the Go compiler doesn't have an optimized version to use, so it falls back to analyzing the time format on every parse.

These two changes ultimately result in a ~71% improvment for every log line parsed.

```
$ benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/canonical/pebble/internals/servicelog
cpu: Apple M1 Max
          │   old.txt    │               new.txt               │
          │    sec/op    │   sec/op     vs base                │
Parser-10   193.25n ± 2%   54.70n ± 1%  -71.69% (p=0.000 n=10)
```

After these changes, the biggest improvement left is the two string copies that turn the service name and message content into strings. It's possible that the byte slices could be safely passed around directly as part of the `servicelog.Entry` struct, but at first sight, it appears that it could produce unsafe behavior because the `Entry` could be processed in one goroutine while the iterator goroutine already proceeds to the next log line, which reuses the buffer.


benhoyt (commented):
Thanks for this. It's quite a lot of fairly fiddly code. I think the concept is a good idea, but I think we can simplify a little. I've played around with this locally, and have come up with a simpler (and slightly faster) approach that validates the whole thing, and then converts decimal to binary more directly:

```go
func parseTime(b []byte) (time.Time, error) {
	valid := len(b) == 24 &&
		isDigit(b[0]) && isDigit(b[1]) && isDigit(b[2]) && isDigit(b[3]) &&
		b[4] == '-' &&
		isDigit(b[5]) && isDigit(b[6]) &&
		b[7] == '-' &&
		isDigit(b[8]) && isDigit(b[9]) &&
		b[10] == 'T' &&
		isDigit(b[11]) && isDigit(b[12]) &&
		b[13] == ':' &&
		isDigit(b[14]) && isDigit(b[15]) &&
		b[16] == ':' &&
		isDigit(b[17]) && isDigit(b[18]) &&
		b[19] == '.' &&
		isDigit(b[20]) && isDigit(b[21]) && isDigit(b[22]) &&
		b[23] == 'Z'
	if !valid {
		return time.Time{}, fmt.Errorf("invalid datetime format: expected timestamp, got %q", b)
	}
	year := int(b[0]-'0')*1000 + int(b[1]-'0')*100 + int(b[2]-'0')*10 + int(b[3]-'0')
	month := int(b[5]-'0')*10 + int(b[6]-'0')
	day := int(b[8]-'0')*10 + int(b[9]-'0')
	hour := int(b[11]-'0')*10 + int(b[12]-'0')
	minute := int(b[14]-'0')*10 + int(b[15]-'0')
	second := int(b[17]-'0')*10 + int(b[18]-'0')
	millisecond := int(b[20]-'0')*100 + int(b[21]-'0')*10 + int(b[22]-'0')

	return time.Date(year, time.Month(month), day, hour, minute, second,
		millisecond*1_000_1000, time.UTC), nil
```

This does change the functionality: it only supports exactly 3 digits for the sub-second part (milliseconds). I think this is okay, though, as it's only for parsing the logs we've formatted into the ring buffer (with 3 digit sub-seconds). Maybe @hpidcock can weigh in too.

I don't think we can avoid an allocation+copy for converting the message bytes to a string (the ring buffer might be overwritten before the Entry.Message is used). However, there's only a limited number of service names, so we could potentially intern those with a mutex+map:

```go
var internPool map[string]string = make(map[string]string)
var internMutex sync.Mutex

// Intern returns a string that is equal to s but that may share storage with
// a string previously passed to Intern.
func Intern(s string) string {
	internMutex.Lock()
	defer internMutex.Unlock()

	pooled, ok := internPool[s]
	if !ok {
		// Clone the string in case it's part of some much bigger string.
		// This should be rare, if interning is being used well.
		pooled = strings.Clone(s)
		internPool[pooled] = pooled
	}
	return pooled
}
```

With my updates `parseTime` across the various versions, here's what I'm seeing (the NoSplit version is avoiding the `bytes.SplitN` but still using `time.Parse`):

```
$ go test -run=^$ -bench=Parse -benchmem ./internals/servicelog/
goos: linux
goarch: amd64
pkg: github.com/canonical/pebble/internals/servicelog
cpu: Intel(R) Core(TM) Ultra 7 255H
BenchmarkParseOriginal-16         	 8129140	       146.1 ns/op	      86 B/op	       3 allocs/op
BenchmarkParseNoSplit-16          	11268946	       100.1 ns/op	       6 B/op	       2 allocs/op
BenchmarkParsePolarSignals-16     	46367496	        25.31 ns/op	       6 B/op	       2 allocs/op
BenchmarkParseInternService-16    	33363584	        34.34 ns/op	       3 B/op	       1 allocs/op
```

So the string-interning version is slightly slower, but does one fewer allocation.

I plan to get Harry's take and then spend some more *time* on this next week.

PR #674, opened 2025-08-08
perf: improve performance of serializing log timestamps

brancz:
Because the timestamp format used is not exactly RFC3339, the Go
standard library falls back to generic formatting code that analyzes the
format on every append, this causes a lot of unnecessary work. Using a
specialized formatter specifically for the format, a ~72% improvement
is achieved.

```
$ benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/canonical/pebble/internals/servicelog
cpu: Apple M1 Max
                   │   old.txt    │               new.txt               │
                   │    sec/op    │   sec/op     vs base                │
AppendTimestamp-10   120.55n ± 0%   33.27n ± 1%  -72.41% (p=0.000 n=10)
```

After this, the only improvement that could still be made is assuming that it's always UTC, which we already know, profililing data shows that could at most improve another ~12% (`time.(*Location).get`).

https://pprof.me/da1dac7424551de8264abc96c430f319/?profileType=profile%3Acpu%3Ananoseconds%3Acpu%3Ananoseconds%3Adelta&color_by=filename&dashboard_items=flamegraph%2Ctable&table_columns=color%2Cflat%2CflatPercentage%2Ccumulative%2Cname

### DIFF
--- a/internals/servicelog/formatter.go
+++ b/internals/servicelog/formatter.go
@@ -108,7 +108,7 @@ func appendTimestamp(b []byte, t time.Time) []byte {
 	b[19] = '.'
 
 	// Write milliseconds (3 digits)
-	b[20] = byte('0' + millisecond/100%10)
+	b[20] = byte('0' + millisecond/100) // millisecond is at most 999, so no need for %10 here
 	b[21] = byte('0' + millisecond/10%10)
 	b[22] = byte('0' + millisecond%10)
 	b[23] = 'Z'
--- a/internals/servicelog/formatter.go
+++ b/internals/servicelog/formatter.go
@@ -55,6 +55,10 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
 	}
 }
 
+func appendTimestamp(buf []byte, t time.Time) []byte {
+	return t.UTC().AppendFormat(buf, outputTimeFormat)
+}
+
 func (f *formatter) Write(p []byte) (nn int, ee error) {
 	f.mut.Lock()
 	defer f.mut.Unlock()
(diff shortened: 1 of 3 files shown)


benhoyt:
Thanks for this @brancz. I've looked over it and done some further tests. I get a similar result on my amd64 machine (-67.35% vs base). I've also tweaked some comments, used `slices.Grow` (new in Go 1.21) instead of a hand-rolled version, and added some more explicit tests.

benhoyt (approved):

benhoyt:
I'll get one more pair of eyes on this and then we can merge.

internals/servicelog/formatter.go
```diff
@@ -55,14 +51,79 @@ func NewFormatWriter(dest io.Writer, serviceName string) io.Writer {
...
+	b[15] = byte('0' + minute%10)
+	b[16] = ':'
+
+	// Write second (2 digits)
+	b[17] = byte('0' + second/10)
+	b[18] = byte('0' + second%10)
+	b[19] = '.'
+
+	// Write milliseconds (3 digits)
+	b[20] = byte('0' + millisecond/100%10)
```
flotter: nit: I think the range is fixed here (0-999, verified from the time.Nanosecond() range * 1000000), unlike years which can grow > 9999 ? So I would say the ```%10``` on this line can be dropped ?

flotter (approved):
Looks great to me.

internals/servicelog/formatter.go
benhoyt: Yeah, good call.

PR #675, opened 2025-08-11
refactor: unify loki and otel packages

IronCore864:


internals/overlord/logstate/opentelemetry/opentelemetry.go
```diff
@@ -126,11 +128,12 @@ type Client struct {
 	resourceAttributes map[string][]keyValue
 }
 
-func NewClient(options *ClientOptions) *Client {
+func NewClient(target *plan.LogTarget, options *ClientOptions) *Client {
```
benhoyt: I think we should revert this change (in both `loki` and `opentelemetry`). I think avoiding the dependency on the `plan` package is a good thing (even if it means the "duplicated" `TargetName` and `Location` fields) -- there's no reason these log forwarders should need to know about or depend on the Pebble plan.

So let's add `TargetName` and `Location` back and remove the *target* argument from `NewClient` in both packages.

internals/overlord/logstate/loki/loki.go
```diff
@@ -44,45 +43,42 @@ type Client struct {
...
-
-// ClientOptions allows overriding default parameters (e.g. for testing)
-type ClientOptions struct {
-	RequestTimeout    time.Duration
-	MaxRequestEntries int
-}
-
-func NewClientWithOptions(target *plan.LogTarget, options *ClientOptions) *Client {
-	options = fillDefaultOptions(options)
+func NewClient(target *plan.LogTarget, options *ClientOptions) *Client {
```
benhoyt: Per comment on `opentelemetry.go`, I'd rather avoid the dependency on the plan package, and have all the options in `ClientOptions` (like we had before with `opentelemetry.NewClient`). That seems cleaner to me -- fewer dependencies, as these packages don't need to know about the Pebble plan.

internals/overlord/logstate/loki/loki.go
```diff
@@ -152,15 +148,15 @@ func (c *Client) Flush(ctx context.Context) error {
...
 	}
 
 	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, c.target.Location, bytes.NewReader(jsonReq))
 	if err != nil {
-		return fmt.Errorf("creating HTTP request: %v", err)
+		return fmt.Errorf("cannot create HTTP request: %v", err)
 	}
 	httpReq.Header.Set("Content-Type", "application/json; charset=utf-8")
-	httpReq.Header.Set("User-Agent", fmt.Sprintf("pebble/%s", cmd.Version))
+	httpReq.Header.Set("User-Agent", c.options.UserAgent)
```
benhoyt, line 158: Great, thanks.

benhoyt (changes requested):
Most of the changes are good, but one change/reversion suggested in comments.

benhoyt (approved):

benhoyt:
Thanks, looks good to me. Please update the PR description and then merge.

PR #686, opened 2025-08-21
fix: concurrent access to executions map

anpep:


flotter (approved):
Looks good to me. Please can you add a small explanation for @benhoyt benefit why we know this works ?

internals/overlord/cmdstate/handlers.go
```diff
@@ -105,14 +105,14 @@ func (m *CommandManager) doExec(task *state.Task, tomb *tomb.Tomb) error {
...
 	m.executions[task.ID()] = e
-	m.executionsMutex.Unlock()
 	m.executionsCond.Broadcast() // signal that Connects can start happening
+	m.executionsCond.L.Unlock()
 	defer func() {
-		m.executionsMutex.Lock()
+		m.executionsCond.L.Lock()
 		delete(m.executions, task.ID())
-		m.executionsMutex.Unlock()
+		m.executionsCond.L.Unlock()
```
dimaqq, line 115: **question:** at the first glance, I would expect every operation that locks to modify data to also notify all waiters.

Here, I feel that we're relying on the fact that the waiters are only waiting for specific elements to appear, and don't wait for any elements to disappear.

So far so good.
(Although I would prefer not to have this intricate semantics encoded and just notifyAll (Broadcast) unconditionally)

However, this brings up a question:
- suppose someone is waiting for this very taskID (to appear)
- and this function runs quickly, inserting the entry, doing stuff and deleting the entry
- the waiter is slow (rare but possible)

Would the waiter then continue waiting forever or until timeout?

benhoyt: I'm not 100% sure what you're asking, but I've played around with this locally, adding time.Sleeps before Connect to try to get failure cases to happen, but I haven't seen any. @anpep now that you're close to the code, do you see any issues with what @dimaqq is suggesting?

flotter: If you look at the ```Connect``` method:

```go
// Connect upgrades the HTTP connection and connects to the given websocket.
func (m *CommandManager) Connect(r *http.Request, w http.ResponseWriter, task *state.Task, websocketID string) error {
        stopWait := make(chan struct{})
        defer func() {
                // So waitExecution wakes up if it's stuck in Wait().
                close(stopWait)
                m.executionsCond.Broadcast()
        }()
        :
```

The universal cure for a receiver Wait()-ing is that the context gets timeout cancelled, and a stop + broadcast is send to release the connection goroutine. We know this works well (although even this code has a race), since this was exactly what happened due to the races fixed in this PR - we timed out after exactly the connection timeout = 5s, because we missed the broadcast.

> and this function runs quickly, inserting the entry, doing stuff and deleting the entry

The requester and receiver code operates in lock step (they wait for each other), so the requested code will not simply delete the entry without an error of some sort, telling it that there is no point to wait for the other side.

```go
        // Store the execution object on the manager (for Connect).
        m.executionsMutex.Lock()
        m.executions[task.ID()] = e
        m.executionsMutex.Unlock()
        m.executionsCond.Broadcast() // signal that Connects can start happening
        defer func() {
                m.executionsMutex.Lock()
                delete(m.executions, task.ID())
                m.executionsMutex.Unlock()
        }()

        // Run the command! Killing the tomb will terminate the command.
        ctx := tomb.Context(context.Background())
        return e.do(ctx, task)
```

In order for delete to cleanup, either a timeout must be reached, or the task (tomb) must be cancelled. The current design has multiple goroutines with a 5s timeout mechanism (also wired into the HTTP connection logic and handshake logic). Any error will likely result in the overall system only cleaning up after 5s. This is the current design, and changing this design will likely require very careful study and testing of the websocket code.

The intention of this PR is not to propose any change to the synchronization system or the error handling mechanism, but simply fixing the use of the conditional variable, to prevent a 5s timeout error for a case which should just work (due to a race in the current conditional code).

Adding a broadcast to the delete will have no impact, as the connection loop will not exit unless the stop channel is closed. Closing the stop channel cannot happen twice, so this would have to be a fundamental redesign of the code, to prevent a double close on exit.

@anpep is out this morning, but we've worked on this together, so I left some input until he is back to add to my view.

@dimaqq let me know if I missed something obvious. The code is solving a very complex problem, so it not that simple to understand all the error paths or triggers that can cancel a context.

benhoyt:
Thanks a lot for the diagnosis and fix, @anpep and @flotter. Merging.

internals/overlord/cmdstate/handlers.go
dimaqq: The explanation makes sense, thank you!

Issue #422, opened 2024-06-05
Ensure websocket endpoint times out if it's never going to happen


benhoyt:
For the problem described at https://github.com/canonical/operator/issues/1246, this issue represents the Pebble side: we should fix the `/v1/tasks/{task_id}/websocket/{websocket_id}` endpoint so that if the Pebble side has already timed out ("timeout waiting for websocket connections"), we should time out (or cancel immediately) attempts to connect to that websocket URL.

Marking as Low Priority given the Ops-side fix at https://github.com/canonical/operator/pull/1247.

benhoyt:
Tiexin's draft PR in #691.

PR #691, opened 2025-09-02
fix: websocket connection timeout

IronCore864:
Fix `CommandManager.Connect` so that if the Pebble side has already timed out (an exec change fails before the execution object is created in the `CommandManager`), the server returns an error, causing the connection to the websocket URL to be timed out immediately.

Also, add a unit test to cover this scenario.

Closes https://github.com/canonical/pebble/issues/422.

### DIFF
--- a/internals/daemon/api_exec_test.go
+++ b/internals/daemon/api_exec_test.go
@@ -482,7 +482,6 @@ func (s *execSuite) TestExecChangeReady(c *C) {
 	rsp.ServeHTTP(rec, req)
 
 	c.Check(rec.Code, Equals, 500)
-	fmt.Printf("Response body: %q\n", rec.Body.String())
 	c.Check(rec.Body.String(), Matches, `.*cannot connect to websocket.*something went wrong.*`)
 }
 
--- a/internals/daemon/api_exec_test.go
+++ b/internals/daemon/api_exec_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -31,6 +32,7 @@ import (
 
 	"github.com/canonical/pebble/client"
 	"github.com/canonical/pebble/internals/logger"
+	"github.com/canonical/pebble/internals/overlord/state"
 	"github.com/canonical/pebble/internals/plan"
 	"github.com/canonical/pebble/internals/reaper"
 )
@@ -38,8 +40,10 @@ import (
 var _ = Suite(&execSuite{})
 
 type execSuite struct {
-	daemon *Daemon
-	client *client.Client
+	daemon         *Daemon
+	client         *client.Client
+	vars           map[string]string
+	restoreMuxVars func()
 }
 
 func (s *execSuite) SetUpSuite(c *C) {
@@ -51,6 +55,7 @@ func (s *execSuite) SetUpTest(c *C) {
 	if err != nil {
 		c.Fatalf("cannot start reaper: %v", err)
 	}
+	s.restoreMuxVars = FakeMuxVars(s.muxVars)
 
 	socketPath := c.MkDir() + ".pebble.socket"
 	daemon, err := New(&Options{
@@ -67,9 +72,15 @@ func (s *execSuite) SetUpTest(c *C) {
 	c.Assert(err, IsNil)
 }
 
+func (s *execSuite) muxVars(*http.Request) map[string]string {
+	return s.vars
+}
+
 func (s *execSuite) TearDownTest(c *C) {
 	err := s.daemon.Stop(nil)
 	c.Check(err, IsNil)
+
+	s.restoreMuxVars()
 
 	err = reaper.Stop()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internals/overlord/cmdstate/manager.go
```diff
@@ -64,42 +66,55 @@ func (m *CommandManager) Connect(r *http.Request, w http.ResponseWriter, task *s
...
+		if err != nil {
+			errCh <- err
+		}
 	}()
 
 	// Wait till the execution object is ready or the request is cancelled.
 	select {
 	case e := <-executionCh:
 		return e.connect(r, w, websocketID)
+	case err := <-errCh:
```
benhoyt: I think you're right to add a new case here, however, I played with this yesterday, and I don't think we need any additional machinery. If we wait for the task/change to be done, that should do it, for example `case <-change.Ready()`. You can get the change using `task.Change()` (though you probably need to lock/unlock state to get the change. Then there's `change.Err()` to get the error. I've played with it locally and arrived at this:

```
	case <-change.Ready():
		// Change unexpected marked ready, probably due to client not sending
		// websocket requests in time.
		st.Lock()
		err := change.Err()
		st.Unlock()
		return fmt.Errorf("change unexpectedly marked ready: %v", err)
```

What do you think?

internals/overlord/cmdstate/manager.go
```diff
@@ -76,12 +76,21 @@ func (m *CommandManager) Connect(r *http.Request, w http.ResponseWriter, task *s
...
+	change := task.Change()
+	st.Unlock()
+
 	// Wait till the execution object is ready or the request is cancelled.
 	select {
 	case e := <-executionCh:
 		return e.connect(r, w, websocketID)
 	case <-r.Context().Done():
 		return r.Context().Err()
+	case <-change.Ready():
```
benhoyt, line 92: Let's add a comment, something like:

```
// Change unexpected marked ready, probably due to client not sending
// websocket requests in time.
```

internals/daemon/api_exec_test.go
```diff
@@ -441,6 +443,53 @@ func (s *execSuite) TestUserGroupError(c *C) {
...
+
+	st := s.daemon.overlord.State()
+	st.Lock()
+	change := st.Change(changeID)
+	c.Assert(change, NotNil)
+	c.Assert(len(change.Tasks()), Equals, 1)
+	// Set the change as failed and set the error on the task.
+	change.SetStatus(state.ErrorStatus)
+	change.Tasks()[0].Errorf("something went wrong")
+	change.Tasks()[0].SetStatus(state.ErrorStatus)
```
benhoyt, line 465: Let's try `change.Abort()` to simplify

internals/daemon/api_exec_test.go
```diff
@@ -441,6 +443,53 @@ func (s *execSuite) TestUserGroupError(c *C) {
...
+// TestExecChangeReady simulates the scenario where the change is ready before the websocket
+// connection is established, so the connection should fail.
+func (s *execSuite) TestExecChangeReady(c *C) {
+	httpResp, execResp := execRequest(c, &client.ExecOptions{
+		Command: []string{"echo", "foo"},
+	})
+	c.Assert(httpResp.StatusCode, Equals, http.StatusAccepted)
+	body := new(bytes.Buffer)
+	err := json.NewEncoder(body).Encode(execResp)
+	c.Assert(err, IsNil)
```
benhoyt: Not sure we need this.

internals/daemon/api_exec_test.go
```diff
@@ -441,6 +443,53 @@ func (s *execSuite) TestUserGroupError(c *C) {
...
+
+	changeID := execResp.Change
+	c.Assert(changeID, Not(Equals), "")
+
+	st := s.daemon.overlord.State()
+	st.Lock()
+	change := st.Change(changeID)
+	c.Assert(change, NotNil)
+	c.Assert(len(change.Tasks()), Equals, 1)
```
benhoyt, line 461: Assuming Abort works, probably don't need this line

internals/daemon/api_exec_test.go
```diff
@@ -441,6 +443,53 @@ func (s *execSuite) TestUserGroupError(c *C) {
...
+	req, err := http.NewRequest("GET", fmt.Sprintf("/v1/tasks/%s/websocket/%s", taskID, "control"), nil)
+	c.Assert(err, IsNil)
+	rsp := v1GetTaskWebsocket(websocketCmd, req, nil).(websocketResponse)
+	rec := httptest.NewRecorder()
+	rsp.ServeHTTP(rec, req)
+
+	c.Check(rec.Code, Equals, 500)
+	err = rsp.connect(req, rec, rsp.task, rsp.websocketID)
+	c.Assert(err, NotNil)
+	c.Check(err.Error(), Matches, `.*cannot perform the following tasks:\n.*something went wrong.*`)
```
benhoyt: rsp.ServeHTTP has already called connect internally. So we should be checking for rec.Result() body to be the error instead.

benhoyt (changes requested):
Per discussion, a few comments.

internals/daemon/api_exec_test.go
IronCore864: I played with `change.Abort`, and it did not work as expected. I found some differences between aborting a change and setting a change to the error status, because what we changed in `CommandManager.Connect` is adding a new case for the change is ready, and if we abort the change, the newly added case will never be reached, so `change.Abort` does not work as expected. Since the change is never ready, it reaches the existing case where the execution object is ready, not our newly added case, where the execution object is deleted because `doExec` is finished and the change is ready.

So, by aborting the change, it's actually testing existing behaviour, not the case where the change is ready (for example, when the command runs too fast before the client connects to the websocket).

Does this make sense?

internals/daemon/api_exec_test.go
```diff
@@ -441,6 +443,49 @@ func (s *execSuite) TestUserGroupError(c *C) {
...
+
+	websocketCmd := apiCmd("/v1/tasks/{task-id}/websocket/{websocket-id}")
+	req, err := http.NewRequest("GET", fmt.Sprintf("/v1/tasks/%s/websocket/%s", taskID, "control"), nil)
+	c.Assert(err, IsNil)
+	rsp := v1GetTaskWebsocket(websocketCmd, req, nil).(websocketResponse)
+	rec := httptest.NewRecorder()
+	rsp.ServeHTTP(rec, req)
+
+	c.Check(rec.Code, Equals, 500)
+	fmt.Printf("Response body: %q\n", rec.Body.String())
```
benhoyt: Leftover debugging?

PR #694, opened 2025-09-08
feat: add "cert" type identity

anpep:
This patch adds a `cert` type identity for storing PEM-encoded x.509 certificates used for mTLS authentication.

### DIFF
--- a/client/identities.go
+++ b/client/identities.go
@@ -29,6 +29,7 @@ type Identity struct {
 	// non-nil.
 	Local *LocalIdentity `json:"local,omitempty" yaml:"local,omitempty"`
 	Basic *BasicIdentity `json:"basic,omitempty" yaml:"basic,omitempty"`
+	Cert  *CertIdentity  `json:"cert,omitempty" yaml:"cert,omitempty"`
 }
 
 // IdentityAccess defines the access level for an identity.
@@ -53,6 +54,13 @@ type LocalIdentity struct {
 type BasicIdentity struct {
 	// Password holds the user's sha512-crypt-hashed password.
 	Password string `json:"password" yaml:"password"`
+}
+
+// CertIdentity holds identity configuration specific to the "cert" type
+// (for mTLS authentication)
+type CertIdentity struct {
+	// Certificate holds the PEM-encoded certificate.
+	Certificate string `json:"certificate" yaml:"certificate"`
 }
 
 // For future extension.
--- a/internals/overlord/state/identities.go
+++ b/internals/overlord/state/identities.go
@@ -34,6 +34,7 @@ type Identity struct {
 	// non-nil.
 	Local *LocalIdentity
 	Basic *BasicIdentity
+	Cert  *CertIdentity
 }
 
 // IdentityAccess defines the access level for an identity.
@@ -57,6 +58,11 @@ type LocalIdentity struct {
 type BasicIdentity struct {
 	// Password holds the user's sha512-crypt-hashed password.
 	Password string
+}
+
+type CertIdentity struct {
+	// Certificate holds the PEM-encoded certificate for the identity.
+	Certificate string
 }
 
 // This is used to ensure we send a well-formed identity Name.
@@ -101,8 +107,14 @@ func (d *Identity) validateAccess() error {
 		}
 		gotType = true
 	}
+	if d.Cert != nil {
+		if d.Cert.Certificate == "" {
+			return errors.New("cert identity must specify certificate (PEM-encoded)")
+		}
+		gotType = true
+	}
 	if !gotType {
-		return errors.New(`identity must have at least one type ("local" or "basic")`)
+		return errors.New(`identity must have at least one type ("local", "basic", or "cert")`)
 	}
 
 	return nil
@@ -115,6 +127,7 @@ type apiIdentity struct {
 	Access string            `json:"access"`
 	Local  *apiLocalIdentity `json:"local,omitempty"`
 	Basic  *apiBasicIdentity `json:"basic,omitempty"`
+	Cert   *apiCertIdentity  `json:"cert,omitempty"`
 }
 
 type apiLocalIdentity struct {
@@ -123,6 +136,10 @@ type apiLocalIdentity struct {
 
 type apiBasicIdentity struct {
 	Password string `json:"password"`
+}
+
+type apiCertIdentity struct {
+	Certificate string `json:"certificate"`
 }
 
 // IMPORTANT NOTE: be sure to exclude secrets when adding to this!
@@ -135,6 +152,9 @@ func (d *Identity) MarshalJSON() ([]byte, error) {
 	}
 	if d.Basic != nil {
 		ai.Basic = &apiBasicIdentity{Password: "*****"}
+	}
+	if d.Cert != nil {
+		ai.Cert = &apiCertIdentity{Certificate: d.Cert.Certificate}
 	}
 	return json.Marshal(ai)
 }
@@ -158,6 +178,9 @@ func (d *Identity) UnmarshalJSON(data []byte) error {
 	}
 	if ai.Basic != nil {
 		identity.Basic = &BasicIdentity{Password: ai.Basic.Password}
+	}
+	if ai.Cert != nil {
+		identity.Cert = &CertIdentity{Certificate: ai.Cert.Certificate}
 	}
 
 	// Perform additional validation using the local Identity type.
--- a/internals/overlord/state/identities_test.go
+++ b/internals/overlord/state/identities_test.go
@@ -45,6 +45,10 @@ func (s *identitiesSuite) TestMarshalAPI(c *C) {
 			Access: state.MetricsAccess,
 			Basic:  &state.BasicIdentity{Password: "hash"},
 		},
+		"olivia": {
+			Access: state.ReadAccess,
+			Cert:   &state.CertIdentity{Certificate: "pem"},
+		},
 	})
 	c.Assert(err, IsNil)
 
@@ -70,6 +74,12 @@ func (s *identitiesSuite) TestMarshalAPI(c *C) {
         "basic": {
             "password": "*****"
         }
+    },
+    "olivia": {
+        "access": "read",
+        "cert": {
+            "certificate": "pem"
+        }
     }
 }`[1:])
 }
@@ -94,6 +104,12 @@ func (s *identitiesSuite) TestUnmarshalAPI(c *C) {
         "basic": {
             "password": "hash"
         }
+    },
+    "olivia": {
+        "access": "read",
+        "cert": {
+            "certificate": "pem"
+        }
     }
 }`)
 	var identities map[string]*state.Identity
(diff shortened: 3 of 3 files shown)


internals/overlord/state/identities.go
```diff
@@ -101,8 +109,25 @@ func (d *Identity) validateAccess() error {
 		}
 		gotType = true
 	}
+	if d.Cert != nil {
+		if d.Cert.Certificate == "" {
+			return errors.New("cert identity must specify certificate (PEM-encoded)")
+		}
+		block, rest := pem.Decode([]byte(d.Cert.Certificate))
+		if block == nil {
+			return errors.New("cert identity must have at least one valid PEM block")
```
flotter, line 118: nit:

```identity user-1 invalid: certificate identity must include PEM encoded x509 certificate```

Can this not be the same error ? As it stands, it sounds like some PEM field exist in the cert identity, but what we mean is the certificate field is not encoded properly?

File: client/identities.go [hunk 2]
```diff
@@ -55,6 +56,13 @@ type BasicIdentity struct {
 	Password string `json:"password" yaml:"password"`
 }
 
+// CertIdentity holds identity configuration specific to the "cert" type
+// (for mTLS authentication)
+type CertIdentity struct {
+	// Certificate holds the PEM-encoded certificate.
+	Certificate string `json:"certificate" yaml:"certificate"`
```
flotter: pem ?

internals/overlord/state/identities.go
```diff
@@ -115,6 +140,7 @@ type apiIdentity struct {
 	Access string            `json:"access"`
 	Local  *apiLocalIdentity `json:"local,omitempty"`
 	Basic  *apiBasicIdentity `json:"basic,omitempty"`
+	Cert   *apiCertIdentity  `json:"cert,omitempty"`
```
flotter, line 131: "Certificate"

internals/overlord/state/identities.go
```diff
@@ -101,8 +109,25 @@ func (d *Identity) validateAccess() error {
 		}
 		gotType = true
 	}
+	if d.Cert != nil {
+		if d.Cert.Certificate == "" {
+			return errors.New("cert identity must specify certificate (PEM-encoded)")
```
flotter: ```identity user-1 invalid: certificate identity must include PEM encoded certificate.```

internals/overlord/state/identities.go
flotter: ```identity user-1 invalid: certificate identity must include PEM encoded certificate.```

internals/overlord/state/identities.go
```diff
@@ -101,8 +109,25 @@ func (d *Identity) validateAccess() error {
...
+		block, rest := pem.Decode([]byte(d.Cert.Certificate))
+		if block == nil {
+			return errors.New("cert identity must have at least one valid PEM block")
+		}
+		if len(rest) > 0 {
+			return errors.New("cert identity cannot have extra data after the PEM block")
+		}
+		_, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return fmt.Errorf("cannot parse certificate from cert identity: %w", err)
```
flotter: cert - > certificate

internals/overlord/state/identities.go
```diff
@@ -59,6 +62,12 @@ type BasicIdentity struct {
 	Password string
 }
 
+type CertificateIdentity struct {
+	// PEM holds the PEM-encoded certificate for the identity.
+	//PEM string
```
flotter: Comments ?

internals/overlord/state/identities.go
```diff
@@ -32,8 +34,9 @@ type Identity struct {
 
 	// One or more of the following type-specific configuration fields must be
 	// non-nil.
-	Local *LocalIdentity
-	Basic *BasicIdentity
+	Local       *LocalIdentity
+	Basic       *BasicIdentity
+	Certificate *CertificateIdentity
```
flotter: I think we still have ```Certificate.Certificate``` ? Does not feel ideal. Taking inspiration from ```Basic -> Password```, how about ```TLS.Client``` ? It feels unlikely this this identity type is just going to be generic ? Perhaps a quick brainstorm tomorrow at the internal arch to get some help ?

client/identities.go
```diff
@@ -29,6 +29,7 @@ type Identity struct {
 	// non-nil.
 	Local *LocalIdentity `json:"local,omitempty" yaml:"local,omitempty"`
 	Basic *BasicIdentity `json:"basic,omitempty" yaml:"basic,omitempty"`
+	Cert  *CertIdentity  `json:"cert,omitempty" yaml:"cert,omitempty"`
```
flotter, line 32: How does this get to the state from the client ? I guess there is also a server story, and ultimately the server calls ```AddIdentities``` and friends ? However, these use the runtime type, I wonder if this is going to end up being a bit confusing with the client sending a PEM, and the server having to convert it to X509 ? What do you think ?

flotter: Are any changed needed in the API end-point handler code in the daemon for identities ?

flotter (commented):
A few more tweaks ? Slightly worried about PEM vs. X509 - see comment about server endpoint.

internals/overlord/state/identities.go
```diff
@@ -59,6 +62,10 @@ type BasicIdentity struct {
 	Password string
 }
 
+type CertificateIdentity struct {
+	Certificate *x509.Certificate
```
flotter: Why not make this ```x509```. This would allow a certificate identity to support other formats in the future ? This resolves the ```x.Certificate.Certificate``` problem ?

internals/overlord/state/identities.go
```diff
@@ -101,8 +108,14 @@ func (d *Identity) validateAccess() error {
 		}
 		gotType = true
 	}
+	if d.Certificate != nil {
+		if d.Certificate.Certificate == nil {
+			return errors.New("certificate identity must include a PEM-encoded certificate")
```
flotter: No longer PEM encoded.

internals/overlord/state/identities_test.go
```diff
@@ -303,7 +380,7 @@ func (s *identitiesSuite) TestAddIdentities(c *C) {
 			Access: "admin",
 		},
 	})
-	c.Assert(err, ErrorMatches, `identity "bill" invalid: identity must have at least one type \(\"local\" or \"basic\"\)`)
+	c.Assert(err, ErrorMatches, `identity "bill" invalid: identity must have at least one type \(\"local\", \"basic\", or \"certificate\"\)`)
```
flotter: I think the double quote escapes are no needed here, see next occurrence without them.

internals/overlord/state/identities_test.go
```diff
@@ -702,3 +779,15 @@ func (s *identitiesSuite) TestIdentityFromInputs(c *C) {
 	identity = st.IdentityFromInputs(&userID, "nancy-wrong-username", "wrong-password")
 	c.Assert(identity, IsNil)
 }
+
+func parseCert(pemBlock string) *x509.Certificate {
```
flotter: Can we not simply pass in ```c``` here, and use c.Assert instead of the panic conditions ?

flotter (approved):
Happy if the last couple of tweaks are considered. Please just update the description to make some of our decisions clear to the other reviewers, Thanks!

client/identities.go
anpep: No, there aren't any

client/identities.go
```diff
@@ -27,8 +27,9 @@ type Identity struct {
 
 	// One or more of the following type-specific configuration fields must be
 	// non-nil.
-	Local *LocalIdentity `json:"local,omitempty" yaml:"local,omitempty"`
-	Basic *BasicIdentity `json:"basic,omitempty" yaml:"basic,omitempty"`
+	Local       *LocalIdentity       `json:"local,omitempty" yaml:"local,omitempty"`
+	Basic       *BasicIdentity       `json:"basic,omitempty" yaml:"basic,omitempty"`
+	Certificate *CertificateIdentity `json:"certificate,omitempty" yaml:"certificate,omitempty"`
```
benhoyt: Maybe this has already been spec'd, but what about shortening "certificate" to "cert"? Seems a common abbreviation, and "certificate" is quite long.

internals/overlord/state/identities.go
```diff
@@ -136,6 +154,13 @@ func (d *Identity) MarshalJSON() ([]byte, error) {
 	if d.Basic != nil {
 		ai.Basic = &apiBasicIdentity{Password: "*****"}
 	}
+	if d.Certificate != nil {
+		pemBlock := &pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: d.Certificate.X509.Raw,
+		}
+		ai.Certificate = &apiCertificateIdentity{PEM: string(pem.EncodeToMemory(pemBlock))}
```
benhoyt: Wouldn't we want to `*****` this out as per the "IMPORTANT NOTE" above, and similar to the basic.password field?

And then the one in `marshalledIdentity` (saved in state) would get the full value.

We should also ensure there are tests of both types of marshalling.

internals/overlord/state/identities.go
```diff
@@ -112,9 +125,10 @@ func (d *Identity) validateAccess() error {
 // for API responses) excludes secrets. The marshalledIdentity type is used
 // for saving secrets in state.
 type apiIdentity struct {
-	Access string            `json:"access"`
-	Local  *apiLocalIdentity `json:"local,omitempty"`
-	Basic  *apiBasicIdentity `json:"basic,omitempty"`
+	Access      string                  `json:"access"`
+	Local       *apiLocalIdentity       `json:"local,omitempty"`
+	Basic       *apiBasicIdentity       `json:"basic,omitempty"`
+	Certificate *apiCertificateIdentity `json:"certificate,omitempty"`
```
benhoyt: I think we also need the corresponding changes to `marshalledIdentity` in state.go -- the one that actually gets saved in the state file.

internals/overlord/state/identities.go
anpep: I thought certificates generally contained a public key and thus no private information

benhoyt: Oh, you're probably right. I wonder if we should `*****` it in any case, to be consistent. @flotter Can you please confirm this is purely public key data, and if so, what do you think of `*****`ing it anyway?

flotter: @benhoyt oops I forgot to update this chat following our conversation here. The x509 certificates are exchanged over the wire in cleartext before the shared encryption key is derived, and a peer sending an X509 certificate has to sign the entire conversation manuscript using the matching private key, so by design because it is a public document, it cannot be used without also owning the private key. Given this, I would argue it is OK to expose it, and it could be useful to have the ability to verify the identities for testing purposes. So I think its OK, but equally, we could start hiding it and postponing the decision to later.

benhoyt: Okay, thanks for the details. I think we should start conservative (and to avoid something that *looks* like it might be concerning) and start be replacing this with `*****` (like we do with passwords), with a code comment explaining the above -- that it's not actually security-sensitive, but we're redacting it for consistency and to avoid the appearance of concern.

anpep: Thanks for your feedback!

PR #696, opened 2025-09-09
fix: update change status before shutting down daemon

IronCore864:


benhoyt:
This is good investigation, but a time.Sleep fix is almost never a good idea. What if the system is busy and it takes 101ms to write the state? Let's brainstorm a more reliable fix.

benhoyt:
Thanks. Yep, it makes sense that we shouldn't change the retry logic in the task runner.

However, what I still don't understand is this: in this case, the service command is exiting early, `exited()` is being called, which sends an error on the `serviceData.started` channel, which will in turn cause `doStart` to return that error and put the task+change into Error state not in Doing (I think).

Then the `exited()` code does a `fallthrough` to handle the on-failure action, and if it's "shutdown" it'll call `Daemon.HandleRestart(RestartServiceFailure)`. This will kick off the daemon stopping via the `d.tomb.Kill(nil)` line. `runDaemon()` at the top level finishes by calling `d.Stop()` which should wait for the daemon to shut down, the overlord to stop (`d.overlord.Stop()`), which presumably will write the updated error state to disk.

Which of my assumptions above is not true?

IronCore864:
> Thanks. Yep, it makes sense that we shouldn't change the retry logic in the task runner.
> 
> However, what I still don't understand is this: in this case, the service command is exiting early, `exited()` is being called, which sends an error on the `serviceData.started` channel, which will in turn cause `doStart` to return that error and put the task+change into Error state not in Doing (I think).
> 
> Then the `exited()` code does a `fallthrough` to handle the on-failure action, and if it's "shutdown" it'll call `Daemon.HandleRestart(RestartServiceFailure)`. This will kick off the daemon stopping via the `d.tomb.Kill(nil)` line. `runDaemon()` at the top level finishes by calling `d.Stop()` which should wait for the daemon to shut down, the overlord to stop (`d.overlord.Stop()`), which presumably will write the updated error state to disk.
> 
> Which of my assumptions above is not true?

I reread the code, and here's what actually happens (I think):

1. Service exits early -> exited() sends error to started channel -> doStart() receives error and returns it.
2. In `exited()`, it falls through, which triggers shutdown.
3. In taskrunner, it receives the error, but here is where your assumption is incorrect, because [the code here](https://github.com/canonical/pebble/blob/master/internals/overlord/state/taskrunner.go#L238-L250) is reached first, whose default case will change any type of error error into retry, so [the logic down below in the default case](https://github.com/canonical/pebble/blob/master/internals/overlord/state/taskrunner.go#L294-L303) which would put the change/task to error state is never reached, since by this time the error is already retry, [the logic in the retry branch](https://github.com/canonical/pebble/blob/master/internals/overlord/state/taskrunner.go#L253-L260) is triggered, and since the task is not aborted, it will schedule a time for retry.
4. The state is persisted to the disk "correctly", and when the daemon restarts, tasks in DoingStatus (which I think is what happens when a task is scheduled for retry) will be picked up and executed again.

I can confirm this theory by adding some prints (see the latest commit).

If we do not take any action [here](https://github.com/canonical/pebble/blob/master/internals/overlord/state/taskrunner.go#L248), the issue will be resolved, but as already mentioned, I believe it's by design. Do we want to change this?

benhoyt:
Thanks for the details -- that's really helpful. Yeah, it's tricky, isn't it. I've asked advice from some other teams that use Changes and Tasks, so we'll see what they say. Worst case we can set tasks of type=start and status=Doing to Error (or Done?) on startup.

benhoyt:
So after discussion on Mattermost with the Snapd team and others, I don't think we want to move `r.stopped`, as that's likely to break things. I think the best approaches are:

- Pebble-specific change: modify `Daemon.stopRunningServices` to wait for all the outstanding "start" changes to finish there, using their `Ready` channels (which will result in all their failed "start" tasks to be set to the ErrorStatus).
- Or, task runner change: introduce a `NoRetry` error that would wrap the service start error and then check for it in the TaskRunner in the `if.stopped` check.

Why don't you push up two PRs, one with a PoC of each approach, to see if both are feasible, and we'll compare?

internals/daemon/daemon.go
```diff
@@ -747,9 +747,34 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
...
 }
 
-// stopRunningServices stops all running services, waiting for a short time
-// for them all to stop.
+// stopRunningServices waits for all running changes to complete then stops all running services,
+// waiting for a short time for them all to stop.
 func (d *Daemon) stopRunningServices() error {
+	st := d.state
+	st.Lock()
+	var runningChanges []*state.Change
```
benhoyt: It seems to me we should restrict this to kind="start" (starting services) and status=DoingStatus, no? And probably call it `startChanges`.

internals/daemon/daemon.go
```diff
@@ -747,9 +747,34 @@ func (d *Daemon) Stop(sigCh chan<- os.Signal) error {
...
+	}
+	st.Unlock()
+
+	if len(runningChanges) > 0 {
+		logger.Noticef("Waiting for %d running changes to complete before stopping services.", len(runningChanges))
+		for _, chg := range runningChanges {
+			select {
+			case <-chg.Ready():
+				logger.Debugf("Change %s completed.", chg.ID())
+			case <-time.After(d.overlord.ServiceManager().StopTimeout()):
```
benhoyt: We should only call `time.After` once before the for loop, so it's not StopTimeout times the number of changes, but only one timeout overall.

```go
timeout := time.After(d.overlord.ServiceManager().StopTimeout())
for _, chg := range runningChanges {
    select {
    case <-chg.Ready(): ...
    case <-timeout: ...
    }
}
```

benhoyt (commented):
I was playing around and testing this locally, and it occurred to me that the state of the service is `stateExited`, so if we add that state to `servstate.servicesToStop` (servstate/manager.go line 441), the existing code in `stopRunningServices` will do this for us. I tried this, and it seems to work nicely -- the original task finishes and is saved as status 9 (ErrorStatus).

So can we just do that much simpler change?

PR #697, opened 2025-09-10
chore: fix failed unit test

IronCore864:
Fix the newly added unit test `TestExecChangeReady`.

Since we merged the other PR that refactors the logging, the regex needs to be updated. Also, revert a change in `api_tasks.go`; it is supposed to be an internal error instead of not found. In the refactor PR, we should only change the format of the error, not the type.

### DIFF
--- a/internals/daemon/api_exec_test.go
+++ b/internals/daemon/api_exec_test.go
@@ -482,7 +482,7 @@ func (s *execSuite) TestExecChangeReady(c *C) {
 	rsp.ServeHTTP(rec, req)
 
 	c.Check(rec.Code, Equals, 500)
-	c.Check(rec.Body.String(), Matches, `.*cannot connect to websocket.*something went wrong.*`)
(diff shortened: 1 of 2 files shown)


internals/daemon/api_exec_test.go
```diff
@@ -482,7 +482,7 @@ func (s *execSuite) TestExecChangeReady(c *C) {
 	rsp.ServeHTTP(rec, req)
 
 	c.Check(rec.Code, Equals, 500)
-	c.Check(rec.Body.String(), Matches, `.*cannot connect to websocket.*something went wrong.*`)
+	c.Check(rec.Body.String(), Matches, `.*cannot perform the following tasks.*something went wrong.*`)
```
benhoyt: Maybe we should just shorten this to just `.*something went wrong.*` as we don't really care about the rest of the message, just that the error text we've set above is coming through.

benhoyt (approved):
Thanks for this. Just one micro comment.

Issue #679, opened 2025-08-15
Avoid fmt.Sprintf and reduce allocations in stderr logging


benhoyt:
There's a fair bit of allocation happening, I believe because the [`Logger` interface](https://github.com/canonical/pebble/blob/a9daf68dc3fdeb1ac241ee4d0c4fe2a088c615ad/internals/logger/logger.go#L38) accepts pre-formatted strings, so the caller uses `fmt.Sprintf`. I think it'd be more efficient to adjust the `Logger` interface to take the format string and args, so the caller can use `fmt.Fprintf` and we can make use of the fmt package’s buffer pool.

The logging allocation did show up in performance profiling, but the proposed solution hasn't been tested. So we'd want to create some benchmarks (and use `-benchmem`) to show this is actually helping.

IronCore864:
My Go version is 1.24.6, and I run it on darwin/arm64. I upgraded to 1.25.1 and the result is more or less identical.

With `go build -gcflags="-m -m" ./logger`:

```
# github.com/ironcore864/logger-test/logger
logger/logger.go:23:6: cannot inline Noticef: unhandled op DEFER
logger/logger.go:25:2: can inline Noticef.deferwrap1 with cost 76 as: func() { (*sync.Mutex).Unlock(.autotmp_2) }
logger/logger.go:29:6: cannot inline SetLogger: function too complex: cost 151 exceeds budget 80
logger/logger.go:40:6: can inline (*byteSliceWriter).Write with cost 13 as: method(*byteSliceWriter) func([]byte) (int, error) { *w.buf = append(*w.buf, p...); return len(p), nil }
logger/logger.go:52:6: cannot inline (*defaultLogger).Noticef: function too complex: cost 368 exceeds budget 80
logger/logger.go:67:6: can inline New with cost 12 as: func(io.Writer, string) Logger { return &defaultLogger{...} }
logger/logger.go:24:17: inlining call to sync.(*Mutex).Lock
logger/logger.go:24:17: inlining call to sync.(*Mutex).Lock
logger/logger.go:25:25: inlining call to sync.(*Mutex).Unlock
logger/logger.go:25:25: inlining call to sync.(*Mutex).Unlock
logger/logger.go:30:17: inlining call to sync.(*Mutex).Lock
logger/logger.go:32:19: inlining call to sync.(*Mutex).Unlock
logger/logger.go:30:17: inlining call to sync.(*Mutex).Lock
[...]

PR #698, opened 2025-09-17
refactor: avoid fmt.Sprintf in logging

IronCore864:
Refactor logger to remove `fmt.Sprintf` for performance improvement.

Closes https://github.com/canonical/pebble/issues/679 (see more discussion and perf test here).

A few things worth mentioning:

1. The interface is a breaking change, but I think it doesn't break anything, it's only used by the default logger and a test logger.
2. I removed an unnecessary check `if len(l.buf) == 0` in `defaultLogger.Noticef()` because it will never be true at that point, since we've already appended at least the timestamp.
3. In the `New()` function, I initialized the buffer with an estimated capacity (`buf: make([]byte, 0, 256)`). In the original code, the buffer was not pre-initialized. I think my approach is a slight improvement because in this case, Go doesn't have to guess the initial capacity and reallocate later. Although 256 is also a guess, I think it is better than no allocation right? In perf test, the alloc numbers are the same, but with the preallocation, it's slightly faster, especially when the logs are long. So I'm including this change in this PR.
4. The test against `os.Stderr` is made optional by using an environment variable, because the output would pollute the bench test output, making it harder to see all the results.

There will be a follow-up PR that ports the `appendTimestamp` function from [this PR](https://github.com/canonical/pebble/pull/674/files) to the logger package.

### DIFF
--- a/internals/logger/logger.go
+++ b/internals/logger/logger.go
@@ -57,9 +57,8 @@ var (
 func Panicf(format string, v ...any) {
 	loggerLock.Lock()
 	defer loggerLock.Unlock()
-	msg := fmt.Sprintf(format, v...)
-	logger.Noticef("PANIC %s", msg)
-	panic(msg)
+	logger.Noticef("PANIC "+format, v...)
+	panic(fmt.Sprintf(format, v...))
 }
 
 // Noticef notifies the user of something
@@ -118,7 +117,7 @@ func securityEvent(level string, event SecurityEvent, arg, description string) {
 		// Should never happen, and not much more we can do here.
 		return
 	}
-	logger.Noticef("%s", buf.String())
+	logger.Noticef("%s", buf.Bytes())
 }
 
 type SecurityEvent string
--- a/internals/logger/logger.go
+++ b/internals/logger/logger.go
@@ -35,15 +35,15 @@ var (
 // A Logger is a fairly minimal logging tool.
 type Logger interface {
 	// Notice is for messages that the user should see
-	Notice(msg string)
+	Noticef(format string, v ...any)
 	// Debug is for messages that the user should be able to find if they're debugging something
-	Debug(msg string)
+	Debugf(format string, v ...any)
 }
 
 type nullLogger struct{}
 
-func (nullLogger) Notice(string) {}
-func (nullLogger) Debug(string)  {}
+func (nullLogger) Noticef(format string, v ...any) {}
+func (nullLogger) Debugf(format string, v ...any)  {}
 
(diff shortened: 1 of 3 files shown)


internals/logger/logger.go
```diff
@@ -120,7 +118,7 @@ func securityEvent(level string, event SecurityEvent, arg, description string) {
 		// Should never happen, and not much more we can do here.
 		return
 	}
-	logger.Notice(buf.String())
+	logger.Noticef("%s", buf.String())
```
benhoyt: I believe you can use `buf.Bytes()` with the `%s` verb, which should avoid an allocation here.

internals/logger/logger_benchmark_test.go
```diff
@@ -0,0 +1,79 @@
...
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			Noticef("Concurrent message from goroutiner: %d and string: %s", 42, "test")
+		}
+	})
+}
+
+// BenchmarkLoggerNoticefStderr tests Noticef with real os.Stderr output.
+// Run with: BENCH_STDERR=1 go test ./internals/logger -bench=BenchmarkLoggerNoticefStderr -v
+func BenchmarkLoggerNoticefStderr(b *testing.B) {
```
benhoyt: Yeah, on second thoughts, I think we should probably drop these Stderr tests -- it's too messy. And we are testing allocations and performance with `io.Discard`, so that's probably enough. Sorry for the churn.

benhoyt (approved):
Thanks. I'll also ask @flotter to review to see if it's going to affect their usage.

internals/logger/logger.go
```diff
@@ -58,24 +58,22 @@ func Panicf(format string, v ...any) {
 	loggerLock.Lock()
 	defer loggerLock.Unlock()
 	msg := fmt.Sprintf(format, v...)
-	logger.Notice("PANIC " + msg)
+	logger.Noticef("PANIC %s", msg)
```
flotter: Can we not do what was done below with:

```go
l.Noticef("DEBUG "+format, v...)
```

To remove the Sprintf here too ?

paul-rodriguez (approved):
Looks good to me. Our logger implementation will have to change but only slightly.

internals/logger/logger.go
IronCore864: Yes but the panic below need the message, that's why I kept it, but after second thought, it's probably better to be consistent with the debug log. Updating now.

PR #699, opened 2025-09-18
fix(plan): empty sections should support defaults

flotter:
The plan library has an important property that should always be maintained - the plan sections (built-in and registered) should always look consistent to the plan consumers getting updates. What this means is that a section should have a well define zero state, which may include default values.

In sort, we cannot simply return nil or zero value fields if a section is missing. This requires special-case code to be added at the point of consumption, and does now allow the section extension to control default values.

Let's take a hypothetical example:

```yaml
pairing:
    mode: disabled
```

In the example above, there is a default rule applied **post-merge**. If no explicit pairing mode was supplied the default is "disabled". In order to support defaults, we **always** have to include the extensions combine call, even in the case where the plan is empty (initially).


internals/plan/plan.go
```diff
@@ -118,13 +118,21 @@ type Plan struct {
...
+		section, err := sectionExtensions[field].ParseSection(yaml.Node{})
 		if err != nil {
 			panic("internal error: ParseSection() of empty node must return a valid section")
 		}
+
+		// Ask the extension to apply all the post-merge rules
+		// (e.g. defaults) on the section instance.
+		p.Sections[field], err = sectionExtensions[field].CombineSections(section)
+		if err != nil {
+			panic("internal error: CombineSections() of empty node must return a valid section")
```
benhoyt: While we're here, can we update these panic messages to include the err string in the message?
```suggestion
			panic("internal error: CombineSections() of empty node must return a valid section: " + err.Error())
```

benhoyt (commented):
Can we add a small test for this?

internals/plan/plan.go
flotter: I updated 3 panics in this file to look the same. Resolving.

benhoyt (approved):

PR #700, opened 2025-09-18
refactor: move `appendTimestamp` from servicelog to logger

IronCore864:
Move the `appendTimestamp` function to `logger.AppendTimestamp` so that it can be used in both logger and servicelog.

Logger benchmark test result (with `os.Stderr`):

- Before: 1867 ns/op	      32 B/op	       1 allocs/op
- After:  1632 ns/op	      32 B/op	       1 allocs/op


benhoyt:
Why does this depend on #698? I would have thought it's completely independent. Can that diff be extracted and this PR just include the refactor?

benhoyt (approved):
Thanks for this, looks good.

PR #702, opened 2025-09-19
feat(pairingstate): add client pairing manager (TLS)

flotter:
The pairing manager controls the client pairing window, a small period of time allowing an incoming client pairing request.


internals/overlord/state/pairing.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Pairing holds state related to client server pairing.
+type Pairing struct {
+	// isPaired reflects if this server instance ever paired
+	// with a client. This flag cannot be cleared once set.
+	isPaired bool
+}
+
+// IsPaired returns true if the server paired with a client before.
+func (s *State) IsPaired() bool {
```
paul-rodriguez: If we pair with a client then unpair, `IsPaired` will still return true, and I think that's a bit confusing.

internals/daemon/daemon_test.go
```diff
@@ -73,6 +75,7 @@ type daemonSuite struct {
 var _ = Suite(&daemonSuite{})
 
 func (s *daemonSuite) SetUpTest(c *C) {
+	plan.RegisterSectionExtension(pairingstate.PairingField, &pairingstate.SectionExtension{})
```
paul-rodriguez, line 78: There are a couple of test packages like this one where the only thing that changes is adding a call to register and unregister in SetUp and TearDown. Why is it necessary? Shouldn't these tests just generally not depend on this extension if there's no pairing stuff in the plan yaml that they use?

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,281 @@
...
+func (c *PairingConfig) Validate() error {
+	switch c.Mode {
+	case ModeUnset, ModeDisabled, ModeSingle, ModeMultiple:
+	default:
+		return fmt.Errorf("cannot support pairing mode %q: unknown mode", c.Mode)
+	}
+	return nil
+}
+
+func (c *PairingConfig) IsZero() bool {
```
paul-rodriguez: It doesn't look like this is used anywhere in this code. Is this part of an interface that PairingConfig must implement? In that case, I think it would be nice to add a comment to say which interface this belongs to, and probably also a static type assertion that the interface is implemented.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+}
+
+type PairingManager struct {
+	state       *state.State
+	mu          sync.Mutex
+	config      *PairingConfig
+	open        bool
+	timer       Timer
+	cancelTimer context.CancelFunc
```
paul-rodriguez: Suggestion to rename `timer` to `cancelTimer` and `cancelTimer` to `cancelFunc`

Edit: actually I understand why you called the function `cancelTimer`, all it does is cancel the timer... I didn't even think about that interpretation, to me it was the timer for cancellation. We have to make sure there is just one interpretation, so adding a "do" prefix or "Func" suffix would help clarify that this is a function.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,281 @@
...
+	defer m.mu.Unlock()
+
+	newConfig := update.Sections[PairingField].(*PairingConfig)
+
+	// If the mode changed, force the pairing window to be reopened
+	// taking the new config into account.
+	if m.config.Mode != newConfig.Mode {
+		m.open = false
+		m.stopTimer()
+	}
```
paul-rodriguez: It looks like the test testing this specific behavior is missing

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,281 @@
...
+		username: newIdentity,
+	})
+	if err != nil {
+		return fmt.Errorf("cannot add identity: %w", err)
+	}
+
+	m.state.SetIsPaired()
+
+	m.open = false
+	m.stopTimer()
```
paul-rodriguez: Shouldn't these two lines go into a defer if the intention is that any failed pairing request should close the window?

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,281 @@
...
+
+	return nil
+}
+
+// generateUniqueUsername generates a unique username following the pattern "user-x"
+// where x starts at 1 and monotonically increments. Users names differently will
```
paul-rodriguez: typo

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+		}
+	}
+
+	return fmt.Sprintf("user-%d", maxUserNumber+1)
+}
+
+// PairingWindowOpen returns whether the pairing window is currently open.
+func (m *PairingManager) PairingWindowOpen() bool {
+	m.mu.Lock()
```
paul-rodriguez: Instead of blocking, shouldn't this just fail if the lock isn't free right now? As in, using TryLock instead of Lock.

This is probably not a big deal, but I worry about a case where an operation would needlessly wait for a long time on this mutex.

internals/overlord/pairingstate/package_test.go
```diff
@@ -0,0 +1,169 @@
...
+	stopped bool
+}
+
+func (ft *fakeTimer) Stop() bool {
+	if ft.stopped {
+		return false
+	}
+	ft.stopped = true
+	return true
+}
```
paul-rodriguez: Is this actually used?

internals/overlord/state/pairing.go
flotter: That was explicitly discussed, and the approved behavior. The reason is because in the single ownership model, simply "unpairing" does not delete your data, and so for that reason you do not want to permit someone else now taking ownership of your data. Since we do not have an unpair at this time, and likely never will, it will actually be true.

OK?

internals/daemon/daemon_test.go
flotter: The problem is simply that daemon tests instantiate the overlord sometimes, which adds the pairing manager, which will receive a PlanChanged event on startup, and expect the extension to be present. Will reach out to discuss.

internals/overlord/pairingstate/manager.go
flotter: I've added a comment. Since this is not compulsory, just like MarhsallYAML and friends, I think it fine to not force compliance. It is needed in cases where you want explicit control over when a data structure is marshaled and when not.

internals/overlord/pairingstate/manager.go
flotter:
```go
        // timer controls the duration of the pairing window.
        timer       Timer
        // cancelTimerFunc cancels a pairing window if in progress. 
        cancelTimerFunc context.CancelFunc
```

Resolving.

internals/overlord/pairingstate/manager.go
flotter: Yes. Fixed, resolving.

internals/overlord/pairingstate/manager.go
flotter: Yes thank you. Also fixed the unit tests to check for this. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Fixed, resolving.

internals/overlord/pairingstate/manager.go
flotter: I would say let's see this in the wild first before we tune it. This is only used during pairing, not for general TLS, so I see the impact as super low. Also, the contention will be between a plan update and another parallel pairing request, not a common occurrence for now.

internals/overlord/pairingstate/package_test.go
flotter:
```go
func (m *PairingManager) stopTimer() {
        if m.timer != nil {
                m.timer.Stop()
                m.timer = nil
        }
:
```


Yes, should be. Resolving.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	timer Timer
+	// cancelTimerFunc cancels a pairing window if enabled.
+	cancelTimerFunc context.CancelFunc
+}
+
+func NewManager(state *state.State) *PairingManager {
+	m := &PairingManager{
+		state: state,
+		config: &PairingConfig{
+			Mode: ModeDisabled,
```
anpep: Shouldn't this be `ModeUnset`? Maybe I'm misunderstanding the differences between `ModeUnset` and `ModeDisabled`

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// PairMTLS adds a client identity with admin permissions to the identity
+// subsystem. A pairing request always closes the pairing window.
+func (m *PairingManager) PairMTLS(clientCert *x509.Certificate) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.open {
+		return errors.New("cannot pair client: pairing is not open")
```
anpep: Shouldn't this simply be

```suggestion
		return errors.New("pairing is not open")
```

and when calling from other places:

```go
if err := m.PairMTLS(clientCert); err != nil {
    return fmt.Errorf("cannot pair client: %w", err)
}

```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+
+	m.state.Lock()
+	defer m.state.Unlock()
+
+	existingIdentities := m.state.Identities()
+
+	for _, identity := range existingIdentities {
+		if identity.Cert != nil && identity.Cert.X509 != nil {
+			if identity.Cert.X509.Equal(clientCert) {
+				return errors.New("cannot pair client: identity already paired")
```
anpep: same as above

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	for name := range existingIdentities {
+		if strings.HasPrefix(name, "user-") {
+			numberStr := strings.TrimPrefix(name, "user-")
+			if number, err := strconv.Atoi(numberStr); err == nil && number > maxUserNumber {
+				maxUserNumber = number
+			}
+		}
+	}
+
+	return fmt.Sprintf("user-%d", maxUserNumber+1)
```
anpep: I know this is super painful, but how would this behave if an identity is added through `add-identities` locally with an user like `user-2147483647`?
Could we forbid the `user-` prefix entirely?

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil
+	}
+
+	// Check the pairing mode
+	switch m.config.Mode {
+	case ModeDisabled, ModeUnset:
+		// Mode is disabled, set Open = False
+		m.open = false
+		m.stopTimer()
+		return errors.New("cannot enable pairing: pairing disabled")
```
anpep: Same comments as above re: `cannot enable pairing:` prefix

internals/overlord/pairingstate/manager_test.go
```diff
@@ -0,0 +1,285 @@
...
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/pebble/internals/overlord/pairingstate"
+	"github.com/canonical/pebble/internals/overlord/state"
+)
+
+// TestEnablePairingDisabledMode tests trying to open a pairing window while
+// the configuration says it is disabled (before the plan update was received).
+func (ps *pairingSuite) TestEnablePairingDisabledMode(c *C) {
+
```
anpep, line 35: super nit: throughout this PR I've seen many instances of spurious line breaks like this. Could we maybe add a rule so this is automatically formatted upon running `make autoformat`?

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	return m
+}
+
+// PlanChanged informs the pairing manager that the plan has been updated.
+func (m *PairingManager) PlanChanged(update *plan.Plan) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	newConfig := update.Sections[PairingField].(*PairingConfig)
```
acalinp: Dangerous type assertion?

internals/overlord/pairingstate/manager.go
anpep: Sounds like a good idea but if we are in the process of pairing, we try to acquire the lock and we fail (as we should), then we fail to actually pair, the second caller might have missed an opportunity to pair if we use `TryLock()` vs the current `Lock()`, so Fred's idea of seeing this in the wild sounds more convincing to me?

internals/overlord/state/pairing.go
flotter: Agreed for now.

internals/daemon/daemon_test.go
flotter: Discussed with Paul, but I have since rediscovered that the plan extensions are not runtime dynamic, by design. The code requires the sections to exist, and will panic if they do not exit. I would like to keep it like this, as at the end of the day, if a section does not exist, you have to treat it as an error (Manager fails to start up) or ignore it. Ignoring it is not an option, as the manager will fundamentally not work. So this is indeed a coding bug, and so I would like to continue treating the absence of an expected plugin as a coding but to make this crystal clear by showing startup failures right away (all the existing uses of extensions follows this philosophy, throughout all projects).

internals/overlord/pairingstate/manager.go
flotter: Agreed comment is ok. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Postponing for now.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	// If the mode changed, force the pairing window to be reopened
+	// taking the new config into account.
+	if m.config.Mode != newConfig.Mode {
+		m.open = false
+		m.stopTimer()
+	}
+	m.config = newConfig
+}
+
+// Ensure implements StateManager.Ensure.
```
acalinp: Nit: Technically implements `overlord.StateManager` does it not?

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+
+	existingIdentities := m.state.Identities()
+
+	for _, identity := range existingIdentities {
+		if identity.Cert != nil && identity.Cert.X509 != nil {
+			if identity.Cert.X509.Equal(clientCert) {
+				return errors.New("cannot pair client: identity already paired")
+			}
+		}
+	}
```
acalinp, line 253: Nit alert!
```suggestion
    for _, identity := range existingIdentities {
	    if identity.Cert == nil || identity.Cert.X509 == nil {
		    continue
	    }
    
	    if identity.Cert.X509.Equal(clientCert) {
		    return errors.New("cannot pair client: identity already paired")
	    }
    }
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,277 @@
...
+
+type PairingManager struct {
+	state  *state.State
+	mu     sync.Mutex
+	config *PairingConfig
+	open   bool
+	// timer controls the duration of the pairing window.
+	timer Timer
+	// cancelTimerFunc cancels a pairing window if enabled.
+	cancelTimerFunc context.CancelFunc
```
acalinp: Do we need the added complexity of `cancelTimerFunc` right now? It doesn't seem like we cancel the pairing window anywhere unless I'm misunderstanding something? It feels like we could simplify this a lot by skipping the cancelfunc, what do you think?

internals/overlord/pairingstate/manager.go
flotter: Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: The way I normally think about this is that public method need to start with ```cannot```, but internal methods do not as long as its always wrapped from a public API perspective.

I see your point, so will reach out to see if we can find a better solution.

internals/overlord/pairingstate/manager.go
flotter: Will reach out to explain, but fundamentally this related to https://github.com/canonical/pebble/pull/702#discussion_r2381780188

internals/overlord/pairingstate/manager.go
flotter: Reworked as per our discussion. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Reworked as per our discussion. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Reworked as per our discussion. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Reworked as per our discussion. Resolving.

internals/overlord/pairingstate/manager_test.go
flotter: Fixed manually. This is Pebble, so I would rather not try to add new validation CI logic as part of this PR. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Fixed, thank you. I also added the overlord.StateStopper, which was missing. Thank you @Ardelean-Calin

flotter: Resolving.

internals/overlord/pairingstate/manager.go
flotter: Much nicer. Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: It is used in ```stopTimer```, which is called from ```startTimer```, which is called from ```EnablePairing```, ```PlanChanged``` and ```Stop```.

flotter: Please let me know if you disagree with my explanation. Resolving for now.

internals/cli/cli_test.go
```diff
@@ -77,6 +78,7 @@ func (s *BasePebbleSuite) TearDownTest(c *C) {
 	cli.ReadPassword = term.ReadPassword
 
 	plan.UnregisterSectionExtension(workloads.WorkloadsField)
+	plan.UnregisterSectionExtension(pairingstate.PairingField)
```
flotter, line 81: Let's document.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+
+// PairingWindowOpen returns whether the pairing window is currently open.
+func (m *PairingManager) PairingWindowOpen() bool {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.open
+}
+
+// stopTimer stops the current timer if it exists
+func (m *PairingManager) stopTimer() {
```
flotter: look if reset.

internals/overlord/state/pairing.go
```diff
@@ -0,0 +1,59 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package state
+
+import (
+	"encoding/json"
+)
+
+// Pairing holds state related to client server pairing.
+type Pairing struct {
```
flotter: Move to Get/Set on state instead of inventing new section here.

internals/daemon/daemon_test.go
flotter: Discussed with Ben, and we agree on the approach of failing early and failing hard if the extensions mechanism is not initialized as intended, as has been the approach in Pebble (workloads) and elsewhere in derivative projects so far. 

I consider this item resolved.

internals/overlord/pairingstate/manager.go
flotter: I consider this item resolved.

internals/overlord/overlord.go
```diff
@@ -642,6 +648,12 @@ func (o *Overlord) TLSManager() *tlsstate.TLSManager {
 	return o.tlsMgr
 }
 
+// PairingManager returns the pairing manager responsible for managing
+// client pairing.
```
acalinp:
```suggestion
// PairingManager returns manager that handles client pairing.
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+
+	// Any success or failure should always close the pairing window.
+	defer func() {
+		m.open = false
+		m.stopTimer()
+	}()
+
+	// Verify that the client certificate is self-signed (the public
+	// key included must verify the signature). We do this here as a
+	// sanity check since we are about to pair this certificat and use
```
acalinp:
```suggestion
	// sanity check since we are about to pair this certificate and use
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+		case <-ctx.Done():
+			return
+		default:
+			m.mu.Lock()
+			m.open = false
+			m.stopTimer()
+			m.mu.Unlock()
+		}
+	})
+}
```
acalinp: So I actually meant this:
```suggestion
func (m *PairingManager) startTimer(timeout time.Duration) {
	if m.timer != nil {
		m.timer.Reset(timeout)
	} else {
		m.timer = time.AfterFunc(timeout, func() {
			m.mu.Lock()
			m.open = false
			m.mu.Unlock()
		})
	}
}
```

Or even: 
```go
func (m *PairingManager) startTimer(timeout time.Duration) {
	if m.timer != nil {
		m.timer.Stop()
	}
	m.timer = time.AfterFunc(timeout, func() {
		m.mu.Lock()
		m.open = false
		m.mu.Unlock()
	})
}
```

Then replace all calls to `m.stopTimer` with `m.timer.Stop()` in `EnablePairing` below and it's the same behavior, is it not?

acalinp: Maybe I'm missing something, but I don't see what the `cancelTimerFunc` does that `Stop` doesn't already do.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+		return fmt.Errorf("cannot enable pairing with unknown pairing mode %q", m.config.Mode)
+	}
+
+	return nil
+}
+
+// generateUniqueUsername finds the first unique username following the pattern
+// "user-x" where x starts at 1 and monotonically increments. Users names not
+// following this pattern will simply not be considered.
+func generateUniqueUsername(existingIdentities map[string]*state.Identity) (string, error) {
```
acalinp, line 281: Suggestion:
```go
func generateUniqueUsername(existingIdentities map[string]*state.Identity) (string, error) {
	// Start checking from "user-1" and find the first available number.
	for i := uint32(1); i <= autoUsernameRangeLimit; i++ {
		username := fmt.Sprintf("user-%d", i)
		
		// If the generated username doesn't already exist, we can use it.
		if _, exists := existingIdentities[username]; !exists {
			return username, nil
		}
	}

	// If the loop finishes, all usernames within the limit are taken.
	return "", fmt.Errorf("user allocation limit '%d' reached", autoUsernameRangeLimit)
}
```

I think this is easier to understand and achieves the same logic as your function. **However**, please note that if you had user-1, user-2 and user-3 and then deleted user-2 somehow, both your code and mine will regenerate user-2.

internals/overlord/pairingstate/manager.go
flotter: Using timer.Reset now. Resolving.

internals/overlord/overlord.go
flotter: Fixed. Thanks. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Reworked, thanks. Please have a look at the new code.

internals/overlord/pairingstate/manager.go
flotter: Fixed, much nicer, thank you. Resolving.

internals/cli/cli_test.go
flotter: Added documentation for the plan schema.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,290 @@
...
+	if m.timer != nil {
+		m.timer.Reset(timeout)
+		return
+	}
+
+	m.timer = timeAfterFunc(timeout, func() {
+		m.mu.Lock()
+		defer m.mu.Unlock()
+		m.open = false
+		m.stopTimer()
```
acalinp: This line does seem unnecessary, since by this point, the timer has expired and is for all intents and purposes, stopped.
```suggestion
```

flotter: Fixed, thank you!

internals/overlord/pairingstate/manager.go
flotter: Resolving.

internals/overlord/state/pairing.go
flotter: Fixed.

internals/overlord/checkstate/manager_test.go
```diff
@@ -329,7 +329,7 @@ func (s *ManagerSuite) TestFailuresBelowThreshold(c *C) {
 		return check.Failures == 0
 	})
 	c.Assert(check.Status, Equals, checkstate.CheckStatusUp)
-	c.Assert(lastTaskLog(s.overlord.State(), check.ChangeID), Matches, `.* INFO succeeded after \d+ failure`)
+	c.Assert(lastTaskLog(s.overlord.State(), check.ChangeID), Matches, `.* INFO succeeded after \d+ failure.*`)
```
flotter, line 332: This is a unit test issue addressed in #707 , which will be removed from this PR once the fix is merged.

flotter: Merging in fix from master. Resolving.

docs/reference/layer-specification.md
```diff
@@ -279,4 +279,18 @@ log-targets:
...
       <label name>: <label value>
+
+# (Optional) Client pairing options controlling the pairing behaviour of the
+# server specifically when HTTPS (mTLS) transport is used.
+pairing:
+
+    # (Optional) Controls the pairing mode of the server
+    #
+    # When 'disabled' (or unset), no client pairing will be possible and as a
+    # result incoming client connections will always be rejected. If set to
```
niemeyer: The sentence probably needs more context. Pairing disabled does not mean every client connection is disabled in Pebble, as this would be a compatibility breakage.  It rather means pairing is disabled, which is what it says on the tin and wouldn't need saying.  So what do we really want to communicate here?

internals/daemon/api_exec_test.go
```diff
@@ -77,6 +79,7 @@ func (s *execSuite) TearDownTest(c *C) {
 	if err != nil {
 		c.Fatalf("cannot stop reaper: %v", err)
 	}
+	plan.UnregisterSectionExtension(pairingstate.PairingField)
```
niemeyer, line 82: Typically we undo in the opposite order that we do on the set up. It's of course all a bit of a gamble, since for us to fail getting here something above is quite wrong, but still it won't hurt to respect the order and undo this first.

internals/overlord/pairingstate/export_test.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"time"
+)
+
+const PairedStateKey = pairedStateKey
+
+var GenerateUniqueUsername = generateUniqueUsername
+
+// FakeAfterFunc replaces the global timeAfterFunc with a fake implementation
+func FakeAfterFunc(fakeAfterFunc func(d time.Duration, f func()) Timer) func() {
```
niemeyer: This is an unusual thing to fake, and most likely a bad idea.  We have no clue about what we are actually faking, because this is low-level operation that can be used for many reasons.  At this point I don't know the rationale yet, but I would recommend considering the actual context and more carefully tuning the desired behavior, instead of a low-level function.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+
+// timeAfterFunc allows faking time.AfterFunc.
+var timeAfterFunc = func(d time.Duration, f func()) Timer {
+	return time.AfterFunc(d, f)
+}
+
+// Timer is used so we can supply a fake timer during testing.
+type Timer interface {
+	Stop() bool
+	Reset(time.Duration) bool
```
niemeyer: Yeah, let's please not do this, for the reasons mentioned above.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+		state: state,
+		config: &PairingConfig{
+			Mode: ModeUnset,
+		},
+	}
+
+	// Load the paired state at startup.
+	m.state.Lock()
+	defer m.state.Unlock()
+	m.state.Get(pairedStateKey, &m.paired)
```
niemeyer: Let's please define a `type pairingState struct` here, as otherwise if we ever need to store a second field, we'll be stuffing it all in the general namespace of the global state as well.  The key used should follow along, and the type names as well.  I believe we have a "pairing-details", PairingDetails, etc convention, as one of the strong arms of this.  The project next door has some recent examples, I believe.

internals/overlord/pairingstate/manager.go
niemeyer: It does feel atypically aggressive since any mistakes are a system crash, potentially unrecoverable as you may not be able to upgrade out of it. If that's okay here, at the very least it sounds like a comment is due.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	_, err := clientCert.Verify(opts)
+	if err != nil {
+		return fmt.Errorf("cannot verify client certificate signature: %w", err)
+	}
+
+	m.state.Lock()
+	defer m.state.Unlock()
+
+	existingIdentities := m.state.Identities()
```
niemeyer, line 236: Unrelated to this PR, but it's really unfortunate that we've allowed this logic to be inserted under the state package.  Feels like every other cycle we need to go back and remove something that was added there because it felt like something central and important, when really it should be on its own subpackage.

@benhoyt Probably a topic for the sprint.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// EnablePairing requests the pairing manager to enable the pairing window.
+func (m *PairingManager) EnablePairing(timeout time.Duration) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// If the pairing window is already enabled, reset the timeout duration
+	if m.enabled {
+		m.startTimer(timeout)
```
niemeyer: There's a race condition here:

1) Test m.enabled, and it's true
2) Previously scheduled timer is called, blocks on mu.Lock 
3) startTimer resets the timer, schedules new function call
4) This function returns, unblocks mu.Lock
5) Previously scheduled timer acquires mu.Lock, sets enabled to false

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+	switch m.config.Mode {
+	case ModeDisabled, ModeUnset:
+		return errors.New("cannot enable pairing with pairing mode disabled")
+
+	case ModeSingle:
+		// Single mode: check if already paired
+		if m.paired {
+			return errors.New("cannot enable pairing when already paired in 'single' pairing mode")
+
+		}
```
niemeyer: Bogus empty line. The comment right above is also unnecessary, as it's just reading the surrounding code.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+
+// generateUniqueUsername finds the first unique username following the pattern
+// "user-x" where x starts at 1 and monotonically increments. Users names not
+// following this pattern will simply not be considered.
+func generateUniqueUsername(existingIdentities map[string]*state.Identity) (string, error) {
+	for i := uint32(1); i <= autoUsernameRangeLimit; i++ {
+		username := fmt.Sprintf("user-%d", i)
+
+		// If the generated username doesn't already exist, we can use it.
+		if _, exists := existingIdentities[username]; !exists {
```
niemeyer, line 285: Same as earlier, the comment is just reading the nice and simple code around it, so it's unnecessary.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,312 @@
...
+			return username, nil
+		}
+	}
+	// No free username found.
+	return "", fmt.Errorf("user allocation limit '%d' reached", autoUsernameRangeLimit)
```
niemeyer: Same as above, the comment is actually even less clear than the code in this case.

I'm nitpicking in an attempt to help in future commenting.

Bonus nitpick: I suggest dropping the quotes around the number.

niemeyer (changes requested):
Thanks for this PR, Fred. Looks quite nice for being such a non-trivial concept.

A few details below.

docs/reference/layer-specification.md
flotter: 1. I've reworded the comment at the top key level (```pairing:```) to better set the context, and highlights that for now pairing is only supported for HTTPS connections.

2. I've reordered this ```mode``` comment to focus first on the difference between ```single``` and ```multiple``` as I think this is the key thing to explain here.

internals/daemon/api_exec_test.go
flotter: I am not sure I follow - let me just double check:

```SetUpTest: register extension; start reaper; start daemon```
```TearDownTest: stop daemon; stop reaper; unregister extension```

So currently the teardown, as you say, are cleaning up in the opposite order to creation.

I think you are proposing moving the unregister first during TearDownTest ? Just checking ?

File: internals/overlord/pairingstate/export_test.go (hunk 31 above)
Comment (flotter) [reply internals/overlord/pairingstate/export_test.go]: I fixed the race you pointed out with a minor change, not yet visible here. Now I feel that using time.AfterFunc is working quite nicely for what I am trying to do. However, the fake version of this is not really testing the code against a real timer, making the coverage a little artificial.

Have you seen this library: https://go.dev/blog/synctest ?

It allows you to run a test inside a properly simulated timer environment. Unfortunately this does not yet work with checks.v1, but I could write the unit tests for this using Go's plain testing library.

This should be an improvement on the current approach ?

Comment (flotter) [reply internals/overlord/pairingstate/export_test.go]: Discussed with Ben. Will skip using synctest for now, but use a time based unit test approach.

Comment (flotter) [reply internals/overlord/pairingstate/export_test.go]: I reworked the unit tests to not fake the time.AfterFunc, and use a simpler time-based approach which works well. Marking this as resolved as I believe this is now fully addressed. Let me know if you want further changes here, then I will reopen the comment.

File: internals/overlord/pairingstate/manager.go (hunk 32 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go]: I removed this, as discussed in a previous comment. Resolving as this code no longer exists.

File: internals/overlord/pairingstate/manager.go (hunk 33 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go]: I fixed this after studying the next door projects.

pairingDetailsAttr => "pairing-details" for the key
pairingDetails for the struct.

Resolving.

File: internals/overlord/pairingstate/manager.go (hunk 34 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go:236]: Discussed with Ben as well. Work scheduled as part of https://github.com/canonical/pebble/issues/710.

Resolving.

File: internals/overlord/pairingstate/manager.go (hunk 35 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go]: I fixed this issue:

When we reach this line and find ```m.enabled``` true, there is a chance that the timeout handler already fired, but is blocking on the mutex (repeating your explanation).

The solution is to inspect the return value of the timer.Reset(...) function (inside ```startTimer```). If it returns ```false```and ```m.enabled``` is true,  we know the condition above is true, and in this case we have to prevent that handler occurrence from disabling the pairing window.

Let me know if you solution works for you.

Comment (flotter) [reply internals/overlord/pairingstate/manager.go]: I also added a unit test that specifically verifies that when the condition described above happens, we do not disable the pairing window, but extend it correctly without interruption.

File: internals/overlord/pairingstate/manager.go (hunk 36 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go]: Fixed, thanks. Resolving.

File: internals/overlord/pairingstate/manager.go (hunk 37 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go:285]: Removed. Resolving.

File: internals/overlord/pairingstate/manager.go (hunk 38 above)
Comment (flotter) [reply internals/overlord/pairingstate/manager.go]: Fixed both points. Resolving.

File: docs/reference/layer-specification.md [hunk 39]
```diff
@@ -279,4 +279,19 @@ log-targets:
     # be substituted using the environment for the corresponding service.
     labels:
       <label name>: <label value>
+
+# (Optional) HTTPS (using mTLS) communication between the client and server
+# requires both sides to be paired first. Pairing is only currently supported
```
benhoyt:
```suggestion
# requires both sides to be paired first. Pairing is currently only supported
```

docs/reference/layer-specification.md
```diff
@@ -279,4 +279,19 @@ log-targets:
...
+# for HTTPS transport (not HTTP or Unix Sockets).
```
benhoyt: Nit: socket wouldn't be capitalised:
```suggestion
# for HTTPS transport (not HTTP or Unix socket).
```

docs/reference/layer-specification.md
```diff
@@ -279,4 +279,19 @@ log-targets:
...
+pairing:
+
+    # (Optional) Controls the pairing mode of the server
```
benhoyt: Nit: others of these end with a `.`, so let's do the same here:
```suggestion
    # (Optional) Controls the pairing mode of the server.
```

docs/reference/layer-specification.md
```diff
@@ -279,4 +279,19 @@ log-targets:
...
+    #
+    # If set to 'single' pairing mode, only a single client can pair
```
benhoyt: I'd prefer this were done as a list -- let's use the format of `on-success`, see "Possible values are" elsewhere in this doc:

```
# (Optional) Controls the pairing mode of the server. Possible values are:
#
# - single: ...
# - multiple: ...
# - disabled (default): ...
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+	"time"
+
+	"github.com/canonical/pebble/internals/overlord/state"
+	"github.com/canonical/pebble/internals/plan"
+)
+
+// autoUsernameRangeLimit determines the maximum number suffix for
+// users auto-allocated by this package when a new certificate is
+// paired.
+const autoUsernameRangeLimit uint32 = 1000
```
benhoyt: It's more common for this sort of thing to be an untyped constant. Also, perhaps "max" rather than "range limit". Maybe:
```suggestion
const maxUsernameSuffix = 1000
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// Mode controls the pairing policy of the pairing manager.
+type Mode string
+
+const (
+	// ModeUnset is the same as ModeDisabled, but this value prevents the
+	// plan from marshalling the Mode explicitly.
+	ModeUnset Mode = ""
+	// ModeDisabled means no pairing is possible
```
benhoyt: Consistency nit:
```suggestion
	// ModeDisabled means no pairing is possible.
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+// pairingConfig contains the options exposed in the plan extension.
+type pairingConfig struct {
+	Mode Mode `yaml:"mode,omitempty"`
+}
+
+func (c *pairingConfig) Validate() error {
+	switch c.Mode {
+	case ModeUnset, ModeDisabled, ModeSingle, ModeMultiple:
+	default:
+		return fmt.Errorf("cannot support pairing mode %q: unknown mode", c.Mode)
```
benhoyt: I know we want top-level error messages to start with "cannot", but for these validation errors, would this be preferable? And perhaps include the possibles:
```suggestion
		return fmt.Errorf("invalid pairing mode %q; should be %q, %q, or %q",
		    c.Mode, ModeDisabled, ModeSingle, ModeMultiple)
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+}
+
+type PairingManager struct {
+	state *state.State
+	mu    sync.Mutex
+	// Plan config of the pairing manager.
+	pairingConfig *pairingConfig
+	// Persisted state of the pairing manager.
+	pairingDetails *pairingDetails
```
benhoyt: Given this is the "pairing manager" struct, can we shorten these field names? (Type names should probably keep "pairing".)
```suggestion
	config *pairingConfig
	// Persisted state of the pairing manager.
	details *pairingDetails
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	m.pairingDetails.Paired = true
+	m.state.Set(pairingDetailsAttr, m.pairingDetails)
+
+	return nil
+}
+
+// PairingWindowEnabled returns whether the pairing window is currently enabled.
+func (m *PairingManager) PairingWindowEnabled() bool {
```
benhoyt: Why does this have "window" in the name, but the enable func is just `EnablePairing`? Can we rename this to `PairingEnabled` to match?

internals/daemon/api_exec_test.go
benhoyt: Hmmm, it looks to me like you do already have them in reverse order in setup/teardown.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+	m.enabled = true
+	m.startTimer(timeout)
+	return nil
+}
+
+// generateUniqueUsername finds the first unique username following the pattern
+// "user-x" where x starts at 1 and monotonically increments. Usernames not
+// following this pattern will simply not be considered.
+func generateUniqueUsername(existingIdentities map[string]*state.Identity) (string, error) {
+	for i := uint32(1); i <= autoUsernameRangeLimit; i++ {
```
benhoyt: Why uint32 instead of plain int?
```suggestion
	for i := 1; i <= autoUsernameRangeLimit; i++ {
```

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,336 @@
...
+		m.stopTimer()
+	}()
+
+	// Verify that the client certificate is self-signed (the public
+	// key included must verify the signature). We do this here as a
+	// sanity check since we are about to persist this certificate and use
+	// it in exactly this way for future client credential checks. Note
+	// that the TLS handshake already proved that the client has access
+	// to the private key by verifying the handshake transcript signature
+	// using the public key in this certificate.
```
benhoyt, line 220: Thanks for the helpful comments like this!

docs/reference/layer-specification.md
flotter: Fixed. Resolving.

docs/reference/layer-specification.md
flotter: Fixed. Resolving.

docs/reference/layer-specification.md
flotter: Fixed. Resolving.

docs/reference/layer-specification.md
flotter: Fixed, thanks. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Yes, thank you. Done. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: This is much cleaner, thanks. Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Much better. Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Yes. Fixed. Resolving.

internals/overlord/pairingstate/manager.go
flotter: Fixed. Resolving.

docs/reference/layer-specification.md
flotter: Made some further tweaks here so this section uses a list, now similar to other sections in the document.

internals/overlord/pairingstate/package_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// 10% jitter
+	testJitter := timeout / 10
+
+	// 1. Window just opened, so should be enabled.
+	c.Assert(f(), Equals, true)
+
+	// 2. Let's wait for the timeout to almost elapse, so that we can
+	//    check the state towards the end of the window.
+	time.Sleep(timeout - testJitter)
```
benhoyt, line 112: I think this is going to be quite fragile. 10ms (10% of 100ms) is an eternity for an unloaded machine, but we often see much longer delays than this on loaded GitHub runners, causing tests to be fragile. Perhaps we can simplify to just "window open at first, wait double the timeout, definitely by then window should be closed".

```go
c.Assert(f(), Equals, true)
time.Sleep(2 * timeout)  // or timeout+100ms, I'm not sure
c.Assert(f(), Equals, false)
```

internals/overlord/pairingstate/manager_test.go
```diff
@@ -0,0 +1,402 @@
...
+// TestEnablePairingSingleModeAlreadyPaired verifies that we will fail to pair in
+// single pairing mode if already paired.
+func (ps *pairingSuite) TestEnablePairingSingleModeAlreadyPaired(c *C) {
+	ps.newManager(c, &pairingstate.PairingDetails{
+		Paired: true,
+	})
+	ps.updatePlan(pairingstate.ModeSingle)
+
+	err := ps.manager.EnablePairing(5 * time.Millisecond)
```
benhoyt: The 5 (actual timeout value) doesn't matter here, right? Perhaps just drop the `5 *` to avoid confusion.

internals/overlord/pairingstate/manager_test.go
```diff
@@ -0,0 +1,402 @@
...
+	c.Assert(ps.manager.PairingEnabled(), Equals, false)
+}
+
+// TestPairMTLSSuccess verifies that a successful pairing request closes the
+// pairing window and updates identities correctly.
+func (ps *pairingSuite) TestPairMTLSSuccess(c *C) {
+	ps.newManager(c, nil)
+	ps.updatePlan(pairingstate.ModeSingle)
+
+	err := ps.manager.EnablePairing(10 * time.Millisecond)
```
benhoyt: I think at least 100ms should be used here, given slow/loaded GitHub runners.

internals/overlord/pairingstate/manager_test.go
```diff
@@ -0,0 +1,402 @@
...
+	c.Assert(err, IsNil)
+	c.Assert(ps.manager.PairingEnabled(), Equals, true)
+
+	clientCert := generateTestClientCert(c)
```
benhoyt: Does generating this take any significant time? Perhaps it should be done before the `EnablePairing` call?

internals/daemon/api_exec_test.go
niemeyer: Sorry, no idea about what I actually read back then. It's clearly in the reverse order.

internals/overlord/pairingstate/manager.go
niemeyer: I don't yet see how that fixes the issue. Sounds like this is assuming that the result is a guarantee of the state of the function having run or not? This doesn't seem to be the case. From the docs:
> When Reset returns false, Reset neither waits for the prior f to complete before returning nor does it guarantee that the subsequent goroutine running f does not run concurrently with the prior one. If the caller needs to know whether the prior execution of f is completed, it must coordinate with f explicitly.

Instead of trying to fight the race, it would be nicer if we could not have the race in the first place, by rethinking how the timeout works.  Having two things just doing "foo = true/false" in parallel is the issue.

internals/daemon/api_exec_test.go
flotter: OK. Resolving.

docs/reference/layer-specification.md
flotter: Resolving.

internals/overlord/pairingstate/manager_test.go
flotter: Done. Resolving.

internals/overlord/pairingstate/manager_test.go
flotter: It is quick, but much safer done up front. Fixed, thank you. Resolving.

flotter: I updated all the tests to make sure client cert generation happens first.

internals/overlord/pairingstate/package_test.go
flotter: The tests have been tested under extreme load (```stress-ng --cpu 20 --io 20``` on my old machine), and under those conditions, I did get one failure within this function (and only this function) every now and then.

I've now increased the **jitter to 50%**, and I cannot simulate any failures anymore. The test is a little less accurate, but it would still capture if a code change breaks the fundamental mechanism.

Would you be comfortable with this change only ?

internals/overlord/pairingstate/manager_test.go
flotter: I changed all the tests to 100ms. This has no outcome on the overall tests time since only a few of them actually wait for the window to expire. Resolved.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,390 @@
...
+				}
+				enabled = false
+			} else {
+				if timer == nil {
+					// Create new on first request.
+					timer = time.NewTimer(*req.timeout)
+					timerCh = timer.C
+				} else {
+					// Reuse timer on subsequent requests.
+					timer.Stop()
```
flotter: remove

internals/overlord/pairingstate/manager.go
flotter: I've re-implemented this so that all access to the ```timer``` and the window ```enabled``` state is exclusively owned by a tomb, with all external accesses taking place through a single unbuffered channel, which includes a result that allows the caller to wait for completion of the request. All events from the unbuffered request channel and unbuffered timer channel is serviced by the same select, removing the possibility for races to exist.

Please let me know what you think :)

internals/overlord/pairingstate/package_test.go
benhoyt: Yep, that sounds reasonable -- thanks for stress testing.

flotter: Resolving.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,295 @@
...
+	// Load the paired state at startup.
+	m.state.Lock()
+	defer m.state.Unlock()
+	err := m.state.Get(pairingDetailsAttr, &m.details)
+	if errors.Is(err, state.ErrNoState) {
+		// Let's make sure the state always reflects the pairing state
+		// explicitly.
+		m.state.Set(pairingDetailsAttr, m.details)
+		err = nil
+	}
```
anpep:
```suggestion
	if err := m.State.Get(pairingDetailsAttr, &m.details); errors.Is(err, state.ErrNoState) {
		// Let's make sure the state always reflects the pairing state
		// explicitly.
		m.state.Set(pairingDetailsAttr, m.details)
	}
```

internals/overlord/pairingstate/manager.go
flotter: @niemeyer Latest implementation tries using the ensure feature, as we discussed. What do you think ?

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// Load the paired state at startup.
+	m.state.Lock()
+	defer m.state.Unlock()
+	err := m.state.Get(pairingDetailsAttr, &m.details)
+	if errors.Is(err, state.ErrNoState) {
+		// Let's make sure the state always reflects the pairing state
+		// explicitly.
+		m.state.Set(pairingDetailsAttr, m.details)
```
niemeyer: Is there a reason for this, considering that the value being stored is the default value anyway?  It seems that everything else should work fine without it, and if so I would probably not put the data in before we have to.

internals/overlord/pairingstate/manager.go
```diff
@@ -0,0 +1,295 @@
...
+
+	now := time.Now()
+	if now.Before(m.expiry) {
+		// The ensure call happened sooner, before the expiry time. We
+		// have to schedule another ensure call to try to reach the
+		// expiry, so we can disable the pairing window.
+		m.state.EnsureBefore(m.expiry.Sub(now))
+	} else {
+		// Expiry time has passed, disable the pairing window
+		m.enabled = false
```
niemeyer, line 158: That's indeed a lot nicer, thanks for working on it.

niemeyer (approved):
Thanks, Fred.  Just an optional trivial, but LGTM now.

Please get another pass from Ben, since this changed significantly.

internals/overlord/pairingstate/manager.go
flotter: Resolving.

internals/overlord/pairingstate/manager.go
flotter: Removed the explicit paired state while its false. Resolving.

PR #703, opened 2025-09-19
chore(golangci-lint): update to latest

flotter:
The current golangci/golangci-lint-action@v3 version used on CI selects an old version of golangci-lint. The problem is that the configuration file in the project is not compatible with the current v2 releases of golangci-lint, which means I cannot run the proposed fix line locally (without downgrading).

Update CI and the configuration to v2 golangci-lint.


benhoyt (approved):
Thanks for this!

PR #708, opened 2025-09-30
feat(daemon): mTLS support [2a/2]

flotter:
The daemon HTTPS server (when enabled) uses TLS v1.3 for the client server handshake.

The traditional TLS handshake does not provide us with a way to exchange identity certificates between client and server, since only the server certificate is sent to the client. This leaves the current TLS connection relying on basic authentication headers for passing a client identity to the server.

This patch migrates the HTTPS server over from basic authentication to using mutual TLS. The client (see https://github.com/canonical/pebble/pull/715) now has to supply a suitable client identity certificate that will be used for verifying the identity during normal API usage, although this requirement is **not enforced for end-points with open access**.

In #702 we added a pairing manager, which integrates with the daemon API permissions system to facilitate client server pairing during a carefully controlled pairing window on a dedicated API end-point ```v1/pairing``` **only**.

A followup task will provide an official event binding mechanism (trigger manager) by which the pairing window can be enabled when certain events are configured. For now, this cannot be enabled and therefore pairing is not possible. The the patch below hooking in the pairing window enable for testing purposes.

## State of HTTPS (mTLS) after this patch

- When Pebble enables the HTTPS server only mTLS connections will be allowed (basic authorization is no longer available on the HTTPS transport, but can still be used with HTTP)
- The Pebble client package adds (see https://github.com/canonical/pebble/pull/715):
     -  The ability to supply the client identity certificate, but this is not yet used by the CLI and therefore Pebble cannot use HTTPS today.
     - The ability to obtain the server identity certificate by making an insecure HTTPS connection. Can be used to implement an SSH-like workflow to implement a CLI warning asking the user to trust the server identity for future connections.
     - The ability to pair with the server by supplying the server identity fingerprint. This workflow requires obtaining the fingerprint by other means, such as DNSSD or via a display on the server side.
- The Pairing Manager provides the EnablePairing method, but this is not yet available, and requires future work to support binding events to actions such as pairing enable.

## Testing

The following test procedure can offer a demonstration how this works, but the process requires a small temporary patch to enable pairing:

1. Apply the following patch as a temporary step to enable the pairing on startup.

```diff
diff --git a/internals/overlord/pairingstate/manager.go b/internals/overlord/pairingstate/manager.go
index 833b979..c933ef9 100644
--- a/internals/overlord/pairingstate/manager.go
+++ b/internals/overlord/pairingstate/manager.go
@@ -131,6 +131,12 @@ func (m *PairingManager) Stop() {
 	m.stopTimer()
 }
 
+// TESTING HACK: Open the pairing window on startup.
+func (m *PairingManager) StartUp() error {
+	fmt.Printf("Pairing Enabled ...")
+	return m.EnablePairing(60 * time.Second)
+}
+
 // PairMTLS adds a client identity with admin permissions to the identity
 // subsystem. A pairing request always closes the pairing window.
 func (m *PairingManager) PairMTLS(clientCert *x509.Certificate) error {
```

2. Generate the client private key and certificate

```sh
~/> openssl genpkey -algorithm ed25519 -out client-key.pem

~/> openssl req -x509 -new -nodes -key client-key.pem -days 365 -out client-cert.pem -subj "/O=Test/CN=client" -addext "keyUsage=critical,digitalSignature" -addext "extendedKeyUsage=clientAuth" -addext "basicConstraints=critical,CA:FALSE"
```

3. Build pebble, run:

```sh
~/pebble> go build ./cmd/pebble
~/pebble> mkdir layers
~/pebble> echo -e "pairing:\n  mode: multiple" > layers/001-default.yaml
~/pebble> PEBBLE=$(pwd) ./pebble run --create-dirs --https=:8443
```

4. Test HTTPS (within the first 60 seconds of starting pebble):

```sh
# Not yet paired.
~/> curl -X GET --cert ./client-cert.pem --key ./client-key.pem --insecure https://localhost:8443/v1/plan?format=yaml

# Pair during open pairing window (will close window).
~/> curl -X POST -H "Content-Type: application/json" -d '{"action": "pair"}' --cert ./client-cert.pem --key ./client-key.pem --insecure https://localhost:8443/v1/pairing

{"type":"sync","status-code":200,"status":"OK"}

# Test HTTPS with mTLS enabled.
~/> curl -X GET --cert ./client-cert.pem --key ./client-key.pem --insecure https://localhost:8443/v1/plan?format=yaml

{"type":"sync","status-code":200,"status":"OK","result":"pairing:\n    mode: multiple\n"}
```

### DIFF
--- a/internals/daemon/api_pairing.go
+++ b/internals/daemon/api_pairing.go
@@ -20,13 +20,6 @@ import (
 )
 
 func v1PostPairing(c *Command, r *http.Request, user *UserState) Response {
-	if transport := RequestTransportType(r); transport != TransportTypeHTTPS {
-		// Note this should not be possible since the
-		// access permission check phase should prevent us getting
-		// here if the POST was received on the wrong transport.
-		return InternalError("cannot pair using transport type %q", transport)
-	}
-
 	var payload struct {
 		Action string `json:"action"`
 	}
--- a/internals/daemon/api_pairing_test.go
+++ b/internals/daemon/api_pairing_test.go
@@ -158,7 +158,7 @@ func (s *apiSuite) TestPairingPairNonHTTPS(c *C) {
 
 	result, ok := rsp.Result.(*errorResult)
 	c.Assert(ok, Equals, true)
-	c.Assert(result.Message, Equals, `cannot pair using transport type "http"`)
+	c.Assert(result.Message, Equals, `cannot find TLS connection state`)
 }
 
 // TestPairingPairMissingTLSState verifies that missing TLS state


internals/daemon/api_pairing.go
flotter, line 1: I need to add API docs to the PR.

internals/daemon/api_pairing.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	decoder := json.NewDecoder(r.Body)
+	if err := decoder.Decode(&payload); err != nil {
+		return BadRequest("cannot decode request body: %v", err)
+	}
+
+	switch payload.Action {
+	case "pair":
+		switch transport := RequestTransportType(r); transport {
+		case TransportTypeHTTPS:
```
benhoyt: I'd prefer to do a sanity check at the top of this function and return an InternalError. Something like:

```go
func v1PostPairing(c *Command, r *http.Request, user *UserState) Response {
    if transport := RequestTransportType(r); transport != TransportTypeHTTPS {
        return InternalError("cannot pair using transport type %q", transport)
    }
    ...
```

internals/daemon/api_pairing.go
```diff
@@ -0,0 +1,58 @@
...
+			if r.TLS == nil {
+				return BadRequest("cannot find TLS connection state")
```
benhoyt: Can this actually happen, or would it be an InternalError for `r.TLS` to be nil when it's a TransportTypeHTTPS?

internals/daemon/api_pairing.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+			// The leaf peer certificate is the client identity certificate.
+			clientCert := r.TLS.PeerCertificates[0]
+
+			pairingMgr := c.d.overlord.PairingManager()
+			if err := pairingMgr.PairMTLS(clientCert); err != nil {
+				return BadRequest("cannot pair client: %v", err)
+			}
+		default:
+			return BadRequest("cannot pair using %q transport type: not supported", transport.String())
```
benhoyt: Nit: %s and %q automatically call `String()` on the object if it supports it, so you can drop the `.String()` here.

internals/daemon/access.go
```diff
@@ -110,3 +110,42 @@ func (ac MetricsAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState)
...
+// PairingAccess is only intended for use as an access checker for the pairing
+// endpoint. This access checker allows a new mTLS client identity to be
+// forwarded to the pairing manager, without identity verification. This access
+// checker will only allow pairing requests while the pairing manager has its
+// pairing window enabled, which typically involves a proof of server ownership
+// procedure, such as a controlled power cycle or button press.
+type PairingAccess struct{}
+
+func (ac PairingAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	// This checker is only for pairing.
```
benhoyt: I would clarify slightly, something like:
```suggestion
	// This should only be called for /v1/pairing, but double-check here just in case.
```

internals/daemon/access.go
```diff
@@ -110,3 +110,42 @@ func (ac MetricsAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState)
...
+// pairingWindowEnabled simplifies testing without a pairing manager.
+var pairingWindowEnabled = (*Daemon).pairingWindowEnabled
+
+// PairingAccess is only intended for use as an access checker for the pairing
+// endpoint. This access checker allows a new mTLS client identity to be
+// forwarded to the pairing manager, without identity verification. This access
+// checker will only allow pairing requests while the pairing manager has its
+// pairing window enabled, which typically involves a proof of server ownership
+// procedure, such as a controlled power cycle or button press.
+type PairingAccess struct{}
```
benhoyt, line 123: I don't really understand the need for this new access checker. We have thorough checks for the transport type and the pairing window in `v1PostPairing` (and we should have thorough tests for those). So can we just not add a new access type here and rely on those?

internals/daemon/access.go
```diff
@@ -110,3 +110,42 @@ func (ac MetricsAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState)
...
+
+func (ac PairingAccess) CheckAccess(d *Daemon, r *http.Request, user *UserState) Response {
+	// This checker is only for pairing.
+	if r.URL.Path != "/v1/pairing" {
+		return Unauthorized(accessDenied)
+	}
+
+	// We only support pairing an mTLS client certificate at this point, so
+	// the transport has to be HTTPS.
+	if RequestTransportType(r) != TransportTypeHTTPS {
```
benhoyt, line 134: What about unix socket access, or user.Access == AdminAccess? Are they allowed?

internals/overlord/tlsstate/manager.go
```diff
@@ -137,13 +137,23 @@ func (m *TLSManager) ListenConfig() *tls.Config {
...
+		ClientAuth:       tls.RequestClientCert,
+		VerifyConnection: m.VerifyClientCertificate,
 	}
 	return tlsConf
 }
 
+func (m *TLSManager) VerifyClientCertificate(state tls.ConnectionState) error {
+	certs := len(state.PeerCertificates)
+	if certs != 1 {
+		return fmt.Errorf("expected one client certificate, received %v", certs)
```
benhoyt: Nit: prefer %d here as we know the type. Also perhaps `certs` -> `numCerts` for clarity:
```suggestion
		return fmt.Errorf("expected one client certificate, received %d", numCerts)
```

internals/overlord/state/identities_test.go
```diff
@@ -688,94 +698,156 @@ func (s *identitiesSuite) TestIdentityFromInputs(c *C) {
...
+		} else {
+			c.Assert(identity, IsNil)
+		}
+	}
+}
 
-	userID = 42
-	identity = st.IdentityFromInputs(&userID, "nancy-wrong-username", "wrong-password")
-	c.Assert(identity, IsNil)
+func uint32Ptr(v uint32) *uint32 {
```
benhoyt: Nit: I think it makes more sense for this to be generic, in case we want to use it elsewhere in tests:

```go
func ptr[T any](v T) *T {
    return &v
}

// And then usage would look like this:
userID: ptr(uint32(100))
```

internals/daemon/daemon.go
```diff
@@ -217,26 +218,42 @@ type Command struct {
...
-	}
-	if identity.Basic != nil {
-		// Prioritize basic type (HTTP basic authentication) and ignore UID in this case.
-		return &UserState{Access: identity.Access, Username: identity.Name}, nil
-	} else if identity.Local != nil {
-		return &UserState{Access: identity.Access, UID: userID}, nil
+	if identity != nil {
+		u := &UserState{
+			Access:   identity.Access,
+			Username: identity.Name,
```
benhoyt, line 251: Where is UserState.UID set now? We seem to have dropped this, and it's used by some of the notices APIs. It looks like in the notices tests we provide mocked `UserState{}` structs with UID set, so this code path isn't tested -- we should address that.

client/client.go
```diff
@@ -630,6 +685,66 @@ func newDefaultRequester(client *Client, opts *Config) (*defaultRequester, error
...
+		return fmt.Errorf("only one TLS server validation option allowed, but %d were provided", tlsOptions)
+	}
+
+	if opts.TLSServerFingerprint != "" {
+		// Client supplied fingerprint must match the server identity.
+		idFingerprint, err := getIdentityFingerprint(serverIDCert)
+		if err != nil {
+			return fmt.Errorf("cannot obtain identity fingerprint: %v", err)
+		}
+		if idFingerprint == opts.TLSServerFingerprint {
```
benhoyt: Should we use https://pkg.go.dev/crypto/subtle#ConstantTimeCompare here?

client/client.go
```diff
@@ -660,3 +775,15 @@ func (rq *defaultRequester) getWebsocket(urlPath string) (clientWebsocket, error
...
 }
+
+// getIdentityFingerprint extracts the public key from the certificate and
+// calculates the fingerprint.
+func getIdentityFingerprint(cert *x509.Certificate) (string, error) {
+	pubKey, ok := cert.PublicKey.(ed25519.PublicKey)
+	if !ok {
+		return "", errors.New("certificate must use Ed25519 public key")
+	}
+	hashBytes := sha512.Sum384(pubKey)
```
benhoyt: Is this always SHA-384?

client/client.go
ghislainbourgeois: In the client it would be unnecessary. `crypto/subtle` is really meant to prevent an attacker from learning anything new by timing the requests. In this particular case, it would really help with that, and can only increase latency.

client/client.go
ghislainbourgeois: This is calculating a new hash in the client from the received public key, to get a fingerprint. This could become an issue when sha-384 is not considered secure enough and we need to change this code. Fingerprints that were already accepted in the client would not match anymore and would fail.

This could be solved by either removing the trust of all previous fingerprints if this changes, or implement a way to upgrade the fingerprints in the future. This is not very concerning to me.

internals/daemon/daemon.go
```diff
@@ -217,26 +218,42 @@ type Command struct {
 	d *Daemon
 }
 
-func userFromRequest(st *state.State, r *http.Request, ucred *Ucrednet, username, password string) (*UserState, error) {
+func userFromRequest(st *state.State, r *http.Request, ucred *Ucrednet) (*UserState, error) {
```
ghislainbourgeois: This never returns an error, maybe the signature should be changed?

flotter:
The client code is now split into a separate PR (https://github.com/canonical/pebble/pull/715). Note that the end-to-end test proposal will only work with both PRs in place.

internals/daemon/api_pairing.go
flotter: Do we want to document pairing API? I think if we do, we need to document HTTPS properly. Could this be a following task for us for 2510 ?

Since the pairing API endpoint would unlikely be used by the normal Pebble workflow, and instead, perhaps the SSH-like workflow where you may have to approve a server identity via CLI. I think this will require a little more CLI work, which would be a good place to design the workflow, and document  it?

internals/daemon/api_pairing.go
flotter: Yes, this feels more explicit. I also like that this is not an internal error, since we cannot actually make this happen if the access permission phase did its job. Fixed, thank you.

flotter: Resolving.

internals/daemon/api_pairing.go
flotter: OK, thank you. Line no longer here, so will just resolve.

internals/daemon/api_pairing.go
flotter: This should not be possible. Demoted to InternalError. Resolving.

internals/daemon/access.go
flotter: Fixed. Resolving.

internals/daemon/access.go
flotter: You have a point.

We could literally leave the "v1/pairing" end-point open, since the HTTP handler can perform all the validity checks.

My thinking was to try and draw some hard lines between "obtaining permission" logic, and the HTTP handlers, which to me feels more like post approval code. I like the idea that the the end-point accesses checker gives an accurate picture of the intended usage, and by whom and under which conditions. The unit tests verifying all those assumptions are in one nice place, and easy to audit, and when you touch code there, you know you are on risky ground.

I will reach out to chat.

internals/daemon/access.go
flotter: The person making this pairing POST much do so using a TLS connection type so that we can extract the client identity certificate from the connection state. Unless I am missing a subtlety with your proposal, I think this makes any non-HTTPS connection type incompatible with pairing in its current design (calling MTLSPair which expects a client certificate) ? I will reach our to discuss.

internals/overlord/tlsstate/manager.go
flotter: Fixed, thank you. Resolving.

internals/overlord/state/identities_test.go
flotter: Yes, nice. Fixed. Resolving.

internals/daemon/daemon.go
flotter: Massive oversight on my side - I only looked as ```userString(...)```.

I added the missing code, with a detailed comment. Would like to discuss with you to explain my thinking.

client/client.go
flotter: We also omitted this in the past TLS PRs for similar reasons - will confirm with Ben.

client/client.go
flotter: This was chosen as per earlier specifications for now. I think there is room to increase this at a later point, easily.

internals/daemon/daemon.go
flotter: Thank you for the feedback!

Will discuss with Ben. I think the signature comes from Snapd, but we have already deviated from that, so we may as well remove the error if its not used.

internals/daemon/api_pairing.go
flotter: TLS docs to be planned for 2510 - Fred to talk to Nick.

internals/daemon/access.go
flotter: Agreed to leave permission related stuff in access checkers. Remove redundant checks in HTTP end-point code.

internals/daemon/access.go
flotter: Discussed, resolving.

internals/daemon/daemon.go
```diff
@@ -217,26 +218,57 @@ type Command struct {
...
-		return &UserState{Access: identity.Access, Username: identity.Name}, nil
-	} else if identity.Local != nil {
-		return &UserState{Access: identity.Access, UID: userID}, nil
+	if identity != nil {
+		u := &UserState{
+			Access:   identity.Access,
+			Username: identity.Name,
+		}
+
+		// TODO: The notices implementation does not yet support
```
flotter: Remove TODO.

client/client.go
flotter: Discussed, resolved.

client/client.go
flotter: Discussed, resolving.

internals/daemon/daemon.go
flotter: Agreed with Ben.

internals/daemon/api_pairing.go
flotter: Discussion started with Nick. Closing this comment.

internals/daemon/daemon.go
flotter: Reviewed this with Ben.

Will add end-to-end test to make sure notices are properly testing using this plumbing.

internals/daemon/daemon.go
flotter: Removed error return. Resolving.

internals/daemon/daemon.go
flotter: Fixed. Resolving.

internals/daemon/access.go
flotter: Removed the transport check from the HTTP end-point handler. Resolving.

internals/daemon/daemon.go
benhoyt: Thanks. Feel free to do that in a follow-up PR if you prefer.

flotter: OK. I actually pushed the unit tests, but I now reverted the commit, and I am going to do that in a follow-up, so this can get merged.

Resolving.

flotter:
Final approval happened for [6d79697](https://github.com/canonical/pebble/pull/708/commits/6d7969736c628ce34b353cddbe4ccd47b3e5cfbb)

I added another commit on top of this, and then reverted it back to this commit (I will move the unit test enhancements into a separate PR).

PR #714, opened 2025-10-20
fix: avoid error log when daemon retries "exec" tasks on startup

### LINKED ISSUE #587: exec tasks continuing after reboot
I have observed this while testing pebble in an environment with a watchdog, where the process can be abruptly killed, and restarted under certain conditions: 

The thing that is interesting for me is that if an exec command (which results in tasks/changes) is active during this time, these will get persisted, and upon restart will be continued.

1. Does this really make sense for exec ?
2. I think one of the task objects are no longer valid: cannot get exec setup object for task "8"

2025-03-12T15:23:06.9434164Z DEBUG 2025-03-12T15:15:46.476Z XXXX Running task 8 on Doing: Execute command "/bin/dlv"
2025-03-12T15:23:06.9434663Z DEBUG 2025-03-12T15:15:46.628Z XXXX 2025-03-12T15:15:46Z ERROR internal error: cannot get exec setup object for task "8"
2025-03-12T15:23:06.9435252Z 2025-03-12T15:15:46.640Z XXXX Change 3 task (Execute command "/bin/dlv") failed: internal error: cannot get exec setup object for task "8"

Maybe the failure is OK. However, in my particular case, if such an operation caused the crash / watchdog reset (which it did by design in this case), I would not want this to be repeated after reboot ?

benhoyt:
Exec tasks are not retryable (by design). However, if the Pebble daemon is hard-killed when an exec task is running, then when the daemon is restarted, the state engine tries to run that task (which was in Doing status) again, and returns an error from the task handler.

Simply log this at debug level instead of returning a task error, as it's not an error.

Before:

```
2025-10-20T22:29:41.286Z [pebble] Started daemon.
2025-10-20T22:29:41.287Z [pebble] Change 1 task (Execute command "sleep") failed: internal error: cannot get exec setup object for task "1"
2025-10-20T22:29:41.287Z [pebble] POST /v1/services 982.809µs 400 (http+unix)
2025-10-20T22:29:41.287Z [pebble] Cannot start default services: no default services
```

After:

```
2025-10-20T22:33:28.862Z [pebble] Started daemon.
2025-10-20T22:33:28.865Z [pebble] POST /v1/services 2.696031ms 400 (http+unix)
2025-10-20T22:33:28.866Z [pebble] Cannot start default services: no default services
```

To repro or test:

- Start `pebble run` in one terminal
- Run a long-running exec like `pebble exec sleep 1000` in another terminal
- `kill -9 <pebble_run_pid>`
- Restart `pebble run` and (before the fix) you'll see the error log on startup

Fixes #587

### DIFF
--- a/internals/overlord/cmdstate/handlers.go
+++ b/internals/overlord/cmdstate/handlers.go
@@ -76,7 +76,11 @@ func (m *CommandManager) doExec(task *state.Task, tomb *tomb.Tomb) error {
 	st.Unlock()
 	setup, ok := setupObj.(*execSetup)
 	if !ok || setup == nil {
-		return fmt.Errorf("internal error: cannot get exec setup object for task %q", task.ID())
+		// A previously-run task was in Doing status when the daemon was killed,
+		// and the state engine retries it when the daemon is next restarted.
+		// Just ignore so we don't retry exec tasks.
+		logger.Debugf("Cannot get exec setup object for task %q (retried after restart)", task.ID())
+		return nil
 	}
 
 	// Set up the object that will track the execution.


internals/overlord/cmdstate/handlers.go
```diff
@@ -76,7 +76,11 @@ func (m *CommandManager) doExec(task *state.Task, tomb *tomb.Tomb) error {
 	st.Unlock()
 	setup, ok := setupObj.(*execSetup)
 	if !ok || setup == nil {
```
tonyandrewmeyer, line 78: If I'm understanding things correctly, `setupObj` will be `nil` in the case we're interested in, and this will cover that case because `nil` can't be converted to an `execSetup`.

Would it be tighter to check if `setupObj` is `nil` in case there are other issues with the type check (that should error out)? Or do we want to just log in either case anyway?

benhoyt: There are actually four cases:

1. missing from cache (`!ok`)
2. in cache, but not an `*execSetup` type (`!ok`)
3. in cache, but a nil `*execSetup` value (`ok && setup==nil`)
4. in cache and a non-nil `*execSetup` value (`ok && setup!=nil`)

So this branch will be executed on 1, 2, or 3. Which I think is fine. Realistically, only 1 will happen because we've only put non-nil `*execSetup`s in the map. So we could handle 2 and 3 more "harshly", but I don't think it hurts to handle them the same way.

PR #715, opened 2025-10-21
feat(client): mTLS support [2b/2]

flotter:
The client side mTLS code was split from #708.

For the complete description of the work, please refer to the description in #708.

Pebble client changes adds:
- The ability to supply the client identity certificate, but this is not yet used by the CLI and therefore Pebble cannot use HTTPS today.
- The ability to obtain the server identity certificate by making an insecure HTTPS connection. Can be used to implement an SSH-like workflow to implement a CLI warning asking the user to trust the server identity for future connections.
- The ability to pair with the server by supplying the server identity fingerprint. This workflow requires obtaining the fingerprint by other means, such as DNSSD or via a display on the server side.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -235,6 +235,21 @@ func New(config *Config) (*Client, error) {
 	// Let's not mutate the input config.
 	localConfig := Config{}
 	if config != nil {
+		// Only one TLS server option is allowed.
+		tlsOptions := 0
+		if config.TLSServerFingerprint != "" {
+			tlsOptions += 1
+		}
+		if config.TLSServerIDCert != nil {
+			tlsOptions += 1
+		}
+		if config.TLSServerInsecure {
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -630,6 +685,66 @@ func newDefaultRequester(client *Client, opts *Config) (*defaultRequester, error
...
+	if opts.TLSServerFingerprint != "" {
+		tlsOptions += 1
+	}
+	if opts.TLSServerIDCert != nil {
+		tlsOptions += 1
+	}
+	if opts.TLSServerInsecure {
+		tlsOptions += 1
+	}
+	if tlsOptions > 1 {
```
benhoyt: Should we do this check only once when the client is created, rather than on every req?

flotter: Yes. Fixed. Resolving.

PR #716, opened 2025-10-22
chore(daemon): improve UserState test coverage

flotter:
The command method for serving end-points

```go
func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {...}
```

is also responsible for looking up named identities if one exists. Otherwise, it checks if a peer credential UID is available, even if no named local username is associated with it.

Following the addition of basic auth and certificate based identities, exactly what is passed to the response function

```go
type ResponseFunc func(*Command, *http.Request, *UserState) Response
```

specifically, ```UserState```, is more complicated to picture.

Add tests to cover all the identity types, to capture what we expect in each identity case.

The tests also highlight that components, such as ```notices```, only supporting UID (and not yet names identities) cannot work with all named identity types.

### DIFF
--- a/internals/daemon/daemon_test.go
+++ b/internals/daemon/daemon_test.go
@@ -1920,13 +1920,15 @@ func (s *daemonSuite) TestServeHTTPUserStateLocal(c *C) {
 	c.Check(rec.Code, Equals, http.StatusOK)
 
 	// Verify UserState for Local identity.
-	c.Check(capturedUser.Username, Equals, "localuser")
-	c.Check(capturedUser.Access, Equals, state.AdminAccess)
+	c.Assert(capturedUser, NotNil)
+	c.Assert(capturedUser.Username, Equals, "localuser")
+	c.Assert(capturedUser.Access, Equals, state.AdminAccess)
+	c.Assert(capturedUser.UID, NotNil)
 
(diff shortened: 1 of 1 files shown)


internals/daemon/daemon_test.go
```diff
@@ -1879,3 +1883,296 @@ func (t *transportTypeSuite) TestTransportTypeContext(c *C) {
...
+	c.Check(rec.Code, Equals, http.StatusOK)
+
+	// Verify UserState for Local identity.
+	c.Check(capturedUser.Username, Equals, "localuser")
+	c.Check(capturedUser.Access, Equals, state.AdminAccess)
+
+	// This specific expectation is only a temporary workaround to
+	// support code not yet supporting named identities, such as
+	// notices. UID should really be nil in this case.
+	c.Check(*capturedUser.UID, Equals, uint32(1000))
```
benhoyt: It would be good to have a `c.Assert(capturedUser.UID, NotNil)` before this line, so that the test fails rather than panics if we break the implementation. Similar throughout.

flotter: Since ```capturedUser``` itself is also a pointer that can be nil, I also added a similar check for it at the start. Also, a couple of other places required the same check as proposed here. Fixed. Resolving.

PR #717, opened 2025-10-29
chore: update doc Makefile and reformat terminal directives in doc source

dwilding:
This PR brings in functionality & fixes from an upcoming version of the doc starter pack. This is a partial upgrade of the starter pack, only bringing in changes related to building the docs. I'll do a full upgrade at a later date.

The format of the `{terminal}` directive is changing. Instead of using `:input:` to denote the terminal command entered, a blank line will separate the terminal command from the output. [More detail about the change](https://github.com/canonical/sphinx-terminal/releases/tag/1.0.0)

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -181,7 +181,7 @@ update: install
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %:
-	$(MAKE) —no-print-directory install
+	$(MAKE) --no-print-directory install
 	. $(VENV); $(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 cli-help:


docs/.sphinx/version
```diff
@@ -1 +1 @@
-1.1.0
+1.1.0+preview
```
dwilding, line 1: This PR brings in a selection of changes from the next starter pack release. I'll do a separate PR for the full upgrade (at a later date), which will bump this version number again.

docs/conf.py
dimaqq, line 1: wow that's a lot of added extensions....
I thought we has filtered_toctree before?

dwilding: These extensions were being loaded before, now we are seeing it explicitly ;) The actual real improvement here is that each extension is maintained as its own project with tests. ie our dependencies are now more rigorously maintained.

I see we're still using sphinx_filtered_toctree in our dependencies. From here:
https://github.com/canonical/sphinx-filtered-toctree

PR #721, opened 2025-11-18
feat: FIPS-140 support

dimaqq:


dimaqq:
@benhoyt I'd appreciate a high level review:
- project structure
- file names / conventions
- anything that stands out...

docs/explanation/security.md
```diff
@@ -49,4 +49,22 @@ Our intention is that projects that build on Pebble can [override how TLS connec
...
+
+The [`pebble` snap](https://snapcraft.io/pebble) provides a separate `fips` channel for FIPS 140-compliant builds:
+
+```{terminal}
+snap install pebble --classic --channel=fips
+```
+
+**FIPS mode restrictions:**
+- HTTPS server disabled: the `--https` flag returns an error
+- Certificate identities disabled: cannot add or authenticate using certificate-based identities
```
dimaqq, line 68: Cannot add because `Unmarshal` code uses x509 which is technically crypto.

Cannot auth is expected, because we avoid crypto.

benhoyt (commented):
Yeah, this looks like the right direction to me.

However, I'm wondering if we can simplify the identities part, so that there's very little code in identities_*fips.go -- perhaps just a couple of small functions that are called by the main code? This is to avoid duplication/drift between fips and nofips.

This is also complicated by the fact that I'm going to be moving the identities code out of the `state` package, but I guess whoever merges first wins (you almost certainly will :-). But that's another reason to keep the identities changes smaller.

internals/daemon/daemon_test.go
```diff
@@ -1172,6 +1171,10 @@ func (s *daemonSuite) TestRestartIntoSocketModePendingChanges(c *C) {
 }
 
 func (s *daemonSuite) TestRestartServiceFailure(c *C) {
+	if !httpsSupported {
+		c.Skip("HTTPS not supported in FIPS builds")
+	}
```
dimaqq: I'm not sure what happened here.
This test doesn't appear to require HTTPS, only HTTP.

HACKING.md
```diff
@@ -29,6 +29,37 @@ go install ./cmd/pebble
...
+- HTTPS health checks are blocked (only HTTP URLs are allowed)
+- HTTPS redirects from HTTP health checks are blocked
+
+To build Pebble with FIPS compliance:
+
+```
+go build -tags=fips ./cmd/pebble
+```
+
+To run Pebble in FIPS mode during development:
```
benhoyt: This is `pebble run` (intentional?), so:

```suggestion
To run the Pebble daemon in FIPS mode during development:
```

HACKING.md
```diff
@@ -29,6 +29,37 @@ go install ./cmd/pebble
...
+
+```
+go run -tags=fips ./cmd/pebble run
+```
+
+When running a FIPS build, only "local" (UID-based) identity authentication and HTTP-only communication is available. The daemon can only listen on Unix sockets and HTTP (not HTTPS), and health checks cannot use HTTPS URLs.
```
benhoyt: The second sentence here just repeats what's in the bullet points above, so:
```suggestion
When running a FIPS build, only "local" (UID-based) identity authentication and HTTP-only communication is available.
```

internals/overlord/checkstate/checkers.go
```diff
@@ -49,8 +49,12 @@ type httpChecker struct {
 }
 
 func (c *httpChecker) check(ctx context.Context) error {
+	if err := checkHTTPSURL(c.url); err != nil {
+		return err
+	}
+
 	logger.Debugf("Check %q (http): requesting %q", c.name, c.url)
-	client := &http.Client{}
+	client := createHTTPClient()
```
benhoyt: This package is nicely done.

internals/overlord/pairingstate/manager_test.go
```diff
@@ -170,6 +170,10 @@ func (ps *pairingSuite) TestEnablePairingUnknownMode(c *C) {
 // TestPairMTLSSuccess verifies that a successful pairing request closes the
 // pairing window and updates identities correctly.
 func (ps *pairingSuite) TestPairMTLSSuccess(c *C) {
+	if !mtlsPairingSupported {
```
benhoyt, line 173: It seems to me we should have something like this in the package itself:

```go
func (m *PairingManager) PairMTLS(clientCert *x509.Certificate) error {
    if !mtlsPairingSupported {
        return errors.New("mTLS pairing not supported in FIPS builds")
    }
    ...
}
```

benhoyt (commented):
I need to review it in detail, but I've skimmed over this again and it looks directionally right. Feel free to do any additional review/cleanup and move out of draft, then re-request review and I'll go over it thoroughly.

.devcontainer/devcontainer.json
dimaqq, line 1: This is temporary and reflect my local setup.
This file should not be merged.

PR #723, opened 2025-11-19
chore: test what gofumpt looks like

benhoyt:
DRAFT: do not merge without further discussion. This is the result of running `gofumpt -l -w .` on the Pebble codebase, to evaluate whether we want to use it.

The `time.Date` arguments, which gofumpt (I think properly) changes from `04` to `0o4`, should be changes to just `4` to avoid this. We clearly intend decimal here -- there will be some manual edits required to fix these. But the other octal changes like `0o777` look fine.

Most of the other changes look good.

The one change I wasn't happy about is the change from `}, {` between struct literals, to `},\n{`. But it seems it allows the one-line format sometimes, so I'm not sure if we can tweak that behaviour, or if it matters enough to worry about.
UPDATE: I discovered gofumpt is fine with this, we just had some inconsistent comments between struct entries. See [these commits](https://github.com/canonical/pebble/pull/723/files/f36d68da981f10aca367505a581d6fbfc1975118..8d6088c613d94a4beed06e2992a074954f7010bd) for the fix (we'll need to fix manually in a few other places too).

### DIFF
--- a/internals/daemon/access_test.go
+++ b/internals/daemon/access_test.go
@@ -40,9 +40,7 @@ func (s *accessSuite) TestAccess(c *C) {
 		userCheckErr    daemon.Response
 		metricsCheckErr daemon.Response
 		pairingCheckErr daemon.Response
-	}{{
-		// API source: Unix Domain Socket
-		//
+	}{{ // API source: Unix Domain Socket
 		// User: nil
 		apiSource:       daemon.TransportTypeUnixSocket,
 		user:            nil,
@@ -87,9 +85,7 @@ func (s *accessSuite) TestAccess(c *C) {
 		userCheckErr:    nil,
 		metricsCheckErr: nil,
 		pairingCheckErr: errUnauthorized,
-	}, {
-		// API source: HTTP
-		//
+	}, { // API source: HTTP
 		// User: nil
 		apiSource:       daemon.TransportTypeHTTP,
 		user:            nil,
@@ -134,9 +130,7 @@ func (s *accessSuite) TestAccess(c *C) {
 		userCheckErr:    errUnauthorized,
 		metricsCheckErr: errUnauthorized,
 		pairingCheckErr: errUnauthorized,
-	}, {
-		// API source: HTTPS
-		//
+	}, { // API source: HTTPS
 		// User: nil
 		apiSource:       daemon.TransportTypeHTTPS,
 		user:            nil,


client/changes_test.go
```diff
@@ -48,12 +48,12 @@ func (cs *clientSuite) TestClientChange(c *check.C) {
...
-			SpawnTime: time.Date(2016, 04, 21, 1, 2, 3, 0, time.UTC),
-			ReadyTime: time.Date(2016, 04, 21, 1, 2, 4, 0, time.UTC),
+			SpawnTime: time.Date(2016, 0o4, 21, 1, 2, 3, 0, time.UTC),
+			ReadyTime: time.Date(2016, 0o4, 21, 1, 2, 4, 0, time.UTC),
 		}},
 
-		SpawnTime: time.Date(2016, 04, 21, 1, 2, 3, 0, time.UTC),
-		ReadyTime: time.Date(2016, 04, 21, 1, 2, 4, 0, time.UTC),
+		SpawnTime: time.Date(2016, 0o4, 21, 1, 2, 3, 0, time.UTC),
+		ReadyTime: time.Date(2016, 0o4, 21, 1, 2, 4, 0, time.UTC),
```
dimaqq, line 56: I'd rather see `4` than `0o4` here.

At the same time, the hidden octal `04` was kind nasty, any change is better than no change.

internals/daemon/access_test.go
```diff
@@ -40,148 +40,142 @@ func (s *accessSuite) TestAccess(c *C) {
...
+		apiSource:       daemon.TransportTypeUnixSocket,
+		user:            nil,
+		openCheckErr:    nil,
+		adminCheckErr:   errUnauthorized,
+		userCheckErr:    errUnauthorized,
+		metricsCheckErr: errUnauthorized,
+		pairingCheckErr: errUnauthorized,
+	}, {
+		// User access: UntrustedAccess
+		apiSource:       daemon.TransportTypeUnixSocket,
```
dimaqq, line 54: I'm happy with the change, though:
- the comments look a bit inconsistent now, how come the "API" comment shares a line with the opening curly, but "Untrusted" doesn't?
- the first comment was meant to apply to the block, and now it doesn't... perhaps the this code needs to be refactored so that variable names or keys are used instead of comments

Something like:
```go
type accessTest struct {...}

var (
	untrustedUser = &daemon.UserState{Access: state.UntrustedAccess}
	metricsUser   = &daemon.UserState{Access: state.MetricsAccess}
...

var unixSocketTests = []accessTest{
	// User: nil
	{
		apiSource:       daemon.TransportTypeUnixSocket,
		user:            nil,
        ...
     },{
        ...
     }
}
...

func ... TestAccess (...) {
	tests := nil
	tests = append(tests, unixSocketTests...)
	tests = append(tests, httpTests...)
    ...
}
```

internals/overlord/state/change_test.go
```diff
@@ -589,21 +589,23 @@ var abortLanesTests = []struct {
...
 }{
-
 	// Some basics.
 	{
 		setup:  "*:do",
 		abort:  []int{},
 		result: "*:do",
-	}, {
+	},
+	{
```
dimaqq, line 598: This change seems to be the opposite of the test change above?

internals/timeutil/schedule.go
```diff
@@ -395,9 +394,7 @@ func randDur(a, b time.Time) time.Duration {
 	return randutil.RandomDuration(dur)
 }
 
-var (
-	timeNow = time.Now
-)
+var timeNow = time.Now
```
dimaqq, line 397: Looks like this only so that it can be mocked out in tests.

Is this a Go idiom, or does this dependency injection point deserve a comment?

internals/timeutil/schedule_test.go
```diff
@@ -350,87 +367,121 @@ func (ts *timeutilSuite) TestParseSchedule(c *C) {
...
+			expected: []*timeutil.Schedule{
+				{
+					WeekSpans: []timeutil.WeekSpan{
+						{Start: timeutil.Week{Weekday: time.Monday}, End: timeutil.Week{Weekday: time.Monday}},
+					},
+				}, {
+					WeekSpans: []timeutil.WeekSpan{
+						{Start: timeutil.Week{Weekday: time.Wednesday}, End: timeutil.Week{Weekday: time.Wednesday}},
+					},
+				},
```
dimaqq, line 472: perhaps a bit too verbose, but these are tests, no biggie.

internals/workloads/workloads_test.go
```diff
@@ -130,7 +130,8 @@ workloads:
     `,
 }, {
 	summary: "override replace policy",
-	layers: []string{`
+	layers: []string{
+		`
```
dimaqq, line 134: Does that technically change the leading whitespace in the multiline string literal?

internals/overlord/state/change_test.go
holmanb: I think that this is one of the spots that Ben mentioned in the PR description:

> The one change I wasn't happy about is the change from }, { between struct literals, to },\n{. But it seems it allows the one-line format sometimes, so I'm not sure if we can tweak that behaviour, or if it matters enough to worry about.
UPDATE: I discovered gofumpt is fine with this, we just had some inconsistent comments between struct entries. See [these commits](https://github.com/canonical/pebble/pull/723/files/f36d68da981f10aca367505a581d6fbfc1975118..8d6088c613d94a4beed06e2992a074954f7010bd) for the fix (we'll need to fix manually in a few other places too).

See the `// Some basics.` comment above. That comment doesn't appear to add significant value and could probably be removed.

internals/workloads/workloads_test.go
holmanb: It doesn't look like it. The same number of newlines exist within the backticks, a newline is just added before.

PR #725, opened 2025-11-20
feat: support forwarding log to syslog

### LINKED ISSUE #711: Add syslog support as a log forwarding target
Currently, Pebble only supports forwarding logs to Loki or OpenTelemetry targets.

As discussed with @benhoyt, our project [SONiC on Ubuntu](https://github.com/canonical/sonic-buildimage/) is adopting Pebble as the ROCKs service manager (to replace supervisord).

To maintain feature parity with the community SONiC version, we need Pebble to support forwarding logs to syslog, which is used by SONiC to collect logs from terminated containers.

We noticed that syslog support was already implemented in a [previous pull request](https://github.com/canonical/pebble/pull/165/files#diff-49f0faa8db587eadcb27b9f855c89ee11c9da76cd5d3bb6d977629a281ad1c3d), but it wasn’t merged into the main branch. It might be a good starting point for implementing this feature in the future.

jy5275:
This pull request adds a new log target type `syslog` to forward service logs to remote syslog servers, similar to the existing `loki` and `opentelemetry` targets. Syslog uses TCP/UDP connections as required by RFC 5424 (syslog) and RFC 5425 (octet framing) formats - that's different from the other two target types which use HTTP to send logs to remote targets. Except that difference, I try to make the implementation and file organization as consistent with those of `loki` and `opentelemetry` as possible.

This is loosely based on [syslog.go from earlier work in PR#165](https://github.com/canonical/pebble/pull/165/files#diff-49f0faa8db587eadcb27b9f855c89ee11c9da76cd5d3bb6d977629a281ad1c3d).

To validate, you can use the test syslog server in [this draft PR](https://github.com/canonical/pebble/pull/726).

Fixes #711

### DIFF
--- a/internals/overlord/logstate/syslog/syslog_benchmark_test.go
+++ b/internals/overlord/logstate/syslog/syslog_benchmark_test.go
@@ -31,6 +31,5 @@ func BenchmarkEncodeEntry(b *testing.B) {
 	b.ResetTimer()
 	for b.Loop() {
 		client.buildSendBuffer()
-		client.sendBuf.Reset()
 	}
 }


internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+	"net/url"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/canonical/pebble/internals/servicelog"
+)
+
+const (
+	defaultMaxRequestEntries    = 100
```
benhoyt: The "default" prefix is accurate, however, I think that's implied given they're `const`, and I'd rather a shorter name that matches the `loki` and `opentelemetry` package version, so let's call remove the `default` prefixes here:
```suggestion
	maxRequestEntries    = 100
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+	defaultSyslogInitialBackoff = 100 * time.Millisecond
```
benhoyt: In names like this, "syslog" isn't necessary either, as this is in the `syslog` package:
```suggestion
	initialBackoff = 100 * time.Millisecond
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+	defaultSyslogMaxBackoff     = 10 * time.Second
+	defaultDialTimeout          = 10 * time.Second
+	canonicalPrivEnterpriseNum  = 28978
+)
+
+// Syslog Priority values - see RFC 5424 6.2.1
+const (
+	FacilityUserLevelMessage = 1
```
benhoyt: Any reason this needs to be exported? And probably just include these in another section inside the above `const` block. 
```suggestion
	facilityUserLevelMessage = 1
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+	SeverityInformational    = 6
```
benhoyt: Similar here, no need to export this. Also, I wonder if we should make it "notice" (normal) rather than "informational", which seems a bit too low.
```suggestion
	severityNotice = 5
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+type ClientOptions struct {
+	MaxRequestEntries    int
+	TargetName           string
+	Location             string
+	SyslogInitialBackoff time.Duration
+	SyslogMaxBackoff     time.Duration
```
benhoyt: Again, because this is in the `syslog` package, there's no need for the "Syslog" prefix here:
```suggestion
	InitialBackoff time.Duration
	MaxBackoff     time.Duration
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+type entryWithService struct {
+	service   string
+	Priority  int
+	Version   int
+	Timestamp string
+	PID       string
+	MsgID     string
+	Message   string
```
benhoyt: I'd like to reduce this to just the values that are not hard-coded. It'll reduce memory consumption, and also make it slightly more efficient to render. On a per-message basis, I think we only need these fields (in addition to `service`):

```suggestion
	Timestamp string
	Message   string
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+
+type Client struct {
+	options *ClientOptions
+
+	// To store log entries, keep a buffer of size 2*MaxRequestEntries with a
+	// sliding window "entries" of size MaxRequestEntries.
+	buffer  []entryWithService
+	entries []entryWithService
+
+	// Store the custom labels(syslog's structured-data) for each service
```
benhoyt:
```suggestion
	// Store the custom labels (syslog's structured-data) for each service
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+		labels:   make(map[string]string),
+	}
+	c.entries = c.buffer[:0]
+	return c, nil
+}
+
+// SetLabels formats the given service's labels into a structured-data section
+// for a syslog message, according to RFC5424 section 6
+func (c *Client) SetLabels(serviceName string, labels map[string]string) {
+	if len(labels) == 0 {
```
benhoyt: This should actually be `labels == nil`, according to the interface spec.
```suggestion
	if labels == nil {
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,300 @@
...
+	c.data.Reset()
+
+	// TODO: do we need to sort all log entries by service name here?
+	for _, entry := range c.entries {
+		structuredData, ok := c.labels[entry.service]
+		if !ok {
+			structuredData = "-"
+		}
+
+		msg := fmt.Sprintf("<%d>%d %s %s %s %s %s %s %s",
```
benhoyt: `fmt.Sprintf` is relatively inefficient, and most of this data is just strings. In fact, much of it is constant strings. And the `fmt.Sprintf` will do a string allocation each log entry. I wonder if we can do something more like this, to do the octet framing without allocation:

```go
const prefix = "<13>1 " // priority 13 = 1*8+5 (facility user, priority notice), version 1
length := len(prefix)+len(timestamp)+1+len(hostname)+1+len(service)+1+len(pid)+3+len(structured)+1+len(message)
fmt.Fprintf(&c.data, "%d", length) // octet framing as per RFC 5425
c.data.WriteByte(' ')
c.data.WriteString(prefix)
c.data.WriteString(timestamp)
c.data.WriteByte(' ')
c.data.WriteString(hostname)
c.data.WriteByte(' ')
c.data.WriteString(service)
c.data.WriteByte(' ')
c.data.WriteString(pid)
c.data.WriteByte(' ')
c.data.WriteString(structured)
c.data.WriteByte(' ')
c.data.WriteString(message)
```

Or something like that. I know it's a bit harder to read, but I think it'll be worth it in terms of efficiency and avoiding allocations. I'm sure there's a more efficient way to do the `fmt.Fprintf` for the octet framing too, perhaps using a pre-allocated fixed-size byte slice on the Client and `strconv.AppendInt`.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+	// Store the custom labels(syslog's structured-data) for each service
+	labels map[string]string
+
+	// connection info
+	conn          net.Conn
+	address       *url.URL
+	waitReconnect time.Duration
+	closed        bool
+	hostname      string
+	data          bytes.Buffer
```
benhoyt: These two aren't "connection info", so should be separated from those with a blank line.

Also, the name "data" is a bit too generic (everything is data), so perhaps change to `sendBuf`.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+		options.SyslogMaxBackoff = defaultSyslogMaxBackoff
+	}
+	if options.DialTimeout == 0 {
+		options.DialTimeout = defaultDialTimeout
+	}
+}
+
+// NewClient creates a syslog client.
+func NewClient(options *ClientOptions) (*Client, error) {
+	hostname, err := os.Hostname()
```
benhoyt: To simplify testing, and avoid a global dependency at this level, I'd prefer if we added `Hostname string` to `ClientOptions` and passed it in from above. And then if it's `""` (empty string) we set it to `"-"` here.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		hostname = "-"
+	}
+
+	opts := *options
+	fillDefaultOptions(&opts)
+
+	u, err := url.Parse(opts.Location)
+	if err != nil || u.Host == "" {
+		u, err = url.Parse("//" + opts.Location)
```
benhoyt: I would prefer we were a bit stricter here (and below). Let's require `tcp://location` or `udp://location`. Something like:

```go
u, err := url.Parse(opts.Location)
if err != nil {
    return nil, fmt.Errorf("invalid syslog server location: %v", err)
}
if u.Scheme != "tcp" || u.Scheme != "udp" || u.Host == "" {
    return nil, fmt.Errorf("syslog server location must be in form 'tcp://host:port' or 'udp://host:port'")
}
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+	// sliding window "entries" of size MaxRequestEntries.
+	buffer  []entryWithService
+	entries []entryWithService
+
+	// Store the custom labels(syslog's structured-data) for each service
+	labels map[string]string
+
+	// connection info
+	conn          net.Conn
+	address       *url.URL
```
benhoyt: Let's call this field `location` as it comes from `opts.Location`.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+
+// SetLabels formats the given service's labels into a structured-data section
+// for a syslog message, according to RFC5424 section 6
+func (c *Client) SetLabels(serviceName string, labels map[string]string) {
+	if len(labels) == 0 {
+		delete(c.labels, serviceName)
+		return
+	}
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "[%s@%d", "pebble", canonicalPrivEnterpriseNum)
```
benhoyt: Due to other Pebble variants, we don't hardcode this anymore. Let's pass `cmd.ProgramName` in using a new field `ClientOptions.SDID` (string), falling back to `"pebble"` if not set.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+		// Zero removed elements to allow garbage collection
+		for i := len(c.entries); i < len(c.buffer); i++ {
+			c.buffer[i] = entryWithService{}
+		}
+	}
+
+	message := entry.Message
+	if len(message) > 0 && message[len(message)-1] == '\n' {
+		message = message[:len(message)-1]
+	}
```
benhoyt: This can be written as `strings.TrimSuffix(entry.Message, "\n")` -- which we can just write inline below.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+
+	c.entries = append(c.entries, entryWithService{
+		Priority:  priorityVal(FacilityUserLevelMessage, SeverityInformational),
+		Version:   1,
+		Timestamp: entry.Time.Format(time.RFC3339), // Format: 2021-05-26T12:37:01Z
+		PID:       "-",
+		MsgID:     "-",
+		Message:   message,
+		service:   entry.Service,
+	})
```
benhoyt, line 159: Per above, I think this will just become:
```suggestion
	c.entries = append(c.entries, entryWithService{
		Timestamp: entry.Time.Format(time.RFC3339), // Format: 2021-05-26T12:37:01Z
		Message:   message,
		service:   entry.Service,
	})
```

benhoyt: Also, we should use `time.RFC3339Nano` here for time accuracy (RFC 5424 allows fractional seconds).

benhoyt (commented):
I also realised that the UDP support here isn't right. For UDP, we should be using [RFC 5426](https://www.rfc-editor.org/rfc/rfc5426.html) instead of octet framing, which means we should be using one datagram per message. This will be a slightly different structure. So for now I think we should just restrict it to `tcp://...` locations only.

internals/overlord/logstate/syslog/syslog.go
benhoyt: Per voice discussion, we just want to support `tcp` for now, as UDP requires a slightly different approach.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (c *Client) ensureConnected(ctx context.Context) error {
+	if c.conn != nil {
+		return nil
+	} else if c.closed {
+		return fmt.Errorf("write to closed SyslogBackend")
+	}
+
+	if c.waitReconnect > 0 {
```
benhoyt: Per discussion with Henry, let's just remove the waitReconnect and expontential backoff.

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -0,0 +1,295 @@
...
+type testSyslogServer struct {
+	listener net.Listener
+	msgChan  chan string
+}
+
+func (s *testSyslogServer) close() error {
+	return s.listener.Close()
+}
+
+func (s *testSyslogServer) run() error {
```
benhoyt, line 272: Ideally we can simplify this similar to https://github.com/canonical/pebble/pull/165/files#diff-4381a454fdf2083e6046457179814146e2d8465dbd33ca69d4fcd0e4458c663eR153

jy5275: Done. 

I found that the code in the deleted PR doesn't work. Calling `listener.Close()` doesn't close the scanner built on top of the listener's connection. Instead, I made the mock server to read from the connection directly and ends when finding an EOF.

internals/overlord/logstate/gatherer.go
```diff
@@ -381,6 +382,11 @@ func newLogClient(target *plan.LogTarget) (logClient, error) {
 			UserAgent:  fmt.Sprintf("%s/%s", cmd.ProgramName, cmd.Version),
 			ScopeName:  cmd.ProgramName,
 		}), nil
+	case plan.SyslogTarget:
+		return syslog.NewClient(&syslog.ClientOptions{
+			TargetName: target.Name,
```
benhoyt: I don't think we're actually using `TargetName` in the syslog client, so we can drop this field from the struct.

My intention is that we'd explicitly pass in the `os.Hostname()` value, and `SDID: cmd.ProgramName` here.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+const (
+	maxRequestEntries          = 100
+	dialTimeout                = 10 * time.Second
+	canonicalPrivEnterpriseNum = 28978
+)
+
+// Syslog Priority values - see RFC 5424 6.2.1
+const (
+	facilityUserLevelMessage = 1
+	severityNotice           = 5
```
benhoyt: With hard-coding this as a string to avoid allocations (see below), we can drop these constants now.

internals/overlord/logstate/syslog/syslog.go
benhoyt: This hasn't been applied.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+	entries []entryWithService
+
+	// Store the custom labels(syslog's structured-data) for each service
+	labels map[string]string
+
+	// connection info
+	conn     net.Conn
+	location *url.URL
+	closed   bool
+	sendBuf  bytes.Buffer
```
benhoyt: `sendBuf` isn't "connection info", so there should be a blank line before this line to separate it from that group.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// priorityVal calculates the syslog Priority value (PRIVAL) from the given
+// Facility and Severity values. See RFC 5424, sec 6.2.1 for details.
+func priorityVal(facility, severity int) int {
```
benhoyt: Given my suggestion of hard-coding this as a string to avoid allocations (see below), we can get rid of the helper function.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+		options.MaxRequestEntries = maxRequestEntries
+	}
+	if options.DialTimeout == 0 {
+		options.DialTimeout = dialTimeout
+	}
+}
+
+// NewClient creates a syslog client.
+func NewClient(options *ClientOptions) (*Client, error) {
+
```
benhoyt: Style nit: let's avoid a blank line here (for consistency with other code and typical Go style).

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+	opts := *options
+	fillDefaultOptions(&opts)
+
+	if len(opts.Hostname) == 0 {
+		hostname, err := os.Hostname()
+		if err != nil {
+			hostname = "-"
+		}
+		opts.Hostname = hostname
+	}
```
benhoyt: Let's drop this block. My intention was that's we'd call `os.Hostname()` in gatherer.go at the level up (dependency injection), rather than fetch it here. Also, let's handle the `"-"` case when encoding the entry, as that's an encoding concern (the hostname itself isn't "-").

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+
+	u, err := url.Parse(opts.Location)
+	if err != nil {
+		return nil, fmt.Errorf("invalid syslog server location: %v", err)
+	}
+	if u.Scheme != "tcp" || u.Host == "" { // may support udp later
+		return nil, fmt.Errorf("invalid syslog server location %s, syslog server location must be in form 'tcp://host:port'", opts.Location)
```
benhoyt: Let's shorten and be consistent about what quotes will be used here: 
```suggestion
		return nil, fmt.Errorf(`invalid syslog server location %q, must be in form "tcp://host:port"`, opts.Location)
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	c := &Client{
+		location: u,
+		options:  &opts,
+		buffer:   make([]entryWithService, 2*opts.MaxRequestEntries),
+		labels:   make(map[string]string),
+	}
+	c.entries = c.buffer[:0]
```
benhoyt, line 86: Let's add `c.sendBuf.Grow(sendBufSize)` here, with `sendBufSize` a const of 4 * 1024. That will avoid allocations the first few log lines and make performance more predictable.

internals/overlord/logstate/syslog/syslog.go
benhoyt: This hasn't been applied yet.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// SetLabels formats the given service's labels into a structured-data section
+// for a syslog message, according to RFC5424 section 6
+func (c *Client) SetLabels(serviceName string, labels map[string]string) {
+	if len(labels) == 0 {
+		delete(c.labels, serviceName)
+		return
+	}
+	var buf bytes.Buffer
```
benhoyt: Let's use a `strings.Builder` here -- it's a drop-in replacement, but it's more efficient as it minimizes data copying.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+
+	if len(c.options.SDID) > 0 {
+		fmt.Fprintf(&buf, "[%s@%d", c.options.SDID, canonicalPrivEnterpriseNum)
+	} else {
+		fmt.Fprintf(&buf, "[%s@%d", "pebble", canonicalPrivEnterpriseNum)
+	}
```
benhoyt: Let's avoid code duplication, like so:

```go
sdid := c.options.SDID
if sdid == "" {
    sdid = "pebble"
}
fmt.Fprintf(&buf, "[%s@%d", sdid, canonicalPrivEnterpriseNum)
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+		// Zero removed elements to allow garbage collection
+		for i := len(c.entries); i < len(c.buffer); i++ {
+			c.buffer[i] = entryWithService{}
+		}
+	}
+
+	entry.Message = strings.TrimSuffix(entry.Message, "\n")
+
+	c.entries = append(c.entries, entryWithService{
+		Timestamp: entry.Time.Format(time.RFC3339Nano), // Format: 2021-05-26T12:37:01Z
```
benhoyt: Nit: update comment:
```suggestion
		Timestamp: entry.Time.Format(time.RFC3339Nano), // Format: 2021-05-26T12:37:01.123456789Z
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+	if c.conn != nil {
+		return nil
+	} else if c.closed {
+		return fmt.Errorf("write to closed SyslogBackend")
+	}
+
+	d := net.Dialer{Timeout: c.options.DialTimeout}
+	conn, err := d.DialContext(ctx, c.location.Scheme, c.location.Host)
+	if err != nil {
+		return fmt.Errorf("cannot connect to %s", c.location)
```
benhoyt: We're losing the original error here. Let's do:
```suggestion
		return fmt.Errorf("cannot connect to %s: %w", c.location, err)
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+		c.sendBuf.WriteByte(' ')
+		c.sendBuf.Write(msgBuf.Bytes())
+	}
+
+	_, err = io.Copy(c.conn, &c.sendBuf)
+	if err != nil {
+		// The connection might be bad. Close and reset it for later reconnection attempt(s).
+		c.conn.Close()
+		c.conn = nil
+		return fmt.Errorf("cannot send syslog: %v", err)
```
benhoyt: Nit:
```suggestion
		return fmt.Errorf("cannot send syslogs: %w", err)
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	c.resetBuffer()
```
benhoyt, line 249: I think we actually need to move this up, just above the `io.Copy`. The reason is because otherwise we've still got logs in the `sendBuf` and *also* in the entries buffer. In the error case, I think we should keep them in the sendBuf but reset the entries buffer.

Do you think we can add a test for that case? Happy to discuss what that might look like.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -0,0 +1,271 @@
...
+	for _, entry := range c.entries {
+		structuredData, ok := c.labels[entry.service]
+		if !ok {
+			structuredData = "-"
+		}
+
+		defaultPriority := priorityVal(facilityUserLevelMessage, severityNotice)
+
+		// Build the message first to calculate length
+		var msgBuf bytes.Buffer
```
benhoyt: This wasn't what I had in mind at all, as it doesn't actually reduce allocations. It avoids fmt.Sprintf, but we're still doing an extra allocation by building this new bytes.Buffer each time around the loop -- that's what the `sendBuf` is for.

See let's use something closer to my [original suggestion in this comment](https://github.com/canonical/pebble/pull/725#discussion_r2548126996). We can also get rid of the `strconv.Itoa`, as those (normally) cause a memory allocation too.

[See a benchmark using `go test -bench` here.](https://github.com/jy5275/pebble/pull/1)

```
BenchmarkEncodeEntryOld-16    	 8141732	       124.0 ns/op	     195 B/op	       3 allocs/op
BenchmarkEncodeEntry-16       	49393112	        24.05 ns/op	       0 B/op	       0 allocs/op
```

benhoyt: It's probably nice to leave this benchmark (just for the new/current version) and keep `encodeEntry` in its own function.

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -0,0 +1,294 @@
...
+type suite struct{}
+
+var _ = Suite(&suite{})
+
+func Test(t *testing.T) {
+	TestingT(t)
+}
+
+func (*suite) TestAddEntries(c *C) {
+	ln, err := net.Listen("tcp", "localhost:0")
```
benhoyt: It's a style nit, but let's prefer slightly more meaningful names, like `listener` here.

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+	serverStarted := make(chan struct{})
+	serverStopped := make(chan struct{})
+	go func() {
+		close(serverStarted)
+		err := srv.run()
+		c.Assert(err, IsNil)
+		close(serverStopped)
+	}()
+	<-serverStarted
```
benhoyt: What's the point of waiting for this? We're basically waiting for the goroutine to start, not synchronising with anything.

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -0,0 +1,294 @@
...
+		c.Assert(err, IsNil)
+	}
+
+	// `Add` and `SetLabels` shouldn't be order-dependent
+	client.SetLabels("svc2", map[string]string{
+		"env":     "production",
+		"version": "1.2.3",
+		"owner":   "team-2",
+	})
+	client.SetLabels("svc4", map[string]string{})
```
benhoyt, line 118: This should be `SetLabels("svc4", nil)`

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -0,0 +1,294 @@
...
+		err := srv.run()
+		c.Assert(err, IsNil)
+		close(serverStopped)
+	}()
+	<-serverStarted
+
+	client, err := syslog.NewClient(&syslog.ClientOptions{
+		Location:   "tcp://" + ln.Addr().String(),
+		TargetName: "test-target",
+		Hostname:   "test-machine",
```
benhoyt: I think we should specify `SDID: "sdid"` or similar here explicitly.

benhoyt (changes requested):
Good progress. A bunch of minor comments, and one significant one about how we encode the entries. Please be sure to go through and resolve all comments when they're done.

internals/overlord/logstate/syslog/syslog.go
jy5275: `Flush` resets send buffer at the beginning anyway. So we don't need to modify here.

internals/overlord/logstate/syslog/syslog.go
jy5275: I found that syslog should add a default label `"[pebble@28978]"`, just similar way as loki (`"pebble_service"`) and opentelemetry (`"service.name"`). So setting nil label is different from setting an empty label map.

internals/overlord/logstate/syslog/syslog_test.go
jy5275: Because for syslog setting nil label differs from setting empty label, and the former is tested by svc3, I still keep the empty map here.

internals/overlord/logstate/syslog/syslog.go
jy5275: In loki and opentelemetry implementation, the method `encodeEntry` was called by `Add`, not `Flush`. To keep syslog implementation consistent with them, I create a new method `buildSendBuffer` to for benchmark testing to call.

internals/overlord/logstate/syslog/syslog.go
jy5275: I think it might be better to call this before `c.sendBuf.Reset()` in `Flush()`? Because the required capacity can be better estimated with awareness of `len(entries)`.

internals/overlord/logstate/gatherer.go
```diff
@@ -383,9 +383,13 @@ func newLogClient(target *plan.LogTarget) (logClient, error) {
 			ScopeName:  cmd.ProgramName,
 		}), nil
 	case plan.SyslogTarget:
+		hostname, err := os.Hostname()
+		if err != nil {
+			return nil, fmt.Errorf("cannot get hostname: %w", err)
```
benhoyt: I suppose it's probably better to not send the hostname and keep going in case of an error here (maybe in some container environments there's no hostname?) So let's log and continue:
```suggestion
			logger.Noticef("Cannot get hostname for syslog: %v", err)
			hostname = ""
```

internals/overlord/logstate/gatherer.go
```diff
@@ -383,9 +383,13 @@ func newLogClient(target *plan.LogTarget) (logClient, error) {
...
+		}
 		return syslog.NewClient(&syslog.ClientOptions{
-			TargetName: target.Name,
-			Location:   target.Location,
+			Location: target.Location,
+			Hostname: hostname,
```
benhoyt, line 393: Per my [earlier comment](https://github.com/canonical/pebble/pull/725#discussion_r2566684674), we should pass in `cmd.ProgramName` here:

```go
SDID: cmd.ProgramName,
```

internals/overlord/logstate/syslog/syslog_benchmark_test.go
```diff
@@ -0,0 +1,36 @@
...
+	client.SetLabels("svc1", map[string]string{
+		"env":     "test",
+		"version": "0.0.1",
+	})
+	client.Add(entry)
+
+	b.ResetTimer()
+	for b.Loop() {
+		client.buildSendBuffer()
+		client.sendBuf.Reset()
```
benhoyt: Thanks for adding/keeping this benchmark. Looks good. However, we can remove this line now as buildSendBuffer() always resets this buffer (at the top of the function) between tests.

benhoyt (commented):
Nice! Just three very minor comments left.

In addition, I think we should do some performance testing against a real-world syslog server. Can you find a production syslog server to test against, and spin it up and then have a "service" which outputs thousands of logs and report the performance and **CPU and memory usage of Pebble** over time?

Perhaps just this as the service, sending logs (almost) as fast as possible.

```python
#!python3
import time
i = 0
while True:
    print(f'This is a log message to be sent via syslog {i}', flush=True)
    time.sleep(0.001)  # 1ms
    i += 1
```

PR #728, opened 2025-11-26
chore: run gopls modernize on the codebase

dimaqq:
This PR contains the changes that the Go language server recommends:

```sh
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix ./...
```

### DIFF
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -533,9 +533,7 @@ type HTTPCheck struct {
 // Copy returns a deep copy of the HTTP check configuration.
 func (c *HTTPCheck) Copy() *HTTPCheck {
 	copied := *c
-	if c.Headers != nil {
-		copied.Headers = maps.Clone(c.Headers)
-	}
+	copied.Headers = maps.Clone(c.Headers)
 	return &copied
 }
 


client/changes.go
```diff
@@ -34,8 +34,8 @@ type Change struct {
 	Ready   bool    `json:"ready"`
 	Err     string  `json:"err,omitempty"`
 
-	SpawnTime time.Time `json:"spawn-time,omitempty"`
-	ReadyTime time.Time `json:"ready-time,omitempty"`
+	SpawnTime time.Time `json:"spawn-time"`
+	ReadyTime time.Time `json:"ready-time"`
```
dimaqq, line 38: @benhoyt I wonder if `omitzero` was originally intended instead of `omitempty`...

Or perhaps this code was written before Go 1.24, so that was not an option.
I wonder what's the safe thing to do here.

Basically `time.Time` is never "empty", but it can be "zero".
Ref: https://stackoverflow.com/questions/32643815/json-omitempty-with-time-time-field

internals/overlord/logstate/loki/loki.go
```diff
@@ -88,9 +89,7 @@ func (c *Client) SetLabels(serviceName string, labels map[string]string) {
 
 	// Make a copy to avoid altering the original map
 	newLabels := make(map[string]string, len(labels)+1)
-	for k, v := range labels {
-		newLabels[k] = v
-	}
+	maps.Copy(newLabels, labels)
```
dimaqq: I reckon this would be more idiomatic:

```suggestion
    newLabels := maps.Clone(labels)
```

client/changes.go
```diff
@@ -61,8 +61,8 @@ type Task struct {
 	Log      []string     `json:"log,omitempty"`
 	Progress TaskProgress `json:"progress"`
 
-	SpawnTime time.Time `json:"spawn-time,omitempty"`
-	ReadyTime time.Time `json:"ready-time,omitempty"`
+	SpawnTime time.Time `json:"spawn-time"`
+	ReadyTime time.Time `json:"ready-time"`
```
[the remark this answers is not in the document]
dimaqq: Copilot is wrong here.

client/changes.go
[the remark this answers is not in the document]
dimaqq: Copilot is wrong here

dimaqq: I figured we should not change semantics in this PR, even if we eventually wanted to.

internals/plan/plan.go
```diff
@@ -543,10 +534,7 @@ type HTTPCheck struct {
 func (c *HTTPCheck) Copy() *HTTPCheck {
 	copied := *c
 	if c.Headers != nil {
-		copied.Headers = make(map[string]string, len(c.Headers))
-		for k, v := range c.Headers {
-			copied.Headers[k] = v
-		}
+		copied.Headers = maps.Clone(c.Headers)
```
benhoyt: I think we get rid of the `c.Headers != nil` check here in the same way as the others, because `maps.Clone` handles that case.

dimaqq: Good catch, fixed!

PR #731, opened 2025-12-01
ci: generate SBOM and run security scan in CI

james-garner-canonical:
This PR adds a workflow based on the one used in `ops` to generate a SBOM and run a security scan in CI.

It's configured to run when releasing Pebble, and can also be run manually.

---

Right now it's also configured to run on PRs to see if it works on this PR, [which it does](https://github.com/canonical/pebble/actions/runs/19876241944). This trigger should be removed before merging.

### DIFF
--- a/.github/workflows/sbom-secscan.yaml
+++ b/.github/workflows/sbom-secscan.yaml
@@ -3,7 +3,6 @@ name: SBOM and secscan
 on:
   push:
     tags: [v*]  # when publishing a release
-  pull_request:  # temporarily for testing in this PR
   workflow_dispatch:
 
 permissions: {}


.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,61 @@
...
+          ./sbomber submit
+      - name: Wait for the scans to finish
+        run: |
+          cd scanner
+          ./sbomber poll --wait --timeout 30
+      - name: Download the reports
+        run: |
+          cd scanner && ./sbomber download
+      - name: Upload reports
+        uses: actions/upload-artifact@v4
```
tonyandrewmeyer: This could be v5.

.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libapt-pkg-dev
+          sudo apt install -y python3-apt
```
benhoyt: What are these deps needed for? Can we get `sbomber` to install its own deps via uv?

Why `apt` in the last line and `apt-get` on others?

.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,61 @@
...
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  scan:
+    strategy:
+      fail-fast: false
+      matrix:
+        manifest: [snap]
```
tonyandrewmeyer: I'm not sure it's worth doing a matrix. It doesn't seem likely that there is going to another type of Pebble artefact in the near future (for FIPS there will be another snap, but I think that's going to be produced by a separate workflow run, and go into a separate track).

In Ops, we had to have this because at the time (maybe still now? I can't remember if this has been improved yet) you couldn't have two artefacts with the same name, and we have that because we have both sdist and wheel. That's not the case with Pebble, so even if there is another artefact it seems likely it will be able to go into the same manifest.

I can see an argument that it's convenient to have the workflow more-or-less the same as Ops, but it's only a couple of lines that change, so I don't think that matters here, and Concierge should also not need two manifests when we end up doing that.

.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,61 @@
...
+          sudo apt-get install -y libapt-pkg-dev
+          sudo apt install -y python3-apt
+      - name: Checkout security scanner
+        uses: actions/checkout@v5
+        with:
+          repository: canonical/sbomber
+          path: scanner
+          persist-credentials: false
+      - name: Install uv
+        uses: astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41  # v7.1.2
```
tonyandrewmeyer: This could be v7.1.4

.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,61 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        manifest: [snap]
+
+    name: SBOM generation
+    runs-on: [self-hosted, self-hosted-linux-amd64-jammy-private-endpoint-medium]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
```
tonyandrewmeyer: This could be v6.

.github/workflows/sbom-secscan.yaml
tonyandrewmeyer: You probably copied this from something I wrote 😊 but it seems odd to have `apt-get` and `apt` here, rather than just consistently use `apt`. Also, I'm not sure that `libapt-pkg-dev` is required any more.

tonyandrewmeyer (commented):
Some small suggestions, but two (connected) larger questions that I'd like to see answers for before approving:

1. Is there a reason to not use the [provided action](https://github.com/canonical/sbomber/blob/main/.github/actions/run/action.yaml) if we're going to use sbomber? This wasn't available at the time that ops was done (it's partly based on the ops one, which was itself based on an earlier one elsewhere in charming), which is why it's not used there. There's [an example of using it](https://github.com/canonical/sbomber/blob/main/.github/workflows/scanner-workflow-test.yaml).
2. If we're not going to use the action, then I'm not convinced that a matrix makes sense here. It was needed in ops because we have sdist and wheel name clashes. That doesn't seem likely here. I could be convinced the matrix makes sense, but would like to hear that argument.

.github/workflows/sbom-secscan.yaml
james-garner-canonical: Makes sense, I wasn't sure initially whether we'd be doing this for both the snap and the raw binaries somehow, but the latter seems redundant and also maybe not supported. I'll drop the matrix.

.github/workflows/sbom-secscan.yaml
james-garner-canonical: Yeah the inconsistency is a copy-paste from `ops`. Looking at the current readme, it does seem like `libapt-pkg-dev` may not be required.

.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,22 @@
+name: SBOM and secscan
+
+on:
+  push:
+    tags: [v*]  # when publishing a release
+  pull_request:  # temporarily for testing in this PR
```
tonyandrewmeyer: Don't forget to remove this :)

.github/workflows/sbom-secscan.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  scan:
+    runs-on: [self-hosted, self-hosted-linux-amd64-jammy-private-endpoint-medium]
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - name: sbomber
+        uses: canonical/sbomber/.github/actions/run@main
```
tonyandrewmeyer: I don't recall why we have a Zizmor conf in this repo but no CI or pre-commit that uses it. However, we should probably go along with what it says, which is that this should be pinned to a hash.

james-garner-canonical: Good call. This action has the same basic issue that we had with the interface tests workflow too, in that there's the action version, and then there's the repo version that it will use. It takes an input for the repo version, so I've set both the workflow version and the repo version to the hash of the current latest commit on `main`.

james-garner-canonical:
The [servstate](https://github.com/canonical/pebble/actions/runs/19913596015/job/57087603956?pr=731) test timed out on this final commit, but passed on a rerun. Definitely unrelated to this PR, which only touches a totally separate workflow.

PR #732, opened 2025-12-01
feat: support forwarding log to syslog via UDP

jy5275:
This PR adds support for forwarding logs to syslog via UDP, in addition to the existing TCP implementation.

Background:
Previous [PR](https://github.com/canonical/pebble/pull/725) implemented log target type `syslog`, but only via TCP connection.
However, [SONiC](https://github.com/canonical/sonic-buildimage/) uses UDP to forward syslog from rock container to host. So we want pebble to support forwarding logs syslog via UDP as well.

UDP syslog forwarding (RFC-5426):

- Raw log message doesn't begin with an octet length prefix.
- Each log entry is sent in a separate UDP datagram.
- No persistent connection is maintained; each flush creates a new UDP socket and sends the datagrams.
- Message size is subject to UDP and network MTU limits (recommended ≤2048 bytes).

### DIFF
--- a/internals/overlord/logstate/syslog/export_test.go
+++ b/internals/overlord/logstate/syslog/export_test.go
@@ -20,10 +20,6 @@ func GetBuffer(c *Client) []servicelog.Entry {
 	return c.buffer
 }
 
-func GetMessage(e servicelog.Entry) string {
-	return e.Message
-}
-
 func ResetBufferToIndex(c *Client, index int) {
 	c.resetBufferToIndex(index)
 }
--- a/internals/overlord/logstate/syslog/syslog_test.go
+++ b/internals/overlord/logstate/syslog/syslog_test.go
@@ -214,7 +214,7 @@ func (*suite) TestBufferFull(c *C) {
 
 			// Otherwise, check buffer message matches string
 			msg := expected[i].(string)
-			c.Assert(syslog.GetMessage(buffer[i]), Equals, msg)
+			c.Assert(buffer[i].Message, Equals, msg)
 		}
 	}
 


docs/reference/layer-specification.md
```diff
@@ -254,8 +254,11 @@ log-targets:
     # - loki: Use the Grafana Loki protocol. A "pebble_service" label is
     #   added automatically, with the name of the Pebble service as its value.
     # - opentelemetry: Use the OpenTelemetry protocol (OTLP). A "service.name"
-    # label is added automatically, with the name of the Pebble service as its
-    # value.
+    #   label is added automatically, with the name of the Pebble service as its
+    #   value.
+    # - syslog: Use the syslog protocol. A "${SDID}@28978" is added as SD-ID.
+    #   28978 is Canonical's enterprise number. ${SDID} has default value
+    #   "pebble" and can be override by custom name.
```
benhoyt: Thanks for updating this. However, let's just mention `pebble` explicitly here, as the *user* can't change that. And also let's mention the service name comes through as the syslog APP-NAME field. Something like:
```suggestion
    # - syslog: Use the syslog protocol. The syslog SD-ID field is set to "pebble@28978"
    #   (28978 is Canonical's enterprise number). The syslog APP-NAME field is set to
    #   the Pebble service name.
```

docs/reference/log-forwarding.md
```diff
@@ -21,7 +21,7 @@ log-targets:
 Required configuration:
 
 - `override`: How this log target definition is combined with other pre-existing definitions with the same name in the plan. Supported values are `merge` and `replace`.
-- `type`: The type of log target. Supported types are `loki` and `opentelemetry`.
+- `type`: The type of log target. Supported types are `loki`, `opentelemetry` and `syslog`.
 - `location`: The URL of the remote log target. For Loki, this needs to be the fully-qualified URL of the push API, including the API endpoint; use the format `http://<ip-address>:3100/loki/api/v1/push`. For OpenTelemetry, include the TCP port (normally 4318) without the API endpoint, for example: `http://<ip-address>:4318`.
```
benhoyt: Let's update the `location` text too, to include "For Syslog, this needs to be either `tcp://<host-or-ip>:<port>` or `udp://<host-or-ip>:<port>`." And we should probably change the other occurrences of `<ip-address>` here to `<host-or-ip>`.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -20,6 +20,8 @@ const (
 	dialTimeout                = 10 * time.Second
 	canonicalPrivEnterpriseNum = 28978
 	initialSendBufferSize      = 4 * 1024
+	syslogRFC5424Prefix        = "<13>1 " // priority 13 = 1*8+5 (facility user, priority notice), version 1
```
benhoyt: This package is all about "syslog" and "RFC 5424", so I'd rename this to something more specific, maybe `priorityAndVersionPrefix`.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -188,7 +190,32 @@ func (c *Client) Close() error {
...
+	if hostname == "" {
+		hostname = "-"
+	}
+
+	structuredData, ok := c.labels[entry.service]
+	if !ok {
+		structuredData = "-"
+	}
+
+	// Message format as per RFC 5424
```
benhoyt:
```suggestion
	// Message format as per RFC 5424: <PRI>VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID STRUCTURED-DATA MSG
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -250,6 +270,26 @@ func (c *Client) Flush(ctx context.Context) error {
...
+	for _, entry := range c.entries {
+		err := c.ensureConnected(ctx)
+		if err != nil {
+			return err
+		}
+
+		entry.message = entry.message[:min(len(entry.message), maxUDPMessageSize)]
+		c.encodeOneEntry(&entry)
+		_, err = io.Copy(c.conn, &c.sendBuf)
+		if err != nil {
```
benhoyt, line 295: Shouldn't we close `c.conn` and set it to nil here, so we'll try to reconnect next time around the loop?
```suggestion
		if err != nil {
			// Error occurred, close and reset connection so we reconnect next time around.
			c.conn.Close()
			c.conn = nil
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -250,6 +270,26 @@ func (c *Client) Flush(ctx context.Context) error {
...
+func (c *Client) flushUDP(ctx context.Context) error {
+	// For UDP, we send each message as a separate datagram (RFC 5426 section 3.1)
+	// UDP connections don't need persistent state, so we create a fresh connection
+	for _, entry := range c.entries {
+		err := c.ensureConnected(ctx)
+		if err != nil {
+			return err
+		}
+
+		entry.message = entry.message[:min(len(entry.message), maxUDPMessageSize)]
```
benhoyt: I would prefer we add a `...` to show that it's been truncated. To avoid allocations, let's add a `messageSuffix string` argument to `encodeOneEntry` and do something like this:

```go
messageSuffix := ""
if len(entry.message) > maxUDPMessageSize {
    messageSuffix = "..."
    entry.message = entry.message[:len(entry.message)-len(messageSuffix)]
}
c.encodeOneEntry(&entry, messageSuffix)

// ...
func (c *Client) encodeOneEntry(entry *entryWithService, messageSuffix string) {
    // ...
	c.sendBuf.WriteString(entry.message)
    if messageSuffix != "" {
    	c.sendBuf.WriteString(messageSuffix)
    }
}
```

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -287,3 +290,91 @@ func (s *testSyslogServer) run() error {
...
+		case <-time.After(2 * time.Second):
+			c.Fatal("timed out waiting for UDP message")
+		}
+	}
+
+	// Test message longer than max UDP size is truncated
+	longMsg := make([]byte, 5000)
+	for i := range longMsg {
+		longMsg[i] = 'A'
+	}
```
benhoyt:
```suggestion
	longMsg := bytes.Repeat([]byte("A"), 5000)
```

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -287,3 +290,91 @@ func (s *testSyslogServer) run() error {
...
+
+	c.Assert(client.Add(servicelog.Entry{
+		Time:    time.Date(2023, 12, 31, 12, 0, 0, 0, time.UTC),
+		Service: "svc2",
+		Message: string(longMsg),
+	}), IsNil)
+	c.Assert(client.Flush(context.Background()), IsNil)
+
+	select {
+	case shortMsg := <-msgChan:
```
benhoyt: It might be clearer to call `shortMsg` -> `truncatedMsg`:
```suggestion
	case truncatedMsg := <-msgChan:
```

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -287,3 +290,91 @@ func (s *testSyslogServer) run() error {
...
+		c.Check(len(longMsg) == 5000, Equals, true)
```
benhoyt: I don't see any need for this assertion. It's guaranteed to be 5000 on creation.

internals/overlord/logstate/syslog/syslog_test.go
```diff
@@ -287,3 +290,91 @@ func (s *testSyslogServer) run() error {
...
+		c.Check(len(shortMsg) < 4100, Equals, true) // +100 for syslog headers
```
benhoyt: Can we check for the exact message content (especially with truncation "..."), something like:

```go
c.Check(truncatedMsg, Equals, "<13>1 headers blah blah" + strings.Repeat("A", 4000-3) + "...")
```

docs/reference/log-forwarding.md
jy5275: Have updated here. Replaced all `<ip-address>` to `<host-or-ip>` in pebble docs (all are in log forwarding section.

internals/overlord/logstate/syslog/syslog_test.go
jy5275: Updated manually

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -250,6 +273,34 @@ func (c *Client) Flush(ctx context.Context) error {
...
+		}
+
+		messageSuffix := ""
+		if len(entry.message) > maxUDPMessageSize {
+			messageSuffix = "..."
+			entry.message = entry.message[:maxUDPMessageSize-len(messageSuffix)]
+		}
+
+		c.encodeOneEntry(&entry, messageSuffix)
+		_, err = io.Copy(c.conn, &c.sendBuf)
```
dimaqq, line 294: Wouldn't `io.Copy` re-chunk the messages as it sees fit?

I'm not sure what's most idiomatic in Go, something like `.Wite(msg)` instead?

jy5275: I just benchmarked the `io.Copy` and `c.conn.Write(c.sendBuf.Bytes())` cdoe. There's no significant performance difference between the two approaches:
```sh
# io.Copy(c.conn, &c.sendBuf)
$ go test -bench=BenchmarkFlushUDP -benchmem ./internals/overlord/logstate/syslog/
OK: 6 passed
goos: linux
goarch: amd64
BenchmarkFlushUDP-16              119164             10182 ns/op           16552 B/op          8 allocs/op
ok      github.com/canonical/pebble/internals/overlord/logstate/syslog  1.246s

# c.conn.Write(c.sendBuf.Bytes())
$ go test -bench=BenchmarkFlushUDP -benchmem ./internals/overlord/logstate/syslog/
OK: 6 passed
goos: linux
goarch: amd64
BenchmarkFlushUDP-16              116054             10693 ns/op           16552 B/op          8 allocs/op
ok      github.com/canonical/pebble/internals/overlord/logstate/syslog  1.273s
```

syslog_benckmark_test.go:
```go
func BenchmarkFlushUDP(b *testing.B) {
	conn, err := net.ListenPacket("udp", "localhost:0")
	...
	go func() {
		for {
			buf := make([]byte, 8192)
			_, _, err := conn.ReadFrom(buf)
		}
	}()

	client, err := NewClient(&ClientOptions{
		Location: "udp://" + conn.LocalAddr().String(),
	})
	...
	entries := []servicelog.Entry{
		{
			Time:    time.Date(2025, 1, 2, 15, 4, 5, 123456789, time.UTC),
			Service: "redis",
			Message: "This is a reasonably long log message to benchmark the flushUDP function.",
		}, {
			Time:    time.Date(2025, 1, 2, 15, 4, 6, 987654321, time.UTC),
			Service: "nginx",
			Message: "Another log message to include in the UDP flush benchmark.",
		},
	}
	client.SetLabels("redis", map[string]string{
		"env":     "test",
		"version": "0.0.1",
	})

	b.ResetTimer()
	for b.Loop() {
		for _, entry := range entries {
			client.Add(entry)
		}
		err = client.Flush(context.Background())
		if err != nil {
			b.Fatalf("Failed to flush entries: %v", err)
		}
	}
}
```

jy5275: I noticed that `io.Copy(conn, &sendBuf)` calls `sendBuf.WriteTo(conn)`, which then calls `conn.Write(sendBuf.buf[...])`. So they're essentially the same way.

benhoyt (approved):
Thanks for this -- looks good to me. As far as I'm concerned, this is ready to merge. Anything more from you @jy5275?

internals/overlord/logstate/syslog/syslog.go
dimaqq: @jy5275 you are quite right!

I've run this under a debugger and learnt so much about Go and interfaces 😅 
Ultimately, `fd_unix.go` knows that this is not a stream socket, and writes in one go.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -250,6 +273,35 @@ func (c *Client) Flush(ctx context.Context) error {
...
+		if err != nil {
+			// Error occurred, close and reset connection so we reconnect next time around.
+			c.conn.Close()
+			c.conn = nil
+			return fmt.Errorf("cannot send syslogs: %w", err)
+		}
```
dimaqq, line 302: **question:**

what happens after?
to my mind, the two cases we'd get an error are roughly:
- kernel buffers full (cpu is faster than the network card; we are sending too much)
- no route to host (ip route to the destination is not available)

I would imagine that we want to drop the entries we have already sent and preserve the entries that we haven't sent yet?

(some other policy is fine too).

When will the next attempt to send data out happen?

benhoyt: Ah yes, that's a good point. Once each entry is successfully sent, we should zero and update the current entry in `c.entries`. @jy5275 Can you please make that update? I'm not sure how best to do it offhand, but put something up and we'll review it.

jy5275: Thanks for pointing that out. I added a method `Client.resetBufferToIndex(int)`. It's similar to `Client.resetBuffer()`. But different from sending by TCP, the log entries are sent one-by-one in separated packet by UDP. So I choose something similar to deal with partial failure. Unit test was also enhanced to cover this method.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -155,7 +155,7 @@ func (c *Client) Add(entry servicelog.Entry) error {
 	entry.Message = strings.TrimSuffix(entry.Message, "\n")
 
 	c.entries = append(c.entries, entryWithService{
-		timestamp: entry.Time.Format(time.RFC3339Nano), // Format: 2021-05-26T12:37:01.123456789Z
+		timestamp: entry.Time,
```
benhoyt: I think at this point we can just get rid of `entryWithService` and use a `[]servicelog.Entry` type directly, as they have exactly the same fields.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -204,7 +204,10 @@ func (c *Client) encodeOneEntry(entry *entryWithService, messageSuffix string) {
 
 	// Message format as per RFC 5424: <PRI>VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID STRUCTURED-DATA MSG
 	c.sendBuf.WriteString(priorityAndVersionPrefix)
-	c.sendBuf.WriteString(entry.timestamp)
+
+	timeBuf := make([]byte, 0, 32)
```
benhoyt:
```suggestion
	timeBuf := make([]byte, 0, len("2006-01-02T15:04:05.999999999Z"))
```

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -232,7 +235,8 @@ func (c *Client) buildSendBufferTCP() {
 		}
 
 		// TCP: Octet framing as per RFC 5425: <length> <message>
-		frameLength := len(priorityAndVersionPrefix) + len(entry.timestamp) + 1 + len(hostname) + 1 +
+		// RFC3339Nano time string ("2023-12-31T12:00:00.123456789Z") has length 30
+		frameLength := len(priorityAndVersionPrefix) + 30 + 1 + len(hostname) + 1 +
```
benhoyt: Unfortunately, RFC3339Nano trims the trailing zeroes in the fractional seconds, so the fixed value of 30 isn't going to work. You can see this if you change one of the tests to 0 nanoseconds instead of 123456789, and the length is incorrect. What I suggest is that we do the `AppendTimestamp` logic in the caller, and pass `timeBuf` as an argument to `encodeOneEntry`. It's a little messy, but it'll work.

We should also change one of the tests to use 0 nanoseconds and fix the frame length, to test this case.

benhoyt (changes requested):
Almost there -- just one more issue with the RFC3339Nano timestamp length, I think.

internals/overlord/logstate/syslog/syslog.go
jy5275: Fixed. I also modified a test data to check if `nanosecond=0 ` can be handled correctly.

internals/overlord/logstate/syslog/export_test.go
```diff
@@ -14,12 +14,16 @@
...
+import "github.com/canonical/pebble/internals/servicelog"
 
-func GetBuffer(c *Client) []EntryWithService {
+func GetBuffer(c *Client) []servicelog.Entry {
 	return c.buffer
 }
 
-func GetMessage(e EntryWithService) string {
-	return e.message
+func GetMessage(e servicelog.Entry) string {
```
benhoyt: We can drop this function now, given that `Message` is an exported field.

benhoyt: Done.

internals/overlord/logstate/syslog/syslog.go
```diff
@@ -250,9 +268,56 @@ func (c *Client) Flush(ctx context.Context) error {
...
+	c.resetBuffer()
+	return nil
+}
+
+func (c *Client) resetBufferToIndex(last int) {
+	for i := range last {
+		c.entries[i] = servicelog.Entry{}
+	}
+	c.entries = c.entries[last:]
+}
```
dimaqq, line 316: **Question**: what's the rationale for this particular implementation?

Would the below be equivalent?

```suggestion
func (c *Client) resetBufferToIndex(last int) {
	c.entries = c.entries[last:]
}
```

benhoyt: It sets the entries we're done with to the zero value so that the Go garbage collector can free up any memory associated with them (assuming nothing else is referencing them). In this case, there are two strings that could potentially be freed, `Service` and `Message`.

jy5275: Here I followed the practice of `Client.resetBuffer()` to reset all entries to default value. Its code was borrowed from loki and opentelemetry, which was contributed by @IronCore864. I think the reason is to avoid memory consumption from outdated logs. If `MaxRequestEntries` is set to a very large value and the flushed logs are not cleared, the growing strings will cause memory leak and longer GC pause when clear them all together

PR #734, opened 2025-12-03
feat: FIPS in a branch

dimaqq:
Don't merge this to `master`.
This should eventually become its own branch.

The alternative to #721 
- rip out anything that uses `crypto/` imports
- rip out the 3rd party password hashing library
- minimum change to enforce HTTP URls in all `http.Client`'s
- inline change to enforce HTTP redirects in all `http.Client`'s
- a couple new tests

What else we may want:
- update `HACKING.md`: in this branch, talk about FIPS, in main branch, update release notes
- update the OpenAPI schema?
- more tests?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -600,7 +600,8 @@ func newDefaultRequester(client *Client, opts *Config) (*defaultRequester, error
 		}
 	}
 
-	requester.doer = &http.Client{Transport: requester.transport,
+	requester.doer = &http.Client{
+		Transport: requester.transport,
 		CheckRedirect: func(req *http.Request, via []*http.Request) error {
 			if req.URL.Scheme != "http" {
 				return errors.New("only HTTP redirects are allowed in FIPS builds")
--- a/internals/overlord/logstate/loki/loki.go
+++ b/internals/overlord/logstate/loki/loki.go
@@ -63,7 +63,8 @@ func NewClient(options *ClientOptions) (*Client, error) {
 	fillDefaultOptions(&opts)
 	c := &Client{
 		options: &opts,
-		httpClient: &http.Client{Timeout: opts.RequestTimeout,
+		httpClient: &http.Client{
+			Timeout: opts.RequestTimeout,
 			CheckRedirect: func(req *http.Request, via []*http.Request) error {
 				if req.URL.Scheme != "http" {
 					return errors.New("only HTTP redirects are allowed in FIPS builds")
--- a/internals/overlord/logstate/opentelemetry/opentelemetry.go
+++ b/internals/overlord/logstate/opentelemetry/opentelemetry.go
@@ -140,7 +140,8 @@ func NewClient(options *ClientOptions) (*Client, error) {
 	fillDefaultOptions(&opts)
 	c := &Client{
 		options: &opts,
-		httpClient: &http.Client{Timeout: opts.RequestTimeout,
+		httpClient: &http.Client{
+			Timeout: opts.RequestTimeout,
 			CheckRedirect: func(req *http.Request, via []*http.Request) error {
 				if req.URL.Scheme != "http" {
 					return errors.New("only HTTP redirects are allowed in FIPS builds")


.github/workflows/binaries.yml
```diff
@@ -64,9 +64,9 @@ jobs:
...
       env:
         ARCHIVE_FILE: ${{ steps.build.outputs.ARCHIVE_FILE }}
-        PEBBLE_VERSION: ${{ steps.build.outputs.PEBBLE_VERSION }}
+        RELEASE_TAG: ${{ github.event.release.tag_name }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       if: ${{ github.event_name == 'release' }}
       run: |
-        echo Uploading $ARCHIVE_FILE to release $PEBBLE_VERSION
-        gh release upload $PEBBLE_VERSION dist/$ARCHIVE_FILE
+        echo Uploading $ARCHIVE_FILE to release $RELEASE_TAG
```
dimaqq: Before, released Pebble version directly matched the tag for release names (tag names) that we use: `vX.Y.Z` without dashes.

In the FIPS branch, we're introducing a small inconsistency: `v1.2.3-fips` release procudes `v1.2.3+fips` snap.

I wonder if we should simplify this, and just use `v1.2.3-fips` with dashes, like Juju does (`
4.1/edge | 4.1-beta1-de8f8db`).

WDYT @benhoyt ?

dimaqq: I've switched versioning to `v1.2.3-fips` for releases and `v1.2.3-fips-1234567` for merges to the branch.

internals/cli/cmd_run.go
benhoyt, line 1: We should remove the `--https` args and the `sharedRunEnterOpts.HTTPS` field.

internals/overlord/checkstate/checkers.go
```diff
@@ -51,7 +52,24 @@ type httpChecker struct {
...
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if req.URL.Scheme != "http" {
+				return errors.New("only HTTP redirects are allowed in FIPS builds")
+			}
+			// Allow HTTP redirects up to 10 times (Go default)
+			if len(via) >= 10 {
+				return errors.New("stopped after 10 redirects")
+			}
+			return nil
+		}}
```
benhoyt: Looks like the braces here aren't matched (with the opening brace of `http.Client{`:
```suggestion
		},
	}
```

client/client.go
```diff
@@ -693,58 +600,23 @@ func newDefaultRequester(client *Client, opts *Config) (*defaultRequester, error
...
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if req.URL.Scheme != "http" {
+				return errors.New("only HTTP redirects are allowed in FIPS builds")
+			}
+			// Allow HTTP redirects up to 10 times (Go default)
+			if len(via) >= 10 {
+				return errors.New("stopped after 10 redirects")
+			}
+			return nil
+		}}
```
benhoyt: Let's put one `http.Client` field per line, and match the braces. Similar elsewhere.
```suggestion
	requester.doer = &http.Client{
		Transport: requester.transport,
		CheckRedirect: func(req *http.Request, via []*http.Request) error {
			if req.URL.Scheme != "http" {
				return errors.New("only HTTP redirects are allowed in FIPS builds")
			}
			// Allow HTTP redirects up to 10 times (Go default)
			if len(via) >= 10 {
				return errors.New("stopped after 10 redirects")
			}
			return nil
		},
	}
```

dimaqq: Done in all three affected http clients, thanks!

internals/overlord/checkstate/checkers.go
dimaqq: done in all four http clients

internals/cli/cmd_run.go
dimaqq: Done, thanks; also updated docs, as there's an automated test that validates that docs match.

PR #736, opened 2025-12-03
ci: upload failed snapcraft build logs

dimaqq:
When the remote build fails, this log file is not pushed to github

```log
025-12-03 05:57:45.458 Build completed.
Log files: snapcraft-pebble-c1af907944ce9386fa3cf069eee5c5b3_ppc64el_2025-12-03T05:57:44.txt
Artifacts:
2025-12-03 05:57:45.458 Cleaning up
```

hopefully, with the extra info, we'd know better what went wrong.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -102,6 +102,7 @@ jobs:
             ~/.cache/snapcraft/log/
             ~/.local/state/snapcraft/log/
             ${{ env.SNAP_NAME }}_*.txt
+            snapcraft-${{ env.SNAP_NAME }}-*.txt
 
       - name: Attach ${{ steps.build-snap.outputs.snap }} to GH workflow execution
         uses: actions/upload-artifact@v4


.github/workflows/snap.yml
```diff
@@ -102,6 +102,7 @@ jobs:
             ~/.cache/snapcraft/log/
             ~/.local/state/snapcraft/log/
             ${{ env.SNAP_NAME }}_*.txt
+            snapcraft-${{ env.SNAP_NAME }}-*.txt
```
tonyandrewmeyer, line 105: Do we also need this on the local build? I had a bit of a search but can't find what code writes to this pattern (but from your log extract, something clearly does). The pattern in the line above can be seen in `fetch_logs()` in craft-application. Maybe something in the remote build adds the `snapcraft-` prefix? Maybe this is the correct name and the previous one is wrong? (It seems odd that both could be the case).

dimaqq: > Maybe something in the remote build adds the snapcraft- prefix?

that's my guess as well.

dimaqq:
Going to merge this in, to see if we get anything out of it.
Can always revert later.

PR #744, opened 2025-12-08
refactor: move identities code out of state package

### LINKED ISSUE #710: Move identities core code out of state package
Per discussion with Gustavo, we'd like to move the `state.*Identities` functions out of the core `state` package and into their own package (`overlord/identitiesstate` or `overlord/identities`) to avoid cluttering the core state package. They'll still be stored in the state database, but via `state.Get` and `state.Set` with an "identities" key.

Note that there will be a data migration associated with this, as currently the "identities" dict lives at the top level, whereas after this it'll live under "data". This means we'll need some legacy code pull it out on startup -- either as part of the `state` package or in a separate migration step.

benhoyt:
This moves the identities state handling out of the main `state` package and into its own `overlord/identities` package, to avoid cluttering the state package with more and more domain concerns. We use `State.Get` and `State.Set` to store the identities data under an "identities" key. We also keep a local copy of the in-memory identities map to avoid having to call Get every time it's read (which unmarshals the data from JSON every time).

In the state JSON file, currently the "identities" dict lives at the top level, whereas after this it lives under the "data" field (Get/Set data). Hence the small amount of code in patch2 to unmarshal `LegacyIdentities` from the top level and move it into data.

Most of the lines of code in this PR are trivial changes due to the refactor, for example `state.AdminAccess` -> `identities.AdminAccess`.

In a follow-up PR I'd like to refactor the Identity types. We currently have `Identity`, `apiIdentity` (without sensitive data, for the API), and `marshalledIdentity` (for state). This is a bit messy. We should probably move the API-related marshalling and validation into the daemon (api_identities.go).

The automated tests cover most cases, I believe. I've also done manual testing to ensure the identities feature is working, and the migration/upgrade works as expected.

Fixes #710.

### DIFF
--- a/internals/overlord/identities/identities_test.go
+++ b/internals/overlord/identities/identities_test.go
@@ -25,7 +25,6 @@ import (
 	. "gopkg.in/check.v1"
 
 	"github.com/canonical/pebble/internals/overlord/identities"
-	"github.com/canonical/pebble/internals/overlord/patch"
 	"github.com/canonical/pebble/internals/overlord/state"
 )
 
@@ -318,94 +317,6 @@ func (s *identitiesSuite) TestUnmarshalState(c *C) {
 			Name:   "mary",
 			Access: identities.AdminAccess,
 			Local:  &identities.LocalIdentity{UserID: 1000},
-		},
-	})
-}
-
-func (s *identitiesSuite) TestUnmarshalStateLegacy(c *C) {
-	data := []byte(`
-{
-    "data": {},
-    "identities": {
-        "bob": {
-            "access": "read",
-            "local": {
-                "user-id": 42
-            }
-        },
-        "mary": {
-            "access": "admin",
-            "local": {
-                "user-id": 1000
-            }
-        }
-    }
-}`)
-
-	st, err := state.ReadState(nil, bytes.NewReader(data))
-	c.Assert(err, IsNil)
-	err = patch.Apply(st)
-	c.Assert(err, IsNil)
-	mgr, err := identities.NewManager(st)
-	c.Assert(err, IsNil)
-
-	st.Lock()
-	defer st.Unlock()
-
-	c.Assert(mgr.Identities(), DeepEquals, map[string]*identities.Identity{
-		"bob": {
-			Name:   "bob",
-			Access: identities.ReadAccess,
-			Local:  &identities.LocalIdentity{UserID: 42},
-		},
-		"mary": {
-			Name:   "mary",
-			Access: identities.AdminAccess,
-			Local:  &identities.LocalIdentity{UserID: 1000},
-		},
-	})
-}
-
-func (s *identitiesSuite) TestUnmarshalStateNewAndLegacy(c *C) {
-	// If both new and legacy are present, it should prefer the new
-	// (and emit a warning log, but we don't test for that).
-	data := []byte(`
-{
-	"data": {
-		"identities": {
-			"bob": {
-				"access": "read",
-				"local": {
-					"user-id": 42
-				}
-			}
-		}
-	},
-    "identities": {
-        "mary": {
-            "access": "admin",
-            "local": {
-                "user-id": 1000
-            }
-        }
-    }
-}`)
-
-	st, err := state.ReadState(nil, bytes.NewReader(data))
-	c.Assert(err, IsNil)
-	err = patch.Apply(st)
-	c.Assert(err, IsNil)
-	mgr, err := identities.NewManager(st)
-	c.Assert(err, IsNil)
-
-	st.Lock()
-	defer st.Unlock()
-
-	c.Assert(mgr.Identities(), DeepEquals, map[string]*identities.Identity{
-		"bob": {
-			Name:   "bob",
-			Access: identities.ReadAccess,
-			Local:  &identities.LocalIdentity{UserID: 42},
 		},
 	})
 }
--- a/internals/overlord/state/state.go
+++ b/internals/overlord/state/state.go
@@ -221,6 +221,7 @@ func (s *State) UnmarshalJSON(data []byte) error {
 
 // LegacyIdentities is exported for use in patch2.go to perform the migration.
 func (s *State) LegacyIdentities() json.RawMessage {
+	s.reading()
 	return s.legacyIdentities
 }
 
--- a/internals/overlord/patch/patch2_test.go
+++ b/internals/overlord/patch/patch2_test.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2026 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package patch_test
+
+import (
+	"bytes"
+
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/pebble/internals/overlord/identities"
+	"github.com/canonical/pebble/internals/overlord/patch"
+	"github.com/canonical/pebble/internals/overlord/state"
+)
+
+type patch2Suite struct{}
+
+var _ = Suite(&patch2Suite{})
+
+func (s *patch2Suite) TestLegacyIdentities(c *C) {
+	restore := patch.FakeLevel(2, 1)
+	defer restore()
+
+	data := []byte(`
+{
+    "data": {"patch-level": 1},
+    "identities": {
+        "bob": {
+            "access": "read",
+            "local": {
+                "user-id": 42
+            }
+        },
+        "mary": {
+            "access": "admin",
+            "local": {
+                "user-id": 1000
+            }
+        }
+    }
+}`)
+
+	st, err := state.ReadState(nil, bytes.NewReader(data))
+	c.Assert(err, IsNil)
+	err = patch.Apply(st)
+	c.Assert(err, IsNil)
+	mgr, err := identities.NewManager(st)
+	c.Assert(err, IsNil)
+
+	st.Lock()
+	defer st.Unlock()
+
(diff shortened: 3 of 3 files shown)


internals/overlord/identities/state.go
```diff
@@ -0,0 +1,76 @@
+package identities
```
benhoyt, line 1: These types and code are directly copied from `state/state.go`.

internals/daemon/daemon_test.go
```diff
@@ -2040,11 +2043,13 @@ func (s *daemonSuite) TestServeHTTPUserStateBasicUnixSocket(c *C) {
 	hashedPassword, err := crypt.Generate([]byte("test"), nil)
 	c.Assert(err, IsNil)
 
+	identitiesMgr := d.overlord.IdentitiesManager()
+	c.Assert(err, IsNil)
 	d.state.Lock()
```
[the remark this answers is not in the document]
benhoyt: True, thanks - fixed.

internals/daemon/api_identities_test.go
```diff
@@ -22,22 +22,29 @@ import (
...
 		"mary": {
-			Access: state.AdminAccess,
-			Local:  &state.LocalIdentity{UserID: 1000},
+			Access: identities.AdminAccess,
+			Local:  &identities.LocalIdentity{UserID: 1000},
+		},
+		"nancy": {
+			Access: identities.MetricsAccess,
+			Basic: &identities.BasicIdentity{
+				Password: "$6$F9cFSVEKyO4gB1Wh$8S1BSKsNkF.jBAixGc4W7l80OpfCNk65LZBDHBng3NAmbcHuMj4RIm7992rrJ8YA.SJ0hvm.vGk2z483am4Ym1", // "test"
```
james-garner-canonical, line 46: Adding a test case for `*****`ing the password here (and below) is nice. Presumably unrelated to the changes in this PR and these tests would pass on `main` right now.

internals/overlord/identities/identities.go
```diff
@@ -213,13 +254,13 @@ func (d *Identity) UnmarshalJSON(data []byte) error {
 
 // AddIdentities adds the given identities to the system. It's an error if any
 // of the named identities already exist.
-func (s *State) AddIdentities(identities map[string]*Identity) error {
-	s.reading()
-
+//
+// The state lock must be held for the duration of this call.
+func (m *Manager) AddIdentities(identities map[string]*Identity) error {
```
james-garner-canonical, line 259: It took me a little bit to understand the changes here in combination with the new comment. Testing my understanding:

The new comment applies equally to the old version, it was just more obvious that `State` was accessed and potentially modified since a reference was taken directly.

Locked access to the state is now enforced by `State.Set`, which calls `writing` internally. Whereas previously `State` was accessed more directly, calling `reading` and `writing` manually.

internals/overlord/state/state.go
```diff
@@ -243,12 +197,19 @@ func (s *State) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	s.data = unmarshalled.Data
+
+	// Load legacy identities if new identities are not in Get/Set data.
+	if !s.data.has("identities") && len(unmarshalled.LegacyIdentities) > 0 {
```
james-garner-canonical: It seems like we shouldn't be able to easily get into a state where we have both the new `data["identities"]` and the old top-level key, and this conditional encodes that assumption by silently ignoring `LegacyIdentities` if we already have `data["identities"]`. Is that the best approach, or would it be worth emitting a warning about a bad database migration if we end up with both?

james-garner-canonical: `s.data.has("identities")` is a second source of truth for `identitiesKey` in `identities.go`. I guess maybe this is fine since it's only for the data migration. To avoid a circular dependency, I guess the only way to dry this would be defining `identitiesKey` in a third place, which may not be worth it.

internals/overlord/state/state.go
```diff
@@ -243,12 +197,19 @@ func (s *State) UnmarshalJSON(data []byte) error {
...
+		logger.Noticef("Loaded legacy identities from state file")
+		if s.data == nil {
+			s.data = make(customData)
+		}
+		s.data["identities"] = &unmarshalled.LegacyIdentities
```
james-garner-canonical: Testing my understanding: this works because we no longer call `unmarshalIdentities` on `LegacyIdentities` here, so `s.data["identities"]` is in the state that the identity manager expects to load it in.

internals/overlord/identities/identities_test.go
```diff
@@ -259,13 +273,55 @@ func (s *identitiesSuite) TestMarshalState(c *C) {
...
+		},
+		"mary": {
+			Name:   "mary",
+			Access: identities.AdminAccess,
+			Local:  &identities.LocalIdentity{UserID: 1000},
+		},
+	})
+}
+
+func (s *identitiesSuite) TestUnmarshalStateLegacy(c *C) {
```
james-garner-canonical: Great to see explicit tests of loading the legacy identities field. Do you think it's worth adding a test for bad data like this:
```
{
  "data": {
    "identities": {
      "bob": {...}
    }
  },
  "identities": {
    "mary": {...}
  }
}
```
Where the currently implemented behaviour is that `mary` would silently be dropped.

internals/overlord/identities/identities.go
benhoyt: Yeah, that understanding's correct -- it was true before, but now that it's not in the `state` package or on the `State` type, I thought it was worth clarifying.

internals/daemon/api_identities_test.go
benhoyt: Yeah, that was unrelated. I probably should have added it in my follow-up PR which moves the API "starring-out" into the daemon/API package. But I'll leave it here now.

internals/overlord/state/state.go
benhoyt: Correct, we shouldn't be able to get into that state. Maybe I'm a coward, but returning an error and stopping felt too strong. :-) But it's a good idea to emit a warning. I've added that.

benhoyt: Yeah, I don't think it's worth it. Plus, this code path is tested in identities_test.go.

internals/overlord/identities/identities_test.go
benhoyt: Good idea to add a test -- I've done that now.

internals/overlord/identities/identities.go
```diff
@@ -26,12 +26,53 @@ import (
...
+	m.state.Lock()
+	defer m.state.Unlock()
+
+	// Read existing identities from state, if any.
+	var marshalled map[string]*marshalledIdentity
+	err := st.Get(identitiesKey, &marshalled)
+	if err != nil && !errors.Is(err, state.ErrNoState) {
+		return nil, err
+	}
+	m.identities, err = unmarshalIdentities(marshalled)
```
dimaqq, line 60: **Observation:**

here, we're relying on the fact/assumption that if the error is ErrNoState, then `marshalled` is left empty, but valid to access.

I think this is a little fragile: for example, is `nil` ok?

internals/overlord/state/state.go
```diff
@@ -243,12 +197,23 @@ func (s *State) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	s.data = unmarshalled.Data
+
+	// Load legacy identities if present (and new identities are not in Get/Set data).
+	if len(unmarshalled.LegacyIdentities) > 0 {
+		if s.data.has("identities") {
+			logger.Noticef("WARNING: both new and legacy identities found in state file, ignoring legacy")
```
dimaqq: **Question:** do we whack the legacy identities from disk after we've loaded them into RAM the first time? After we've loaded them into RAM and certainly saved them in the new format?

internals/overlord/identities/identities.go
benhoyt: Yeah, we are, but I think that's okay. It's the only thing that would make sense, as `ErrNoState` means it wasn't even in the data, so `State.Get` won't be touching `marshalled`. I've confirmed this by reading the code.

internals/overlord/state/state.go
benhoyt: Yes (though implicitly). As soon as any state data is written to the next time, the new data format will be written to disk (via `State.Unlock`). That will normally happen right away (due to say the autostart change on startup), but even if it doesn't and you exit Pebble, the state file will be untouched and just load it from the legacy position next time.

dimaqq: So in summary, when we do write, we'll both remove the legacy key and add the new key?
Sounds good, because that's atomic.

dimaqq (approved):

dmitry-lyfar (approved):
@benhoyt LGTM. It's a bit pity that we cannot use the overlord's patch system for that kind of one-shot migration and will have to live with `LegacyIdentities` in `marshalledState` (we'd have to leave something like `Identities json.RawMessage` as part of the `State` for a patch to work I suppose). After some thinking, your approach looks more reasonable. 

Maybe one of the hidden advantages of this PR would also be an ability to migrate your marshaled identities if needed similarly to how snapd patches do it for the snap setup struct.

internals/overlord/state/state.go
```diff
@@ -264,29 +219,9 @@ func (s *State) UnmarshalJSON(data []byte) error {
...
-			cert, err := x509.ParseCertificate(block.Bytes)
-			if err != nil {
-				return fmt.Errorf("cannot parse certificate from cert identity: %w", err)
-			}
-			s.identities[name].Cert = &CertIdentity{X509: cert}
-		}
-	}
-	return nil
+// LegacyIdentities is exported for use in patch2.go to perform the migration.
+func (s *State) LegacyIdentities() json.RawMessage {
```
dmitry-lyfar, line 223:
```suggestion
func (s *State) LegacyIdentities() json.RawMessage {
  s.reading()
```
?

internals/overlord/identities/identities_test.go
```diff
@@ -259,15 +274,58 @@ func (s *identitiesSuite) TestMarshalState(c *C) {
...
```
dmitry-lyfar: Maybe this and the next one can go to patch2_test.go? snapd uses this approach by mocking current patch levels.

I'm not entirely sure it is possible.

internals/overlord/state/state.go
benhoyt: Done, thanks.

internals/overlord/identities/identities_test.go
benhoyt: Yeah, fair point. Done (with `patch.FakeLevel` similar to snapd tests).

PR #746, opened 2025-12-09
ci: upload errors better

dimaqq:
I've simplified the CI changes I made when debugging the s390x and ppc64el failures.

I believe these changes will come in handy next time something breaks.

### DIFF
--- a/docs/reference/layer-specification.md
+++ b/docs/reference/layer-specification.md
@@ -254,17 +254,22 @@ log-targets:
     # - loki: Use the Grafana Loki protocol. A "pebble_service" label is
     #   added automatically, with the name of the Pebble service as its value.
     # - opentelemetry: Use the OpenTelemetry protocol (OTLP). A "service.name"
-    # label is added automatically, with the name of the Pebble service as its
-    # value.
+    #   label is added automatically, with the name of the Pebble service as its
+    #   value.
+    # - syslog: Use the syslog protocol. The syslog SD-ID field is set to "pebble@28978"
+    #   (28978 is Canonical's enterprise number). The syslog APP-NAME field is set to
+    #   the Pebble service name.
     type: loki
 
     # (Required) The URL of the remote log target.
     # For Loki, this needs to be the fully-qualified URL of the push API,
     # including the API endpoint, for example:
-    #     http://<ip-address>:3100/loki/api/v1/push
+    #     http://<host-or-ip>:3100/loki/api/v1/push
     # For OpenTelemetry, this needs to include the TCP port (normally 4318)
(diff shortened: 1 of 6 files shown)


.github/workflows/snap.yml
```diff
@@ -94,14 +94,19 @@ jobs:
         with:
           snapcraft-args: remote-build --build-for=${{ matrix.arch }} --launchpad-accept-public-upload
 
+      - name: Rename the logs file, for colons are not allowed in the artifact name
+        if: always()
+        run: cat snapcraft-${{ env.SNAP_NAME }}-*.txt > snapcraft-logs.txt || echo "no log files"
```
tonyandrewmeyer, line 99: I remember when you were showing me the logs previously it wasn't completely obvious when one file ended and another started, with them merged. It would obviously be less clean and simple, but do you think it would be worth starting each one with the name of the file? If you think it worths fine as a plain concat that's fine with me.

.github/workflows/snap.yml
```diff
@@ -94,14 +94,19 @@ jobs:
...
+
       - uses: actions/upload-artifact@v4
-        if: failure()
+        if: always()
```
tonyandrewmeyer, line 102: Were there cases where the job succeeded but we needed the logs? It seems like this uploads a lot of noise in cases where we don't generally need it. We don't really see the noise and I guess GitHub covers the cost, but it does seem a bit inefficient.

tonyandrewmeyer (approved):
Opps, I meant to select approve, not request changes!

.github/workflows/snap.yml
benhoyt: I agree -- I wouldn't think we need this in success cases.

Also, isn't `if: always()` equivalent to not having the `if:` line at all?

tonyandrewmeyer: I think `if: always()` includes cancellations and failures, and `if: success()` is the default.

.github/workflows/snap.yml
dimaqq: In this case there's actually only one file.

What I was showing you were two different sources, that are located in different directories.

.github/workflows/snap.yml
dimaqq: The default is not to run this step when some step above (snapcraft remote-build) fails.

Getting logs on failures requires some kind of custom `if: ...` clause.

What I found useful was to contrast logs from failing vs good runs.

These artifacts are not persisted very long (weeks?), so I figured that the cost is low.
I can change this to `if: failure()` in both steps if that's better.

PR #748, opened 2025-12-10
ci: temporarily disable ppc64el and s390x snap builds

dimaqq:
ppc64el and s390x remote build are broken in our infrastructure at the moment.
This PR disables these architectures for the time being.

ref: https://chat.canonical.com/canonical/pl/mg7gxzii5pydjjx3wjnutpqpne

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -63,6 +63,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
+        # TODO: temporarily pause ppc64el and s390x builds while builders are broken
+        # arch: ["amd64", "arm64", "armhf", "ppc64el", "riscv64", "s390x"]
         arch: ["amd64", "arm64", "armhf", "riscv64"]
     steps:
       - name: Checkout Pebble repo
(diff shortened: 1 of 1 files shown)


.github/workflows/snap.yml
```diff
@@ -63,7 +63,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        arch: ["amd64", "arm64", "armhf", "ppc64el", "riscv64", "s390x"]
+        arch: ["amd64", "arm64", "armhf", "riscv64"]
```
benhoyt, line 68: Because we still need these builds, and this change is only temporary, let's comment them out and add a TODO comment. Something like this (here and below):
```suggestion
        # TODO: temporarily pause ppc64el and s390x builds as those builders are broken
        # arch: ["amd64", "arm64", "armhf", "ppc64el", "riscv64", "s390x"]
        arch: ["amd64", "arm64", "armhf", "riscv64"]
```

dimaqq: Done!

PR #754, opened 2025-12-11
client: improve error message when pulling non-existent files

### LINKED ISSUE #722: Provide a clearer error message when pulling a non-existent file
With the Pebble snap (1.25):

```
$ pebble pull /tmp/file-does-not-exit.txt /tmp/test
error: expected first field name to be "files", got "response"
```

If it's the destination that doesn't exist, the error is nicer:

```
$ pebble pull /tmp/file.txt /tmdsfhj/sdfjhjsdf/sdfhj
error: open /tmdsfhj/sdfjhjsdf/sdfhj: no such file or directory
```

It seems like we should provide a similar error when the remote file doesn't exist.

Nurysso:
Fixed error message when attempting to pull a file that doesn't exist on the remote system.

Previously, when a remote file didn't exist, the Pull function would return: as mentioned in #722
```
error: expected first field name to be "files", got "response"
```

This was confusing because the API returns a "response" part (without a "files" part) when the file doesn't exist, but the client expected "files" first.

Now the client checks if the first multipart part is "response" and extracts the actual error from the API, resulting in a clearer error message:
```
error: stat /tmp/file-does-not-exist.txt: no such file or directory
```

This fix also refactors the response processing logic into a shared `processResponsePart()` helper function to eliminate code duplication between the error-first and success paths.

Fixes #722

### DIFF
--- a/client/files.go
+++ b/client/files.go
@@ -599,7 +599,7 @@ func (client *Client) Pull(opts *PullOptions) error {
 		return processPullResponsePart(firstPart)
 	}
 	if firstPart.FormName() != "files" {
-		return fmt.Errorf(`expected first field name to be "files" or "response", got %q`, firstPart.FormName())
+		return fmt.Errorf(`expected first field name to be "files", got %q`, firstPart.FormName())
 	}
 	if _, err := io.Copy(opts.Target, firstPart); err != nil {
 		return fmt.Errorf("cannot write to target: %w", err)


client/files.go
```diff
@@ -575,28 +575,30 @@ func (client *Client) Pull(opts *PullOptions) error {
 	if mediaType != "multipart/form-data" {
 		// If it's not multipart, it's likely an API error, so try parsing it as such.
 		var serverResp response
-		err := decodeInto(resp.Body, &serverResp)
-		if err != nil {
+		if err := decodeInto(resp.Body, &serverResp); err != nil {
```
benhoyt: Let's please avoid making stylistic changes like this in this PR (here and elsewhere). It makes the actual changes harder to review.

client/files.go
```diff
@@ -606,33 +608,42 @@ func (client *Client) Pull(opts *PullOptions) error {
...
-		return fmt.Errorf(`expected second field name to be "response", got %q`, responsePart.FormName())
+		return fmt.Errorf("expected second field name to be \"response\", got %q", responsePart.FormName())
 	}
 
-	// Process response metadata (see defaultRequester.Do)
+	return processResponsePart(responsePart)
+}
+
+// processResponsePart decodes the response part and extracts any errors
+func processResponsePart(part io.Reader) error {
```
benhoyt: Because this is in the `client` package, let's include the word "pull" in the name. Let's call it `processPullResponsePart`.

client/files.go
```diff
@@ -575,28 +575,30 @@ func (client *Client) Pull(opts *PullOptions) error {
...
 	}
-	defer filesPart.Close()
+	defer firstPart.Close()
 
-	if filesPart.FormName() != "files" {
-		return fmt.Errorf(`expected first field name to be "files", got %q`, filesPart.FormName())
+	// Check if we got "response" first (error case) or "files" (success case)
+	if firstPart.FormName() == "response" {
+		// Decode and process the response to extract the actual error
+		return processResponsePart(firstPart)
```
benhoyt: This seems reasonable. But after this `if` statement I think we should keep the `if firstPart.FormName() != "files"` block in case the field name is something other than `files` or `response`.

client/files.go
```diff
@@ -606,33 +608,42 @@ func (client *Client) Pull(opts *PullOptions) error {
 	}
 	defer responsePart.Close()
 	if responsePart.FormName() != "response" {
-		return fmt.Errorf(`expected second field name to be "response", got %q`, responsePart.FormName())
+		return fmt.Errorf("expected second field name to be \"response\", got %q", responsePart.FormName())
```
benhoyt: Again, let's avoid unnecessary stylistic changes in this PR. In this case, the backticks were there for a reason, as they avoid the escaped quotes inside the string.

client/files.go
```diff
@@ -606,33 +608,42 @@ func (client *Client) Pull(opts *PullOptions) error {
...
+	return processResponsePart(responsePart)
+}
+
+// processResponsePart decodes the response part and extracts any errors
+func processResponsePart(part io.Reader) error {
 	var multipartResp response
-	if err := decodeInto(responsePart, &multipartResp); err != nil {
-		return err
+	if err := decodeInto(part, &multipartResp); err != nil {
+		return fmt.Errorf("cannot decode response: %w", err)
```
benhoyt: Let's avoid this error wrapping and revert this change. `decodeInto` already adds "cannot decode ...", which I think is enough here.

client/files.go
```diff
@@ -606,33 +608,42 @@ func (client *Client) Pull(opts *PullOptions) error {
...
 	}
+
+	// Check for error in response
 	if err := multipartResp.err(); err != nil {
 		return err
 	}
+
```
benhoyt: Let's avoid the additional line breaks here and below. These code sections were nicely divided into "paragraphs" already -- it's less clear to have each if statement as its own "paragraph".

benhoyt (changes requested):
Thanks. Left some comments: mostly minor stylistic changes.

In addition, we should add a test of this code path. Perhaps you can do something similar to `TestPullFailsWithErrorResponse`, added [here](https://github.com/canonical/pebble/pull/452/files#diff-411ef57155896d34b3c3bfbfe080e0ab13e4cd0242359972d282583d0e2daf65).

client/files.go
Nurysso: okay got it will make the necessary changes

client/files.go
```diff
@@ -609,23 +614,29 @@ func (client *Client) Pull(opts *PullOptions) error {
...
-	// Process response metadata (see defaultRequester.Do)
+	return processPullResponsePart(responsePart)
+}
+
+// processPullResponsePart decodes the response part and extracts any errors
+func processPullResponsePart(part io.Reader) error {
 	var multipartResp response
-	if err := decodeInto(responsePart, &multipartResp); err != nil {
+	if err := decodeInto(part, &multipartResp)
+	err != nil {
```
benhoyt: There's some funny formatting changes here, with the `err != nil {` split onto the next line (here and two places below). You can see this in the "Format" workflow failures here: https://github.com/canonical/pebble/actions/runs/20191040984/job/58022154319

client/files.go
```diff
@@ -587,16 +587,21 @@ func (client *Client) Pull(opts *PullOptions) error {
...
-	if filesPart.FormName() != "files" {
-		return fmt.Errorf(`expected first field name to be "files", got %q`, filesPart.FormName())
+	// Check if we got "response" first (error case) or "files" (success case)
+	if firstPart.FormName() == "response" {
+		// Decode and process the response to extract the actual error
+		return processPullResponsePart(firstPart)
 	}
-	if _, err := io.Copy(opts.Target, filesPart); err != nil {
+	if firstPart.FormName() != "files" {
+		return fmt.Errorf(`expected first field name to be "files" or "response", got %q`, firstPart.FormName())
```
benhoyt: I'd prefer we keep this error message the same as before. Because we shouldn't really get "response" except in case of error, which is handled above.
```suggestion
		return fmt.Errorf(`expected first field name to be "files", got %q`, firstPart.FormName())
```

client/files.go
```diff
@@ -634,5 +645,6 @@ func (client *Client) Pull(opts *PullOptions) error {
 	if fr[0].Error != nil {
 		return fr[0].Error
 	}
+
```
benhoyt: Nit: per earlier message, let's avoid extraneous formatting changes in this PR.

client/files.go
Nurysso: so something like this? 
``` go
if err := decodeInto(part, &multipartResp); err != nil {
	return err
}
```

benhoyt: Yes, exactly. If you look at the diff here on GitHub you'll see that's what the original was.

Nurysso: ohh, ok everything else looks good then pushing the code

client/files.go
benhoyt: @Nurysso You missed making this update. Committing now.

benhoyt (approved):

benhoyt:
> okay so it did worked, looks like I need to improve my code formating/review skills a little more lol.

Yes, I think that would be good. :-) I always encourage my team members to do a thorough self-review using a git diff tool, or just using the GitHub diff UI itself. That would have caught a number of these follow-up issues.

> Also on a side note can i mail you sometime not regarding this project but for guidance on career/stuff i am working at?

Sure.

PR #756, opened 2025-12-15
tests: validate that all http.Client's are instantiated with CheckRedirect argument

dimaqq:
These are AI-generated tests that validate that all http.Client's instantiated in Pebble have use the `CheckRedirect` argument.

This way, if a new HTTP client is introduced in the master branch, and the changes are merged into the FIPS branch, we're sure that the new client was patched up.

The value of the parameter is not tested. We're assuming that if the parameter is used, the developer made a conscious choice.

### DIFF
--- a/internals/testutil/http_client_check_test.go
+++ b/internals/testutil/http_client_check_test.go
@@ -1,0 +1,191 @@
+// Copyright (c) 2025 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package testutil_test
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestHTTPClientCheckerSelf is a self-test that verifies the static analyzer
+// correctly identifies http.Client instances with and without CheckRedirect.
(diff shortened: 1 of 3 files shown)


internals/testutil/http_client_check_test.go
```diff
@@ -0,0 +1,191 @@
...
+	var violations []string
+
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		// Skip non-Go files, test files, testdata, and vendor/generated code
+		if !strings.HasSuffix(path, ".go") ||
+			strings.Contains(path, "/vendor/") ||
```
dimaqq, line 117: There are pros and cons here... we don't actually use `go mod vendor`, so maybe it's better to remove this?

wdyt @benhoyt

internals/testutil/testdata/bad_client.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2025 Canonical Ltd
+//
+// Licensed under GPLv3.
```
dimaqq, line 3: Other files use a longer license blurb here, though the effect is the same.

internals/testutil/http_client_check_test.go
```diff
@@ -0,0 +1,191 @@
...
+			strings.Contains(path, "/.git/") ||
+			strings.Contains(path, "/dist/") ||
+			strings.Contains(path, "/testdata/") {
+			return nil
+		}
+
+		// Parse the file
```
benhoyt, line 124: Could just call the `checkHTTPClientsInFile` helper here.

benhoyt (commented):
Happy to discuss further, but I don't think it's worth it.

For one thing, this isn't checking nearly enough:

- we need to check that no https URLs are getting sent
- the body of CheckRedirect needs to be correct (not checked at all here): no https URLs in redirect, max 10 redirects

But it's also checking too much:

- we could create a client with `c := http.Client{}` but then assign `c.CheckRedirect = ...` afterwards, which would be fine, but be flagged by this

I think we'd be better just manually being aware of what we merge into `fips` and check all the other things too. Plus, we only add new `http.Client`s very rarely for significant new features.

dimaqq:
Happy to close.

PR #766, opened 2025-12-17
ci: fix up download-artefact path

dimaqq:
The newer download-artefact action looks into the download zip file, and if there's a single file inside, it unpacks that single file.

Thus, we no longer need `f*o*/` (directory) pattern.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -130,26 +130,8 @@ jobs:
     steps:
       - uses: actions/download-artifact@v7
 
-      - id: FIXME-temp-1
-        run: |
-          echo debugging FIXME
-          pwd
-          echo $GITHUB_WORKSPACE
-          ls -halt .
-          ls -halt $GITHUB_WORKSPACE
-          echo old location
-          echo ${{ env.SNAP_NAME }}*${{ env.TEST_ON }}*snap/* || echo nope
-          ls ${{ env.SNAP_NAME }}*${{ env.TEST_ON }}*snap/* || echo nope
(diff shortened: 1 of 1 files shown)


.github/workflows/snap.yml
```diff
@@ -131,7 +131,7 @@ jobs:
       - uses: actions/download-artifact@v7
 
       - id: get-snap
-        run: echo "filename=$(find ${{ env.SNAP_NAME }}*${{ env.TEST_ON }}*snap/*)" >> $GITHUB_OUTPUT
+        run: echo "filename=$(echo ${{ env.SNAP_NAME }}*${{ env.TEST_ON }}*.snap)" >> $GITHUB_OUTPUT
```
dimaqq, line 134: This case is simple: one artefact to download, which contains one file.

.github/workflows/snap.yml
```diff
@@ -158,7 +158,7 @@ jobs:
       - uses: actions/download-artifact@v7
 
       - id: get-snap
-        run: echo "filename=$(find ${{ env.SNAP_NAME }}*${{ matrix.arch }}*snap/*)" >> $GITHUB_OUTPUT
+        run: echo "filename=$(echo ${{ env.SNAP_NAME }}*${{ matrix.arch }}*.snap)" >> $GITHUB_OUTPUT
```
dimaqq, line 161: This case is more complex: N artefacts to download, each containing one file.
I think that this should work correctly, but we won't find out until:
- we merge this to `master`
- RISC-V builders are up

PR #768, opened 2025-12-17
ci: use named artefacts per architecture

dimaqq:
Since we've merged download-artefact version bump and I tried to fix the local build, the remote build began to fail, because when multiple artefacts are available, download would download all of them, but then unpack into a directory.

Set exact names on the artefacts, so that the test and release step download a single snap, for the specific architecture.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -149,9 +149,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     needs: [test]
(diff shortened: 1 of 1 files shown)


dimaqq:
This works!
- the `test` step works :)
- the `release` step seems like it would work when merged

Run: https://github.com/canonical/pebble/actions/runs/20321161196/job/58379564027
See "FIXME debug" step.

benhoyt:
Thanks for this, merging (as hopefully it'll help me move ahead with the release).

PR #773, opened 2025-12-18
ci: move cluttery files out of root directory

benhoyt:
This moves the sbomber config file into .github.

Also deletes the staticcheck.conf and uses the defaults. Staticcheck doesn't have a way to configure the config file location, plus I think these errors are pretty reasonable, so just use //lint:ignore comments in the specific places.


james-garner-canonical (approved):
Looks reasonable to me.

PR #783, opened 2026-01-12
ci: run sbomber after publishing snap, and pull from edge

### LINKED ISSUE #755: Correct the SBOM workflow/config
Introduced in #731: 

https://github.com/canonical/pebble/blob/d2838bbda1029479ca592d77bad2c1861c4b34cc/.sbomber-manifest-snap.yaml#L12-L15

On push to master, pebble snap is pushed to `latest/edge`.
One release, the snap is pushed to `latest/candidate`.

As far as I know, candidate is promoted to stable manually (seems undocumented, cc @benhoyt).
Which means that the workflow `.github/workflows/sbom-secscan.yaml` automated via `on: push: tags: ["v*"]` won't see the newest release.

The current work-around is to run the workflow manually via `on: workflow_dispatch:`.

james-garner-canonical:
@dimaqq  identified in #755 that the SBOM workflow was not running against the correct version of the snap.

Previously, the tool was configured to run against the stable version of the snap (running whenever we publish). We publish to edge, so this PR updates the config to run against the edge version of the snap instead.

As @tonyandrewmeyer noted, with the channel corrected, there would be a race condition with publishing the new snap and downloading the snap to scan. This PR addresses this, by moving the `sbom-secscan` workflow invocation to a job in the `snap` workflow, which runs only after publishing succeeds.

Resolves #755.

### DIFF
--- a/.github/workflows/sbom-secscan.yaml
+++ b/.github/workflows/sbom-secscan.yaml
@@ -1,0 +1,21 @@
+name: SBOM and secscan
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  scan:
+    runs-on: [self-hosted, self-hosted-linux-amd64-jammy-private-endpoint-medium]
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: sbomber
+        uses: canonical/sbomber/.github/actions/run@cf7a76e810497ec932e8285b2c9d6b373d225e79  # main -- the sbomber-ref input below should match
+        with:
+          manifest: '${{ github.workspace }}/.github/.sbomber-manifest-snap.yaml'
+          artifact-name: secscan-report-upload-snap
+          sbomber-ref: cf7a76e810497ec932e8285b2c9d6b373d225e79  # should match the ref in 'uses' above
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Thanks for this. Changes look fine to me. Have you / is it possible to test before the next release? If not, I'm fine just waiting till then.

james-garner-canonical:
I've refactored to retain the separate worfklow file rather than inlining it. This actually makes the diff cleaner too. Should be good to merge, and hopefully the scan runs successfully after the next release.

.github/.sbomber-manifest-snap.yaml
```diff
@@ -12,4 +12,4 @@ artifacts:
     ssdlc_params:
         name: 'pebble'
         version: ''  # auto-detect
-        channel: 'stable'
+        channel: 'edge'
```
dimaqq, line 15: I guess this is OK, however note that "promote" job is running in parallel with "sbom-secscan".

We could order "sbom-secscan" after promote, in which case we could use `channel: candidate` here.

james-garner-canonical: It's fine for them to run in parallel because they're both only reading from `edge`, right? I guess reading from `candidate` does feel a bit more correct though. Reckon it's worth changing? I can make a PR if so.

PR #787, opened 2026-01-13
chore: merge master info fips (SBOM automation)

dimaqq:
Merge master branch into fips, specifically the SBOM automation.

### DIFF
--- a/.github/.sbomber-manifest-snap.yaml
+++ b/.github/.sbomber-manifest-snap.yaml
@@ -9,6 +9,7 @@ clients:
 artifacts:
   - name: 'pebble'
     type: 'snap'
+    channel: 'fips/edge'
     ssdlc_params:
         name: 'pebble'
         version: ''  # auto-detect


.github/.sbomber-manifest-snap.yaml
dimaqq, line 1: @james-garner-canonical I wonder how to specify the "fips" track so that the full snap channel becomes `fips/edge`.

dimaqq: Perhaps that's not something we can express with ssdlc params:
https://github.com/canonical/sbomber/blob/cf7a76e810497ec932e8285b2c9d6b373d225e79/src/state.py#L126

So we'd be relying on the version string being something like `v1.2.3-fips`, right?

james-garner-canonical: AFAIK the tool only supports specifying the risk component of the channel.

I guess this means that the action won't be able to scan the fips channel releases? Maybe we need to rethink how we're approaching the sbom/secscan work. I feel like it makes more sense for it to be operating with our locally built artifacts directly than pulling them from the store anyway.

tonyandrewmeyer: > AFAIK the tool only supports specifying the risk component of the channel.

sbomber will take a full channel for the source to download a snap. The secscan tool's SSDLC params have a "channel" argument but it expects only the risk part (I don't know why and didn't get around to trying to dig into that with the security team). It does have a "version" argument, which would presumably include "fips" in this case. So I think you can do the appropriate thing here with the existing tools (for sbomber, it's also very straightforward in my experience to get changes accepted upstream). *However*,

> I feel like it makes more sense for it to be operating with our locally built artifacts directly than pulling them from the store anyway.

I agree with this, and I think I ended up going in the wrong direction with the ops packages in that regard, partly because that was how sbomber was oriented at the time (but I contributed changes upstream so that's not a valid excuse). Pulling them from the 'store' has a nice side-effect that you're guaranteeing that the report is for the exact thing that ends up in the store, but I don't think that outweighs the cleanliness of using the thing you just produced, and the avoidance of races.

dimaqq: @tonyandrewmeyer I don't understand what I should do for the fips branch.

tonyandrewmeyer: I think you'd want a manifest like:

```yaml
clients:
  sbom:
      service_url: https://sbom-request.canonical.com
      department: charm_engineering
      email: dima.tisnek@canonical.com
      team: charm_tech
  secscan: {}

artifacts:
  - name: 'pebble'
    type: 'snap'
    channel: 'fips/edge'
    ssdlc_params:
        name: 'pebble'
        version: ''  # auto-detect
        channel: 'edge'
```

This will definitely do the right thing for the security scanning and the SBOM generation. In terms of automatically submitting the results, it would have the right cycle, the right name, the right channel (risk), and a version like `1738`. I think the revision plus name and risk is enough to distinguish the artefact, but I'm not totally sure on that. If you want to be confident it's doing the right thing, I would ask the SSDLC team what they expect to see in the `id-params` arguments when a product has multiple tracks.

There are two other simpler options:

1. Drop the `ssdlc_params` section from the above YAML, and just do the manual process of submitting the results. It's not compulsory at this time, as far as I know, and is meant to be a step towards getting rid of the manual process.
2. Drop the workflow entirely. The current SSDLC requirements are (more than) satisfied by what's happening in master, as far as I understand. It would definitely be nice to have the SBOM for the FIPS branch, and somewhat nice to do the security scan (but we are already scanning elsewhere, and given the nature of the branch, it seems quite unlikely that there would be issues not in master), but that's more us trying to outdo what the requirements are.

dimaqq: Done, thanks!

james-garner-canonical (approved):
LGTM. If/when I get to updating the SBOM/secscan workflow to use the local files, I'll try to remember to update this branch too.

dimaqq:
No worries, we always merge the master branch into the fips branch before the releases, thus sbom changes won't get left out.

PR #797, opened 2026-01-30
chore: add a test to prevent crypto being reintroduced in the fips branch

dimaqq:
Adding a test that uses `go list -deps -json` to scan direct imports and 3rd party library imports and bails if someone accidentally merges in some code that uses cryptographic primitives.

### DIFF
--- a/tests/dependencies_test.go
+++ b/tests/dependencies_test.go
@@ -19,7 +19,11 @@ import (
 	"os/exec"
 	"strings"
 	"testing"
+
+	. "gopkg.in/check.v1"
 )
+
+func Test(t *testing.T) { TestingT(t) }
 
 // packageInfo represents the relevant fields from `go list -json` output
 type packageInfo struct {


tests/dependencies_test.go
```diff
@@ -0,0 +1,170 @@
...
+	cmd := exec.Command("go", "list", "-deps", "-json", "../...")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("Failed to run go list: %v\nOutput: %s", err, output)
+	}
+
+	decoder := json.NewDecoder(strings.NewReader(string(output)))
+
+	allowedCryptoDeps := map[string]bool{
+		"github.com/gorilla/websocket":       true, // Uses crypto/tls, crypto/rand, crypto/sha1
```
dimaqq, line 111: The websocket dep is fine, because we only use `ws://` URLs and the library doesn't handle redirects, thus we're guaranteed to stay on `ws://` and not use any crypto.

tests/dependencies_test.go
```diff
@@ -0,0 +1,170 @@
...
+		"github.com/canonical/x-go/randutil": true, // Uses crypto/rand, crypto/sha256
```
dimaqq, line 112: The `randutil` dep shows how easy it is to introduct crypto use by accident. We use:
- `randutil.RandomString`
- `randutil.RandomDuration`
- `randutil.RandomKernelUUID`

So we're safe for now.

dimaqq:
P.S. I have tested that the new test works:
- runs OK in this branch
- fails correctly in the master branch

dimaqq:
I'll merge it and Ben can always review later.

PR #798, opened 2026-01-30
feat(plan): allow supplying a base plan to plan.ReadDir and PlanManager.Load

anpep:
This patch refactors `plan.ReadDir` and `planstate.PlanManager` to support an optional "base plan" during plan loading. This allows applications extending Pebble to supply a set of base layers on top of which the on-disk plan will be stacked.

```diff
// internals/plan/plan.go
- func ReadDir(layersDir string) (*Plan, error)
+ func ReadDir(layersDir string, base *Plan) (*Plan, error)

// internals/overlord/planstate/manager.go
- func (m *PlanManager) Load() error
- func (m *PlanManager) Init(*plan.Plan) error
+ func (m *PlanManager) Load(base *plan.Plan) error
```

### DIFF
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -1566,7 +1566,7 @@ func ReadDir(layersDir string, base *Plan) (*Plan, error) {
 
 	if base != nil {
 		baseLayer := &Layer{
-			// XXX: The "000-layer.yaml" file name is forbidden, but there's no
+			// The "000-layer.yaml" file name is forbidden, but there's no
 			// logic preventing us from having a sub-1 order for layers. If we
 			// use order "1", we would be clashing with a subsequent layer that
 			// would logically use this order number. A workaround for this would
@@ -1574,7 +1574,7 @@ func ReadDir(layersDir string, base *Plan) (*Plan, error) {
 			// other layers' orders by 1000 (which is what PlanManager.AppendLayer)
 			// does internally in order to support layer sub-directories.
 			Order:      0,
-			Label:      "base-layer",
+			Label:      "pebble-base",
 			Services:   base.Services,
 			Checks:     base.Checks,
 			LogTargets: base.LogTargets,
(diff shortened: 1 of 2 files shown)


internals/plan/plan.go
```diff
@@ -1543,19 +1543,46 @@ func configLayerEntries(configDir string, dirOK bool) (configs []*configEntry, e
...
+	if base == nil && len(layers) == 0 {
+		return NewPlan(), nil
 	}
 
-	layers, err := ReadLayersDir(layersDir)
-	if err != nil {
-		return nil, err
+	if base != nil {
+		baseLayer := &Layer{
+			// XXX: The "000-layer.yaml" file name is forbidden, but there's no
```
benhoyt: I guess XXX means something like TODO? In any case, I think we should just remove it and leave this comment.

internals/plan/plan.go
```diff
@@ -1543,19 +1543,46 @@ func configLayerEntries(configDir string, dirOK bool) (configs []*configEntry, e
...
+			// logic preventing us from having a sub-1 order for layers. If we
+			// use order "1", we would be clashing with a subsequent layer that
+			// would logically use this order number. A workaround for this would
+			// be to use order "1" and, before prepending this layer, offset all
+			// other layers' orders by 1000 (which is what PlanManager.AppendLayer)
+			// does internally in order to support layer sub-directories.
+			Order:      0,
+			Label:      "base-layer",
```
benhoyt: I think it's be better to use `"pebble-base"` here, as this is a special layer, and `ParseLayer` prevents normal layers from starting with the `pebble-` prefix (see `Layer.Validate`).

Or maybe that doesn't work because `Layer.Validate` is called on the base layer? If so, I'd recommend `"_base"` or similar.

anpep: Thanks! I picked `"pebble-base"`; as the label is lost during `CombineLayers()`, it should not be affected by the `Layer.Validate` forbidding the `"pebble-"` prefix.

PR #799, opened 2026-01-30
feat(planstate): add support for immutable sections

anpep:
This patch adds support for immutable plan sections. This is a generalization of the behavior previously exhibited by the `workloads` section, in an attempt to move the logic out of the `WorkloadsSectionExtension.ValidatePlan()` method, and remove the need to cache a copy of the previously loaded plan.

`WorkloadsSection` now implements the `ImmutableSection` interface, which makes it non-changeable at runtime:

```go
package plan

type ImmutableSection interface {
    Section

    Immutable()

    Equal(other plan.Section)
}
```

### DIFF
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -298,7 +298,7 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 	for name, oldSection := range m.plan.Sections {
 		if immutable, ok := oldSection.(plan.ImmutableSection); ok {
 			newSection := p.Sections[name]
-			if !immutable.Equals(newSection) {
+			if !immutable.Equal(newSection) {
 				return nil, &plan.FormatError{
 					Message: fmt.Sprintf("cannot change immutable section %q", name),
 				}
--- a/internals/plan/plan.go
+++ b/internals/plan/plan.go
@@ -65,8 +65,8 @@ type ImmutableSection interface {
 	// Immutable marks the section as immutable. It must implement no behavior.
 	Immutable()
 
-	// Equals returns false if the section differs from other.
-	Equals(other Section) bool
+	// Equal reports whether this section is equal to other.
+	Equal(other Section) bool
 }
 
 const (
--- a/internals/workloads/workloads.go
+++ b/internals/workloads/workloads.go
@@ -92,7 +92,7 @@ func (ws *WorkloadsSection) IsZero() bool {
 	return len(ws.Entries) == 0
 }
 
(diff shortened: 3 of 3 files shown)


internals/plan/plan.go
```diff
@@ -59,6 +59,15 @@ type Section interface {
 	IsZero() bool
 }
 
+type ImmutableSection interface {
+	Section
+
+	Immutable()
```
paul-rodriguez, line 66: This needs to be documented: it's a marker that should only be implemented on actually immutable sections

paul-rodriguez (approved):

internals/overlord/planstate/manager.go
```diff
@@ -294,6 +294,18 @@ func (m *PlanManager) updatePlanLayers(layers []*plan.Layer) (*plan.Plan, error)
 	if err != nil {
 		return nil, err
 	}
+
+	for name, oldSection := range m.plan.Sections {
+		if immutable, ok := oldSection.(plan.ImmutableSection); ok {
+			newSection := p.Sections[name]
```
benhoyt, line 300: What if the new plan doesn't have this section? Do we need a `newSection, ok` and `if !ok || !immutable.Equals(...)` or similar? (And if so, a test for that.)

internals/plan/plan.go
```diff
@@ -59,6 +59,16 @@ type Section interface {
 	IsZero() bool
 }
 
+type ImmutableSection interface {
+	Section
+
+	// Immutable marks the section as immutable. It must implement no behavior.
+	Immutable()
+
+	// Equals returns false if the section differs from other.
```
benhoyt: Nit: this is a reverse way to comment it. Perhaps use "reports" as is common for boolean-returning functions in Go docs:
```suggestion
	// Equals reports whether this section is equal to other.
```

internals/plan/plan.go
```diff
@@ -59,6 +59,16 @@ type Section interface {
...
+	Equals(other Section) bool
```
benhoyt: Naming nit: these sorts of functions are normally spelled `Equal` in Go, for example `reflect.DeepEqual`, `time.Equal`, `bytes.Equal`, and so on.
```suggestion
	Equal(other Section) bool
```

internals/overlord/planstate/manager.go
anpep: If the new plan doesn't have this section, `newSection == nil` and `ImmutableSection.Equal(nil)` should return `false`, meaning the immutable section was changed (there was one, now there is not!) and the error should be triggered.

This being said, there is no way currently in Pebble that this can be triggered. Layers can be appended but never removed, so there's no way of triggering this logic. However, if there was, it would still be fine!

Issue #801, opened 2026-02-01
Moving identities to own package may have broken the read-only mode


dimaqq:
@alesancor Alejandro Santisteban Corchos said

Hello team!

I'm trying to run a rock with --read-only flag, but I'm getting this error:

`cannot run daemon: cannot create identity directory: mkdir /var/lib/pebble/default/identity: read-only file system`
I tried using `PEBBLE_PERSIST=never`, but that only affects `/var/lib/pebble/default/.pebble.state`.

If I try running with `--tmpfs /var/lib/pebble/default/identity:rw,mode=700` then:

`cannot run daemon: directory "/var/lib/pebble/default" not writable`
And I cannot tmpfs the entire default directory since I'd loose the layers configuration. 

Is there no way of running pebble inside a read-only fs? 

@dimaqq Dima Tisnek said

It's possible that we have a bug.
Pebble converts existing state to new format on startup (?) and maybe we forgot about the read only mode

#744 introduced auto-conversion from previous "legacy" to new format.
That doesn't write if there were no legacy identities, IIRC.
However, here, we trip on creating a directory, which must be happening before the conversion.

anpep:
My two cents:
 - Currently, we have two "modes" of running the Pebble daemon: HTTP and HTTPS. However, we attempt to load/generate an mTLS identity regardless.
 - Overlord creates a `tlsstate.TLSManager` regardless of Pebble being in "HTTP-only mode", which is completely unused if no HTTPS endpoint is set on the `run` command.

I propose we:
 - honor `PEBBLE_PERSIST=never` in `idkey.Generate()` and don't write the newly generated identity back to disk (nor attempt to create any missing directories), and
 - don't instantiate `TLSManager`, nor attempt to load/create the identity iff the HTTPS endpoint is not set (TLS disabled in Pebble)

PR #802, opened 2026-02-02
fix: don't writeback identity keys when PEBBLE_PERSIST=never

anpep:
Partially fixes, but not completely #801 by not writing to disk when `PEBBLE_PERSIST=never` is set.

Additionally, TLS manager is not created when no HTTPS endpoint is provided as an option on run.


hpidcock:
Thank-you @anpep for this, we're moving forward with #896, as it handles this slightly differently by avoiding the indentity key generation when it is not required/not valuable.

PR #813, opened 2026-03-11
feat(tlsstate): more generic certificate customization

lp35:
This PR introduces a callback system to the TLS manager, allowing callers to customize identity and TLS certificates at creation time without exposing internal certificate construction logic.
This augment (and replace) the available SetX509Templates() function which can only touch specific fields (Subject, Issuer, DNSName and EmailAddresses).

This allows us to manipulate fields internally set by the tlsstate manager, such as the NotBefore and the NotAfter fields of certificate, which where not exposed before this PR.

TLS configuration callback are passed at instantiation time of the manager, through the ```tlsstate.Options``` parameter.

This PR breaks backward compatibility with the SetX509Templates() function, but it only affects "internals" functions, so we're okay with that.

### DIFF
--- a/internals/overlord/tlsstate/manager.go
+++ b/internals/overlord/tlsstate/manager.go
@@ -95,7 +95,7 @@ type Options struct {
 	// exist. The leaf directory will be created with 0o700 permissions.
 	TLSDir string
 	// Signer is a private key representing the identity of a Pebble
-	// instance (machine, container or device), which implements the
+	// instance, which implements the
 	// tlsstate.Signer interface (allowing it to sign digests).
 	Signer IDSigner
 	// ConfigureIDCertificate, if non-nil, is called after the default identity


internals/overlord/tlsstate/manager.go
```diff
@@ -82,6 +82,12 @@ type IDSigner interface {
 	Fingerprint() string
 }
 
+// Holds configurable options functor for the certificate.
```
paul-rodriguez:
```suggestion
// A functor to apply options to certificates at creation.
```

internals/overlord/tlsstate/manager.go
```diff
@@ -94,10 +100,13 @@ type TLSManager struct {
 	signer IDSigner
 
 	// The identity and tls certificate optionally allows a
-	// select number of fields to be supplied from externally
+	// number of fields to be supplied from externally
 	// supplied X509 templates (see SetX509Templates).
-	idTemplate  *x509.Certificate
-	tlsTemplate *x509.Certificate
```
paul-rodriguez: Suggestion:

```
// When creating identity and TLS certificates, the following CertOption functors will be applied to 
// assign default values to certificate fields.
// The collection of functors acts as a template for certificates.
```

internals/overlord/tlsstate/manager.go
```diff
@@ -94,10 +100,13 @@ type TLSManager struct {
...
+	// Note: CertOption will be applied in the order encountered
+	// in this array, meaning there is 2 options manipulating the
+	// same fields, only the last one will be taken in consideration.
+	idTemplate  []CertOption
+	tlsTemplate []CertOption
```
paul-rodriguez: These two things are only templates in a very conceptual sense, so I would be in favor of renaming them to `idCertOptions` and `tlsCertOptions` respectively.

internals/overlord/tlsstate/manager.go
```diff
@@ -122,8 +190,26 @@ func NewManager(tlsDir string, signer IDSigner) *TLSManager {
 func (m *TLSManager) SetX509Templates(idTemplate, tlsTemplate *x509.Certificate) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.idTemplate = idTemplate
-	m.tlsTemplate = tlsTemplate
+
+	if idTemplate != nil {
+		m.idCertOptions = append(m.idCertOptions, WithLegacyTemplate(idTemplate))
```
acalinp: Shouldn't this also replace, not append? It does feel a bit weird, if the caller called `SetX509IDCertificateOptions` first and then `SetX509Templates`, the legacy template will be appended on top of the new options, was this the intent?

internals/overlord/tlsstate/manager.go
```diff
@@ -122,8 +190,26 @@ func NewManager(tlsDir string, signer IDSigner) *TLSManager {
...
+	}
+}
+
+func (m *TLSManager) SetX509IDCertificateOptions(opts ...CertOption) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.idCertOptions = opts
+}
+
+func (m *TLSManager) SetX509TLSCertificateOptions(opts ...CertOption) {
```
acalinp: Missing function comments similar to `SetX509Templates`. These functions should presumably also be called before `GetCertificate`? The comment should clarify when the options list is replaced vs appended to.

internals/overlord/tlsstate/manager.go
```diff
@@ -93,11 +99,14 @@ type TLSManager struct {
...
-	// The identity and tls certificate optionally allows a
-	// select number of fields to be supplied from externally
-	// supplied X509 templates (see SetX509Templates).
-	idTemplate  *x509.Certificate
-	tlsTemplate *x509.Certificate
+	// When creating identity and TLS certificates, the following CertOption
+	// functors will be applied to  assign default values to certificate fields.
+	// The collection of functors acts as a template for certificates.
+	// Note: CertOption will be applied in the order encountered
+	// in those arrays, meaning if there is 2 CertOption manipulating the
```
acalinp:
```suggestion
	// in those arrays, meaning if there are 2 CertOptions manipulating the
```

internals/overlord/tlsstate/manager.go
```diff
@@ -82,6 +82,12 @@ type IDSigner interface {
 	Fingerprint() string
 }
 
+// A functor to apply options to certificates at creation.
+// The option will receive the certificate to modify
+// as a parameter, as well as a deep copy of the parent
+// certificate and will return an error if a failure occured.
```
acalinp:
```suggestion
// certificate and will return an error if a failure occurred.
```

internals/overlord/tlsstate/manager.go
```diff
@@ -93,11 +99,14 @@ type TLSManager struct {
 	// The identity key used for signing TLS certificates.
 	signer IDSigner
 
-	// The identity and tls certificate optionally allows a
-	// select number of fields to be supplied from externally
-	// supplied X509 templates (see SetX509Templates).
-	idTemplate  *x509.Certificate
-	tlsTemplate *x509.Certificate
+	// When creating identity and TLS certificates, the following CertOption
+	// functors will be applied to  assign default values to certificate fields.
```
acalinp:
```suggestion
	// functors will be applied to assign default values to certificate fields.
```

internals/overlord/tlsstate/manager.go
acalinp, line 1: Nit: I think `functor` is a non-idiomatic term in Go. Typically Go uses the name "Functional Option" https://golang.cafe/blog/golang-functional-options-pattern

acalinp, line 1: You need to add some tests in `manager_test.go` to cover:
- options being applied in order
- parentCopy being passed correctly to TLS options
- parentCopy being nil for ID cert options
- the case where an option returns an error

internals/overlord/tlsstate/manager.go
```diff
@@ -114,6 +123,65 @@ func NewManager(tlsDir string, signer IDSigner) *TLSManager {
...
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return nil, err
+	}
+
+	return serialNumber, nil
+}
+
+func WithDefaultIDTemplate(signer IDSigner) CertOption {
```
acalinp: Since this and `WithDetaultTLSTemplate` are always applied automatically, maybe you shouldn't export them so that a consumer of this package doesn't accidentally append them twice?

internals/overlord/tlsstate/manager.go
```diff
@@ -114,6 +123,65 @@ func NewManager(tlsDir string, signer IDSigner) *TLSManager {
...
+		c.NotAfter = c.NotBefore.Add(tlsCertValidity)
+		c.KeyUsage = x509.KeyUsageDigitalSignature
+		c.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}
+		c.BasicConstraintsValid = true
+		return nil
+	}
+}
+
+// If an externally supplied TLS certificate template was provided,
+// use it to update the supported fields.
```
acalinp: The comment says "TLS certificate" but this function is also used for ID certificates. Broaden the comment or make it generic.

internals/overlord/tlsstate/manager_test.go
```diff
@@ -390,6 +391,110 @@ func (ts *tlsSuite) TestTLSServerClientIDKeyChange(c *C) {
...
+	// the identity certificate (second in chain).
+	c.Assert(tlsCert.DNSNames, DeepEquals, []string{"second.local"})
+	c.Assert(idCert.DNSNames, DeepEquals, []string{"second.local"})
+
+	// Identity cert options receive nil: the cert is self-signed, no parent.
+	c.Assert(idParentCopy, IsNil)
+
+	// TLS cert options receive a deep copy of the identity certificate as parent.
+	c.Assert(tlsParentCopy, NotNil)
+	c.Assert(tlsParentCopy.Equal(idCert), Equals, true)
```
paul-rodriguez: You should probably move these assertions inside the implementations of each CertOption. That way you wouldn't need the intermediate variable.

lp35:
Hey @benhoyt !

I had some time to work again on this. I wonder if you apply fine-grained control over the ```daemon.Options``` passed to the overlord. If yes, I'll add the new parameters introduced by this PR:
```go
    // ConfigureIDCertificate, if non-nil, is called after the default identity
	// certificate fields are set and may override any of them.
	ConfigureIDCertificate ConfigureCertificateFunc
	// ConfigureTLSCertificate, if non-nil, is called after the default TLS
	// certificate fields are set and may override any of them.
	ConfigureTLSCertificate ConfigureCertificateFunc
```

Otherwise I'll simply change the ```daemon.Options``` in ```overlord.go``` like this:
```go
type Options struct {
	// PebbleDir is the path to the pebble directory. It must be provided.
	PebbleDir string
	// LayersDir is the path to the layers directory. It defaults to "<PebbleDir>/layers" if empty.
	LayersDir string
	// RestartHandler is an optional structure to handle restart requests.
	RestartHandler restart.Handler
	// ServiceOutput is an optional output for the logging manager.
	ServiceOutput io.Writer
	// Extension allows extending the overlord with externally defined features.
	Extension Extension
	// Configuration to pass to the tlsstate manager.
	TLSOptions tlsstate.Options // <---- New field, TLSDir and IDSigner is removed
	// Persist specifies whether the state should be persisted to disk.
	Persist PersistMode
}
```

internals/overlord/tlsstate/manager.go
```diff
@@ -93,37 +117,70 @@ type TLSManager struct {
...
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return nil, err
+	}
+
+	return serialNumber, nil
+}
+
+func withDefaultIDTemplate(signer IDSigner) ConfigureCertificateFunc {
```
benhoyt: The func return values of these two `withFoo` helpers are always immediately called, so with the simplifications it's unnecessary to have them return funcs now. Let's just make them normal functions that take the arguments they need:

```go
func setIDCertDefaults(signer IDSigner, c *x509.Certificate) error { ... }

func setTLSCertDefaults(signer IDSigner, c *x509.Certificate) error { ... }
```

I'm not stuck on those names. `initIDCert` / `initTLSCert` would be fine too.

internals/overlord/tlsstate/manager.go
```diff
@@ -93,37 +117,70 @@ type TLSManager struct {
...
+	return func(c *x509.Certificate, parentCopy *x509.Certificate) (err error) {
+		c.SerialNumber, err = generateSerialNumber()
+		if err != nil {
+			return err
+		}
```
benhoyt: Errors will be extremely unlikely here, but it's probably a better pattern to only do the assignment in the non-error case (same below):
```suggestion
		serialNumber, err := generateSerialNumber()
		if err != nil {
			return err
		}
		c.SerialNumber = serialNumber
```

internals/overlord/tlsstate/manager_test.go
```diff
@@ -458,60 +544,60 @@ func (ts *tlsSuite) TestDefaultCertSubject(c *C) {
 }
 
 // TestTLSServerClientCustomTemplates checks that we can provide custom
-// X509 certificate templates for the identity and tls certificates.
+// X509 certificate fields for the identity and TLS certificates via Options.
```
benhoyt, line 547: Maybe we should rename this test slightly now that we don't have "templates"?

internals/overlord/tlsstate/manager_test.go
```diff
@@ -526,19 +612,11 @@ func (ts *tlsSuite) TestTLSServerClientCustomTemplates(c *C) {
...
 	// Check the TLS certificate.
-	c.Assert(tlsCert.Subject.String(), Equals, tlsTemplate.Subject.String())
-	if !slices.Equal(tlsCert.DNSNames, tlsTemplate.DNSNames) {
-		c.Fail()
-	}
-	if !slices.Equal(tlsCert.EmailAddresses, tlsTemplate.EmailAddresses) {
-		c.Fail()
-	}
+	c.Assert(tlsCert.Subject.String(), Equals, tlsSubject.String())
+	c.Assert(slices.Equal(tlsCert.DNSNames, tlsDNSNames), Equals, true)
```
benhoyt: I believe you can use DeepEquals for this kind of check (which would give better error messages on failure).

benhoyt (commented):
Yeah, I really like this approach now. Just a few minor comments.

Regarding the daemon.Options question. I'm on the fence, but I think I prefer the simplification and just having a daemon.Options.TLSOptions field. Would it make sense for this to be a `*tlsstate.Options` (pointer) and if nil, not set up TLS at all? I'm not sure.

benhoyt:
Oh, and let's update the PR description to match the latest design (nit: we don't need the word "Summary" in big letters at the top).

lp35:
> Would it make sense for this to be a `*tlsstate.Options` (pointer) and if nil, not set up TLS at all? I'm not sure.

We don't have this use case right now on our side, but it's a really good idea on the paper. Maybe this is out of scope of this PR for now?

benhoyt:
If you merge from master you'll get https://github.com/canonical/pebble/pull/839 which updates to Go 1.26.2 to fix the govulncheck issues.

internals/overlord/tlsstate/manager.go
```diff
@@ -82,6 +81,31 @@ type IDSigner interface {
...
+// certificate being created, and parentCopy is a deep copy of the signing
+// (parent) certificate, or nil when the certificate is self-signed.
+// Returning an error aborts certificate creation.
+type ConfigureCertificateFunc func(cert *x509.Certificate, parentCopy *x509.Certificate) error
+
+// Options holds the configuration for creating a TLSManager.
+type Options struct {
+	// TLSDir is the directory used to persist identity certificate files. It
+	// must be a directory of which only the leaf element, at most, does not
+	// exist. The leaf directory will be created with 0o700 permissions.
```
paul-rodriguez, line 95: This is a strange requirement. If the directory hierarchy above the leaf `TLSDir` does not exist, what's the rationale to _not_ create it as required?

internals/overlord/tlsstate/manager.go
```diff
@@ -82,6 +81,31 @@ type IDSigner interface {
...
+	TLSDir string
+	// Signer is a private key representing the identity of a Pebble
+	// instance (machine, container or device), which implements the
```
paul-rodriguez:
```suggestion
	// instance, which implements the
```

internals/overlord/tlsstate/manager.go
lp35: TBH it's a legacy comment, I just copy/pasted here. If you think it's an issue, maybe you can open a
 ticket?

PR #814, opened 2026-03-11
fix(servstate): prevent deadlock in TestStartFastExitCommand

mohqas:
During a test run in CI/CD I noticed twice that this test panics and exits after 10 minutes:
```
2026-03-09T20:09:00.680Z [test] Service "test4" on-success action is "restart", waiting ~30s before restart (backoff 24)
panic: test timed out after 10m0s
	running tests:
		Test (10m0s)
```

It lasts 10 minutes since it goes into a deadlock and times out. Looking through the logs, the reason is a panic while accessing the logs in the code below, as we use `c.Check` which carries on when the check fails for the number of logs, then attempts to access a non-existing item panicing and causing go.check to recover the test and go into cleanup:
```
	c.Check(chg.Tasks()[0].Log(), HasLen, 2)
	c.Check(chg.Tasks()[0].Log()[0], Matches, `(?s).* INFO Most recent service output:\n    too-fast\n    second line`)
```
Since we already have the lock before we panic, when trying to clean up, we can not acquire the state lock, and we go into a deadlock, timing out after 10 minutes.

Switching the check to assert solves the deadlock; an additional change to read the logs and release the lock before we do the checks is also done to avoid deadlocking elsewhere.

It seems the test fail as the service is not starting within the configured "OkayWait" of 50 ms. Therefore, the timeout has also been increased to 2 seconds to avoid future failures altogether.

### DIFF
--- a/internals/overlord/servstate/manager_test.go
+++ b/internals/overlord/servstate/manager_test.go
@@ -706,6 +706,9 @@ services:
 }
 
 func (s *S) TestStartFastExitCommand(c *C) {
+	restore := servstate.FakeOkayWait(2 * time.Second)
+	defer restore()
+
 	s.newServiceManager(c)
 	layer := `
 services:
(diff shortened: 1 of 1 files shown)


internals/overlord/servstate/manager_test.go
```diff
@@ -719,13 +722,21 @@ services:
 	chg := s.startServices(c, [][]string{{"test4"}})
 
 	s.st.Lock()
-	c.Check(chg.Status(), Equals, state.ErrorStatus)
-	c.Check(chg.Err(), ErrorMatches, `(?s).*\n- Start service "test4" \(service start attempt: exited quickly with code 0, will restart\)`)
-	c.Check(chg.Tasks()[0].Log(), HasLen, 2)
-	c.Check(chg.Tasks()[0].Log()[0], Matches, `(?s).* INFO Most recent service output:\n    too-fast\n    second line`)
-	c.Check(chg.Tasks()[0].Log()[1], Matches, `.* ERROR service start attempt: exited quickly with code 0, will restart`)
```
holmanb, line 726: Based on the description of the deadlock I expected that manually introducing a panic between the Lock/Unlock calls would recreate the issue. However it exists immediately without the deadlock. Am I missing something?

mohqas: Sorry, I was missing one additional piece of the puzzle. We don't always attempt to acquire the lock during cleanup. By introducing a panic, we intermittently deadlock still, but with a higher probability. 

To consistently reproduce the issue, force the service to go into backoff, which forces us to attempt to acquire the lock during cleanup. This is what happens when the test fails, as can be seen in the logs. Here is a full repro from the current test in main. It has two changes:
* The service now sleeps for 1 second, forcing it to go into backoff, which requires the lock during cleanup
* There is a panic as we attempt to access the log line at index 2, which does not exist:

```
func (s *S) TestStartFastExitCommand(c *C) {
	s.newServiceManager(c)
	layer := `
services:
    test4:
        override: replace
        command: echo -e 'too-fast\nsecond line' && sleep 1
`
	s.planAddLayer(c, layer)
	s.planChanged(c)

	chg := s.startServices(c, [][]string{{"test4"}})

	s.st.Lock()
	c.Check(chg.Status(), Equals, state.ErrorStatus)
	c.Check(chg.Err(), ErrorMatches, `(?s).*\n- Start service "test4" \(service start attempt: exited quickly with code 0, will restart\)`)
	c.Check(chg.Tasks()[0].Log(), HasLen, 2)
	c.Check(chg.Tasks()[0].Log()[0], Matches, `(?s).* INFO Most recent service output:\n    too-fast\n    second line`)
	c.Check(chg.Tasks()[0].Log()[2], Matches, `.* ERROR service start attempt: exited quickly with code 0, will restart`)
	s.st.Unlock()

	svc := s.serviceByName(c, "test4")
	c.Assert(svc.Current, Equals, servstate.StatusBackoff)
}
```

benhoyt (approved):
@mohqas Thanks for your analysis and the fix! And thanks @holmanb for review.

PR #816, opened 2026-03-18
style: simplify slice definitions

holmanb:
Slices of structs don't require interior types. Simplify it.

This enables pebble code to pass `gofmt -s`, which is used to lint snapd.


benhoyt (approved):
Works for me, especially if we're doing it in other projects.

PR #818, opened 2026-03-18
chore: use Go 1.26 now that 1.24 is end-of-life

### LINKED ISSUE #817: Go 1.24 EOL
Just a heads-up that Go 1.24 has been EOL for a month or so, and static code analysis is reporting impact related to CVE-2026-25679 and CVE-2025-68121.

I believe that CVE-2025-68121 is fixed in Go 1.24.13, but CVE-2026-25679 won't be fixed for 1.24.

On a related note, I'm not sure why the Trivy workflow here doesn't seem to be triggering against these. It has been when checking against the gobinary, for example in this workflow: https://github.com/canonical/mediawiki-k8s-operator/actions/runs/23269118479/job/67658725081?pr=19

tonyandrewmeyer:
This PR bumps the Go version from 1.24 to the latest 1.26.1, as 1.24 was made end-of-life when 1.26.0 was released (Go's "end of life as soon as there are two newer minor versions" policy).

The go.mod value is changed, and the rest is the result of running `go fix`.

Also fixes the Go setup in the workflows to pull the Go version from the mod file correctly.

I've read through the 1.25 and 1.26 release notes, and I can't see anything else that should be an issue for Pebble. There are security/crypto/tls changes that should be improvements but should not impact the FIPS branch, from what I can tell. The `GOMAXPROCS` change in 1.25 might impact how Pebble behaves, but it's not really Pebble itself that's changing, and it ought to be for the better anyway.

Fixes #817

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
 
     - name: Test
       run: |
-        PEBBLE_TEST_USER=runner PEBBLE_TEST_GROUP=runner sudo -E -H go test ./...
+        PEBBLE_TEST_USER=runner PEBBLE_TEST_GROUP=runner sudo -E -H "$(which go)" test ./...
 
   format:
     runs-on: ubuntu-latest


.github/workflows/lint.yml
```diff
@@ -9,7 +9,7 @@ jobs:
 
       - uses: actions/setup-go@v6
         with:
-          go-mod-file: 'go.mod'
+          go-version-file: 'go.mod'
```
tonyandrewmeyer, line 12: From what I can tell, `go-mod-file` has never been an option for this action, at least going back to [v4](https://github.com/actions/setup-go/blob/v4/action.yml) when it was added to this workflow. I think we have been ignoring it the whole time?

It seems very odd that it wouldn't have been noticed, but other code (e.g. Concierge) uses `go-version-file`, and both Copilot and Claude want it to be `go-version-file` and say that `go-mod-file` doesn't exist, and hasn't ever, and reading through the releases and looking at old versions of the action's code, I can't find it.

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
     - name: Test
       run: |
-        PEBBLE_TEST_USER=runner PEBBLE_TEST_GROUP=runner sudo -E -H go test ./...
```
benhoyt, line 41: Why is this change needed? Wouldn't bash (or whatever does the PATH handling) choose the `which go` version of go automatically?

tonyandrewmeyer: Without that change, the test fails like this:

```
Run PEBBLE_TEST_USER=runner PEBBLE_TEST_GROUP=runner sudo -E -H go test ./...
  shell: /usr/bin/bash -e {0}
  env:
    GOTOOLCHAIN: local
go: go.mod requires go >= 1.26.1 (running go 1.24.13; GOTOOLCHAIN=local)
Error: Process completed with exit code 1.
```

Do you remember a while back working through something with Pebble (I think - unless it was Concierge?) where the wrong Go binary was being picked up? I can't remember where we talked about it and where you shared your eventual finding (I think in Matrix, and the search there in private chat is awful). I think this is similar to that? It's what made me try `which`, anyway, although I don't remember all the specifics.

PR #820, opened 2026-03-21
ci: add Zizmor workflow

tonyandrewmeyer:
This PR adds workflow static analysis using Zizmor. The standard Charm Tech configuration is used (and was already present, for manual scanning).

In addition, some small findings are addressed:
* Don't cache when building
* Default to empty permissions
* Don't persist Git credentials 
* Pin actions
* Use environment variables to avoid injection


.github/workflows/snap.yml
```diff
@@ -70,14 +73,18 @@
       - name: Checkout Pebble repo
         uses: actions/checkout@v6
         with:
+          persist-credentials: false
           fetch-depth: 0
 
       - name: Set LP credentials for remote build
+        env:
+          LP_CREDENTIALS: ${{ secrets.LP_CREDENTIALS }}
```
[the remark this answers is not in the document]
tonyandrewmeyer: @benhoyt I believe for this (and the other similar one) it can be fixed by setting up an environment in settings. I don't think that can be done with the repo automation currently, it has to be done manually.

.github/workflows/binaries.yml
```diff
@@ -22,12 +26,14 @@ jobs:
...
       uses: actions/checkout@v6
       with:
+        persist-credentials: false
         fetch-depth: 0
 
     - name: Set up Go
       uses: actions/setup-go@v6
       with:
         go-version-file: "go.mod"
+        cache: false
```
benhoyt, line 36: I'm not necessarily against it, but why is `cache: false` desired?

.github/workflows/scanning.yml
```diff
@@ -1,10 +1,10 @@
 name: Vulnerability scanning
 
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+# TODO: Re-enable once trivy-action supply chain compromise is fully resolved.
+# See: https://github.com/aquasecurity/trivy/discussions/10425
+on: []
```
benhoyt, line 5: Probably best to comment out the previous `on` configuration so it's easier to reinstate when we need to.

.github/workflows/snap.yml
```diff
@@ -162,10 +171,12 @@ jobs:
           pattern: pebble-snap-${{ matrix.arch }}
 
       - id: get-snap
-        run: echo "filename=$(echo ${{ env.SNAP_NAME }}*${{ matrix.arch }}*.snap)" >> $GITHUB_OUTPUT
+        env:
+          ARCH: ${{ matrix.arch }}
```
benhoyt, line 174: Can you clarify why having it here is better than having `${{ matrix.arch }}` inline below?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78  # v7.6.0
+
+      - name: Run zizmor
+        run: uvx zizmor@v1.23.1 --format=sarif . > results.sarif
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
benhoyt, line 28: Why does this step need the GitHub token? Isn't it just doing static checks on the .github workflow files? Or is it for the upload step below?

benhoyt (commented):
Couple of questions, but I'm happy with adding this.

.github/workflows/binaries.yml
tonyandrewmeyer: In build/release workflows, there is a risk of [cache poisoning](https://docs.zizmor.sh/audits/#cache-poisoning). Used in the Ultralytics attack, and although GitHub have made changes since, my understanding is that you can still make of of this, and that there are research tools that will do all the hard work for you (so presumably also non-research tools).

.github/workflows/snap.yml
tonyandrewmeyer: The `${{ }}` expansion doesn't have the same shell quoting/expansion rules as using an environment variable. More details [in the zizmor docs](https://docs.zizmor.sh/audits/#template-injection).

.github/workflows/zizmor.yml
tonyandrewmeyer: Zizmor has multiple [operating modes](https://docs.zizmor.sh/usage/#operating-modes). In 'online' mode, it can do extra checks like looking for imposter commits. The token avoids hitting rate limits, even though everything it would check with Pebble is public.

.github/workflows/scanning.yml
```diff
@@ -13,9 +13,11 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Run Github Trivy FS Action
-        uses: aquasecurity/trivy-action@master
+        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1  # v0.35.0
```
dimaqq, line 25: **side question**

Since we're pinning after the compromise, where let's assume for the sake of the argument the attacker could have rewritten git history, how do we know that this hash is a good hash?

It feels like we're relying on what's provided in the [blog post](https://www.aquasec.com/blog/autonomous-runtime-security-turning-runtime-intelligence-into-agentic-response-2/) but how do we know that's kosher?

dimaqq: One other project in the company used this hash on March 13th, which is a good week before the blog post / disclosure: https://github.com/canonical/lxd-ui/commit/345d576e1f73a8d6c1ed8aa632c907a1d1e06665 

Circumstantial, but better than nothing?

.github/workflows/snap.yml
```diff
@@ -133,12 +140,14 @@ jobs:
           pattern: pebble-snap-amd64
 
       - id: get-snap
-        run: echo "filename=$(echo ${{ env.SNAP_NAME }}*${{ env.TEST_ON }}*.snap)" >> $GITHUB_OUTPUT
+        run: echo "filename=$(echo ${SNAP_NAME}*${TEST_ON}*.snap)" >> $GITHUB_OUTPUT
```
dimaqq: **off-topic**

We need a blessed 3rd party action that copies secrets and environ to github output... doing this by hand feels error-prone 😅 

OT: why is it `env.SNAP_NAME` in the name of the step above this but `SNAP_NAME` here? I mean, should we change that too, or is that a GitHub limitation?

dimaqq (approved):

.github/workflows/scanning.yml
tonyandrewmeyer: I feel that realistically, *if* we enable Trivy scanning again (and that is quite questionable), then we would only do so once Trivy and whoever we're going to trust on this (a Canonical security team, perhaps) gave an all-clear, and that the all-clear would have to include either a new version that we'd pin to.

For this PR, where it's disabled, it just feels safer to avoid `@master`, in case someone did mistakenly enable it again too early. For that, this hash, which is in the blog post but also elsewhere, including in the case you reference, seems a reasonable backstop.

.github/workflows/snap.yml
tonyandrewmeyer: The inconsistency isn't great, no. However, I think we should just get rid of this, it's always building the Pebble snap; we should just say "pebble" and make everything simpler. I don't believe we're trying to sync this file with one somewhere else.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,34 @@
+name: Workflow static checks
```
benhoyt, line 1: Nit: zizmor.yaml -> zizmor.yml for consistency with the others.

.github/workflows/scanning.yml
```diff
@@ -13,9 +18,11 @@ jobs:
...
         with:
           scan-type: 'fs'
           scan-ref: '.'
```
benhoyt, line 28: per discussion, let's add the `exit-code: 1` fix in this PR.

PR #833, opened 2026-04-01
fix(overlord): support racing Loop and Stop correctly

mohqas:
TestRacingLoopAndStop would fail the race detector with the old code.

---------

(cherry picked from snapd commit e0ea2182f84d6daa08c2432080282e7f74dbf08f, https://github.com/canonical/snapd/pull/16709)


internals/overlord/overlord.go
```diff
@@ -474,32 +477,38 @@ func (o *Overlord) ensureBefore(d time.Duration) {
...
-				st.Prune(o.startOfOperationTime, pruneWait, abortWait, pruneMaxChanges, pruneMaxNotices)
-				st.Unlock()
-				serviceMgr := o.ServiceManager()
-				serviceMgr.Prune(pruneWait, pruneMaxInactiveServices)
-			}
+	// Proceed with the loop.
+	close(o.loopProceed)
+}
+
+func (o *Overlord) loop() error {
```
benhoyt, line 481: This doesn't look correct -- perhaps a bad merge / conflict resolution? Should this either call, or fall through to `loop`?

benhoyt (changes requested):
Thanks for this. However, this doesn't look quite right as `loop` is never called. In addition, it introduces a data race -- [see logs](https://github.com/canonical/pebble/actions/runs/23844799836/job/69509312664?pr=833).

internals/overlord/overlord.go
mohqas: `loop` is being called during `New` as a go routine, which is blocked on the select statement until either the loop is marked as dying, in which case we just return from the loop without running it, or we close the `loopProceed` channel and fall into the loop itself. We do that in the `Loop` call.

The goal here is to have the tomb in a running state early and control the loop through channels, and eliminate any race conditions between New and Loop, communicating through channels

mohqas (commented):
Thanks for the review! Sorry was intending to open this as a draft for now, need to address the additional datarace or try to reproduce it in snapd if it needs fixing.

Edit: reverted to draft*

benhoyt:
@mohqas Any update on this?

mohqas:
> @mohqas Any update on this?

not for this week. If anyone has capacity and would like to take over happy to help; otherwise will get to it early next week (:

mohqas:
The datarace is intermittent but frequent in `pebble`. I couldn't reproduce it in snapd. 

The detection happens due to the newer toolchain in `pebble`. In particular, versions 1.24+ detect it while 1.23 or lower versions never do. 

The race can be fixed by guarding the faking function from the stack trace. However, I am looking for an alternative to ensure the go routine in the new test does not outlive its test (where the read of the faked value happens, causing the race condition)

mohqas:
With additional improvements to the runtime in 1.24, it seems Go is yielding the runtime faster to the next test, causing the intermittent race condition with the overlord started in the routine. I've added a change to the test to ensure the overlord in the routine is stopped before the next test starts

Edit: The failures of docs and zizmor are unrelated to the change

benhoyt:
@hpidcock Could you please review this one? In particular, that it looks correct and is cherry-picking correctly from snapd (https://github.com/canonical/snapd/pull/16709). See also comments on https://github.com/go-tomb/tomb/issues/21 and

benhoyt (approved):

internals/overlord/overlord.go
```diff
@@ -137,10 +138,12 @@ type Overlord struct {
...
 	o := &Overlord{
-		pebbleDir: opts.PebbleDir,
-		inited:    true,
-		extension: opts.Extension,
+		pebbleDir:   opts.PebbleDir,
+		inited:      true,
+		extension:   opts.Extension,
+		loopProceed: make(chan struct{}),
 	}
+	o.loopTomb.Go(o.loop)
```
hpidcock: This is too early in the `New`, since we have multiple return paths that do not have cleanup.

hpidcock: Easier to just place this at the end of the function, where errors are never returned.

mohqas: Thanks for the pointer, makes sense, I will adjust it with a sidenote; we are diverting from the cherry-picked commit.

hpidcock (approved):

PR #834, opened 2026-04-02
ci: use a dedicated GitHub environment for the snap build

tonyandrewmeyer:
This PR sets the `snap` workflow to use a dedicated `snap-build` GitHub environment. Primarily, this lets us put the required secrets in that environment, so they are available to this workflow but not any others.

I have created the environment already, so everything should be good to go.


benhoyt (approved):
Sorry, must have missed this notification last week. Looks good, thanks!

PR #837, opened 2026-04-08
feat(cli): add structured output formats to `services` & `identities`

Snehil-Shah:
adds a new `--format` flag to support structured output formats like json & yaml.

Towards #824


Snehil-Shah:
@benhoyt I only added it to the `services` command to start with. Once we are aligned, we can add to all the other commands too.

I ran it on some dummy services:

**JSON**:

```bash
$ pebble services --format=json
{"services":[{"name":"redis","startup":"enabled","current":"active","current-since":"2026-04-08T16:37:27.453591135Z"},{"name":"webapp","startup":"enabled","current":"active","current-since":"2026-04-08T16:37:27.45483726Z"},{"name":"worker","startup":"disabled","current":"inactive","current-since":"0001-01-01T00:00:00Z"}]}
```

**YAML**:

```bash
$ pebble services --format=yaml
services:
    - name: redis
      startup: enabled
      current: active
      current-since: 2026-04-08T16:37:27.453591135Z
    - name: webapp
      startup: enabled
      current: active
      current-since: 2026-04-08T16:37:27.45483726Z
    - name: worker
      startup: disabled
      current: inactive
      current-since: 0001-01-01T00:00:00Z
```

Filtering and everything works as expected too.

client/services.go
```diff
@@ -96,10 +96,10 @@ type ServicesOptions struct {
...
 // ServiceInfo holds status information for a single service.
 type ServiceInfo struct {
-	Name         string         `json:"name"`
-	Startup      ServiceStartup `json:"startup"`
-	Current      ServiceStatus  `json:"current"`
-	CurrentSince time.Time      `json:"current-since"`
+	Name         string         `json:"name" yaml:"name"`
+	Startup      ServiceStartup `json:"startup" yaml:"startup"`
+	Current      ServiceStatus  `json:"current" yaml:"current"`
+	CurrentSince time.Time      `json:"current-since" yaml:"current-since"`
```
benhoyt: This fields needs to be marked as "omitzero" to avoid the following output if a service hasn't been started yet:

```
$ go run ./cmd/pebble services --format=yaml
services:
    - name: svc1
      startup: disabled
      current: inactive
      current-since: 0001-01-01T00:00:00Z
```

Thankfully we're on Go 1.24 so the encoding/json package has "omitzero", and for the YAML package "omitempty" works for time.Time values too. So this works:

```suggestion
	CurrentSince time.Time      `json:"current-since,omitzero" yaml:"current-since,omitempty"`
```

internals/cli/cmd_services.go
```diff
@@ -54,6 +59,18 @@ func (cmd *cmdServices) Execute(args []string) error {
...
+	var writeOutput func(services []*client.ServiceInfo) error
+	switch cmd.Format {
+	case "", "text":
+		writeOutput = cmd.writeText
+	case "json":
+		writeOutput = cmd.writeJSON
+	case "yaml":
+		writeOutput = cmd.writeYAML
+	default:
+		return fmt.Errorf(`invalid output format (expected "text", "json", or "yaml", not %q)`, cmd.Format)
```
benhoyt: I realized we already have code like this in cmd_identities.go, so I go back on what I said before -- I think it would be good to find a way to make this common. Let's add a new `formatMixin` to the bottom of format.go and try the following. Also, can we use the `choice` tag here to let the command line parser validate for us?

```go
type formatMixin struct {
    Format string `long:"format" default:"text" choice:"text" choice:"json" choice:"yaml"`
}

var formatArgsHelp = map[string]string{
	"--format": `Output format: "text" (default), "json", or "yaml".`,
}

func (mx formatMixin) formatNonText(result any) error {
    switch mx.Format {
    case "json":
        // ... use json.Marshal on result
    case "yaml":
        // ... use yaml.Marshal on result
    }
    panic(fmt.Sprintf("internal error: invalid format option %q", mx.Format))
}
// ... and then in code which uses it ...

if cmd.Format == "text" {
    // use "services" variable
    // if len==0 print "No services." to Stderr
    // else use tabwriter
    return nil
}

type servicesMap struct {
	Services map[string]*client.ServiceInfo `json:"services" yaml:"services"`
}
servicesMap := /* convert "services" to map */
err := cmd.formatSmart(servicesMap{Services: servicesMap})
if err != nil { ... }
```

benhoyt: Note that this implies we don't print "No services" or "No identities." or similar if len==0 for JSON and YAML, but instead print `{"services": {}}` -- which I think is better than what we have no for identities.

internals/cli/cmd_services.go
```diff
@@ -84,3 +105,25 @@ func (cmd *cmdServices) Execute(args []string) error {
 	}
 	return nil
 }
+
+type servicesList struct {
+	Services []*client.ServiceInfo `json:"services" yaml:"services"`
```
benhoyt: I think this should be `map[string]*client.ServiceInfo`, similar to how we do it in `cmd_identities.go`.

internals/cli/cmd_services_test.go
```diff
@@ -112,6 +112,69 @@ foo      enabled   inactive  -
 	c.Check(s.Stderr(), check.Equals, "")
 }
 
+func (s *PebbleSuite) TestServicesJSON(c *check.C) {
```
benhoyt, line 151: Thanks for adding tests. It would be good to add test cases for the no-services JSON and YAML cases.

internals/cli/cmd_services_test.go
```diff
@@ -112,6 +112,69 @@ foo      enabled   inactive  -
...
+		{"name": "svc1", "current": "inactive", "startup": "enabled", "current-since": "2022-04-28T17:05:23+12:00"},
+		{"name": "svc2", "current": "inactive", "startup": "enabled"},
+		{"name": "svc3", "current": "backoff", "startup": "enabled"}
+	]
+}`)
+	})
+	rest, err := cli.ParserForTest().ParseArgs([]string{"services", "--format", "json"})
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
+	c.Check(s.Stdout(), check.Equals, `{"services":[{"name":"svc1","startup":"enabled","current":"inactive","current-since":"2022-04-28T17:05:23+12:00"},{"name":"svc2","startup":"enabled","current":"inactive","current-since":"0001-01-01T00:00:00Z"},{"name":"svc3","startup":"enabled","current":"backoff","current-since":"0001-01-01T00:00:00Z"}]}`+"\n")
```
benhoyt: This shows the zero-time behaviour. We should update this to ensure the current-since field is omitted in this case.

benhoyt (changes requested):
Thanks for taking a stab at this. I've suggested a slightly different approach (though I realise that's going back on what I said earlier) as this is too similar to cmd_identities. If you could please refactor and also update cmd_identities to use this approach, that would be great.

internals/cli/cmd_services_test.go
Snehil-Shah: Yeah.. looks ugly😬. I have updated to omit empty fields.

internals/cli/cmd_services.go
Snehil-Shah: I have updated the PR to reflect it. I was hoping for further abstraction so we could have a single `formatMixin.format(result)` function that would also work for text, and each command could simply call it without any `if` conditions, but text outputs seem heavily command-specific, so no generalisation is possible, I suppose.🥲

Snehil-Shah:
@benhoyt I have also added flag-level validation to the `--format` flag in `logs` command (the implementation is untouched, see https://github.com/canonical/pebble/pull/837/changes/2309f46da36d78193e769489a023e2dc0baaefab), just so we have a consistent public API surface (see https://github.com/canonical/pebble/pull/837/changes/a6a9292214b4f0e8110a8b5d6dc7169380a0084b). WDYT?

internals/cli/format.go
```diff
@@ -148,6 +150,35 @@ func tabWriter() *tabwriter.Writer {
 	return tabwriter.NewWriter(Stdout, 5, 3, 2, ' ', 0)
 }
 
+type formatMixin struct {
+	//lint:ignore SA5008 "choice" tag is intentionally duplicated
+	Format string `long:"format" default:"text" choice:"text" choice:"json" choice:"yaml"`
+}
+
+var formatArgsHelp = map[string]string{
+	"--format": `Output format: "text" (default), "json", or "yaml".`,
```
benhoyt: I see the choices and the default are listed in the help already:

```
$ go run ./cmd/pebble identities --help
...
[identities command options]
      --format=[text|json|yaml]   Output format: "text" (default), "json", or "yaml". (default: text)
```

So I think we should truncate this to just:
```suggestion
	"--format": `Output format`,
```

internals/cli/cmd_logs.go
```diff
@@ -81,22 +82,18 @@ func (cmd *cmdLogs) Execute(args []string) error {
...
 		}
-
 	case "json":
 		encoder := json.NewEncoder(Stdout)
 		encoder.SetEscapeHTML(false)
 		writeLog = func(entry client.LogEntry) error {
 			return encoder.Encode(&entry)
 		}
-
-	default:
```
benhoyt, line 98: For defensive programming (in case of programmer error), let's put the `default` case back, but panic with an internal error:

```go
default:
    panic(fmt.Sprintf("internal error: invalid format option %q", cmd.Format))
```

benhoyt (commented):
Yes, this looks nice! And thanks for using "choice" in `pebble logs` -- that makes sense. Just a couple of minor comments.

If you'd like to add this to the other commands where relevant, go ahead.

internals/cli/format.go
Snehil-Shah: @benhoyt One small thing I noticed is that some flag descriptions end with a period (`.`), some don't. Is there a convention we like to follow here?

internals/cli/cmd_ls.go
Snehil-Shah, line 1: @benhoyt For the `ls` command, I added a manual step where we convert `client.FileInfo` type to our `fileEntry` type (as it has accessor methods for all fields and is not directly serializable). The `fileEntry` shape references:

https://github.com/canonical/pebble/blob/692c2f5c3ea98d71d40dfbf12c9915a2323ef616/client/files.go#L155-L166

WDYT?

internals/cli/cmd_changes_test.go
```diff
@@ -30,6 +30,69 @@ import (
...
+  "id":   "uno",
+  "kind": "foo",
+  "summary": "...",
+  "status": "Do",
+  "ready": false,
+  "spawn-time": "2016-04-21T01:02:03Z",
+  "tasks": [{"kind": "bar", "summary": "some summary", "status": "Do", "progress": {"done": 0, "total": 1}, "spawn-time": "2016-04-21T01:02:03Z"}]
+}}`
+
+var fakeChangesJSON = `{"type": "sync", "result": [
```
Snehil-Shah: @benhoyt nit: moved some of the mock variables to the top for cleaner separation of shared variables. Let me know if it goes against any conventions.

Snehil-Shah:
@benhoyt implemented formatting for the remaining relevant commands. Some decisions I took:

- The commands `check`, `identity`, `notice`, and `tasks` don't return an empty JSON/YAML when nothing is found and instead error out (plain text error), or just give an empty output. This makes sense because the above commands need a target, and for invalid or non-existent targets (non-existent identity, for example), returning an empty structure can be seen as incorrect. On the other hand, it might also not be ideal in cases where the caller always expects the structure mentioned. WDYT?
- The commands `check`, `identity`, and `notice` had outputs in YAML format by default. I still added `formatMixin` to these from the perspective that the default `text` output happens to be YAML too. This keeps it extendible if we wish to have a different textual output format in the future, while YAML & JSON would always be structured. WDYT?

internals/cli/format.go
benhoyt: Hmm, yeah, it's a bit of a mix, isn't it? I'd say let's not end with a period to avoid the noise. No need to change existing ones, but let's drop it on this one ("Output format" is not a complete sentence anyway, so it wouldn't have a full stop / period at the end.)

benhoyt:
Sorry to be a pain, but now that I see how big this PR is with all commands, let's revert commit b902e6c and just get this PR merged with only `services` and `identities` converted, leaving the rest to subsequent PRs (one command per PR). That'll make this and the subsequent PRs much easier to review.

> The commands check, identity, notice, and tasks don't return an empty JSON/YAML when nothing is found and instead error out (plain text error)

I think that makes sense -- you're requesting a specific ID/name for those, so it should be an error if it doesn't exist.

> The commands check, identity, and notice had outputs in YAML format by default. I still added formatMixin to these from the perspective that the default text output happens to be YAML too. This keeps it extendible if we wish to have a different textual output format in the future, while YAML & JSON would always be structured. WDYT?

Agreed, that sounds consistent.

Snehil-Shah:
@benhoyt Understood. I have reverted the changes (& kept a stash for future PRs). This PR is now limited to just the `services` and `identities` commands.

internals/cli/cmd_ls.go
benhoyt: This seems reasonable. We can discuss further if/when we do the ls command.

internals/cli/cmd_logs.go
```diff
@@ -51,7 +52,7 @@ func init() {
 		Description: cmdLogsDescription,
 		ArgsHelp: map[string]string{
 			"--follow": "Follow (tail) logs for given services until Ctrl-C is\npressed. If no services are specified, show logs from\nall services running when the command starts.",
-			"--format": "Output format: \"text\" (default) or \"json\" (JSON lines).",
+			"--format": "Output format.",
```
benhoyt:
```suggestion
			"--format": "Output format",
```

internals/cli/format.go
```diff
@@ -148,6 +150,35 @@ func tabWriter() *tabwriter.Writer {
 	return tabwriter.NewWriter(Stdout, 5, 3, 2, ' ', 0)
 }
 
+type formatMixin struct {
+	//lint:ignore SA5008 "choice" tag is intentionally duplicated
+	Format string `long:"format" default:"text" choice:"text" choice:"json" choice:"yaml"`
+}
+
+var formatArgsHelp = map[string]string{
+	"--format": `Output format`,
```
benhoyt: Nit:
```suggestion
	"--format": "Output format",
```

benhoyt (approved):
Great, thanks for this. Will merge shortly!

Snehil-Shah:
@benhoyt Would you like a myriad of PRs for each of the remaining commands? Or some grouping based on complexity?

benhoyt:
@Snehil-Shah More or less one PR for each, but you could group the related commands together, like `check` and `checks`.

PR #839, opened 2026-04-10
chore: bump Go version to 1.26.2 to address govulncheck CVEs

benhoyt:
This addresses several CVEs raised by govulncheck due to the fact that we're still running CI with Go 1.26.1. The CVEs are fixed in Go 1.26.2.


benhoyt:
Bump patch version of Go for CI -- merging without further review.

PR #841, opened 2026-04-14
feat(cli): add structured output formats to `identity`

Snehil-Shah:
adds `--format` flag to support structured output formats to the `identity` command.

Towards #824


benhoyt (approved):
Thanks!

PR #843, opened 2026-04-14
feat(cli): add structured output formats to `ls`

Snehil-Shah:
adds `--format` flag to support structured output formats to the `ls` command.

Towards #824

### DIFF
--- a/internals/cli/cmd_ls.go
+++ b/internals/cli/cmd_ls.go
@@ -17,8 +17,10 @@ package cli
 import (
 	"errors"
 	"fmt"
+	"os"
 	pathpkg "path"
 	"strings"
+	"time"
 
 	"github.com/canonical/go-flags"
 	"github.com/canonical/x-go/strutil/quantity"
@@ -36,6 +38,7 @@ type cmdLs struct {
 	client *client.Client
 
 	timeMixin
+	formatMixin
 	Directory  bool `short:"d"`
 	LongFormat bool `short:"l"`
 	Positional struct {
@@ -48,7 +51,7 @@ func init() {
 		Name:        "ls",
 		Summary:     cmdLsSummary,
 		Description: cmdLsDescription,
-		ArgsHelp: merge(timeArgsHelp, map[string]string{
+		ArgsHelp: merge(timeArgsHelp, formatArgsHelp, map[string]string{
 			"-d": "List matching entries themselves, not directory contents",
 			"-l": "Use a long listing format",
 		}),
@@ -56,6 +59,51 @@ func init() {
 			return &cmdLs{client: opts.Client}
 		},
 	})
+}
+
+type fileEntry struct {
+	Path         string `json:"path" yaml:"path"`
+	Name         string `json:"name" yaml:"name"`
+	Type         string `json:"type" yaml:"type"`
+	Size         *int64 `json:"size,omitempty" yaml:"size,omitempty"`
+	Permissions  string `json:"permissions" yaml:"permissions"`
+	LastModified string `json:"last-modified" yaml:"last-modified"`
+	UserID       *int   `json:"user-id" yaml:"user-id"`
+	User         string `json:"user" yaml:"user"`
+	GroupID      *int   `json:"group-id" yaml:"group-id"`
+	Group        string `json:"group" yaml:"group"`
+}
+
+type lsResult struct {
+	Files []fileEntry `json:"files" yaml:"files"`
+}
+
+func fileInfoToEntry(fi *client.FileInfo) fileEntry {
+	entry := fileEntry{
+		Path:         fi.Path(),
+		Name:         fi.Name(),
+		Permissions:  fmt.Sprintf("%03o", fi.Mode().Perm()),
+		LastModified: fi.ModTime().Format(time.RFC3339),
+		User:         fi.User(),
+		Group:        fi.Group(),
+	}
+	switch {
+	case fi.Mode().IsDir():
+		entry.Type = "directory"
+	case fi.Mode()&os.ModeSymlink != 0:
+		entry.Type = "symlink"
+	default:
+		entry.Type = "file"
+		size := fi.Size()
+		entry.Size = &size
+	}
+	if uid := fi.UserID(); uid != nil {
+		entry.UserID = uid
+	}
+	if gid := fi.GroupID(); gid != nil {
+		entry.GroupID = gid
+	}
+	return entry
 }
 
 func (cmd *cmdLs) Execute(args []string) error {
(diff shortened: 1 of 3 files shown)


internals/cli/cmd_ls.go
```diff
@@ -58,6 +61,51 @@ func init() {
...
+	Name         string `json:"name" yaml:"name"`
+	Type         string `json:"type" yaml:"type"`
+	Size         *int64 `json:"size,omitempty" yaml:"size,omitempty"`
+	Permissions  string `json:"permissions" yaml:"permissions"`
+	LastModified string `json:"last-modified" yaml:"last-modified"`
+	UserID       *int   `json:"user-id" yaml:"user-id"`
+	User         string `json:"user" yaml:"user"`
+	GroupID      *int   `json:"group-id" yaml:"group-id"`
+	Group        string `json:"group" yaml:"group"`
+}
```
Snehil-Shah, line 75: @benhoyt Continuing the conversation, I added a manual step where we convert `client.FileInfo` type to our `fileEntry` type (as it has accessor methods for all fields and is not directly serializable). The `fileEntry` shape references:

https://github.com/canonical/pebble/blob/692c2f5c3ea98d71d40dfbf12c9915a2323ef616/client/files.go#L155-L166

Should we export the `fileInfoResult` struct from the client and reuse it here?

benhoyt: This looks reasonable to me. No, I don't think we should export `fileInfoResult` from client, as that's too internal. It's a pity we go back and forth like this, but I think what you have is reasonable.

internals/cli/cmd_ls.go
```diff
@@ -58,6 +61,51 @@ func init() {
...
 }
 
+type fileEntry struct {
+	Path         string `json:"path" yaml:"path"`
+	Name         string `json:"name" yaml:"name"`
+	Type         string `json:"type" yaml:"type"`
+	Size         *int64 `json:"size,omitempty" yaml:"size,omitempty"`
+	Permissions  string `json:"permissions" yaml:"permissions"`
+	LastModified string `json:"last-modified" yaml:"last-modified"`
+	UserID       *int   `json:"user-id" yaml:"user-id"`
```
benhoyt: I think these (UserID and GroupID) should be "omitempty".

internals/cli/cmd_ls_test.go
```diff
@@ -145,3 +145,106 @@ func (s *PebbleSuite) TestLsInvalidPattern(c *C) {
...
+		}
+	]
+}`)
+	})
+
+	rest, err := cli.ParserForTest().ParseArgs([]string{"ls", "--format", "json", "/"})
+	c.Assert(err, IsNil)
+	c.Assert(rest, HasLen, 0)
+	c.Check(s.Stdout(), Matches, `\{"files":\[.*\]\}\n`)
+	c.Check(s.Stderr(), Equals, "")
```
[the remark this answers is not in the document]
benhoyt: Yeah, Copilot is definitely right here -- we should just assert on an exact string like we do with the other tests. Both here and below in the YAML test.

internals/cli/cmd_ls.go
```diff
@@ -77,6 +125,18 @@ func (cmd *cmdLs) Execute(args []string) error {
 		return err
 	}
 
+	if cmd.Format == "text" {
+		return cmd.writeText(files)
+	}
+
+	entries := make([]fileEntry, len(files))
+	for i, fi := range files {
+		entries[i] = fileInfoToEntry(fi)
```
benhoyt: I slightly prefer this idiom (and think we should settle on it) as it doesn't require the `i` index variable:

```suggestion
	entries := make([]fileEntry, 0, len(files))
	for _, fi := range files {
		entries = append(entries, fileInfoToEntry(fi))
```

internals/cli/cmd_ls.go
```diff
@@ -58,6 +61,51 @@ func init() {
...
+		entry.Type = "file"
+		size := fi.Size()
+		entry.Size = &size
+	}
+	if uid := fi.UserID(); uid != nil {
+		entry.UserID = uid
+	}
+	if gid := fi.GroupID(); gid != nil {
+		entry.GroupID = gid
+	}
```
benhoyt: Here, can we just set the fields directly? Actually, we could move this to the `entry := fileEntry{ ... }` struct initialization above.
```suggestion
	entry.UserID = fi.UserID()
	entry.GroupID = fi.GroupID()
```

internals/cli/cmd_ls.go
```diff
@@ -58,6 +61,51 @@ func init() {
...
+	switch {
+	case fi.Mode().IsDir():
+		entry.Type = "directory"
+	case fi.Mode()&os.ModeSymlink != 0:
+		entry.Type = "symlink"
+	default:
+		entry.Type = "file"
+		size := fi.Size()
+		entry.Size = &size
```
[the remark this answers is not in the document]
benhoyt: Copilot's right here. But let's copy what we have in api_files.go instead, and use a `fileModeToType` helper (but without the type and named consts):

```go
func fileModeToType(mode os.FileMode) string {
	switch {
	case mode&os.ModeType == 0:
		return "file"
	case mode&os.ModeDir != 0:
		return "directory"
	case mode&os.ModeSymlink != 0:
		return "symlink"
	case mode&os.ModeSocket != 0:
		return "socket"
	case mode&os.ModeNamedPipe != 0:
		return "named-pipe"
	case mode&os.ModeDevice != 0:
		return "device"
	default:
		return "unknown"
	}
}
```

benhoyt (changes requested):

internals/cli/cmd_ls.go
```diff
@@ -58,6 +61,62 @@ func init() {
...
+	Path         string `json:"path" yaml:"path"`
+	Name         string `json:"name" yaml:"name"`
+	Type         string `json:"type" yaml:"type"`
+	Size         *int64 `json:"size,omitempty" yaml:"size,omitempty"`
+	Permissions  string `json:"permissions" yaml:"permissions"`
+	LastModified string `json:"last-modified" yaml:"last-modified"`
+	UserID       *int   `json:"user-id,omitempty" yaml:"user-id,omitempty"`
+	User         string `json:"user" yaml:"user"`
+	GroupID      *int   `json:"group-id,omitempty" yaml:"group-id,omitempty"`
+	Group        string `json:"group" yaml:"group"`
```
benhoyt: Sorry for missing this earlier, but the User and Group fields should be "omitempty" as well, because if the API has an error when fetching the user/group info, these can be empty strings.

internals/cli/cmd_ls_test.go
```diff
@@ -145,3 +145,117 @@ func (s *PebbleSuite) TestLsInvalidPattern(c *C) {
...
+			"path": "/bar",
+			"name": "bar",
+			"type": "file",
+			"permissions": "644",
+			"last-modified": "2021-04-21T01:02:03Z",
+			"user-id": 600,
+			"user": "toor",
+			"group-id": 600,
+			"group": "toor",
+			"size": 1024
```
benhoyt, line 178: Let's add one more entry here with user-id/group-id = "null" and user/group = "" from the API, to ensure those are omitted in that case.

benhoyt (changes requested):
Thanks. This is very close, just a couple of very minor changes requested.

benhoyt (approved):
Excellent, thanks

PR #844, opened 2026-04-14
feat(cli): add structured output formats to `changes` & `tasks`

Snehil-Shah:
adds `--format` flag to support structured output formats to `changes` and `tasks` commands.

Example of `pebble changes` in YAML format:

```
$ pebble changes --format yaml
changes:
    - id: four
      kind: install
      summary: '...'
      status: Do
      tasks:
        - id: ""
          kind: bar
          summary: some summary
          status: Do
          progress:
            label: ""
            done: 0
            total: 1
          spawn-time: 2015-02-21T01:02:03Z
          ready-time: 2015-02-21T01:02:04Z
      ready: false
      spawn-time: 2015-02-21T01:02:03Z
      ready-time: 2015-02-21T01:02:04Z
    ...
```

Towards #824

### DIFF
--- a/internals/cli/cmd_changes.go
+++ b/internals/cli/cmd_changes.go
@@ -34,6 +34,7 @@ type cmdChanges struct {
 	client *client.Client
 
 	timeMixin
+	formatMixin
 	Positional struct {
 		Service string `positional-arg-name:"<service>"`
 	} `positional-args:"yes"`
@@ -49,6 +50,7 @@ type cmdTasks struct {
 	client *client.Client
 
 	timeMixin
+	formatMixin
 	changeIDMixin
 }
 
@@ -57,7 +59,7 @@ func init() {
 		Name:        "changes",
 		Summary:     cmdChangesSummary,
 		Description: cmdChangesDescription,
-		ArgsHelp:    timeArgsHelp,
+		ArgsHelp:    merge(timeArgsHelp, formatArgsHelp),
 		New: func(opts *CmdOptions) flags.Commander {
 			return &cmdChanges{client: opts.Client}
 		},
@@ -66,7 +68,7 @@ func init() {
 		Name:        "tasks",
 		Summary:     cmdTasksSummary,
 		Description: cmdTasksDescription,
-		ArgsHelp:    merge(changeIDMixinArgsHelp, timeArgsHelp),
+		ArgsHelp:    merge(changeIDMixinArgsHelp, timeArgsHelp, formatArgsHelp),
 		New: func(opts *CmdOptions) flags.Commander {
 			return &cmdTasks{client: opts.Client}
 		},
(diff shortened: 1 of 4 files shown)


Snehil-Shah:
@benhoyt clubbed them together as they touch the same files.

internals/cli/cmd_changes.go
```diff
@@ -152,6 +168,10 @@ func (c *cmdTasks) Execute([]string) error {
 	return c.showChange(chid)
 }
 
+type taskResult struct {
+	Change *client.Change `json:"change" yaml:"change"`
```
benhoyt: I think we should move the tasks to the top level, like so, to make it more consistent with "changes" and other commands:
```suggestion
	Tasks []*client.Task `json:"tasks" yaml:"tasks"`
```

benhoyt (commented):
For commands that are inherited from snapd, I'm going to check that they are okay with this too -- ideally we'd do it there too.

internals/cli/cmd_changes.go
Snehil-Shah: @benhoyt Makes sense, much cleaner.

internals/cli/cmd_changes.go
```diff
@@ -116,12 +118,26 @@ func (c *cmdChanges) Execute(args []string) error {
 		return err
 	}
 
-	if len(changes) == 0 {
-		return fmt.Errorf("no changes found")
+	sort.Sort(changesByTime(changes))
+
+	if c.Format == "text" {
+		if len(changes) == 0 {
+			return fmt.Errorf("no changes found")
```
benhoyt, line 125: We shouldn't change it in this PR, but I noticed that "snap changes" has updated this (correctly) to not be an error. I've opened https://github.com/canonical/pebble/issues/846 to track.

internals/cli/cmd_changes.go
```diff
@@ -169,10 +189,23 @@ func (c *cmdTasks) showChange(chid string) error {
 		return err
 	}
 
+	tasks := chg.Tasks
+	if tasks == nil {
```
benhoyt: In other commands we've put this ==nil check just before the `formatNotText` call (it's not needed for `--format text`). Let's move it there for consistency.

benhoyt (changes requested):
Just one tiny change requested, but looks good.

benhoyt (approved):

internals/cli/cmd_changes.go
Snehil-Shah: @benhoyt makes sense, looks like a small change, will create a PR once this is merged.

internals/cli/cmd_changes.go
```diff
@@ -152,6 +168,10 @@ func (c *cmdTasks) Execute([]string) error {
 	return c.showChange(chid)
 }
 
+type taskResult struct {
+	Tasks []*client.Task `json:"tasks" yaml:"tasks"`
+}
+
 func queryChange(cli *client.Client, chid string) (*client.Change, error) {
 	chg, err := cli.Change(chid)
```
pedronis, line 176: in snapd we have decided to ignore the fact that the command is called "tasks" and deal with the fact that historically is really about querying the state of a change, so the json formatting will be of full chg and not just chg.Tasks

pedronis (commented):
comment

internals/cli/cmd_changes.go
Snehil-Shah: @benhoyt could you weigh in on this? Happy to revert if we wish to be consistent with snap's signature.

benhoyt: @Snehil-Shah Let me confirm with Samuele (pedronis) that returning the full change is definitely their intention for `snap tasks --format=json` and if so, we can push up a small PR to revert.

benhoyt: @Snehil-Shah I've confirmed with the snap team, and they're definitely going to return the full change object. If you still would like to push up a tiny PR with that change, go ahead, otherwise let me know and I'll open an issue to track.

benhoyt: Actually, I've opened https://github.com/canonical/pebble/issues/849 to track in any case, so we don't forget.

PR #847, opened 2026-04-16
fix(cli): exit normally if "pebble changes" has no changes

### LINKED ISSUE #846: "pebble changes" should return exit code zero if there are no changes
As per https://github.com/canonical/snapd/pull/10245, when you run `pebble changes` and there are no changes yet, we should return exit code 0 (instead of 1). It's not an error, it's just that nothing significant has happened yet.

Snehil-Shah:
Updated as per https://github.com/canonical/snapd/pull/10245 to not return an error if no changes found. Now we return normally and just print an info message to `stderr`.

Resolves #846


benhoyt (approved):
Thanks for this fix.

PR #848, opened 2026-04-16
chore: prepare a CI-changes-only v1.30.1 release

tonyandrewmeyer:
Prepare for a v1.30.1 release.

This release doesn't include any non-CI changes compared with v1.30.0. At the time 1.30 was released, our snap building functionality was broken while we were rotating credentials and adjusting the workflow to use a dedicated environment. That's working now, but re-running the 1.30 release action doesn't work because at that time the workflow didn't point to the environment where the secrets now are located.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -61,6 +61,7 @@ jobs:
 
   remote-build:
     runs-on: ubuntu-latest
+    environment: snap-build
     if: github.event_name != 'pull_request'
     strategy:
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
The linting issues would be solved by moving to 1.26.2, but that doesn't need necessary in this PR, and sticking to the current version is more true to the nature of this release.

benhoyt (approved):
Thanks

PR #853, opened 2026-04-23
docs: remove reference to promtail release which no longer exists

benhoyt:
This should fix broken docs CI.


tonyandrewmeyer (approved):
Happy to just remove -- should we be considering mentioning Alloy or something else in the future, or do you think this small tip isn't really necessary anyway?

benhoyt:
Thanks. I reckon it's not really necessary anyway. People can find other tools that fit their needs if the simple suggestion (tail) doesn't apply.

PR #854, opened 2026-04-24
chore(daemon): rename InternalError to ServerError

holmanb:
The name InternalError used for HTTP 500 status codes
causes confusion. The semantics between the two differ.

Rename to clarify.


benhoyt (approved):
Fine with me, thanks.

hpidcock (approved):
Thank-you.

PR #860, opened 2026-05-10
refactor: move to canonical/tc test harness

hpidcock:
This is a mostly mechanical change from gocheck to [canonical/tc](https://github.com/canonical/tc).

The reason behind this change is to adopt learnings from Juju on best practices around
testing with suites, while also gaining use of `testing.T` benefits (e.g. fuzz testing, result caching etc.).

Since tc is a derivative of gocheck, the change is relatively boring as most continues to work the same.
Key changes are:
- `check.Suite(&mySuite{})` -> `func TestMySuite(t *testing.T) { tc.Run(t, &mySuite{}) }`
- Remove `check.TestingT` entrypoint.
- Use `tc.ErrorIsNil` for typed nil error checking.
- `Assert` errors instead of `Check`ing them.
- Use `c.Cleanup` instead of bespoke `testutils.BaseSuite`.
- Add `c.Cleanup` calls that reset suite state to zero to prevent use after test bugs.


hpidcock:
Will come back to this in the future.

PR #862, opened 2026-05-11
chore: update Go version to 1.26.3

tromai:
This PR bumps the Go version to 1.26.3 to resolve the issues reported by `govulncheck` in our CI (see [this CI run](https://github.com/canonical/pebble/actions/runs/25673337908/job/75364234870)).

`govulncheck` revealed 2 issues for Go version 1.26.2, which are fixed in Go version 1.26.3 (see [Go release log](https://go.dev/doc/devel/release#go1.26.minor)).

```
Vulnerability #1: GO-2026-4971
    Panic in Dial and LookupPort when handling NUL byte on Windows in net
...

Vulnerability #2: GO-2026-4918
    Infinite loop in HTTP/2 transport when given bad SETTINGS_MAX_FRAME_SIZE in
    net/http/internal/http2 in golang.org/x/net
...
```

We have validated that bumping the Go version to 1.26.3 fixed a similar issue in the concierge repo https://github.com/canonical/concierge/pull/193.

PR #865, opened 2026-05-20
ci: remove the (disabled) trivy scanning

tonyandrewmeyer:
We have `govulncheck` running in CI, which does a better job of detecting vulnerabilities than Trivy, and we don't really trust Trivy after the issues earlier this year. This PR removes the (currently disabled) Trivy scanning workflow, and the related configuration file.

A Trivy scan is still done as part of the `secscan` process during releasing.


benhoyt (approved):
Thanks!

tonyandrewmeyer:
Ironic that the `govulncheck` check fails for the PR 😂. I'll merge master into the branch after #862 lands and that should fix it.

benhoyt:
Hehe, yes, a bit ironic. I think we can go ahead and merge this one though, no need to wait till #862 is merged. I'll push it through.

PR #867, opened 2026-05-24
docs: document pebble enter command

### LINKED ISSUE #823: docs: "pebble enter" missing from documentation
Hi,
Simply put, `pebble enter` is not listed in the documentation, despite being available on the command line.
Could it be added? Is there some reason for why you would want it "hidden"?

Thanks!

---
*Reported from: https://documentation.ubuntu.com/pebble/reference/cli-commands/*

tonyandrewmeyer:
Adds `enter` to the CLI reference, with the standard auto-generated content, and a little bit of surrounding context to explain how it behaves in different contexts (like `layer` has) and examples of how it should be used (as many commands have).

[Preview](https://canonical-ubuntu-documentation-library--867.com.readthedocs.build/pebble/reference/cli-commands/#reference-pebble-enter-command)

Fixes #823

### DIFF
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -6,7 +6,7 @@ tocdepth: 2
 
 The `pebble` command has the following subcommands, organized into logical groups:
 
-* Run: [run](#reference_pebble_run_command)
+* Run: [run](#reference_pebble_run_command), [enter](#reference_pebble_enter_command)
 * Info: [help](#reference_pebble_help_command), [version](#reference_pebble_version_command)
 * Plan: [add](#reference_pebble_add_command), [plan](#reference_pebble_plan_command), [replan](#reference_pebble_replan_command)
 * Services: [services](#reference_pebble_services_command), [logs](#reference_pebble_logs_command), [start](#reference_pebble_start_command), [restart](#reference_pebble_restart_command), [signal](#reference_pebble_signal_command), [stop](#reference_pebble_stop_command)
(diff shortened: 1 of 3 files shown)


benhoyt (commented):
Was the help text manually generated, or did you use docs/generate_command_docs.py to generate it? I think we should add `enter` to the `get_all_commands` list in that script, otherwise it'll get out of date with the help text.

tonyandrewmeyer:
> Was the help text manually generated, or did you use docs/generate_command_docs.py to generate it? I think we should add `enter` to the `get_all_commands` list in that script, otherwise it'll get out of date with the help text.

Ah, I did run `generate_command_docs` to make sure there wasn't a mis-match, but I missed that it's excluded because it's not in the help output. Will fix, thanks!

hpidcock (approved):
Thank-you.

docs/reference/cli-commands.md
```diff
@@ -167,6 +167,92 @@ arguments.
...
+`pebble enter` is intended to be the container's entrypoint. With no subcommand it behaves like [`pebble run`](#reference_pebble_run_command): the daemon starts and runs the configured services in the foreground until interrupted.
+
+When a subcommand is given, `enter` runs that subcommand against the same Pebble environment, choosing the right runtime behavior for it:
+
+* `help`, `version` — run without starting the service manager.
+* `plan`, `services`, `ls` — start the service manager but skip default-service autostart (so the inspection is quick and quiet); logs are suppressed on stdout unless `-v`/`--verbose` is passed.
+* `exec` — start the service manager (without autostarting default services) and run a one-shot remote command; logs are suppressed.
+* `start`, `stop` — start the service manager and keep it running after the subcommand returns, so the container stays up.
+
+This makes `enter` a convenient way to share one container image across run, inspection and exec entrypoints without invoking the daemon separately each time.
```
dwilding: I found the concept of "sharing" a container image confusing. Also, the above breakdown of subcommands suggests that the daemon does start - if I understand correctly that "service manager" == "daemon". How about something like this instead:

```suggestion
This makes `enter` a convenient way for a container image to provide inspection and exec modes that don't start the default services.
```

docs/reference/cli-commands.md
```diff
@@ -167,6 +167,92 @@ arguments.
...
+
+```{terminal}
+pebble enter exec -- echo hello
+```
+
+Inspect the plan without starting any services:
+
+```{terminal}
+pebble enter plan
+```
```
dwilding: Would it be better to (also?) show examples from outside the container? Such as:

```
docker run --rm <image> plan
```

docs/reference/cli-commands.md
```diff
@@ -167,6 +167,92 @@ arguments.
 <!-- END AUTOMATED OUTPUT FOR checks -->
 
 
+(reference_pebble_enter_command)=
+## enter
+
+The `enter` command is used to run Pebble as a container entrypoint, optionally invoking another Pebble subcommand inside the same environment.
```
dwilding: The command/subcommand terminology varies throughout the doc and I don't think it's possible to make it 100% precise without becoming unreadable. But I'd go with "command" here as we're comparing to the same level as `enter`.


```suggestion
The `enter` command is used to run Pebble as a container entrypoint, optionally invoking another Pebble command inside the same environment.
```

dwilding (commented):
Looks great, thanks! Some small questions/suggestions

dwilding (approved):

docs/reference/cli-commands.md
tonyandrewmeyer: Discussed this in a call and we feel that we should just drop the whole examples section.

docs/reference/cli-commands.md
```diff
@@ -167,6 +167,92 @@ arguments.
...
+
+```{terminal}
+pebble enter exec -- echo hello
+```
+
+Inspect the plan without starting any services:
+
+```{terminal}
+pebble enter plan
+```
```
tonyandrewmeyer:
```suggestion
```

PR #874, opened 2026-06-01
feat: log authz_fail security events on client-cert verify failure

tonyandrewmeyer:
This was flagged in the TIOBE security review, which mostly produced false positive noise, but this seems reasonable to do.

The PR adds emission of a `SecurityWarn(SecurityAuthzFail, ...)` audit record when a client certificate fails self-signed verification, both during identity lookup (`identities.IdentityFromInputs`) and during pairing (`pairingstate.PairMTLS`).

These provide a SIEM-friendly audit trail distinct from the generic error returned to the caller. It follows the pattern of the security event logging added a while back.

### DIFF
--- a/internals/overlord/identities/identities.go
+++ b/internals/overlord/identities/identities.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/GehirnInc/crypt/sha512_crypt"
 
+	"github.com/canonical/pebble/internals/logger"
 	"github.com/canonical/pebble/internals/overlord/state"
 )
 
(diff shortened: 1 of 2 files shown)


benhoyt (commented):
Thanks. I'll ask the downstream project team to review.

paul-rodriguez (approved):
Seems entirely reasonable to me. It will affect us, but only in the sense that the intended message will appear in the intended situation.

Implementation-wise I would have expected the use of `Sprintf` instead of string concatenation but that hardly matters.

benhoyt (approved):

hpidcock (approved):
Thank-you.

PR #876, opened 2026-06-02
fix: mitigate retry after cancellation

hpidcock:
Mitigates the retry loop in the client from dispatching after context
cancellation when `retry.C` and `ctx.Done()` fire simultaneously.

Additionally this fixes two other flaky tests:
- Fixes a flaky test in `TestEnterSubCommandWaits` that compared timestamps to
assert ordering, replaced with a sentinel file so the assertion is ignores
scheduling delays.
- Removes the brittle `osutil.IsTestBinary` regex helper in favour of the
standard `testing.Testing` introduced in go 1.21.

QA steps are to compile the test binaries, then run with `stress`.


benhoyt (approved):
Nice, thanks! Probably worth posting this PR on the "Cross-project snapd and derivative codebases" channel.

PR #877, opened 2026-06-02
test: goroutineleak profile for tests

hpidcock:
Using Go's experimental `goroutineleak` pprof profile, this PR adds into all
test packages goroutine leak detection using a new `PrintGoroutineLeaks` helper
in the testutil package.

When built with `GOEXPERIMENT=goroutineleakprofile`, any leaked goroutine stacks
are reported and the entry point test fails. If the test binary was compiled
without the `goroutineleakprofile` experiment flag, it is a no-op.

All `Test(t *testing.T)` entry points are updated to wrap their calls to
`check.TestingT` with the helper.

There is some complexity here to a) self-test the leak detection and b) ensure
all go routines have likely been garbage collected, but otherwise it is rather
straight-forward.


internals/testutil/goleak.go
```diff
@@ -0,0 +1,77 @@
...
+	"bytes"
+	"runtime"
+	"runtime/pprof"
+	"testing"
+)
+
+func PrintGoroutineLeaks(t *testing.T, f func(t *testing.T)) {
+	t.Helper()
+
+	grl := pprof.Lookup("goroutineleak")
```
benhoyt: I'd suggest a more meaningful name -- doesn't need to be super-short. Perhaps:

```suggestion
	leakProfile := pprof.Lookup("goroutineleak")
```

internals/testutil/goleak.go
```diff
@@ -0,0 +1,77 @@
...
+		// mark the test as failed if any leaked go routines was found
+		if leaked {
+			t.Fail()
+		}
+	}()
+
+	f(t)
+}
+
+func leakSentinal(done chan struct{}) {
```
benhoyt: I believe the spelling is "sentinel" (similar elsewhere):
```suggestion
func leakSentinel(done chan struct{}) {
```

internals/testutil/goleak.go
```diff
@@ -0,0 +1,77 @@
...
+	done := make(chan struct{})
+	go leakSentinal(done)
+	defer func() {
+		// leak the sentinal
+		done = nil
+
+		// find the sentinal in the go routine leak profile
+		out := &bytes.Buffer{}
+		sentinalBytes := []byte("leakSentinal")
+		for {
```
benhoyt, line 42: For defensive programming, do we need a timeout of max number of times for this loop? Or can that never happen?

internals/testutil/goleak.go
```diff
@@ -0,0 +1,77 @@
...
+			out.Reset()
+			runtime.GC()
+		}
+
+		// find any leaked go routines other than the sentinal
+		leakedBytes := []byte("(leaked)")
+		leaked := false
+		for stack := range bytes.SplitSeq(out.Bytes(), []byte("\n\n")) {
+			if bytes.Contains(stack, sentinalBytes) ||
+				!bytes.Contains(stack, leakedBytes) {
```
benhoyt: I think this would be clearer on one line (I thought this second line was in the body at first).

internals/testutil/goleak.go
```diff
@@ -0,0 +1,77 @@
...
+				continue
+			}
+			if leaked {
+				_, _ = t.Output().Write([]byte("\n\n"))
+			}
+			_, _ = t.Output().Write(stack)
+			leaked = true
+		}
+
+		// mark the test as failed if any leaked go routines was found
```
benhoyt:
```suggestion
		// mark the test as failed if any leaked goroutines were found
```

internals/testutil/goleak.go
```diff
@@ -0,0 +1,77 @@
...
+	grl := pprof.Lookup("goroutineleak")
+	if grl == nil {
+		f(t)
+		return
+	}
+
+	done := make(chan struct{})
+	go leakSentinal(done)
+	defer func() {
+		// leak the sentinal
```
benhoyt: We're not fully consistent, but the preferred style for comments is sentences: start with a capital letter and (normally) finish with a full stop:
```suggestion
		// Leak the sentinel.
```

benhoyt (commented):
Thanks for this. It's a pity the Go test runner doesn't have an option to enable this, but this seems good for now. Just a few minor questions/comments.

internals/testutil/goleak.go
hpidcock: go test handles the timeout, but for that to happen, the garbage collector would need to be misbehaving.

benhoyt (approved):

benhoyt:
I note that this doesn't turn on the new profile in CI, but I guess we'll do that once the leaks are fixed?

PR #880, opened 2026-06-03
feat: structured output for version and health commands

hpidcock:
This adds format to both `pebble version` and `pebble health` such that they
would be useful for machine interaction.

Version command is rather simple, but the health command removes the non-zero
exit code when health check fails for JSON or YAML format.

Closes #824.

### DIFF
--- a/internals/cli/cmd_health.go
+++ b/internals/cli/cmd_health.go
@@ -28,11 +28,14 @@ The health command queries the health of configured checks.
 
 It returns an exit code 0 if all the requested checks are healthy, or
 an exit code 1 if at least one of the requested checks are unhealthy.
+If --format is passed either json or yaml, an exit code of 0 is returned for
+both healthy and unhealthy requested checks.
 `
 
 type cmdHealth struct {
 	client *client.Client
 
+	formatMixin
 	//lint:ignore SA5008 "choice" tag is intentionally duplicated
 	Level      string `long:"level" choice:"alive" choice:"ready"`
 	Positional struct {
@@ -49,11 +52,15 @@ func init() {
 		Name:        "health",
 		Summary:     cmdHealthSummary,
 		Description: cmdHealthDescription,
-		ArgsHelp:    cmdHealthArgsHelp,
+		ArgsHelp:    merge(cmdHealthArgsHelp, formatArgsHelp),
 		New: func(opts *CmdOptions) flags.Commander {
 			return &cmdHealth{client: opts.Client}
 		},
 	})
+}
+
+type healthResult struct {
+	Healthy bool `json:"healthy" yaml:"healthy"`
 }
 
 func (cmd *cmdHealth) Execute(args []string) error {
@@ -70,14 +77,23 @@ func (cmd *cmdHealth) Execute(args []string) error {
 		return err
 	}
 
-	status := "unhealthy"
-	if health {
-		status = "healthy"
-	}
-	fmt.Fprintln(Stdout, status)
+	if cmd.Format == "text" {
+		status := "unhealthy"
+		if health {
+			status = "healthy"
+		}
+		fmt.Fprintln(Stdout, status)
 
-	if !health {
-		panic(&exitStatus{1})
+		if !health {
+			panic(&exitStatus{1})
+		}
+
+		return nil
+	}
+
+	err = cmd.formatNonText(healthResult{Healthy: health})
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/internals/cli/cmd_health_test.go
+++ b/internals/cli/cmd_health_test.go
@@ -170,7 +170,7 @@ func (s *PebbleSuite) TestHealthUnhealthyYAML(c *check.C) {
 	exitCode := cli.PebbleMain()
 	// Ensure an unhealthy check does not fail with YAML output.
 	c.Check(exitCode, check.Equals, 0)
-	c.Check(s.Stdout(), check.Equals, `healthy: false`+"\n")
+	c.Check(s.Stdout(), check.Equals, "healthy: false\n")
 	c.Check(s.Stderr(), check.Equals, "")
 }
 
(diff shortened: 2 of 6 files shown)


internals/cli/cmd_health.go
```diff
@@ -28,11 +28,14 @@ The health command queries the health of configured checks.
 
 It returns an exit code 0 if all the requested checks are healthy, or
 an exit code 1 if at least one of the requested checks are unhealthy.
+If --format is passed either json or yaml, an exit code of 0 is returned for
+both healthy and unhealthy requested checks.
```
benhoyt, line 32: Are we sure about this? At first it seems odd to me that the exit code changes based on output format. And if you're piping to `jq`, for example, shell swallows the error code (though I guess that depends on the pipefail setting?). In any case, do you mind replying with a bit more justification for the change?

internals/cli/cmd_health_test.go
```diff
@@ -99,3 +99,113 @@ func (s *PebbleSuite) TestHealthBadLevel(c *check.C) {
...
+	restore := fakeArgs("pebble", "health", "--format", "foobar")
+	defer restore()
+
+	exitCode := cli.PebbleMain()
+	c.Check(exitCode, check.Equals, 1)
+	c.Check(s.Stdout(), check.Equals, "")
+	c.Check(s.Stderr(), check.Matches, "error: Invalid value .* Allowed values are: text, json or yaml\n")
+}
+
+func (s *PebbleSuite) TestHealthInvalidResponseYAML(c *check.C) {
```
benhoyt, line 187: I think we could drop these two tests, or at least one of them.

internals/cli/cmd_health_test.go
```diff
@@ -99,3 +99,113 @@ func (s *PebbleSuite) TestHealthBadLevel(c *check.C) {
...
+		}`)
+	})
+
+	restore := fakeArgs("pebble", "health", "--format", "yaml")
+	defer restore()
+
+	exitCode := cli.PebbleMain()
+	// Ensure an unhealthy check does not fail with YAML output.
+	c.Check(exitCode, check.Equals, 0)
+	c.Check(s.Stdout(), check.Equals, `healthy: false`+"\n")
```
benhoyt:
```suggestion
	c.Check(s.Stdout(), check.Equals, "healthy: false\n")
```

internals/cli/cmd_version.go
```diff
@@ -39,34 +40,50 @@ func init() {
...
 func (cmd cmdVersion) Execute(args []string) error {
 	if len(args) > 0 {
 		return ErrExtraArgs
 	}
 
 	if cmd.ClientOnly {
+		if cmd.Format != "text" {
+			return cmd.formatNonText(
+				versionResult{Client: version.Version},
+			)
```
benhoyt: Easily fits on one line:
```suggestion
			return cmd.formatNonText(versionResult{Client: version.Version})
```

internals/cli/cmd_version.go
```diff
@@ -39,34 +40,50 @@ func init() {
...
 		fmt.Fprintln(Stdout, version.Version)
 		return nil
 	}
 
-	return printVersions(cmd.client)
+	return printVersions(cmd.client, &cmd.formatMixin)
 }
 
-func printVersions(cli *client.Client) error {
+func printVersions(cli *client.Client, format *formatMixin) error {
```
benhoyt: Could this just be a method of `cmdVersion` and avoid the need for this arg entirely?

internals/cli/cmd_version.go
```diff
@@ -39,34 +40,50 @@ func init() {
...
 	serverVersion := "-"
 	sysInfo, err := cli.SysInfo()
 	if err == nil {
 		serverVersion = sysInfo.Version
 	}
+	if format != nil && format.Format != "text" {
+		return format.formatNonText(versionResult{
+			Client: version.Version,
+			Server: serverVersion,
```
benhoyt, line 83: In the JSON and YAML version, do we want this to be "" (and omitempty'd) rather than "-" when the server times out or errors? And perhaps add a test for this case.

internals/cli/cmd_health_test.go
hpidcock: Right, the importance with these two is that some bad response still produces a non-zero exit code. But if we decide to not make the change above regarding json/yaml output changing the exit code, then we don't need this.

internals/cli/cmd_health.go
hpidcock: The justification is that, `pebble health --format json` is for machine consumption (as in, they asked for json/yaml not for the status code), but I can understand the subtle behaviour difference might be lost.

An alternative is to add another argument like, `--no-fail` or `--exit-0`, which disables the failure status reporting as an exit code (as in only genuine connectivity issues/panics etc cause a non-zero exit code).

Lastly, we could just drop the yaml/json formatting for this command.

internals/cli/cmd_version.go
hpidcock: printVersions is used by the global command, so I'd need to duplicate version printing, happy to do so.

internals/cli/cmd_version.go
hpidcock: Yep, good choice.

internals/cli/cmd_health.go
benhoyt: Makes sense, and yeah, I think your initial instinct is probably right. Let's make it always return 0 in the JSON/YAML case then, as you have it. If I were using `subprocess.run(['pebble', 'health', '--format=json')` in Python and then parsing the JSON, I wouldn't want to have to worry about the exit code.

internals/cli/cmd_version.go
benhoyt: Ah, I see, I didn't realise it was used globally. Re duplication, up to you: whatever's clearest/simplest.

benhoyt (approved):
Looks good to me, thanks.

PR #884, opened 2026-06-12
ci: enforce Conventional Commits on PR titles

tonyandrewmeyer:
This PR ports canonical/operator's validate-pr-title.yaml workflow, which validates that pull request titles follow the Conventional Commits specification.

The workflow uses actions/checkout@v6.0.2 (the only action) and runs the local .github/check-conventional-pr-title.py script. There is no third-party action. The script's type list and optional component scopes are adapted to match the local convention from HACKING.md rather than operator's stricter no-scopes set.

This is a stdlib only script, run with plain `python3` from the runner image, but if it's too Python for here, I could probably make a bash (or Go) version. Or, if pebble doesn't need to align with the rest of Charm Tech, we could leave this as only enforced by reviewers manually.

### DIFF
--- a/.github/check-conventional-pr-title.py
+++ b/.github/check-conventional-pr-title.py
@@ -1,0 +1,75 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Check that a PR title follows the Conventional Commits specification.
+
+Reads the PR title from the PR_TITLE environment variable.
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
Seems reasonable to me, thanks. But we should get @hpidcock's take too.

hpidcock (approved):
Huge fan of this.

PR #886, opened 2026-06-16
build: armhf + re-enable other architectures

hpidcock:
This builds pebble binaries for armv6/armhf and re-enables snap building
for ppc64el, riscv64 and s390x.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         goos: [linux]
-        goarch: [amd64, arm64, armhf, ppc64le, riscv64, s390x]
+        goarch: [amd64, arm64, armv6, ppc64le, riscv64, s390x]
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer (approved):
As far as I can tell, this should be fine. Most of the issues I had publishing were with auth to the remote builds and snap store, so I expect this should just work (although moving away from remote building in the future still seems valuable, and I think these changes will still just work then). Thanks!

tonyandrewmeyer:
FYI the [failure here](https://github.com/canonical/pebble/actions/runs/27724018212/job/82016080717) is the same wording as when I was having all the troubles building the snaps in April. Not sure if you can see this but [this is the launchpad support ticket I opened](https://answers.launchpad.net/launchpad/+question/824057). At the time the conclusion was that it was a combination of the attacks happening at the time (maybe still?) and some dependency that had been updated on the remote build side. I don't know if either of those are what's happening here or if this is entirely different, but just FYI in case it's useful.

PR #888, opened 2026-06-18
fix: notify change listeners after AppendLayer and SetServiceArgs

hpidcock:
The inner `newPlan, err := m.appendLayer(...)` declarations shadowed
the outer `newPlan` variable read by the deferred
`callChangeListeners(newPlan)` call, leaving the outer value nil.
Because callChangeListeners ignores a nil plan, listeners were never
notified of successful AppendLayer or SetServiceArgs operations.

Adds regression tests covering both paths (success notifies, failure
does not).

### DIFF
--- a/internals/overlord/planstate/manager.go
+++ b/internals/overlord/planstate/manager.go
@@ -129,7 +129,8 @@ func (m *PlanManager) AppendLayer(layer *plan.Layer, inner bool) error {
 		return &LabelExists{Label: layer.Label}
 	}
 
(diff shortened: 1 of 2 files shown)


benhoyt (changes requested):
The fixes look good, thanks. However, when I revert the changes to manager.go locally, the regression tests still pass, so I don't think they're testing what they should be.

hpidcock:
Never-mind, this is not a problem.

PR #889, opened 2026-06-22
fix: allow running on read-only filesystem

lczyk:
runs pebble on a read-only rootfs. addresses #801.

two things get written to the pebble dir on startup and break when it's read-only: the state file and the identity key file. when `PEBBLE_PERSIST` is set, the state and the key are now stored in memory. 

btw, here are the things which still need a writable path:
  - the unix socket. could switch to abstract sockets maybe, but i'm not sure what would be the repercussions.. 
  - the TLS identity certificate on the first https connection. so https still breaks.
  
for both i thought it better to keep this PR small and fix this separately.

for now, "demo.sh":

```bash
rootfs="$PWD/rootfs"
socket="$PWD/pebble.socket"

trap 'kill ${pid:-} 2>/dev/null; umount rootfs 2>/dev/null; rm -rf "$rootfs"; rm -f "$socket"' EXIT

mkdir -p "$rootfs" && mount -t tmpfs -o size=1m tmpfs "$rootfs" && mount -o remount,ro "$rootfs"

env \
    PEBBLE="$PWD/rootfs" \
    PEBBLE_SOCKET="$socket" \
    PEBBLE_PERSIST=never \
    ./pebble run --hold &
pid=$!

for _ in $(seq 100); do [[ -S "$socket" ]] && break; sleep 0.1; done
[[ -S "$socket" ]] || { echo "FAIL: socket never appeared"; exit 1; }
```

from repo root:

```bash
$ sudo bash ./demo.sh
2026-06-22T12:03:03.560Z [pebble] Not persisting identity key; using an ephemeral identity.
2026-06-22T12:03:03.560Z [pebble] {"type":"security","datetime":"2026-06-22T12:03:03Z","level":"WARN","event":"sys_startup:0","description":"Starting daemon","appid":"pebble"}
2026-06-22T12:03:03.560Z [pebble] Started daemon.
2026-06-22T12:03:03.660Z [pebble] Exiting on terminated signal.
2026-06-22T12:03:03.660Z [pebble] {"type":"security","datetime":"2026-06-22T12:03:03Z","level":"WARN","event":"sys_shutdown:0","description":"Shutting down daemon","appid":"pebble"}
```

### DIFF
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -194,10 +194,7 @@ func runDaemon(rcmd *cmdRun, ch chan os.Signal, ready chan<- func()) error {
 	plan.RegisterSectionExtension(workloads.WorkloadsField, &workloads.WorkloadsSectionExtension{})
 	plan.RegisterSectionExtension(pairingstate.PairingField, &pairingstate.SectionExtension{})
 
-	// When state isn't persisted (in-memory backend, e.g. on a read-only
-	// rootfs) the identity key is ephemeral too -- it's never written to disk.
 	persist := os.Getenv("PEBBLE_PERSIST") != "never"
-
 	idPath := filepath.Join(rcmd.pebbleDir, "identity")
 	idSigner, err := idkey.Get(idPath, persist)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


lczyk:
@anpep pinging you since you appear as the assignee for #801

hpidcock:
Thank-you @lczyk, there is a roadmap item to handle this issue, I'll need to review this within that context and get back to you.

hpidcock:
Thank-you @lczyk for this, we're moving forward with #896, as it handles this slightly differently by avoiding the indentity key generation when it is not required/not valuable.

PR #896, opened 2026-06-30
fix: only generate identity key when HTTPS is enabled

### LINKED ISSUE #801: Moving identities to own package may have broken the read-only mode

hpidcock:
Previously `pebble run` always loaded or generated an identity
key under `$PEBBLE/identity`, even when HTTPS was not configured and
the key would never be used.

Skip the call to `idkey.Get` entirely when `--https` is unset. When
`--https` is set with `PEBBLE_PERSIST=never`, fail with an error since
the identity key cannot be generated without persistent state.

Fixes #801

### DIFF
--- a/internals/cli/cmd_run.go
+++ b/internals/cli/cmd_run.go
@@ -166,6 +166,26 @@ func sanityCheck() error {
 	return nil
 }
 
+// setupTLSOptions returns the TLS options for the daemon based on the
+// configured HTTPS address. If HTTPS is not requested no identity key is
+// loaded. If HTTPS is requested, an identity key is loaded (or generated)
+// from pebbleDir; this requires persistent state, so an error is returned
+// when persist is overlord.PersistNever.
+func setupTLSOptions(pebbleDir, httpsAddress string, persist overlord.PersistMode) (tlsstate.Options, error) {
+	if httpsAddress == "" {
+		return tlsstate.Options{}, nil
+	}
+	if persist == overlord.PersistNever {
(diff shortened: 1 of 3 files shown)


benhoyt (commented):
I'm happy with the change, thanks. Couple of questions:

- Does this apply for `pebble enter` too?
- I prefer us testing using the public or higher-level APIs where possible, rather than than un-exported `setpuTLSOptions` function. Is that doable?

benhoyt (approved):
Per discussion, we're going ahead with the tests as is for now.

PR #899, opened 2026-07-01
chore: merge master into fips branch for 1.32.0

hpidcock:
Merge master into fips branch. Resolved conflicts by dropping tlsstate,
pairing state and tls option setup.


tonyandrewmeyer (approved):
Checked by comparing with main (I ignored the docs since they're not used anyway) and seems identical other than the bits mentioned in the PR description.

I assume you'll figure out whatever the linting complaints are.

Thanks for doing this! I look forward to the day we can be fipsy in main and not need to do this :)

PR #904, opened 2026-07-01
chore: bump x/sys, x/net, x/term

tonyandrewmeyer:
Bumps to the latest versions to pull in upstream fixes. As far as I can tell, none of the security fixes apply to Pebble's use, so this is just to silence the scanner warnings.


hpidcock (approved):
Thank-you.

PR #912, opened 2026-07-09
build: add basic workshop with Pi.dev agent harness

### LINKED ISSUE #873: Setup Workshop for Pebble
We want to setup [workshop](https://documentation.ubuntu.com/canonical-workshop/latest/) to bootstrap Pebble's dev environment.

benhoyt:
What it says on tin: adds a very basic workshop "dev" with Go installed and the [Pi](https://pi.dev/) agent harness.

```
$ workshop launch  # this takes a couple of minutes the first time
$ workshop exec go test ./...  # run tests inside the workshop
$ workshop shell
> pi
/login
/model
...
```

Fixes #873


hpidcock:
I think the dev workshop should not have a built in agent. The dev workshop should be the bare essentials, then a dev-pi or dev-agent for the agentic workshop.

benhoyt:
I disagree, because then `workshop launch` won't work anymore, as there will be multiple workshops you'll have to specify the name `workshop launch dev-pi` or whatever, so it makes the dev experience a bit worse. I'd like to settle on an agent harness by default (but that doesn't prevent any individual dev from using a different one). We could have `pi` in the workshop be an alias to lazily install pi if it's not already installed.

I've also taken a look at how other workshop SDKs (in the store) install their tooling. It's a bit better than `curl https://example.com/install.sh | sh` -- they download a versioned file like https://github.com/canonical/copilot-sdk/blob/a02430240018bb14d3b061a986f7b5258b70f4e6/sdkcraft.yaml#L31-L36, and some of them do a checksum on it, like: https://github.com/canonical/opencode-sdk/blob/main/sdkcraft.yaml

Dmitry said there's also a Pi one that Hartmut put together (https://github.com/canonical/pi-coding-agent-sdk). He's going to check that it's best practice and then publish it properly. I suggest that once that's published we use that in the default workshop here.

benhoyt:
Reference reference, for reference: https://github.com/canonical/reference-sdks

hpidcock:
OK, we can wait to see what happens with the pi SDK.

hpidcock (approved):

benhoyt:
Decision is to merge this as is, and iterate from there -- merging.

PR #928, opened 2026-09-01
ci: refactor release process with automation of each stage

hpidcock:
Releasing Pebble is currently a sequence of manual steps, including lengthy PR
reviews. This adds a new `Release` workflow that drives an entire release,
including FIPS releases, from a single manual trigger.

Release handles the following:
1. Bumps `Version` in `cmd/version.go` on `master`, or, for a FIPS release,
merges an already-published release tag into `fips` and appends the `-fips`
suffix. Merge conflicts on the `fips` branch are pre-resolved by opencode,
following the branch's usual rules (drop incoming TLS/crypto, keep FIPS
specific CI config).
2. Opens a PR for that change with auto-merge enabled (squash for `master`,
merge commit for `fips`), then polls until it's approved and merged.
3. Tags the tip of the branch and drafts a GitHub release, with the title and
notes generated by opencode using the previous few releases of the same kind as
style examples.
4. Calls `binaries.yml`, `snap.yml` and `sbom-secscan.yaml` to build and attach
binaries, snaps and the secscan report to the draft release. Snaps go straight
to `latest/candidate`.

The person handling the release must manually:
1. Review the generated version bump PR or FIPS merge PR.
2. Review the generated draft GitHub release.
3. Wait for the builds to finish.
4. Publish the draft GitHub release.

Publishing it triggers `snap.yml`, which promotes the candidate build to
`latest/stable`.

The release process is mostly configurable (e.g. skip amd64, don't build snaps
etc.). Most importantly, the release process tries to be idempotent where it can.


.github/workflows/release.yml
```diff
@@ -0,0 +1,678 @@
...
+            if git merge --no-ff --no-commit "$VERSION"; then
+              echo "conflicts=false" >> "$GITHUB_OUTPUT"
+            else
+              echo "conflicts=true" >> "$GITHUB_OUTPUT"
+            fi
+          fi
+
+      - name: Install opencode
+        if: inputs.fips && steps.current.outputs.done != 'true' && steps.existing.outputs.found != 'true' && steps.merge.outputs.conflicts == 'true'
+        run: curl -fsSL https://opencode.ai/install | bash -s -- --version 1.18.25
```
tonyandrewmeyer: I don't love that this is curl->bash inside of a release workflow, and it's also version pinned somewhere nothing checks, I assume.

What about `snap install opencode`? The snap is published by us and is only a little behind, and then we'd always be roughly up-to-date, and presumably someone at Canonical is verifying that the install process is safe.

.github/workflows/release.yml
```diff
@@ -0,0 +1,678 @@
...
+
+      - name: Resolve conflicts with opencode
+        if: inputs.fips && steps.current.outputs.done != 'true' && steps.existing.outputs.found != 'true' && steps.merge.outputs.conflicts == 'true'
+        env:
+          OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
```
tonyandrewmeyer: I think we should have some mention somewhere (HACKING.md maybe?) that this key needs to get added for releasing to work, with some suggestion about expiry/budget limits.

.github/workflows/release.yml
```diff
@@ -0,0 +1,678 @@
...
+          PROMPT: |
+            This repository checkout is in the middle of a git merge with
+            conflicts: merging tag ${{ inputs.version }} into the fips
+            branch.
+
+            Run `git status` to find every conflicted file, and resolve all
+            of them, using this guidance:
+
+            - The fips branch intentionally excludes standard TLS/crypto
+              functionality in favour of FIPS-approved alternatives. Whenever
```
tonyandrewmeyer: It's hard to know for sure how well a prompt will work without actually exercising it, but I wonder if this will mislead a model, because the FIPS branch is intended to do this, but actually it doesn't favour alternatives, it just rips out the crypto instead (as the prompt says a sentence later). Maybe have both sentences align on the "drop it" approach rather than leading with "find alternatives"?

.github/workflows/release.yml
```diff
@@ -0,0 +1,678 @@
...
+        if: steps.current.outputs.done != 'true' && steps.existing.outputs.found != 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          LABEL: ${{ steps.vars.outputs.label }}
+        run: |
+          gh label create "$LABEL" \
+            --color 0E8A16 \
+            --description "Automated pull request for a Pebble release" \
+            --force
+
```
tonyandrewmeyer: Why create a label as part of the workflow? If the label is useful, it seems like we should just add it to the repo (via canonical-repo-automation) and have it always present rather than create it inside of the release process. If it didn't exist then it could error or just skip the label with a warning.

If we leave it like this it'll flip-flop because CRA will remove the label each time it runs then each release will add it back.

I think this may also fail because there isn't the permission listed for creating labels. But I'd rather not fix that and remove this instead.

.github/workflows/release.yml
```diff
@@ -0,0 +1,678 @@
...
+          set -eu
+          if [[ -n "$PREV_TAG" ]]; then
+            git --no-pager log --no-merges --pretty=format:'- %s (%h)' "${PREV_TAG}..${VERSION}" > .release/commits.txt
+          else
+            git --no-pager log --no-merges --pretty=format:'- %s (%h)' "${VERSION}" > .release/commits.txt
+          fi
+
+      - name: Install opencode
+        if: steps.existing.outputs.exists != 'true'
+        run: curl -fsSL https://opencode.ai/install | bash -s -- --version 1.18.25
```
tonyandrewmeyer: Same as above, maybe `sudo snap install opencode`?

HACKING.md
```diff
@@ -269,27 +269,27 @@ Recommended tone:
...
-- Open a second PR on the `fips` branch to bump the `Version` in `cmd/version.go` to the FIPS version, for example `v1.27.0-fips`, and have it reviewed and merged.
-- Publish a FIPS GitHub release (for example `v1.27.0-fips`) targeting the tip of the `fips` branch.
-- As with the main release, monitor the release [GitHub Actions](https://github.com/canonical/pebble/actions), check that the FIPS snap is uploaded, and promote it to stable.
+Do this after you've published the main release.
+
+1. Run the [Release](https://github.com/canonical/pebble/actions/workflows/release.yml) workflow again with the same version tag, for example `v1.27.0`, this time checking the `fips` checkbox. It merges that tag into `fips` and bumps the version to `v1.27.0-fips`.
+2. Review the pull request it opens carefully. AI resolves the merge conflicts for you, so check each resolution, especially anything touching TLS or crypto. Then approve the CI actions to run and approve the pull request. Don't merge it yourself -- auto-merge is already enabled, and the release workflow carries on as soon as it merges.
+4. Follow steps 3 to 6 above to publish the `v1.27.0-fips` draft release.
```
tonyandrewmeyer: Nit, which I think rendering would mask out anyway:

```suggestion
3. Follow steps 3 to 6 above to publish the `v1.27.0-fips` draft release.
```

.github/workflows/snap.yml
```diff
@@ -167,73 +282,199 @@ jobs:
...
+          STABLE_REV=$(echo "$STATUS" | awk 'NR==1 { pos = index($0, "Revision") } /stable/ { print 0+substr($0, pos) }')
+          echo "Candidate revision: $CANDIDATE_REV, stable revision: $STABLE_REV"
+          echo "revision=${CANDIDATE_REV}" >> "$GITHUB_OUTPUT"
+          if [[ "$CANDIDATE_REV" == "$STABLE_REV" ]]; then
+            echo "already-promoted=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "already-promoted=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Promote snap to ${{ env.DEFAULT_TRACK }}/${{ env.STABLE_RISK }}
```
tonyandrewmeyer, line 542: I might have missed it, sorry, but does this handle the FIPS case? It looks like it's always going to promote to latest/stable, even if it's a FIPS release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,678 @@
...
+            BODY="Automated merge of ${VERSION} into fips, in preparation for the ${VERSION}-fips release.
+
+          Conflicts (if any) were resolved by dropping TLS and crypto usages from the merged code; please review carefully."
+          else
+            BODY="Automated version bump for the ${VERSION} release.
+
+          Once this is approved and merged, a draft GitHub release will be created automatically."
+          fi
+
+          PR_URL=$(gh pr create \
```
tonyandrewmeyer, line 352: Claude tells me that because this is opened with `GITHUB_TOKEN` no workflow runs are triggered and required status checks don't complete, blocking auto-merge. I don't know if that's an issue, or whether you tested this part in your fork, seems like Claude might be wrong here, but mentioning it in case.

tonyandrewmeyer (approved):
I didn't try this on my fork, so I'm assuming that the e2e testing you've done covers things, or that we roll forward with fixes as needed - testing publish/release workflows is a pain generally.

I'd rather that we did not create a label in this workflow, and I'd prefer that we didn't do an opencode install as curl->bash. Also, the security issues need resolving or ignoring as you mentioned you'd do in chat.

Other than that, some minor notes/questions, but approving optimistically.

.github/workflows/release.yml
hpidcock: I think I'll break this out into its own job and snap install it. Keeps it low on permissions and from a Canonical source (rather than the canonical source :smile:).

.github/workflows/release.yml
hpidcock: I'll find an alternative, labels are a bit iffy.

HACKING.md
hpidcock: Would you believe that I manually messed this part up. Humans...

.github/workflows/release.yml
hpidcock: I need to check, since the release was made by a github actions token, but the publisher is a user.

hpidcock: For PRs, it works fine since the user manually pokes it with a button.

PR #933, opened 2026-09-04
fix(reaper): remove race window between start and signal listening

### LINKED ISSUE #932: Race condition in reaper startup code
I frequently see flaky test failures in `TestStartFastExitCommand` (state is "Done", expected "Error"). Some test environments (such as when running the test as root) cause the test to fail more frequently.

I have traced the issue to a race condition in the reaper.

`reaper.Start()` sets up a SIGCHLD relay channel inside of `reapChildren`, which is spawned inside of a goroutine. This means that `reaper.Start()` can exit before the channel is set up. If the spawned process completes before the signal relay channel is set up, then the SIGCHLD is ignored and `reaper.WaitCommand()` is blocked waiting on the channel that returns the exit code. Eventually the okay-wait timer elapses in the "do" handler, so the service is considered running successfully. Since the test is checking for fast failures, and the reaper looses the failure signal, the test fails.

I created a reproducer (LLM assisted):

```go
package reaper_test

import (
	"os/exec"
	"runtime"
	"sync"
	"sync/atomic"
	"testing"
	"time"

	. "gopkg.in/check.v1"

	"github.com/canonical/pebble/internals/reaper"
)

func Test(t *testing.T) { TestingT(t) }

type reaperSuite struct{}

var _ = Suite(&reaperSuite{})

func (s *reaperSuite) SetUpTest(c *C) {
	err := reaper.Start()
	c.Assert(err, IsNil)
}

func (s *reaperSuite) TearDownTest(c *C) {
	err := reaper.Stop()
	c.Assert(err, IsNil)
}

// TestFastExitBeforeHandlerInstalled reproduces the mechanism behind the
// occasional TestStartFastExitCommand failure in overlord/servstate.
//
// On an idle machine the race window is small (the reaper goroutine is
// scheduled almost immediately). This test widens the window with
// GOMAXPROCS(1), CPU-bound spinner goroutines queued ahead of the
// reaper goroutine keep the single process busy, so that the handler is
// not installed until after the fast-exiting child has died.
func (s *reaperSuite) TestFastExitBeforeHandlerInstalled(c *C) {
	// Serialize goroutines to delay the reaper goroutine.
	oldProcs := runtime.GOMAXPROCS(1)
	defer runtime.GOMAXPROCS(oldProcs)

	// Queue CPU-bound spinners ahead of the reaper goroutine: they will
	// each hold the single P for ~10ms (until async preemption), keeping
	// the SIGCHLD handler uninstalled for tens of milliseconds.
	const spinners = 4
	var stopSpinners atomic.Bool
	var wgSpinners sync.WaitGroup
	wgSpinners.Add(spinners)
	for i := 0; i < spinners; i++ {
		go func() {
			defer wgSpinners.Done()
			for !stopSpinners.Load() {
				// Busy-wait until preempted.
			}
		}()
	}

	// Start the reaper; its goroutine is now queued behind the spinners,
	// so the SIGCHLD handler is NOT yet installed.
	err := reaper.Start()
	c.Assert(err, IsNil)

	// Start a child that exits as fast as possible, like the "echo"
	// service in TestStartFastExitCommand. It will fork, exec, and die
	// while the handler is still not installed.
	cmd := exec.Command("true")
	err = reaper.StartCommand(cmd)
	c.Assert(err, IsNil)

	// The exit must be delivered even though the child died before the
	// reaper goroutine installed the SIGCHLD handler. When the exit is
	// lost, WaitCommand blocks forever, hence the timeout.
	waitDone := make(chan int, 1)
	go func() {
		exitCode, err := reaper.WaitCommand(cmd)
		if err != nil {
			exitCode = -1
		}
		waitDone <- exitCode
	}()
	select {
	case exitCode := <-waitDone:
		c.Assert(exitCode, Equals, 0)
	case <-time.After(5 * time.Second):
		c.Fatal("timed out in WaitCommand: the child's exit was never delivered (lost SIGCHLD)")
	}

	// Let the spinners finish so the deferred GOMAXPROCS restore and
	// test teardown aren't starved.
	time.Sleep(10 * time.Second)
	stopSpinners.Store(true)
	wgSpinners.Wait()
}
```

A proposed fix: #933

holmanb:
Prevent a signal from arriving before the handler is
configured by moving handler setup into `Start()`. 

Fixes https://github.com/canonical/pebble/issues/932

### DIFF
--- a/internals/reaper/reaper.go
+++ b/internals/reaper/reaper.go
@@ -52,7 +52,9 @@ func Start() error {
 	}
 
 	started = true
-	reaperTomb.Go(reapChildren)
+	sigChld := make(chan os.Signal, 1)
+	signal.Notify(sigChld, unix.SIGCHLD)
+	reaperTomb.Go(func() error { return reapChildren(sigChld) })
 	return nil
 }
 
@@ -100,10 +102,8 @@ func setChildSubreaper(set int) error {
 
 // reapChildren "reaps" (waits for) child processes whose parents didn't
 // wait() for them. It stops when the reaper tomb is killed.
-func reapChildren() error {
+func reapChildren(sigChld chan os.Signal) error {
 	logger.Debugf("Reaper started, waiting for SIGCHLD.")
-	sigChld := make(chan os.Signal, 1)
-	signal.Notify(sigChld, unix.SIGCHLD)
 	for {
 		select {
 		case <-sigChld:


hpidcock (approved):
LGTM, will land after the release has finalised.

PR #934, opened 2026-09-06
ci: swap attest-build-provenance for actions/attest

tonyandrewmeyer:
`actions/attest-build-provenance` has been a thin wrapper around `actions/attest` since its v4.0.0, and its release notes recommend using `actions/attest` directly, so this swaps the action in both places it's used: the binaries workflow and the snap one.

With no predicate inputs, `actions/attest` defaults to build provenance, so `subject-path` is the only input either action needs and the job permissions are unchanged. Both were still pinned to v4.1.1; the new pin is v4.2.2, the current release.


hpidcock (approved):
LGTM, will land after the release has been finalised.

PR #935, opened 2026-09-07
ci: allow snaps to optionally be built via qemu

hpidcock:
Due to issues with riscv builds via launchpad taking a very long time, this
adds an optional fallback to build via qemu for any specified architecture.


tonyandrewmeyer (approved):
Seems fine to me. I assume at some point we will want to simplify all of this to just use emulated or the self-hosted runners or remote-but-it's-actually-reliable-and-good or something.

PR #936, opened 2026-09-07
ci: generate build matricies and fix qemu image download

hpidcock:
This fixes the qemu image downloading as well as changes the binaries and snaps jobs
to run a matrix generate action first then run the required actions based on that rather
than consuming action runners for skipped jobs.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -48,8 +48,11 @@ jobs:
   # it has nothing to do.
   plan:
     runs-on: ubuntu-latest
+    # contents: write (not just read) is required for `gh release view` to
+    # see a draft release -- drafts are only visible to callers with push
+    # access to the repo.
     permissions:
-      contents: read
+      contents: write
     outputs:
       goarch: ${{ steps.plan.outputs.goarch }}
     steps:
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
Reviewing this with my old human eyes it seems fine, although I obviously didn't try any of it myself.

For what it's worth, my agent told me this stuff, none of which I'm confident about, but passing it along just in case:

1. .github/workflows/binaries.yml:49 — plan has no actions/checkout and no GH_REPO, so gh release view fails ("no git remotes found"), silently swallowed by 2>/dev/null. The idempotency branch can never be taken → re-running a release rebuilds amd64 and hits gh release upload without --clobber ("asset already exists"). Fix: GH_REPO: ${{ github.repository }} (the repo already does this at snap.yml:620/708/771, release.yml:725).
2. .github/workflows/snap.yml:191 — same bug in plan-snap-builds; every arch always reports "not yet built", so a retry re-runs all six Launchpad builds and republishes all six to candidate.
3. binaries.yml:52 and 4. snap.yml:197 — both plan jobs use permissions: contents: read, but release.yml:744 creates a draft release, and draft listings require push access. Even after fixing 1/2, the check 404s. Needs contents: write.
5. snap.yml:629 — release is now skipped when built-arch == '[]', but record-skipped-architectures gates on needs.release.result == 'success', so the pebble-snap-skip-arch / dry-run markers never get written. promote then finds no markers and promotes stale candidates to stable for all six arches. Suggest gating on result != 'failure'.
6. snap.yml:257 — new qemu → skip → remote precedence means qemu-<arch>: true overrides skip-<arch>: true, so that arch gets published, yet record-skipped-architectures still writes a skip marker for it from inputs.skip-*. The two jobs disagree on an unusual input combination.

PR #938, opened 2026-09-07
fix(metrics): write each metric family once in the metrics response

lenamonj:
With two or more services, or two or more checks, `/v1/metrics` is not valid Prometheus text exposition: each service and check writes its own `# HELP` and `# TYPE` lines, so the same family appears several times with other families in between. The Prometheus text parser (`prometheus/common` `expfmt.TextParser`) rejects the response with a second-HELP-line error, so nothing is scraped.

`OpenTelemetryWriter` now buffers samples by metric name and `Flush` writes one HELP and TYPE line per family followed by that family's samples, in first-seen order. `v1GetMetrics` calls `Flush` before writing the response. The `metrics.Writer` interface and both `WriteMetrics` callers are unchanged; the one-service, one-check example in the docs still matches the output.

The servstate `TestMetrics` expectation asserted the duplicated, interleaved form; it now asserts the grouped form with no assertion removed. Both test suites call `Flush`.

`go test -race ./...`, `go build ./cmd/pebble`, `gofmt`, `staticcheck` pass.

### DIFF
--- a/internals/daemon/api_metrics.go
+++ b/internals/daemon/api_metrics.go
@@ -44,6 +44,13 @@ func v1GetMetrics(c *Command, r *http.Request, _ *UserState) Response {
 			return
 		}
 
+		err = metricsWriter.Flush()
+		if err != nil {
+			logger.Noticef("Cannot write metrics: %v", err)
+			http.Error(w, "# internal server error", http.StatusInternalServerError)
+			return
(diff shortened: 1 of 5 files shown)


hpidcock:
Thank-you for the PR. The fix has been absorbed into #1075.

PR #943, opened 2026-09-09
docs: document NSS user and group resolution limitation

### LINKED ISSUE #868: Pebble returns incorrect fileinfo on NSS-managed systems
Many of us are now using NSS-managed systems. Pebble is built with `CGO_ENABLED=0`, which means Go's os/user package uses its pure-Go fallback that only reads `passwd` and `group` directly — it cannot query NSS/SSSD/LDAP.

This is unlikely to be an issue for charms, as Pebble is used to manage Kubernetes workload containers where it runs as root, rather than an NSS-managed user. This was noticed when running Pebble locally for testing purposes in `charmlibs` (https://github.com/canonical/charmlibs/issues/487), but this is resolved by updating the test code to expect this limitation.

FWIW Concierge faced the same problem, which was resolved by using `getent` (https://github.com/canonical/concierge/pull/183). That makes sense for Concierge (a classic confinement snap managing installation of other packages on arbitrary systems), but would not be the right approach for Pebble.

Perhaps resolving this is ultimately out of scope for Pebble's use cases, in which case it may be worth documenting the limitation.

chapman-von-monty:
Because Pebble is statically compiled with CGO_ENABLED=0, user and group
lookups in file operations (such as resolving user and group names in
GET /v1/files and pebble ls -l) use the pure-Go resolver in Go's os/user
package. This resolver reads /etc/passwd and /etc/group directly rather
than querying system NSS plugins, SSSD, or LDAP.

On NSS-managed systems, file metadata queries return numeric IDs (user-id
and group-id) accurately, but usernames and group names may resolve as
empty strings.

This documents this known limitation across:
1. The API reference (reference/api.md under ## Data formats).
2. The CLI reference for the ls command (reference/cli-commands.md).
3. The OpenAPI specification (docs/specs/openapi.yaml on FileInfo).
4. The Go client FileInfo methods (client/files.go on User() and Group()).
5. The docs spelling wordlist (.custom_wordlist.txt).

Closes #868

### DIFF
--- a/client/files.go
+++ b/client/files.go
@@ -105,17 +105,11 @@ func (fi *FileInfo) GroupID() *int {
 }
 
 // User is the string representing the owner user name.
-//
-// Because Pebble is built with CGO_ENABLED=0, user lookups rely on /etc/passwd
-// and will not resolve usernames managed via NSS, SSSD, or LDAP.
 func (fi *FileInfo) User() string {
 	return fi.user
 }
 
 // Group is the string representing the owner user group.
-//
-// Because Pebble is built with CGO_ENABLED=0, group lookups rely on /etc/group
-// and will not resolve group names managed via NSS, SSSD, or LDAP.
 func (fi *FileInfo) Group() string {
 	return fi.group
 }
--- a/client/files.go
+++ b/client/files.go
@@ -105,11 +105,17 @@ func (fi *FileInfo) GroupID() *int {
 }
 
 // User is the string representing the owner user name.
+//
+// Because Pebble is built with CGO_ENABLED=0, user lookups rely on /etc/passwd
+// and will not resolve usernames managed via NSS, SSSD, or LDAP.
 func (fi *FileInfo) User() string {
 	return fi.user
 }
 
 // Group is the string representing the owner user group.
+//
+// Because Pebble is built with CGO_ENABLED=0, group lookups rely on /etc/group
+// and will not resolve group names managed via NSS, SSSD, or LDAP.
 func (fi *FileInfo) Group() string {
 	return fi.group
 }
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -518,6 +518,14 @@ pebble run --verbose
 
 The `ls` command is used to list path contents.
 
+```{note}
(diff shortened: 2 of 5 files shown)


client/files.go
```diff
@@ -105,11 +105,17 @@ func (fi *FileInfo) GroupID() *int {
...
+// Because Pebble is built with CGO_ENABLED=0, user lookups rely on /etc/passwd
+// and will not resolve usernames managed via NSS, SSSD, or LDAP.
 func (fi *FileInfo) User() string {
 	return fi.user
 }
 
 // Group is the string representing the owner user group.
+//
+// Because Pebble is built with CGO_ENABLED=0, group lookups rely on /etc/group
+// and will not resolve group names managed via NSS, SSSD, or LDAP.
```
benhoyt: I don't think these comments are worth it here.

docs/reference/cli-commands.md
```diff
@@ -518,6 +518,14 @@ pebble run --verbose
 
 The `ls` command is used to list path contents.
 
+```{note}
+Pebble is statically compiled with `CGO_ENABLED=0`, which means Go's user
+and group lookup uses a pure-Go implementation reading directly from
+`/etc/passwd` and `/etc/group`. On systems where users and groups are managed
+by NSS, SSSD, or LDAP, usernames and group names may not resolve in `pebble ls -l`
+output and will display as numeric IDs or empty fields.
```
benhoyt: Which is it, numeric IDs or empty fields?

client/files.go
chapman-von-monty: Removed these comments to keep the client method docs focused.

docs/reference/cli-commands.md
chapman-von-monty: In `pebble ls -l`, the table formatter outputs `fi.User()` and `fi.Group()` directly, so unresolved names display as empty fields (numeric IDs are only returned in JSON/YAML output or via the API). Updated the note to clarify that they display as empty fields.

chapman-von-monty:
@hpidcock This PR is ready for final review. The spelling issue in the docs check has been fixed, and all documentation changes are up to date.

hpidcock (approved):

PR #944, opened 2026-09-09
ci: do not change workflows in automerge

hpidcock:
This fixes FIPS merge resolution PR creation when there are actions workflow changes.
As it turns out, it is impossible for a GitHub Actions workflow to make changes to
workflows with the default tokens.

To side step this without introducing another bot account or GitHub app, we simply
give the user a patch to optionally apply on top of the merge PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -171,13 +171,13 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
-      actions: write
     outputs:
       target-branch: ${{ steps.vars.outputs.base-branch }}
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):
Interesting workaround to the workflow problem. I'm looking forward to the day the FIPS branch goes away :).

Reading over this with my human eyes, it makes sense and I don't see anything needing fixing. I think the small amount of duplication is fine within the workflow and not worth over engineering by factoring it out. I'm trusting your fork testing (and the imminent testing here) in terms of it actually working e2e.

PR #850, opened 2026-04-22
revert: return full change object for `tasks` cmd

Snehil-Shah:
Roughly reverts https://github.com/canonical/pebble/pull/844/changes/9c7f7453360676b5b5114748ee96e7192abc337f as discussed in https://github.com/canonical/pebble/pull/844#pullrequestreview-4127211909.

Fixes #849

PR #947, opened 2026-09-14
ci: allow revert type for conventional commit

dariuszd21:
I've noticed that this is not possible to use `revert` as a type of conventional commit.

Looking at the commit history, I've found the PR #850, which would fail with the current checker.

The conventional commit spec doesn't have a clear definition of how to handle reverts according to [docs](https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits), but
they mention a recommendation that `revert` might be a valid type.


hpidcock (approved):
LGTM, thank-you.

PR #948, opened 2026-09-15
fix(daemon): stop overlord before HTTP server on shutdown

### LINKED ISSUE #682: One second shutdown delay when exec command is running
The Polar Signals analysis pointed out that Pebble has a 1s delay on shutdown due to websockets (that is, when an `exec` command is still running). Pebble should send `SIGTERM` to the process and terminate it gracefully (and quickly if the process is behaving well), so there shouldn't be a full 1s delay. Try to repro and determine what's going on.

If this happens when the process doesn't cleanly exit, I think that's working as intended, but otherwise we should look at fixing it.

chapman-von-monty:
Fixes #682: stopping the daemon while an `exec` command is running used to
take about a second (the HTTP server's shutdown timeout), and would make
`pebble run` exit with a spurious `context deadline exceeded` error.

## Root cause

`pebble exec` keeps a long-poll `GET /v1/changes/{id}/wait` HTTP request
open on the daemon while the remote command runs (this is how the CLI
learns the command has finished). That connection only becomes idle once
the change is marked ready, which happens once `Overlord.Stop` kills the
running exec's process/tomb.

`Daemon.Stop` called `d.serve.Shutdown(ctx)` (bounded by the 1s
`shutdownTimeout`) *before* `d.overlord.Stop()`. Since `http.Server.Shutdown`
blocks until all active connections go idle, and the long-poll connection
couldn't go idle until the (not-yet-stopped) overlord killed the running
command, `Shutdown` always ran out its full 1s timeout whenever an exec was
in flight, returning `context.DeadlineExceeded`, which then became
`Daemon.Stop`'s (and `pebble run`'s) fatal error.

## Fix

Stop the overlord (which kills any concurrently-running task, including
`exec` commands) *before* calling `serve.Shutdown`. This way, any open
long-poll or websocket connections for running commands are already being
torn down by the time `Shutdown` starts waiting for connections to become
idle, so it returns almost immediately instead of hitting its timeout.

## Verification

Manually reproduced on the release current when the issue was filed
(v1.24.0, from 2025-08-27, close to the issue's 2025-08-15 creation date)
and on current `master` before the fix: starting `pebble run`, running
`pebble exec -- sleep 30` against it, then sending `SIGTERM` to the daemon,
consistently showed:
- ~1.0s delay before the process exited.
- `cannot run daemon: context deadline exceeded` printed to stderr.

After the fix, the same scenario shuts down in ~25ms with no error.

Added a regression test, `execSuite.TestStopWhileExecRunning` in
`internals/daemon/api_exec_test.go`, which starts a long-running `exec`,
calls `Daemon.Stop` directly, and asserts it returns within 500ms with a
nil error. Confirmed this test fails (reproducing the exact `1.00...s`
delay and `context deadline exceeded` error) when reverting the
`daemon.go` change, and passes with the fix applied.

Ran the full test suite (`go test ./...`); no new failures introduced (a
handful of pre-existing failures remain, all requiring
`PEBBLE_TEST_USER`/`PEBBLE_TEST_GROUP` env vars to be set, unrelated to
this change).

Fixes #682


hpidcock:
@chapman-von-monty The root tests are failing now (I re-ran it once). Additionally, we need to make sure there is test coverage to ensure that the http server is no longer accepting new requests before telling the overlord to stop.

chapman-von-monty:
Addressed the shutdown ordering feedback in commit `752bccc`.

- `http.Server.Shutdown` now starts first and closes the HTTP listeners.
- The daemon waits for listener closure before stopping the overlord, so no new requests can begin while the overlord is stopping.
- Active requests remain available to finish after overlord cancellation, and shutdown completion is awaited afterward.
- Added regression coverage proving listeners close before overlord stop.

Verification passed locally:
- `go build ./...`
- `go vet ./...`
- `PEBBLE_TEST_USER=ubuntu PEBBLE_TEST_GROUP=ubuntu go test ./... -count=1`
- `PEBBLE_TEST_USER=ubuntu PEBBLE_TEST_GROUP=ubuntu go test -race ./internals/daemon/... -count=2`

The earlier Root Tests failure was `TestPostChecksRefresh`, which is a pre-existing flaky failure: the same failure occurred on master in workflow run `34425306239`, root job `102709165125`. No checks are currently reported for the updated branch yet.

hpidcock (approved):

PR #1071, opened 2026-09-18
refactor(cmd): embed version file instead of generating Go source

### LINKED ISSUE #962: cmd: escape version before generating Go source
`cmd/mkversion.sh` interpolates the selected version directly inside a Go string literal in generated source. A legal Git tag or command-line value containing a quote or backslash can produce invalid Go or alter/inject generated source. Serialize or escape the version as a Go string before writing `version_generated.go`.

Found during cmd package review.

### LINKED ISSUE #963: cmd: preserve git describe failures in mkversion fallback
`cmd/mkversion.sh` pipes `git describe` through `sed` while selecting a version. Without pipefail, a failed `git describe` can be masked by successful `sed`, causing empty output to be accepted and potentially falling through to an unrelated changelog version. Validate the command and output explicitly (or use pipefail where supported) before fallback.

Found during cmd package review.

chapman-von-monty:
Updates Pebble version handling to embed a plain version file rather than generating Go source during builds, and preserves failures while deriving versions from Git.

## Summary & Motivation

The generated Go version source made version handling dependent on generated files and required special-case packaging. Embedding a plain version file keeps the source file stable while preserving build-time version generation. The version fallback also needs to distinguish a failed `git describe` from an empty result so failures do not get masked by `sed`.

## Technical Details / Changes

- Updated `cmd/mkversion.sh` to write the resolved version to `cmd/version` without a trailing newline.
- Reworked `cmd/version.go` to use `go:embed` while preserving `MockVersion`.
- Added the default `cmd/version` file and removed generated-file ignore rules.
- Updated release and snap workflows to update and package the embedded version file.
- Explicitly checks `git describe` before transforming its output, allowing the changelog fallback only when Git cannot provide a version.

## QA / Verification

- `go build ./cmd/pebble`
- `go generate ./cmd`
- `go run ./cmd/pebble version --client`
- Tested `mkversion.sh` with Git available and with a failing Git executable.
- Verified the generated version file has no trailing newline.

## Links

Fixes #962
Fixes #963


hpidcock:
@chapman-von-monty change the `cmd/version` file to `cmd/VERSION` so it more cleanly works with the snap building.

chapman-von-monty:
Updated the embedded version file to `cmd/VERSION`.

Changes include:
- Renamed `cmd/version` to `cmd/VERSION`.
- Updated `cmd/mkversion.sh` to write to `cmd/VERSION`.
- Updated `cmd/version.go` to use `//go:embed VERSION`.
- Updated references across workflows (`release.yml`, `snap.yml`) and `snapcraft.yaml`.
- Verified with `go generate ./cmd`, `go build ./cmd/pebble`, and `PEBBLE_TEST_USER=ubuntu PEBBLE_TEST_GROUP=ubuntu go test ./...`.

hpidcock (approved):

hpidcock (approved):

PR #1079, opened 2026-09-18
fix(cmdstate): reject and close duplicate websocket connection after upgrade

### LINKED ISSUE #990: Concurrent cmdstate websocket connects can panic and leak connections
`execution.connect` checks for an existing websocket under a mutex, unlocks while upgrading, then unconditionally stores the new connection. Concurrent requests for the same execution and websocket ID can both pass the check and upgrade. The later connection overwrites/leaks the first; for control/readiness channels both handlers can close the same channel, causing a `close of closed channel` panic. Reserve the slot atomically or reject/close the loser after reacquiring the lock.

chapman-von-monty:
Summary & Motivation

Concurrent requests for the same execution websocket could both upgrade successfully, after which the later connection overwrote the first and could cause double channel closes. This rejects the losing connection instead.

Technical Details / Changes

- Re-check the websocket slot while holding the mutex after the upgrade.
- Close the newly upgraded websocket and return the existing websocket-already-connected error when another connection won the race.
- Add a regression test verifying the duplicate connection is rejected, closed, and does not replace the original.

QA / Verification

- `PEBBLE_TEST_USER=ubuntu PEBBLE_TEST_GROUP=ubuntu go test ./internals/overlord/cmdstate/...`

Links: Fixes #990


hpidcock (dismissed):

internals/overlord/cmdstate/handlers_test.go
```diff
@@ -0,0 +1,50 @@
...
+		second.Close()
+		t.Fatal("duplicate websocket unexpectedly connected")
+	}
+	if err := <-errors; err == nil || err.Error() != "control websocket already connected" {
+		t.Fatalf("unexpected duplicate connection error: %v", err)
+	}
+	if e.websockets[wsControl] != connected {
+		t.Fatal("duplicate websocket replaced the connected websocket")
+	}
+}
```
hpidcock: @chapman-von-monty Re-write this test to follow the test writing conventions of pebble.

chapman-von-monty:
I've rewritten the regression test in `internals/overlord/cmdstate/manager_test.go` to follow Pebble test writing conventions using `gocheck` (`gopkg.in/check.v1`), adding the test case to `managerSuite`, using `c.Assert`, and adding helper exports (`AddFakeExecution`, `ExecutionWebsocket`) following the project's testing conventions. Both `./internals/overlord/cmdstate/...` and the full test suite pass cleanly.

hpidcock (approved):

PR #1084, opened 2026-09-18
fix(state): copy user ID and data map in AddNotice

### LINKED ISSUE #1021: AddNotice retains caller-owned user ID and data
`State.AddNotice` stores the caller's user ID pointer and `options.Data` map directly in the notice. Mutating those values after AddNotice changes already-recorded state without a state API; changing the user ID can even make the map key inconsistent with the notice value. The method should copy the scalar and data map.

chapman-von-monty:
State.AddNotice retained pointers and maps supplied by callers, allowing mutations after recording a notice to alter persisted state.

## Summary & Motivation

Prevent callers from mutating already-recorded notice state by copying the user ID and occurrence data during AddNotice.

## Technical Details / Changes

- Copy the user ID value into a new pointer when creating a notice.
- Shallow-copy the occurrence data map before storing it.
- Add regression coverage for mutations to both inputs after AddNotice returns.

## QA / Verification

- `PEBBLE_TEST_USER=ubuntu PEBBLE_TEST_GROUP=ubuntu go test ./internals/overlord/state/...`

## Links

Fixes #1021


internals/overlord/state/notices.go
```diff
@@ -246,9 +247,14 @@ func (s *State) AddNotice(userID *uint32, noticeType NoticeType, key string, opt
 	if !ok {
 		// First occurrence of this notice userID+type+key
 		s.lastNoticeId++
+		var noticeUserID *uint32
+		if userID != nil {
+			copiedUserID := *userID
+			noticeUserID = &copiedUserID
```
benhoyt: I believe we can use `new` now:

```suggestion
			noticeUserID = &new(userID)
```

hpidcock: More specifically, this would be the most idiomatic way:
```go
		if userID != nil {
			userID = new(*userID)
		}
```

@chapman-von-monty make this change.

chapman-von-monty:
Updated to use `userID = new(*userID)` when copying non-nil user IDs. Tests continue to pass.

hpidcock (approved):

